PR #9, opened 2019-03-28
Fix select() example

keskitalo:
This pull request adds a necessary keyword to the select() method example. It also formats config.py with black.

### DIFF
--- a/sotodlib/hardware/config.py
+++ b/sotodlib/hardware/config.py
@@ -151,10 +151,10 @@ class Hardware(object):
             26 which have "A" polarization and are located in pixels 20-29
             (recall the "." matches a single character)::
 
-                new = hw.select({"wafer": ["25", "26"],
-                                 "band": "MF.1",
-                                 "pol": "A",
-                                 "pixel": "02."})
+                new = hw.select(match={"wafer": ["25", "26"],
+                                "band": "MF.1",
+                                "pol": "A",
+                                "pixel": "02."})
 
         Args:
             telescopes (str): A regex string to apply to telescope names or a


tskisner:
Thanks for fixing that docstring.  

The source file was already PEP8 compliant- if we are going to use the particular formatting enforced by the "black" tool (https://github.com/ambv/black), then we should be sure to have consensus on this, and also we should apply that to all source files.

@mhasself and @smsimon , do you have any strong opinions about using this format everywhere:

https://github.com/ambv/black#the-black-code-style

?  By default this formatter uses 88 characters per line, but this is configurable.  We have been or are planning to use this tool for some of our other projects in order to never have to think about formatting again.  I am fine if we want to use it here, but just wanted to get feedback first.

mhasself:
From the dev guide: "Large reformatting jobs should not be undertaken unless there is an immediate intention to then make extensive modifications to the code.  In cases where extensive reformatting is necessary, commits carrying the reformatting must not also include any changes to the functionality or interface."  That's taken slightly out of context, but I think it's a good practice to keep large reformatting jobs  separate from non-behaviorally-inert changes.

Adoption/recommendation of Black is a separate issue, which can be discussed elsewhere... but I'm not sure I'm ready to "cede control over minutiae of hand-formatting".  I think a general recommendation to refer to PEP8, and for more experienced users to monitor PRs for any egregious formatting practices, is enough for most SO code.

tskisner:
Ok.  PEP8 line length guidelines are one of those much debated topics ;-)  For now, lets just fix this docstring.  For new source files or during the next overhaul we could use the black formatter with a 79 character threshold (or longer if our dev style guide allows it).  I push a change shortly to this branch.

PR #11, opened 2019-04-03
Add Carlos Sierras' LAT noise fits to the example bands

keskitalo:


tskisner:
Looks good, as long as you think the names "A" and "C" are unambiguous enough.  Merge away.

PR #14, opened 2019-04-18
change crate and card indexing to match hardware plan

kmharrington:
I noticed that the current `get_example` doesn't index the crates / cards for the readout system like we'd expect. Crates can't be split between different telescopes. 

This update splits the crates by telescope and makes sure the cards in a crate all come from wafers in the same telescope.

### DIFF
--- a/sotodlib/hardware/config.py
+++ b/sotodlib/hardware/config.py
@@ -279,28 +279,6 @@ def get_example():
     """
     cnf = OrderedDict()
 
-    cards = OrderedDict()
-    crates = OrderedDict()
-
-    cdindx = 0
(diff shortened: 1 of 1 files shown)


tskisner:
Hello @kmharrington , not sure if you were looking for a review of this, but the changes look fine to me...  

One tiny request- could you update the single quotes to double to match the surrounding code?  PEP8 does not give a recommendation for this (except for triple-quoted strings, where double quotes are preferred).  I'm happy to convert all the source code to single quotes if everyone prefers that.  Mainly I just care about consistency through the code base.

Aside from that, merge away :-)

PR #16, opened 2019-05-28
Initial Data Processing

kmharrington:
This is a basic framework for timestream data processing with SO. 

It's mostly built off a base class `DataG3Module` that is set up to allow anyone to quickly build G3Modules that will a parse through a pipeline, find frames that are `G3FrameType.Scan` and run a function `process` on every detector timesteam under a give flag. 

I've used this `Data3GModule` to build example filters (`sotodlib.data.filter`) and tod operations like detrending (`sotodlib.data.condition`). I've also used it to build data simulation examples like `MakeNoiseData` (`sotodlib.data.sim`).

Includes modest updates to the documentation and test cases. I plan to talk about this more at the meeting tomorrow and wanted to advertise the update now.


tskisner:
One thought- usually the __init__.py has just namespace imports and other lightweight things.  Here is a nice description:

https://docs.python-guide.org/writing/structure/#packages

Would it be possible to move the base class into its own file?

kmharrington:
alright, I've moved the DataG3Module out of `__init__.py` and into a new file called `core.py`

PR #20, opened 2019-06-13
Per-detector bandpass and noise parameters

keskitalo:
This PR introduces a pipeline script to generate nominal and perturbed hardware configurations.

The perturbation means that detectors can no longer reference the band-averaged NET, fknee, bandwith etc. Instead, the script adds these parameters to the individual detector dictionaries. I have checked that these additional keywords get saved in the output TOML files and become available to future scripts that load the modified TOML file.


keskitalo:
Merging this branch to make way for more development. If the individual detector extension to the TOML files goes against the design we can revisit.

PR #21, opened 2019-06-20
Split schedule

keskitalo:


keskitalo:
Self-merging as this PR only concerns the toast_so_sim.py pipeline script.

PR #23, opened 2019-06-26
support custom SO models for PySM 3

zonca:
Check the command line help for the default string that should be used for the simulations.

This requires:

* TOAST 2.3
* PySM 3 integration https://github.com/hpc4cmb/toast/pull/277
* PySM 3 master from https://github.com/healpy/pysm
* so_pysm_models from https://github.com/simonsobs/so_pysm_models

This is currently untested, I'll work on testing this in the next couple of days.

### DIFF
--- a/pipelines/toast_so_sim.py
+++ b/pipelines/toast_so_sim.py
@@ -31,6 +31,13 @@ import toast.todmap as ttm
 
 import sotodlib.hardware
 
+try:
+    import pysm
+    import so_pysm_models
+except:
+    pysm = None
+    so_pysm_models = None
+
 if tt.tidas_available:
     from toast.tod.tidas import OpTidasExport, TODTidas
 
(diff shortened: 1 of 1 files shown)


zonca:
It looks like `toast_so_sim.py` is not yet updated for 2.3, if I try to run it I get:

```
  File "/global/cscratch1/sd/zonca/simonsobs_prefix/bin/toast_so_sim.py", line   8, in <module>    from toast.mpi import MPI, finalize 
ImportError: cannot import name 'finalize'
```

I'll wait for this to be ported then I'll proceed with testing.

pipelines/toast_so_sim.py
```diff
@@ -1458,19 +1470,80 @@ def simulate_sky_signal(args, comm, data, schedules, subnpix, localsm):
...
         focalplanes=[s[3] for s in schedules],
         nside=args.nside,
         subnpix=subnpix,
         localsm=localsm,
         apply_beam=args.apply_beam,
-        coord=args.coord,
+        coord="G", # setting G doesn't perform any rotation
+        map_dist=map_dist,
     )
+    assert args.coord == "Q", "Input SO models are always in Equatorial coordinates"
```
zonca: notice we can only run in Equatorial for now, `so_pysm_models` defaults to Equatorial as well, so all input templates for PySM are in Equatorial, PySM doesn't perform any rotation (we set "G" for output and PySM defaults to input "G" templates as well).
If this is a problem please let me know and we'll think a better solution.

PR #26, opened 2019-07-25
Per wafer datadump

keskitalo:
This PR modifies the TOAST/so3g facilities to
* support breaking the so3g files by wafer, card or other detector characteristic listed in the hardware configuration
* support FLAC compression of the G3TimeStreams. The compression requires casting the floating point-valued TOD into integers. The `recode` method determines the optimal gain and offset to meet the desired number of significant bits within the TOD RMS. The applied gain and offset are recorded in the so3g file and used when decompressing the TOD.
* support loading all flavors of the TOD recorded in the so3g file
* fix a bug in recording the time stamps in the so3g files and add missing method to expand them into the TOAST cache upon load
* Fix a bug in not propagating the common flag mask from `ToastExport` all the way to `tod_to_frames`
* Fix an off-by-one error in reading and writing the flags
* significantly expand the toast_load unit tests to verify the signal, timestamps and flags in the output files
* ensure that the cache names of the TOD objects agree between exported and imported data
* fixes a bug in loading so3g files that only recovered the first frame in each file

### DIFF
--- a/sotodlib/data/toast_frame_utils.py
+++ b/sotodlib/data/toast_frame_utils.py
@@ -34,7 +34,7 @@ FIELD_TO_CACHE= {
 }
 
 
-def recode_timestream(ts, rmstarget=2 ** 10):
+def recode_timestream(ts, params, rmstarget=2 ** 10, rmsmode="white"):
     """ts is a G3Timestream.  Returns a new
     G3Timestream for same samples as ts, but with data
     scaled and translated with gain and offset,
@@ -42,11 +42,22 @@ def recode_timestream(ts, rmstarget=2 ** 10):
 
     Args:
         ts (G3Timestream) :  Input signal
-        rmstarget (float) :  Scale the input signal to have this RMS.
+        compresspar (bool or dict) :  if True, compress with default
+            parameters.  If dict with 'rmstarget' member, override
+            default `rmstarget`.  If dict with `gain` and `offset`
+            members, use those instead.
+        params (None, bool or dict) :  If None, False or an empty dict,
+             no compression or casting to integers.  If True or
+             non-empty dictionary, enable compression.  Expected fields
+             in the dictionary ('rmstarget', 'gain', 'offset', 'rmsmode')
+             allow overriding defaults.
+        rmstarget (float) :  Scale the iput signal to have this RMS.
             Should be much smaller then the 24-bit integer range:
             [-2 ** 23 : 2 ** 23] = [-8,388,608 : 8,388,608].
             The gain will be reduced if the scaled signal does
             not fit within the range of allowed values.
+        rmsmode (string) : "white" or "full", determines how the
+            signal RMS is measured.
     Returns:
         new_ts (G3Timestream) :  Scaled and translated timestream
             with the FLAC compression enabled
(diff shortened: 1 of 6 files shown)


keskitalo:
Thanks for the thoughtful comments. I'll add a guard for RMS=0. The reason for not using the full dynamic range for the compressor gain is that even one outlier data point could completely change the level of re-digitization noise and the resulting compression factor. For consistent compression performance, the number of digitizer bins per RMS should be roughly constant.

I like the idea of using G3MapDouble for the gains and offsets. I'll implement it and let you know.

The unit test I implemented creates files with and without compression under
`sotodlib_test_output/test_toast_load/test_02/sat3_uncompressed_0000*g3`
and
`sotodlib_test_output/test_toast_load/test_02/sat3_compressed_0000*g3`
You just need to check out the `per_wafer_datadump` branch and run `python3 setup.py test`

keskitalo:
10e579af012fa5bc3795b178a8c413020ea69e38 adds the guard against zero RMS and moves the compressor gains and offsets into G3MapDouble dictionaries.

Here are the first 10 samples of the uncompressed and compressed TOD from the unit test:
```
#              uncompressed                compressed                difference
0    -0.0013000782818187858    -0.0012997967132239623    2.8156859482354746e-07
1      0.010655187726935461      0.010655020799081913   -1.6692785354856243e-07
2      0.009219722572911496      0.009219726427861412     3.854949916573536e-09
3    -0.0047861003592696485     -0.004786101946965888    -1.587696239237213e-09
4      0.006463696785600746      0.006463465356064655   -2.3142953609056605e-07
5     -0.009842009639588777     -0.009842002128889924     7.510698852783437e-09
6    -6.542931884586079e-05    -6.560884699213093e-05    -1.795281462701391e-07
7     -0.013299653963396834      -0.01329938108451704    2.7287887979440817e-07
8       0.01744570290484415      0.017445808946986968     1.060421428167313e-07
9      0.008900035222409295      0.008900159926783528    1.2470437423320324e-07
```

We can easily adjust the digitization error using the `rmstarget` argument to `recode`. Larger `rmstarget` will lead to a smaller error and less compression.

mhasself:
Ah, right.  I was wrong to focus only on precision.  I had forgotten about the main goal, which is storing fewer bits per sample.  What kind of compression ratio do you get with this configuration?  I would expect it to be fairly poor for the lowest frequency bands.

I think we want to try to isolate the white noise RMS here, rather than taking the RMS of the whole signal.  For example, compute np.std(np.diff(v))/2**.5 instead of np.std(v).  If we state our acceptable level of precision in units of the white noise RMS, rather than the all-freq RMS, then this will have a more direct correlation with the compression factor achieved (because FLAC is probably quite good at compressing 1/f noise to almost nothing, while white noise is inherently difficult to compress and thus benefits from careful truncation.

My earlier suggestion of truncating at roughly 30 uK was based on keeping about 8 bits of white noise.  Any reasonable estimate of how much noise we should keep suggests we don't even need nearly that many bits -- so it's a safe upper limit that no one will bother us about.

keskitalo:
18b64fef46ca84d8eac72a24bd98645dc36e88b8 re-interprets the `compress` keyword argument to `ToastExport`. If it is `True`, compression will be done with default parameters. If it is a non-empty dictionary, the user can set `rmstarget`, `rmsmode` ("white" or "full") or fix the "gain" and "offset" altogether. When applying user-specified gain and offset, `RuntimeError` is raised if the scaled and translated timestream overflows 24-bit signed integers.

The unit test now compresses the simulated noise with the default parameters (`rmstarget=2**10`) and `rmstarget=2**8`. The achieved compression ratios wrt. the input 64-bit floats are 0.25 and 0.22. The latter choice fails the comparison to input signal with `rtol=1e-3` and `atol=1e-5`.

sotodlib/data/toast_frame_utils.py
```diff
@@ -20,26 +20,118 @@
...
+            rms = np.std(v)
+        else:
+            raise RuntimeError("Unrecognized RMS mode = '{}'".format(rmsmode))
+        if rms == 0:
+            gain = 1
+        else:
+            gain = rmstarget / rms
+            # If the data have extreme outliers, we have to reduce the gain
+            # to fit the 24-bit signed integer range
+            while amp * gain >= 2 ** 23:
```
mhasself: This while loop should be un-indented by one level, to catch the case of zero RMS, no offset, large data value.  Yes, that case is unlikely.

sotodlib/data/toast_frame_utils.py
```diff
@@ -20,26 +20,118 @@
...
+    G3Timestream for same samples as ts, but with data
+    scaled and translated with gain and offset,
+    rounded, and with FLAC compression enabled.
+
+    Args:
+        ts (G3Timestream) :  Input signal
+        compresspar (bool or dict) :  if True, compress with default
+            parameters.  If dict with 'rmstarget' member, override
+            default `rmstarget`.  If dict with `gain` and `offset`
+            members, use those instead.
```
mhasself: compresspar is not a parameter any more.

keskitalo:
c3e7990ff36ed4445c251334ad12ff57f0483bdc fixes that last issues in this PR. Merging.

PR #30, opened 2019-08-14
TOD loading code that leverages an ObsFileDB.

mhasself:
Does not require TOAST, but optionally will populate a toast.TOD.  No
MPI yet, though.  Requires an ObsFileDB, which will live in sotodlib [oops, I meant sotoddb],
and the script to create that DB file, which will be in pwg-scripts.

### DIFF
--- a/sotodlib/data/load.py
+++ b/sotodlib/data/load.py
@@ -1,0 +1,317 @@
+"""Supports loading of TOD data from files disk.  Does not assume you
+have TOAST, but can populate TOAST structures if you do.
+
+For large datasets, we assume that time and detector information can
+be mapped to filenames using an ObsFileDB, implemented in sotoddb.
+Such database structures provide a guide for finding requested data
+without doing expensive searches.
+
+We also support loading streams from specified G3 files, in the
+absence of a database.
+
+There will inevitably be a few different "file formats" out there, at
+the very least some variations in the schema used on top of G3.
+Presently the loader targets the pipe-s0001 "1day" TOD sim for the LAT.
+
+"""
+
+import so3g
+from spt3g import core
+
+import numpy as np
+
+from collections import OrderedDict
+
+class FieldGroup(list):
+    """This is essentially a roadmap for decoding data from a
+    G3FrameObject.  Each entry in this list is either a string, giving
+    the key of a vector that should be loaded from the frame object,
+    or a FieldGroup requesting a recursive decent into a "map" defined
+    in the FrameObject.
+
+    It is used to formulate decoding requests for a frame.  For
+    example, a frame structured like this::
+
+      Frame (Scan) [
+        "boresight" (spt3g.core.G3TimestreamMap) => Timestreams from 3 detectors
+        "signal" (spt3g.core.G3TimestreamMap) => Timestreams from 794 detectors
+        "site_velocity" (spt3g.core.G3VectorDouble) => 3327 elements
+      ]
+
+    might have a roadmap defined like this:
+
+      request = FieldGroup('root', [
+        FieldGroup('boresight', ['az', 'el', 'roll']),
+        FieldGroup('signal', ['det1', 'det2', 'det3']),
+        'site_velocity'])
+
+    Note that the name 'root', of the outermost FieldGroup, is not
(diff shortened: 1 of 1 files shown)


sotodlib/data/load.py
```diff
@@ -0,0 +1,317 @@
...
+    #
+    from toast.tod import TOD
+
+    detquats = []
+    nsamp = len(streams['boresight']['az'])
+    dets = list(streams['signal'].keys())
+    detranks = 1
+    sampsizes = [(0, nsamp)]
+    meta = {}
+    tod = TOD(mpicomm, dets, nsamp, detranks=detranks, meta=meta)
```
kmharrington: These might be more just questions on TOAST than comments on this. But the comments in `toast.TOD` say to never directly instantiate it like this ( https://github.com/tskisner/toast/blob/refactor_2.3/src/toast/tod/tod.py#L34   ) because of all the `NotImplementedError` exceptions that are there. Are those going to be a problem here?

There's also another TOD object made there? https://github.com/simonsobs/sotodlib/blob/toast_refactor/sotodlib/data/toast_load.py#L38 

It seems like we should have one base TOD object for SO, at minimum one base TOD object for loading from SO "data." Do we have one of these yet?

tskisner:
Yes, a TOD base class should never be instantiated.  Instead, I had assumed that we would modify the existing load function to query this obs DB for metadata, rather than traversing the directories of frame files and reading observation frames to get that metadata.

sotodlib/data/load.py
```diff
@@ -0,0 +1,317 @@
...
+
+    """
+    if prefix is None:
+        prefix = db.prefix
+        if prefix is None:
+            prefix = './'
+
+    # Regardless of what dets have been asked for (maybe none), get
+    # the list of detsets implicated in this observation.
+    c = db.conn.execute('select distinct DS.name, DS.det from detsets DS '
```
tskisner: This DB query is what could be used when loading data into toast, rather than the current traversal of the filesystem to find observations:  https://github.com/simonsobs/sotodlib/blob/toast_refactor/sotodlib/data/toast_load.py#L574

sotodlib/data/load.py
```diff
@@ -0,0 +1,317 @@
...
+        
+    FieldGroup.hstack_result(streams)
+
+    if not toast:
+        return streams
+
+    #
+    # Re-process streams into standard TOD.cache info.
+    #
+    from toast.tod import TOD
```
tskisner: Is this frame unpack-and-concatenate on a single process intended to replace the existing frame broadcast and unpack for toast?  We could consider how to best share code for this.  For example, the existing toast loading function could broadcast the frames and then use your unpacking methods on every process.

In other words, if you have code which does "unpack SO3G frames into dictionaries of arrays, keeping all the things we want" (which is what I think you are intending with this code), then I can modify this code:

https://github.com/simonsobs/sotodlib/blob/toast_refactor/sotodlib/data/toast_load.py#L201

To call your functions here on each process to unpack the data and store it appropriately in the SO toast TOD class.  However, I don't think we want to add toast stuff here.  We should just update the existing code in sotodlib.data.toast_load to use this.

mhasself:
Ya, apologies for not actually reading about the toast.TOD much.  The reason I stayed away from the SOTOD class is indeed because the constructor triggers data load.  How does one instantiate an SOTOD without triggering a load from disk?  files = []?  How does one create an SOTOD in other contexts, such as when making a simulation?

Over the years I have come to prefer that data container objects have a fairly minimal constructor, with I/O (or other useful constructors) being defined in class functions or non-class factory functions.  This may seem sort of anti-RAII... but I think it isn't, provided that the (perhaps rather basic) constructor and all such factory/class functions produce objects that are internally consistent.

tskisner:
Hello @mhasself , just to clarify, I think we should remove all reference to toast from this source file.  Then we can merge this PR and open a separate PR that modifies the toast functions to use the code here.

mhasself:
@tskisner  Yup, that makes sense.  Will do.

mhasself:
Any objections to merging this into toast_refactor prior to merging toast_refactor into master... please let me know in the next day or so.

PR #37, opened 2019-09-30
Fix versioneer configuration.

mhasself:
Use tags of the form vX.Y.Z.  Set the docs version code through
`__version__`.


tskisner:
Sure, this is fine.  You can also use an empty string for the tag prefix if you like.

PR #39, opened 2019-10-22
intro.rst: use pip instead of setup.py directly

ickc:
Uses pip to manage the package directly. Useful when in the future there are other dependencies. See e.g. <https://stackoverflow.com/questions/30306099/pip-install-editable-vs-python-setup-py-develop/30306403>


mhasself:
Thanks!

PR #40, opened 2019-10-24
Automatically select the models with small scales if we are simulating high nside

zonca:
`so_pysm_models` has 2 versions for each model,
e.g. `SO_d0` is the same as PySM 2, `SO_d0s` has small scales signal added.

We want to automatically select the models with small scales if we are running at high nside


zonca:
If you think this is good, please merge, I don't have write access (and I don't need it)

keskitalo:
The unit test failure seems like transient network issue and only affected one of the builds. Merging.

PR #43, opened 2020-02-12
Modify matplotlib backend to PDF inside plot function

### LINKED ISSUE #32: Remove matplotlib.use
I get some warnings when I build the docs for `mapsims` due to the line:

    matplotlib.use("pdf")

at https://github.com/simonsobs/sotodlib/blob/master/sotodlib/hardware/vis.py#L9

Would it be possible to remove that line and have the user choose the matplotlib backend before importing the library?

zonca:
Closes #32

### DIFF
--- a/sotodlib/hardware/vis.py
+++ b/sotodlib/hardware/vis.py
@@ -4,12 +4,8 @@
 """
 
 import numpy as np
-
-import matplotlib
-matplotlib.use("pdf")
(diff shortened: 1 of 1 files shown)


tskisner:
Thanks for adding this workaround.  If it really becomes too annoying in the future for interactive use, we can then remove it completely and tell people in the docs to use PDF backend for nice plots.

PR #64, opened 2020-04-30
Fix stray bad import

mhasself:


mhasself:
This is basic... self-merging.

PR #70, opened 2020-05-14
Adding FlagManager and various TOD operations

kmharrington:
This adds a `FlagManager` extension to `AxisManager` that's specifically designed to deal with cuts and flags in the data. I imagine we'll expand it more as we work with it but it's a good set of baseline operations. I add them to the tod by default in `load_observation`. 

Also adding nice/easy FFT-ing and frequency-space filtering for timestreams (meant to be "pretty good" but might not be for every application). `flags.py` and `sim_flags.py` are meant to hold detection and creation algorithms for flags of all sorts. More sorting / organization of them will occur once we have enough for that to matter.


sotodlib/core/flagman.py
```diff
@@ -0,0 +1,163 @@
...
+        elif method == 'intersect':
+            op = lambda x, y: x*y
+        else:
+            op = method
+        out = reduce(op, to_reduce)
+        return out
+    
+    def collapse_all(self, method='union'):
+        """same as collapse but for all fields, to be more verbose"""
+        return self.collapse(new_flag, list(self._fields.keys()), method)
```
mhasself, line 132: This is typo... new_flag not defined.  Also it seems like this is intended to behave exactly like self.collapse() so I'm not sure you need it.

sotodlib/core/flagman.py
```diff
@@ -0,0 +1,163 @@
...
+        # drop the fields if needed
+        if not keep: 
+            for f in flags: 
+                self.move(f, None)
+        
+        self.wrap(new_flag, new)
+        
+    def reduce_all(self, new_flag, method='union', keep=False):
+        """same as reduce but for all fields, to be more verbose"""
+        return self.reduce(new_flag, list(self._fields.keys()), method, keep)
```
mhasself, line 157: Once again... I think the verbosity would be better spent in the docstring of collapse() and reduce()!  For example, state what kind of object they return, and whether they modify self or not.

sotodlib/core/flagman.py
```diff
@@ -0,0 +1,163 @@
...
+        Arguments:
+        
+            n_buffer: number of samples to buffer the samps cuts
+            flags: List of flags to buffer. Uses their names
+        """
+        new = self.copy()
+        new.buffer(n_buffer, flags)
+        return new
+        
+    def collapse(self, flags=None, method='union'):
```
mhasself, line 98: I don't get a great semantic vibe from "collapse" vs. "reduce" ...  So the difference between them is only in whether the result is stored back in the FlagManager... but their names do not help me remember which of them does that (nor does the docstring, really).  I would be more partial to keyword activation.  E.g.  fm.reduce(save_as='uber_flag'), with default behavior to not save it but only return it.

sotodlib/flags.py
```diff
@@ -0,0 +1,83 @@
...
+    tod: AxisManager object
+    qlim: percentile used to find turnaround
+    merge: If true, merge into tod.flags
+    name: name of flag when merged into tod.flags
+
+    Returns:
+    --------
+    Ranges object
+
+    """
```
mhasself, line 24: This docstring is a different style from all the others.

sotodlib/core/flagman.py
```diff
@@ -0,0 +1,163 @@
...
+    are True
+    """
+    
+    def __init__(self, det_axis, samp_axis):
+        super().__init__(det_axis, samp_axis)
+                
+        if not 'dets' in self._axes:
+            raise ValueError('FlagManagers require a dets axis')
+        if not 'samps' in self._axes:
+            raise ValueError('FlagManagers require a samps axis')
```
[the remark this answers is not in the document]
kmharrington: I think the issue is with my docstring, which I'll update. I expect the `FlagManager` to always be created with both axes (just like I always expect a TOD that has a samps axis to also have a detector axis, are there times where we wouldn't have that _and_ want flags?) The individual flags in the manager can have any combination of those axes.

I had left out the "exclusivity" checks just in case someone wanted to do something "stupid" that wasn't going to interfere with the functioning of the manager.

sotodlib/core/flagman.py
```diff
@@ -0,0 +1,163 @@
...
+                s = data.shape
+            except:
+                ### catches if a detector mask is just a list
+                s = np.shape(data)
+            if len(s) == 1:
+                if s[0] == self.dets.count:
+                    ## detector only flag. Turn into RangesMatrix
+                    axis_map=[(0,'dets')]
+                elif s[0] == self.samps.count:
+                    axis_map=[(0,'samps')]
```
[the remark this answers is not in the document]
kmharrington: I'm going to do a couple things here:

1) I'll raise a "cannot auto determine type" error if `dets.count==samps.count`, that will deal with that possible edge case. 
2) We're going to enforce the dets/samps ordering for 2D flags. Because I realize other parts of this code assumes that exists anyway.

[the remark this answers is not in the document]
kmharrington: And 3. Going to define `wrap_samps`, `wrap_dets`, `wrap_dets_samps` too, since if you're building a pipeline type thing those could be used by default.

kmharrington:
@mhasself  I think this fixes everything you commented on?

PR #72, opened 2020-06-14
Run unit tests directly on github, using the so3g docker container.

tskisner:
github actions have worked well for us in other contexts.  This small PR aims to fix our unit tests by running them inside the latest so3g container and pip installing toast.  Still a work in progress.


tskisner:
Ok, this is ready for review.  Since some of our dependencies (boost + spt3g + so3g) take a while to install, this work uses the upstream so3g docker container as the starting point for running tests.  Then it pip installs toast and sotoddb.

This means we essentially only have one entry in our test matrix (Ubuntu 18.04 with python 3.6), but that is better than nothing.  If we wanted to expand to python 3.7 and 3.8 that would require alternate so3g containers (for example based on the official upstream python debian ones).  To test on MacOS would require something like pip wheels for so3g, which would be a massive undertaking.

tskisner:
Good point, will update the badge before merge.

mhasself:
Installing the latest software isn't always the best idea -- can lead to unnecessary version creep in the dependencies.  I had originally hoped we could skate along being compatible with the versions of major python packages available in Ubuntu 18.04.  But... pixell now requires (due to astropy, due to...) a numpy that is beyond 18.04.

My preference here would be to specify particular versions for key packages like scipy/numpy rather than just getting the latest.  But that's not our biggest problem right now, so don't let that hold this up.

tskisner:
When using the default numpy / scipy in the so3g container and running the sotodlib unit tests, I get this error:

https://stackoverflow.com/questions/59474533/modulenotfounderror-no-module-named-numpy-testing-nosetester

Caused by scipy using a deprecated internal interface in `numpy.testing`.  The solution is to update scipy to a newer version that fixes that.

Here is a concrete recommendation (I can open this against `so3g` if it sounds good):

- Use the `python-slim-buster` official python containers as the base.  This gives you a debian system with a specific python version (3.6, 3.7, 3.8).

- Apt install boost and dependencies.  This is the latest debian, so it has fairly modern versions (gcc 8.3, boost 1.67).

- Use pip to install specific python package versions (these images ship with a "bare" python installation)

- Install spt3g and so3g

So this lets us make an so3g image for each python version.  Then in sotodlib we can test against 3.6, 3.7, and 3.8.

Another option would be to build 2 images for so3g, one of them the "oldest" supported environment (say CentOS6 with python-3.6) and one of them the newest (say ubuntu 20.04 with python-3.8) and then test against those.

We really do need to use docker for this obviously rather than installing on the fly.  The spt3g install with all the cereal-templated object files is 1.5GB and takes a while to build...

tskisner:
Ok, merging this now, and then we can consider other flavors of so3g containers on the so3g repo.

PR #73, opened 2020-06-18
use the new TOAST el-nod facility

keskitalo:
Modest changes to the TOAST interface to make use of the el-nod facilities.


keskitalo:
Will self-merge this PR once the checks complete. It only touches the TOAST interface.

PR #81, opened 2020-07-10
tod_ops: gap-filling

mhasself:
Introduces simple polynomial (linear, really) gap filling, as well as "model-based" gap-filling that can use the results of a PCA (or common mode).

The Extract and ExtractMatrix classes (the later of which is not quite compatible with storage in AxisManager, but could be moved in that direction) help with storing sparse subsets of a TOD signal efficiently.

### DIFF
--- a/sotodlib/tod_ops/gapfill.py
+++ b/sotodlib/tod_ops/gapfill.py
@@ -1,0 +1,331 @@
+import numpy as np
+
+
+class Extract:
+    """Container for storage of sparse sub-segments of a vector.  This
+    ties together a Ranges object and a data array with the same
+    length as the total length of the Ranges segments.
+
+    This is useful in cases where a relatively small number of samples
+    need to be excised from a signal vector, but stored and perhaps
+    restored at a later time.
+
+    In docstrings below a "full data vector" is a 1-d array with
+    self.ranges.count elements; the "tracked samples" are a subset
+    consisting of self.ranges.mask.sum() elements.
+
+    Attributes:
+        ranges: The so3g.proj.Ranges object mapping tracked samples
+            into full data vector.
+        n_ex: The number of tracked samples.
+        data: The data vector containing tracked samples.
+    """
+    def __init__(self, ranges, init_data=True):
+        """Constructor.
+
+        Arguments:
+            ranges: a Ranges object; all positive ranges will be
+                tracked.
+            init_data: array or boolean.  If False, self.data is
+                initialized to None.  If True, self.data is
+                initialized to zeros.  Otherwise, init_data is stored
+                in self.data.  Note this is a reference, not a copy.
+
+        """
+        self.ranges = ranges.copy()
+        rr = self.ranges.ranges()
+        self.n_ex = rr[:, 1].sum() - rr[:, 0].sum()
+        if init_data is False or init_data is None:
+            self.data = None
+        elif init_data is True:
+            self.data = np.zeros(self.n_ex)
+        else:
+            self.data = init_data
+
+    @property
+    def dtype(self):
+        return self.data.dtype
+
+    @property
+    def shape(self):
+        return (self.ranges.count,)
+
+    def offset_iter(self):
+        """Returns an iterator for use in expanding / collapsing extracts.
+        Each iteration returns a tuple (ex_lo, ex_hi, full_lo,
+        full_hi) where ex_lo:ex_hi is a range in extracted data
+        vector, and full_lo:full_hi is a range in the full data
+        vector.
+
+        """
+        offset = 0
+        for lo, hi in self.ranges.ranges():
+            yield offset, offset + hi - lo, lo, hi
+            offset += hi - lo
+
+    def expand(self, fill_value=0):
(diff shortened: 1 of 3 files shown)


mhasself:
If there are no volunteers to review this, I will merge it in 24 hours.

sotodlib/tod_ops/gapfill.py
```diff
@@ -0,0 +1,331 @@
...
+        for o, i in zip(signal, self.items):
+            i.accumulate(o, scale)
+        return signal
+
+
+def get_gap_fill_single(data, flags, nbuf=10, order=1, inplace=False):
+    """Computes samples to fill the gaps in data identified by flags.
+    Each flagged segment is modeled with a polynomial of the order
+    specified, based on up to nbuf points on each side of the segment.
+
```
tskisner, line 188: General questions about the use case of this gap-filling:  If you apply this kind of gap-filling technique to data which contains signal + noise, it will put some amount of fake signal in the gap which could then leak back into unflagged samples when applying a filter.  The only kind of gap-filling I have experience with is the case of first estimating a noise timestream (by subtracting a signal estimate) and then using that to build a constrained noise realization across all gaps simultaneously (to ensure that long-timescale correlations are included).

Of course if your timestream is completely noise dominated, then you can skip the subtraction of the signal estimate.  So are you just treating atmosphere as "noise" in this case?

Even in this noise dominated regime, are you not concerned about gap-filling individual gaps versus all of them simultaneously for the entire CES?

Zhilei:
One overview question is whether we can gap-fill with some functions other than polynomials. I am asking since if we have the SAT time streams, which are modulated, it may be great if we can gap-fill with a sinusoidal function, matching the overall tend (frequency, phase, amplitude) of the unflagged data.

kmharrington:
> One overview question is whether we can gap-fill with some functions other than polynomials. I am asking since if we have the SAT time streams, which are modulated, it may be great if we can gap-fill with a sinusoidal function, matching the overall tend (frequency, phase, amplitude) of the unflagged data.

The SAT timestreams will have the HWPSS in them, which for the most part won't be sinusoidal? They will have shapes depending on the position across the focal plane right?

tskisner:
I guess my point is, "what are you doing with the gap filled data"?  If you are just making a plot or a binned map, or something then that is fine.  If you are applying a filter to the gap-filled timestream, then any signal in the gap will impact nearby samples, and one would need to expand the flagged region afterwards to avoid those contaminated regions.

Zhilei:
> > One overview question is whether we can gap-fill with some functions other than polynomials. I am asking since if we have the SAT time streams, which are modulated, it may be great if we can gap-fill with a sinusoidal function, matching the overall tend (frequency, phase, amplitude) of the unflagged data.
> 
> The SAT timestreams will have the HWPSS in them, which for the most part won't be sinusoidal? They will have shapes depending on the position across the focal plane right?

You are probably right, I just naively guessed that it will be a sinusoidal function... I think it will probably depend on the position of the focal plane.

To answer @tskisner 's question, I was more thinking along the line of the ringing effects from FFT if the boundaries between the gap-filling part and the original data are not smooth.

sotodlib/tod_ops/gapfill.py
mhasself: I agree that, because of filtering, filling a single sample potentially has long-range consequences.  The ACT map-maker (IIRC) actually solves for the value of flagged samples, so the gap-filled values are just starting points in the iterative solution.

However, the gap-filled data _are_ used to estimate a noise model for the max-L solution.  As you imply, the thing to worry about is the atmosphere; it is 1/f "noise" with strong correlations between detectors.  We've actually seen that naive gap-filling can lead to noise models that have higher map noise, i.e. are highly sub-optimal, especially when the data contain gaps >= 1/f_knee.  Our modern approach does two-passes -- at first you just fill single time-stream gaps, then you do a PCA to constrain the low-frequency correlated modes and use that for a superior gap fill.  Then you estimate a (correlated) noise model.

But I agree with your statements (especially below) that some care should be taken with mapping of gap-filled time-streams.  The present routines will get us through basic troubles such as "I can't plot or fft this timestream because there are huge glitches", and similar.

mhasself:
> > > One overview question is whether we can gap-fill with some functions other than polynomials. I am asking since if we have the SAT time streams, which are modulated, it may be great if we can gap-fill with a sinusoidal function, matching the overall tend (frequency, phase, amplitude) of the unflagged data.
> > 
> > 
> > The SAT timestreams will have the HWPSS in them, which for the most part won't be sinusoidal? They will have shapes depending on the position across the focal plane right?
> 
> You are probably right, I just naively guessed that it will be a sinusoidal function... I think it will probably depend on the position of the focal plane.

I think in general the only safe thing to do with non-trivial structured signal (whether it's atmosphere or HWPSS) is to model it as best you can using unflagged samples (and trivially gap-filled flagged samples if need be).  Then subtract that first model estimate from the original data; gap fill those residuals; and add back in the model (if you want it to look like the original signal).  Then iterate -- in cases where where your model fit itself requires some gap-filling.

PR #82, opened 2020-07-15
docs: rebrand SOTODLib -> sotodlib; set some autodoc options

mhasself:


mhasself:
I should also mention that I added a bunch of stuff to the mocks list in conf.py, but pushed those directly to master (accidentally... but also readthedocs builds are hard to debug unless they're on master)

PR #85, opened 2020-07-28
Improve obsdb

mhasself:
Completes the API for ObsDb, including a "tag" feature.  Documentation.

This is part 1 of a 4-5 part series that will explain the Context system fully and properly...

### DIFF
--- a/docs/context.rst
+++ b/docs/context.rst
@@ -15,9 +15,11 @@ documented at all.
 Context [empty]
 ---------------
 
---------------------------------------
- DetDb: Detector Information Database
---------------------------------------
+.. py:module:: sotodlib.core.metadata
+
+------------------------------------
+DetDb: Detector Information Database
+------------------------------------
 
 The purpose of the ``DetDb`` class is to give analysts access to
 quasi-static detector metadata.  Some good example of quasi-static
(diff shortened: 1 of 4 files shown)


docs/context.rst
```diff
@@ -300,28 +302,200 @@ DetDb auto-documentation
...
+The column definitions must be specified in a format compatible with
+sqlite; see :py:meth:`ObsDb.add_obs_columns`.  When the user adds data
+using :py:meth:`ObsDb.update_obs`, the first argument is the
+``obs_id``.  This is the primary key in the ObsDb and is also used to
+identify observations in the ObsFileDb.  When we write the database
+using :py:meth:`ObsDb.to_file`, using a .gz extension selects gzip
+output by default.
+
+
+Using an ObsDb
```
tskisner, line 388: Your ObsDB and ObsFileDB are essentially "Object Relational Managers" (ORMs).  In Python, there are several popular tools that represent a class as a database table and each row is an instance of that class.  For example, with Django you define a "Model" (https://docs.djangoproject.com/en/3.0/topics/db/models/) which is a class representing a table and then add rows (instances) and make queries (https://docs.djangoproject.com/en/3.0/topics/db/queries/) while also applying filters on the values.  And you can do joins between models as well.  Most of these tools support multiple backends (sqlite, postgresql, etc).

Of course you can "roll your own" as you have done here, but just wondering if there was a particular reason you rejected an off-the-shelf tool for this?

mhasself: For ObsFileDb you are probably right... I have not looked at that one closely, lately, but the table structure is ~completely specified in code.  But for ObsDb I'm less sure.

The ObsDb schema is deliberately under-specified.  The code that reads an ObsDb will not know, in advance, what the columns of the obs table are (other than obs_id).  Perhaps the ORMs can handle that (not knowing column names / instance variables in advance) gracefully, but it does not seem to be their focus.  I also like that it so easy to work with the table of results one gets back; the ResultSet exposes the data as numpy arrays and dicts, so they are ready to go for whatever.

tskisner: Sounds good.  And these are relatively straightforward classes representing small databases.  If you needed to also target things like postgres due to the database size, that might be a stronger argument for using off-the-shelf packages.

PR #87, opened 2020-07-31
HWP demodulation

keskitalo:
This PR implements a TOAST demodulation operator. It processes the HWP-modulated signal, pointing, timestamps, noise PSDs and flags into demodulated and downsampled vectors. The process is irreversible, so it is not possible to perform Monte Carlo loops that involve demodulation.

A low pass filter used in demodulation is set automatically, but it is possible to control the filter width and set the window function.

For the time being, flags, pixel numbers and timestamps are simply sampled from the full resolution vectors. A more advanced approach may be developed later.

### DIFF
--- a/sotodlib/io/toast_frame_utils.py
+++ b/sotodlib/io/toast_frame_utils.py
@@ -768,7 +768,7 @@ def tod_to_frames(
     nnz = 1
     if cache_common_flags is None:
         if rankdet == 0:
-            cflags = tod.read_common_flags(local_start=cacheoff, n=ncache)
+            cflags = np.array(tod.read_common_flags(local_start=cacheoff, n=ncache))
             cflags &= mask_flag_common
     else:
         cflags, nnz, mtype = get_local_cache(0, cache_common_flags, cacheoff,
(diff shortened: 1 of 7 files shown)


keskitalo:
Presently filtering and demodulation are ordered as
1. Apply polynomial filter (to remove subharmonic modes and enable Fourier analysis)
2. Demodulate
3. Apply ground filter (to remove polarized scan-synchronous signal)
However, one might also consider applying the ground filter before demodulation to prevent scan synchronous signal from leaking into the demodulated data. Any thoughts, @satoru99 ?

sotodlib/io/toast_frame_utils.py
```diff
@@ -768,7 +768,7 @@ def split_field(data, g3t, framefield, mapfield=None, g3units=units,
     nnz = 1
     if cache_common_flags is None:
         if rankdet == 0:
-            cflags = tod.read_common_flags(local_start=cacheoff, n=ncache)
+            cflags = np.array(tod.read_common_flags(local_start=cacheoff, n=ncache))
```
keskitalo, line 771: `read_common_flags` should already by returning an array. Why do you think this is needed?

satoru99: Without this, I have sometimes encountered memory error at clean up process. It may be due to some memory allocation issue between toast and spt3g libraries. Copying into numpy array makes it simple. The flatten() above should be for the same reason.

keskitalo: Perhaps @tskisner can take a look at these modifications to the `so3g` interface?

tskisner: Considering that these tools are changing anyway in the toast3 port, I suggest we just merge this before that refactor.

tskisner (approved):
This is all fine, and we should merge this before moving toast code into a `toast` subpackage.

keskitalo:
Fair enough. I'll merge.

PR #98, opened 2021-01-20
Reduce size of simulated hardware test.

### LINKED ISSUE #62: tests.test_hardware takes too long
I think we should use smaller test cases, by default, so the tests are light enough to be run frequently.  This will encourage use of the tests.

In cases where performance "at scale" needs to be checked, individual tests can be run with additional options set; but that need not be the default.

tskisner:
Running the tests in this branch locally takes about 20s on my laptop:
```
python setup.py test
<snip>
Ran 39 tests in 20.255s

OK
```
If I disable plotting, I get:
```
SOTODLIB_TEST_DISABLE_PLOTS=1 python setup.py test
<snip>
Ran 39 tests in 10.738s

OK
```
Is this acceptable?  If so, then this closes #62

### DIFF
--- a/.github/workflows/run_tests.sh
+++ b/.github/workflows/run_tests.sh
@@ -8,6 +8,8 @@ cp -a /home/sotodlib .
 
 pushd sotodlib >/dev/null 2>&1
 
+export SOTODLIB_TEST_DISABLE_PLOTS=1
+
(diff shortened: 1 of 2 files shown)


mhasself (approved):
Thanks, I think that will be much more manageable.

PR #99, opened 2021-01-21
When drawing plot labels, include labels on the wafer slots.

tskisner:
When using the `--labels` option to `so_hardware_plot`, the wafer_slot is plotted at a location that is the average of all detectors in that wafer on the plot.  Here is an example:
sat1-f090-p5

### DIFF
--- a/sotodlib/vis_hardware.py
+++ b/sotodlib/vis_hardware.py
@@ -103,6 +103,27 @@ proceeding with the default matplotlib backend"""
         hmin = -half_height
         hmax = half_height
 
+    wafer_centers = dict()
+    if labels:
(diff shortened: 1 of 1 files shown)


keskitalo:
Awesome!

Zhilei:
Oh, wow! This is amazing!

tskisner:
Just to be clear, these are not "new" tools.  Not even the commandline options have changed.  I'm just plotting an extra string.  The tools themselves are documented in the sotodlib docs:

https://sotodlib.readthedocs.io/en/latest/cli.html#hardware-configuration

But those do need to be updated from the recent name changes within the hardware model.

smsimon (approved):

PR #101, opened 2021-01-28
Add module for simulating HWPSS

keskitalo:


keskitalo:
@satoru99 would you be open to testing this simulation module against your demodulation codes?

tskisner (approved):
Thanks for getting this work in so we can port it to the toast3 API.

PR #103, opened 2021-02-01
Sso sim

keskitalo:
This branch was developed with @aamirmali and @Zhilei. It adds Solar System Object (SSO) simulation capability to `toast_so_sim.py`.


tskisner:
This branch needs a major rebase against master.  Looking at the network graph, it has diverged quite a bit.

keskitalo:
@tskisner Indeed, I would like to merge this branch and start anew with less convoluted branch history. Since this branch is functional (if not feature-complete) it seems like a good time to merge.

pipelines/get_wafer_offset.py
```diff
@@ -0,0 +1,104 @@
...
+        match["wafer_slot"]  = args.wafer_slots.split(",")
+    elif args.tube_slots is not None:
+        tube_slots = args.tube_slots.split(",")
+
+    hw = hw.select(tube_slots=tube_slots, match=match)
+    ndet = len(hw.data["detectors"])
+
+    # print(f"tube_slots = {tube_slots}, match = {match} leaves {ndet} detectors")
+
+    # Average detector offset
```
tskisner, line 81: Note that I do a duplicate calculation in the visualization tools (https://github.com/simonsobs/sotodlib/blob/f560e3196c36ebdd890516e58bac94da24f7708d/sotodlib/vis_hardware.py#L108).  We should make a function that does this and can be used in both places.  That can wait until after this is merged, but let's not forget about it.

sotodlib/observe/sim_sso.py
```diff
@@ -0,0 +1,271 @@
...
+    )
+    parser.add_argument(
+        "--beam-file",
+        required=False,
+        help="Pickle file containing instrumental beam",
+    )
+    return
+
+
+def apply_sim_sso(args, comm, data, mc, totalname, verbose=True):
```
tskisner, line 50: It looks like this file is duplicate code that was split up into the other correct source files.  If so, can we delete it from this branch before merging?  Note that the `observe` directory does not even exist in the sotodlib package.

tskisner (changes requested):
Looks fine after deleting that stale file (assuming everything in there has been moved to other sources)

pipelines/get_wafer_offset.py
keskitalo: Agreed. We don't want duplicated code. The context for this pipeline is that we needed to automate the offset estimation for shell scripting.

sotodlib/observe/sim_sso.py
keskitalo: Indeed! Thanks for catching this.

tskisner (approved):
Thanks for the fixes.  Fine to merge after resolving the conflicts.

PR #106, opened 2021-02-11
Export corotator angle

keskitalo:


smsimon:
I think from our discussion with Mike, it was rotated 90 deg CCW from the instrument design orientation at all elevations except when the co-rotator is dis-engaged, which would only be in the 1 deg elevation nod scan scenario at this point. As in this: https://docs.google.com/presentation/d/11_-LYjxMPzrz0FkwqTFZJo3nFpVuLr3Ajg8HcJial1o/edit?usp=sharing

keskitalo:
@smsimon You are right! I got so hung up on setting up the infrastructure that I forgot about that. Now the receiver is automatically rotated to compensate for the observing elevation but not for the elevation nods. The corotation can also be completely disabled from the command line.

smsimon:
`        if args.corotate_lat:
            obs["corotator_angle_deg"] = -ces.el
        else:
            obs["corotator_angle_deg"] = 0`

The co-rotoation case looks good. I think the else case depends on what you're doing. I think it'd be -60 deg if there is no co-rotation at all since the LAT boresight is at the 90 deg when the elevation is 60 deg. In the case of the elevation oscillations, I think this would be -the base elevation of the scan since you are oscillating the elevation around that.

mhasself (approved):
Looks good... thanks!  I second Sara's preferences with regards to the "else" case, though.

keskitalo:
Thanks for your diligent checking, @smsimon ! @mhasself once this PR is merged, I imagine the axismanager objects need to be made aware of the vector of corotator angles written into every LAT so3g file.

mhasself:
Yes good point... that will be a quick fix in case you want to include it immediately.  See how it is done for hwp_angle:
https://github.com/simonsobs/sotodlib/blob/5c643505a1224adf5088a3c3d107daad72286d6d/sotodlib/io/load.py#L283-L290
https://github.com/simonsobs/sotodlib/blob/5c643505a1224adf5088a3c3d107daad72286d6d/sotodlib/io/load.py#L416-L423

Any consequences on pointing aside, those are the only places that need modification to load the new field.

keskitalo:
@mhasself done!

smsimon (approved):

PR #107, opened 2021-02-25
Improvements feb21

mhasself:
This is mostly documentation, including:
- AxisManager sphinx docs (accidentally left out of original work)
- coords module sphinx docs and docstrings

Also has bugfixes:
- Allow ResultSet to be indexed with an np.integer, whatever that is.
- iir_filter argument order is better

New features:
- AxisManager.wrap_new(...) makes building new AxisManagers much easier.
- AxisManager.get(key, default=None) makes AxisManager more dict-like
- Logging system is initialized and docs describe how it should be used.
- coords module internal functioning is adjusted slightly, in preparation for more demanding

### DIFF
--- a/docs/axisman.rst
+++ b/docs/axisman.rst
@@ -1,0 +1,181 @@
+===========
+AxisManager
+===========
+
+The AxisManager is a container class for numpy arrays and similar
+structures, that tracks relationships between axes.  The object can be
+made aware, for example, that the 2nd axis of an array called "data"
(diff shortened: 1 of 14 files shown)


mhasself:
Requesting partial reviews from @tskisner - is this logging approach sensible and @kmharrington - is the new AxisManager stuff clear, especially wrap_new?  Any other comments welcome.

kmharrington (approved):
Glad to see the AxisManager docs in there. I'm happy with the filtering changes.

PR #108, opened 2021-03-03
Adding a tiny function to sotodlib/toast/sim_sso.py to calculate the planet temperature 

nad-dch:
The function takes  sso_name and frequency of interest and outputs a float. Also adding sso_freq as an attribute to class OpSimSSO.


Zhilei (approved):
Looks good!

sotodlib/toast/sim_sso.py
```diff
@@ -118,6 +119,29 @@ def exec(self, data):
...
+        """
+        Get the thermodynamic planet temperature given
+        the frequency
+        """
+        freqs = [30,44,70,100,143,217,353]
+
+        t_jupyter = [144.5, 159.1, 171.9, 172.6, 174.1, 175.8, 167.4]
+        t_saturn = [138.9, 147.3, 150.6, 145.7, 147.1, 145.1, 141.6]
+        t_mars = [183, 187, 183, 194.3, 198.4, 201.9, 209.9]
+        t_uranus = [190, 230, 138, 120.5, 108.4, 120.5, 98.5, 86.2]
```
mhasself, line 133: Just a reminder that the 44 GHz value for Uranus should be checked.  I suspect it's a typo.

mhasself (changes requested):

mhasself (approved):

keskitalo:
Would you entertain convolving the SSO SED if I provided a prescription to get the detector bandpass? That would make `--sso-freq` redundant and would support bandpass mismatch studies as well.

Zhilei:
@keskitalo That's a great point! I have a feeling that the bandpass will be out of the scope of this pull request. 

Can we merge this PR as it is and maybe start another issue to implement the bandpass development?

keskitalo:
@Zhilei I'm fine either way.

Zhilei:
Great! New issue opened with No. [109](https://github.com/simonsobs/sotodlib/issues/109)

nad-dch:
@keskitalo Sure, I would be happy to do this. Let me know how I can get bandpass information and I will include it in this or the next pull request (whatever @Zhilei and you prefer).

keskitalo:
@nad-dch Please see https://github.com/hpc4cmb/toast/blob/master/src/toast/pipeline_tools/atm.py#L417-L437 and please don't hesitate to ask if it is unclear. I'll link the same snippet to the new issue @Zhilei opened.

Zhilei:
It seems like Nadia incorporated the bandpass treatment in the latest commits. I was wondering whether we can just ignore the newly-opened issue and get things sorted out under this pull request. @keskitalo ?

keskitalo:
Really nice @nad-dch !

@zhilei the proper way to handle the new issue is to close it once the PR is merged with a message saying so.

nad-dch:
@keskitalo  Thanks. I made a small addition to setup.py and MANIFEST.in in order to include the data file (I remember we had some discussion about how to do this in the bcp channel). Not sure if this is the most optimal way to do it but it is necessary for the code to work now. Let me know when you take a look if there is something you would like me to change and I will merge the pull request afterwards if everything is ok.

Zhilei:
Thanks for the suggestion, Reijo! Will do!
We decided to add the planet information data file into the sotodlib, I was wondering whether @keskitalo @mhasself thinks that's a good thing to do.
But I did not see the .h5 file, am I missing anything @nad-dch ?

nad-dch:
@Zhilei yes, you can follow the commit ec5ada9. The file is in sotodlib/toast/data and I have slightly modulated setup.py to include this file.

keskitalo:
Looks great. Seems ready to merge.

nad-dch:
Great, thanks. I will give it one more day in case anybody else wants to check it and, if not, I will merge it tomorrow.

mhasself:
Sorry to be slow to revisit...

Major comments:
1) It looks as though planet_data.h5 is still present in the root of the repository; please remove it!

2) These data are the thermodynamic brightness temperatures?  We will need to convert to CMB uK units at some point; where is that done?  (This is effect is large -- factors of 1 - 3 for our bands.)

Minor suggestions for the .h5 file format --
- The SSO names in the h5 file are capitalized, like 'Jupiter'.  How do folks feel about changing these to lower case?  If This is TOAST standard already, then fine.
- Consider renaming freqs to freqs_ghz, or freq_ghz to be more concordant with, e.g., the focalplane field names.  It is helpful to inform future-user what the units of things are... for frequencies, in particular, GHz is not obvious.

nad-dch:
Hey Matthew. Thank you very much for the comments. [Major Point 1]: Done. [Major Point 2]: These are thermodynamic temperatures, so in the .h5 file they are already stored in Kelvin. As far as I saw, before using this file, Zhilei was also hardcoding the temperature in K units but if uK is preferred then I can also store them like this directly. [Suggestion 1] I used capitalized names in the .h5 file to follow the SSO names convention but I think pyephem() calls do not work with small letter planet names, right ? [Suggestion 2] Sure ! That makes sense, will do.

Zhilei:
Thanks for following up on this! As long as the pyephem is not case sensitive, I think we can follow the planet naming convention @mhasself prefers.
Also, did you mean the difference between Rayleigh-Jeans temperature vs the CMB temperature? To my understanding, the CMB temperature is the thermodynamic temperature, but my understanding may be completely wrong.

keskitalo:
TOD units internal to TOAST are always K_CMB so uK_CMB would just complicate things. I believe `pyephem` is case sensitive, but you can manipulate the SSO name in the `pyephem` call to be capitalized, regardless of its case.

mhasself:
Yes, my concern is that we're not yet in K_CMB; we're still in the planet's "brightness temperature" K.  (Correct me if I'm wrong about this.)  So we're sort of two steps away from K_CMB... we have to use a linear brightness temperature (such as RJ) before diluting, then convert the diluted output to CMB temperature units.  I'll email you all some things and we can discuss there / in Slack.

jegudmunds:
Nadia and I can go over the code to make sure things are in the proper units and derived using accurate conversion factors between RJ temperature (input to Planck bb function in RJ limit), thermodynamic temperature (input to Planck bb function). The term brightness temperature, which (unless I'm mistaken) is used by WMAP and ACT to mean RJ temperature can be quite confusing as different research groups use it to mean different things. See for example first few paragraphs in Section 2 of https://doi.org/10.1051/0004-6361/201630311

Zhilei:
Hi!

Thank you so much for the email @mhasself and all the explanation! 

First of all, I would suggest we stay with the planet naming scheme so that there are no additional interfaces to pyephem.

Then for the brightness temperature, it is really nice of Mattehw to explain everything in the email. I think it makes sense for radio astronomy to use T_RJ since it is linear and the result is close enough to the rigorous BB calculation. Given the limited computation power in early times, T_RJ definitely wins in radio astronomy. And I think when we go up to THz, the T_RJ vs T_CMB difference becomes significant enough so that T_RJ is almost irrelevant, so they naturally assume brightness temperature as T_bb.
The CMB frequency lies between those two cases so I think that's where the confusion comes from. Right now, I have been assuming everything is in T_CMB, but what I did incorrectly was that I have been 'diluting' the temperature directly (as Matthew pointed out) and I will correct that.
Therefore, I suggest we use T_bb altogether for SO. In the future, we won't have this confusion once for all. I can go change the code following the calculation philosophy as T_bb -> spectral radiance -> solid angle diluted spectral radiance -> corrected T_bb.

Please let me know if you think that makes sense! Thanks!

mhasself:
That sounds fine, and the present PR doesn't have to wait for it; I just wanted to make sure it was on your radar.

Zhilei:
Sounds great! Thank you so much for the help! I think @nad-dch should probably merge this and I will start a new PR for the brightness temperature handling

nad-dch:
Great, thanks ! I updated the name convention (freqs --> freqs_ghz) and I will merge the pull request. I can add in the future to this branch the additional changes implemented from Zhilei. Thank you all for the comments, this discussion was really educating.

PR #116, opened 2021-03-23
Filter upgrades

mhasself:
Key changes:
- New tests, covering some (but not much) of tod_ops module.
- The fourier filter framework also exposes an "apply" method, that can be faster in some applications.
- fourier_filter will pad to any nice FFT number, not necessarily 2^n.
- New names for timeconstant (de-)convolution filters; deprecates "tau_filter".
- Made some important filter arguments non-optional; better arg names.
- More docs.


kmharrington (approved):
Thanks for setting this all up! I'm going to need to test it with all our new beam map data :-D

guanyilun (approved):
The changes in filters.py look good to me by inspection. Thanks for the improvements!

PR #117, opened 2021-04-01
Adding G3tSmurf Data Loading for working with lab data

kmharrington:
I'd like to bring this branch into master to keep it from getting too far off. (Especially because I've started wanting to make edits to other parts of sotodlib). Summary of changes:

### io.load_smurf
The `G3Smurf` class and it's associated databases are used to index and load SMuRF data out of SO style `.g3` files. Eventually the plan is to merge this databasing into the sotodlib context system. We aren't quite there, but it is working at a level for people to submit issues and feature requests. Nothing in other parts of `sotodlib` depend on things in this file.

### core.axisman + test.test_core
A small fix to make parts of the AxisManager work for classes inheriting from AxisManager (like FlagManager). Updates to AxisManage had made it so that wrapping a FlagManager into an AxisManager converted the FlagManager into an AxisManager. I've also added a test case to check for this type of error. 

### core.flagman
Larger changes here to deal with detector channels sometimes being called `'dets'` and sometimes being called `'channels'`. Solved by allowing users to specify any name for the two axes you want FlagManager to manage.  Should not affect any existing code. ( SMuRF data is using `'channels'` instead of `'dets'` because the mapping between resonator channels and pixel locations is not known a priori ).

### flags
Adding `get_trending_flags` function to look for large slopes in data. Since this seems to be the primary way SMuRF channels "unlock".

### tod_ops.detrend
Adding additional keywords to the `detrend_tod` function. Shouldn't affect any existing code.


sotodlib/flags.py
```diff
@@ -83,3 +83,46 @@ def get_glitch_flags(tod, params={}, signal='signal', merge=True,
...
     return flag
+
+def get_trending_flags(aman, max_trend=np.pi, n_pieces=1, signal_name='signal',
+                 flag_name='trends', overwrite=True):
+    """ Flag Detectors with trends larger than max_trend.
+    
+    Args:
+        aman (AxisManager): the tod 
+        max_trend: maxmium amount to always the detectors to change by
+                   default is pi for use in phase units.
```
mhasself, line 94: This could use a rewrite.  The second line here is not a new sentence -- sphinx will interpret it as a continuation of the first line.

sotodlib/flags.py
```diff
@@ -83,3 +83,46 @@ def get_glitch_flags(tod, params={}, signal='signal', merge=True,
             tod.flags.wrap(name, flag)
         
     return flag
+
+def get_trending_flags(aman, max_trend=np.pi, n_pieces=1, signal_name='signal',
+                 flag_name='trends', overwrite=True):
```
mhasself, line 88: This further entrenches the old aman API conventions (i.e. signal_name='signal' instead of signal=None).  I know you plan to update everything ... someday ... but consider just starting to develop with the new API.

mhasself (changes requested):
I haven't commented on the load_smurf bit, because I know it's a WIP and is isolated from everything else.  But some of the other code is in the general "fct" space so please take a look at the comments.

In FlagsManager you've generalized the "dets" and "samps" axis, replacing them with self._det_name and self._samp_name.  I get why you might want to do this, but I want to suggest naming those variables differently.  Since the defaults / standards are "dets" and "samps", depluralizing to "det" and "samp" makes the name space more complicated.  Please use "dets" and "samps" everywhere.

Good finds on the AxisManager.  It is tricky to make such a complex, self-referential object subclassable so we'll have to be vigilant to not break this going forward.

mhasself:
One more thing -- please put a minimal place-holder into the docs/, even if it's just one sentence of text and a link to the source code (e.g. autoclassing the G3tData class).

sotodlib/core/flagman.py
```diff
@@ -116,6 +121,12 @@ def copy(self, axes_only=False):
             out._assignments[k] = v.copy()
         return out
     
+    def get_zeros(self):
+        """
+        Return a correctly sized RangesMatrix for building cuts for the FlagManager
+        """
+        return RangesMatrix([Ranges(self[self._samp_name].count) for det in self[self._det_name].vals])
+
```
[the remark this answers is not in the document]
kmharrington: That's a good idea, I'll add that.

kmharrington:
@mhasself  Thanks for the review. I think I've fixed all the pieces you were worried about. Added some placeholder docs for FlagManager (in addition to G3tSMuRF) while I was at it.

mhasself (approved):
Thanks ... docs look great!

To get them to compile on readthedocs you'll probably have to work around sqlalchemy dependency -- add this:

```
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -96,6 +96,7 @@ autodoc_default_options = {
 autodoc_mock_imports = []
 for missing in ('numpy', 'matplotlib', 'healpy', 'astropy',
                 'quaternionarray', 'yaml', 'toml', 'sqlite3',
+                'sqlalchemy',
                 'pyfftw', 'toast', 'spt3g', 'so3g', 'pixell'):
     try:
         foo = import_module(missing)
```

PR #122, opened 2021-04-30
Load .g3 file

kmharrington:
This PR does a fair bit of updates / rearranging to the data loading functions in load_smurf. The goal is to expand functionality and move toward full `sotodlib` integration without breaking any existing analysis code. 

Main updates: 

- `load_file` function that takes any SO style `.g3` file (or list of files) and returns a correctly formatted AxisManager
- `get_channel_mask` function that connects a user's request for different channels to what exists in a file or observation. When the `dets` kwarg is passed to either `load_file` or `G3tSmurf.load_data` only the requested channels will be loaded into memory. 
- `G3tSmurf.load_data` now uses `load_file` under the hood. Right now this is slightly less optimized than before, because `sotodlib.io.load.unpack_frames` doesn't have Seek capability, likely a good thing to put on the todo list.
- New timing function that calculates timestamps off the whole loaded dataset instead of frame-by-frame
- Expand docs to include new docstrings
- Add in logging for this module

### DIFF
--- a/sotodlib/io/load_smurf.py
+++ b/sotodlib/io/load_smurf.py
@@ -16,6 +16,7 @@ from collections import namedtuple
 from enum import Enum
 
 from .. import core
+from . import load as io_load
 
 
 Base = declarative_base()
@@ -855,7 +856,7 @@ class G3tSmurf:
             show_pb : bool, optional: 
                 If True, will show progress bar.
             load_biases : bool, optional 
-                If True, will return biases.
+                If T'darue, will return biases.
 
         Returns
         --------
@@ -1334,7 +1335,7 @@ def get_channel_mask(ch_list, status, archive=None, ignore_missing=True):
         session.close()
     return msk
 
-def get_channel_info(status, mask=None, detset=None, ch_name_type='sch'):
+def get_channel_info(status, mask=None, detset=None):
     """Create the Channel Info Section of a G3tSmurf AxisManager
     
     This function returns an AxisManager with the following properties    
@@ -1345,6 +1346,7 @@ def get_channel_info(status, mask=None, detset=None, ch_name_type='sch'):
             * band : Smurf Band
             * channel : Smurf Channel
             * frequency : resonator frequency
+            * rchannel : readout channel
     
     Args
     -----
@@ -1353,10 +1355,6 @@ def get_channel_info(status, mask=None, detset=None, ch_name_type='sch'):
         mask of which channels to use
     detset : DetSet instance (optionl)
         detector set for the data
-    ch_name_type : string
-        if 'sch' the channel names will be in the for sch_TUNEFILE_BAND_CHANNEL and
-        TUNEFILE will be None if unavaliable. Use 'rch' to simply list names by 
-        absolute readout channel.
         
     Returns
     --------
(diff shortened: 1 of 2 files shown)


sotodlib/io/load_smurf.py
```diff
@@ -896,169 +899,43 @@ def load_data(self, start, end, stream_id=None, detset=None, show_pb=True, load_
...
-        if len(timestamps) > 0:
-            status = self.load_status(timestamps[0])
-        else:
+        
+        scan_start = session.query(Frames.start).filter(Frames.start > start,
+                                                        Frames.type_name=='Scan')
+        scan_start = scan_start.order_by(Frames.start).first()
+        
+        try:
+            status = self.load_status(scan_start[0])
```
jlashner, line 923: One thing I noticed when trying to debug why some queries failed to load the status, is that it's actually possible for the first timestamp of a scan frame to occur *before* the first Status frame.  This is actually how I intended the Smurf SessionManager to work, but in some cases it results in an Observation, Wiring, and Scan frame all to be written at time `t`, but the first data sample actually being at `t-1`. 

tldr: we should be loading the Status frame at something like `Frame.time+1` to be safe.

jlashner (commented):
I think this looks good, but before we merge I think we should run a timing comparison between this and the previous version. 

I see the benefit of using io.load to load the detector data, but the old SmurfArchive code was like extremely optimized, with only one np.array allocation (with the exception of that bug that copied the entire array to change the units to phi0) . The io.load function is a bit hard to follow, but it looks like it's doing something like creating an array for each frame, which I think can be pretty time consuming (based on personal experience).

Part of the reason I'm concerned is that it takes a very long time to load high sample-rate hk data, so I'm not convinced we should be using the same system for smurf data unless the load function has been seriously optimized. Part of the benefit of using the Frame database is that we can easily calculate the number of chans / samps in a range so we could allocate data for the whole range without actually loading the data, which has sped things up immensely

jlashner:
Ok did a quick timing  test on an hours worth of UFM data and seems like they're pretty similar! Old version took 40 sec and new version took 45 sec, so I probably just don't understand how the FieldGroups method actually works, or what the bottleneck is of hk loading.

One thing that I do miss in this branch is the existence of a progress bar! Any chance we can somehow add this back in? Having a timing estimate for how long it'll take to load is really nice

jeffiuliano:
+1 Vote for the progress bar.

kmharrington:
Thanks for the reviews and the timing checks! I'm glad to hear it doesn't make too massive a difference. This is one of those things where I'm going a little backward to try and make the future a bit easier. I'd like to add Seek capability into the frame unpacking (https://github.com/simonsobs/sotodlib/issues/123) and then add it back in when we load data via the database. The `sotodlib` context loading should also be able to do this frame Seeking thing and I was hoping to get both it and G3tSmurf doing them the same way. Since it doesn't seem to be adding too much overhead I'd like to go this direction. 

Trouble with the progress bar is you need to know the total number of frames to load, and I've make the loading blind to that right now. Would you all hate me if that gets added to the "Add Seek to unpack_frames" issues? It's all along the same vane.

jlashner:
What if we add a progress bar based on the number of files for the time being instead of the number of frames? Should still give us a rough estimate

kmharrington:
yeah, sounds like a good middle ground. I'll add that in

PR #124, opened 2021-05-11
Flag sso

keskitalo:
Add a facility to flag solar system objects. For example `--flag-sso Jupiter,30` will flag 30 arc minutes around Jupiter everywhere in the data. Multiple instances of `--flag-sso` are allowed and, notably, `Sun` and `Moon` are supported.

### DIFF
--- a/pipelines/toast_so_sim.py
+++ b/pipelines/toast_so_sim.py
@@ -96,6 +96,7 @@ def parse_arguments(comm):
     toast_tools.add_debug_args(parser)
     so_tools.add_import_args(parser)
     so_tools.add_sim_sso_args(parser)
+    so_tools.add_flag_sso_args(parser)
     so_tools.add_sim_hwpss_args(parser)
 
     parser.add_argument(
@@ -268,6 +269,8 @@ def main():
         signalname = so_tools.simulate_sky_signal(args, comm, data, focalplanes)
     else:
         signalname = toast_tools.scan_sky_signal(args, comm, data)
+
+    so_tools.apply_flag_sso(args, comm, data)
 
     memreport("after PySM", comm.comm_world)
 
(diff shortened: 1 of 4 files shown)


Zhilei:
Looks great! I am curious to know how the functions are called from a higher scope.

smsimon:
Are multiple units supported for the radius of the flagging (e.g. arcmin vs. degrees)? I could see needing to convert degrees into arcmin for larger radii as potentially annoying (but not hard to do).

keskitalo:
@smsimon 

> Are multiple units supported for the radius of the flagging (e.g. arcmin vs. degrees)? I could see needing to convert degrees into arcmin for larger radii as potentially annoying (but not hard to do).

I can only support one unit on the command line. TOAST3 uses astropy units so we can revise this when we port it.

smsimon (approved):

smsimon:
Sounds good

aiolasimone:
Sorry to sneak into the conversation, but I have a question. 
Is it possible to dump the cut flags to a file and interface those with the Flag Manager (that @kmharrington developed)? I think those are useful if for example we want to map the plant TODs with the SO planet mapper or even the ML one.

tskisner:
I think the eventual flow would be:

- Simulation creates toast observations in memory, including data ranges and flags
- Simulated data (including flags) is written to G3 files
- OR simuated data can be passed directly to other tools (map maker, wrapped "low-level processing")
- G3 files on disk (simulated or real) can be loaded into AxisManager and friends
- AxisManager data can be passed to low-level tools and anything working with that representation
- OR AxisManager data can be converted to toast observations and passed to map maker, wrapped low-level processing

aiolasimone:
I agree, I think that would cover the spectrum of possible uses.

PR #129, opened 2021-05-21
fixing up inputs to load_data so that if a string is passed it gives …

jeffiuliano:
…an obvious error instead of a confusing one about not being able to find information in the databases.  Added a docstring to the relevant function while at it.

### DIFF
--- a/sotodlib/io/load_smurf.py
+++ b/sotodlib/io/load_smurf.py
@@ -371,11 +371,13 @@ class G3tSmurf:
         ----
             datetime: datetime of x if x is a timestamp
         """
-        if isinstance(x,float) or isinstance(x,int):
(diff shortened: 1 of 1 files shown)


kmharrington (changes requested):
I've found that type checking like this isn't robust enough for pulling things out of numpy arrays, which I expect will be done here at some point. 

Check out the type checking in the[ `get_channel_mask` function ](https://github.com/simonsobs/sotodlib/blob/master/sotodlib/io/load_smurf.py#L1226). Those versions are more robust, they'll work with both standard python types and numpy types.

jeffiuliano:
Yeah, that's a useful example.  I think this should be more robust/general.

kmharrington (approved):
Works for me!

PR #130, opened 2021-05-24
Cadence map

keskitalo:
Added a new operator to the `toast_so_sim.py` pipeline. It produces a boolean table, written into a single HDF5 file. The table indicates which sky pixels are hit on which modified Julian dates across the data span. Some further development may be needed in the future to break the file into smaller sections without exhausting memory.


keskitalo:
Self-merging this PR after a brief review window. It only affects TOAST-based workflows and is fully backwards compatible.

Issue #120, opened 2021-04-16
Reconcile G3tSmurf -- Expand Context to include optional Channels table


kmharrington:
The overall thought here is that I would like for `G3tSmurf` to work more completely with Context / sotodlib fairly soon, because I am worried about "building too far away" from the main direction. Particularly as there's multiple versions of work heading toward building metadata from SMuRF work. 

I've gone through and verified that, with one minor tweak, the databases built by G3tSmurf can go directly into `ObsDb` and `ObsFileDb`. The piece that is harder is Channels vs Dets. 

I could technically rename some things and turn Channels into Dets, but that seems like a bad idea long term since channels are inherently not detectors. It's also obvious that SO wanting to do Channel-level analysis is going to persist when we're on-sky.

Looking through the code, I think its fairly possible to add some `or`s to the Context code to deal with having either Channels or Dets. Dealing with the combination (when Dets map to a set of channels) will be more subtle, but I think if we always build assuming one detector will map to many channels then we'll be able to deal with that. 

Looking to see if anyone is totally opposed to this idea or has thoughts about where/how this should be implemented.

mhasself:
Trying to summarize a discussion we had offline (though I inevitably added some stuff, sorry):

- The single database that G3tSmurf uses can be passed for ObsDb and ObsFileDb -- great!
- There is nomenclature conflict, in that a distinction needs to be made between readout "channels" and the specific physical "detectors" or pseudo-detectors to which they correspond (a thing that is not easy to determine).  Although G3tSmurf is careful about this terminology, it is more annoying to maintain this terminology distinction in Context / AxisManager.
- The Context DetDb can't be mapped into some table(s) in the G3tSmurf database, because DetDb will try to look at all the tables in the database.  Therefore it makes sense for the archive update code to simply write the DetDb information into a separate database file.  The information copied out can be some useful but incomplete subset of the full web of tuning info that G3tSmurf tracks.  This has a nice side-effect of not cluttering up G3tSmurf with unwelcome "dets" terminology!
- The channel -> detector mapping is a complicated map that probably should not live in G3tSmurf database.  If we load it as metadata, you can easily look up, in an AxisManager indexed by Channel names, what the corresponding "detector" name (and other info) is.

So I believe our conclusions were:
- Write out a separate "DetDb" for context use
- Frown but accept that the axis will be called ".dets" even if it indexes "Channels".
[...]

PR #131, opened 2021-05-25
Reconcile G3tSmurf to enable use with Context

kmharrington:
Pretty large update to the G3tSmurf system that solves https://github.com/simonsobs/sotodlib/issues/120. This also includes many updates to use new pysmurf and smurf-streamer features to make the matching of actions, tunes, and channel assignments more deterministic. If these updates are not found in files it should fall back on previous methods. 

## Summary of Context Updates

- Context databases (ObsFileDb, ObsDb, and DetDb) now have read-only options. The previous default was to load the whole databases into memory. The new default is to create a read-only persistent connection to the sqlite files.
- DetDb updated to use the common loading/connection options
- `load_smurf` now registers a loader for Context

## Summary of G3tSmurf Updates
**General Updates**
- `logging` now implemented in many functions (usage not quite complete yet)
- `SmurfStatus.from_file` now stops when it finds the `dump` frame.
- Use tune file (now in SmurfStatus) to check we're loading the correct list of channels

**Table Updates** 
- Many table name changes to match Context expectation. 
- Dump DetDb function that creates a "Detector" database from the Channels database
- Removal of usage the the Bands table, `stream_id` and `band` are now just known by several more tables
- DetSets --> TuneSets. TuneSets table collects unique combinations of <= 8 ChanAssignements 
- New Tunes table. Tunes will match directly to all tune files. (All TuneSets are in Tunes, not all Tunes are in TuneSets)
- Try to find channel assignments inside tune files (new pysmurf update)
- Use tune file and pysmurf action (now in SmurfStatus) to add checks in matching observations to .g3 files and TuneSets

**Data Loading Updates** 
- `load_file` can now be called with an obsfiledb instance, a G3tSmurf instance, or neither. The returned AxisManager should be identical if using either obsfiledb or G3tSmurf. If called with neither, the channel names will be less informative but band/channel information should still be there and useful.
- `G3tSmurf.load_data` still works as well for loading between two times.

## Update your Databases for usage
Because I needed to do so much table reorganization, I created an update script for existing databases. This is in pwg-scripts here: https://github.com/simonsobs/pwg-scripts/tree/master/pwg-duf/g3tsmurf_rec 

Usage:
 1. Use script to make new database `> sh main.sh /path/to/old/database /path/to/new/database`  This will print out some updates and does take some amount of time, ~15 minutes for LATRt's database. I'm bad at bash, you have to run it from the `g3tsmurf_rec` folder. 
 2. Update sotodlib install
 3. Make G3tSmurf instance using new database
 4. Only remove / rename database files *after* you've verified the new one works.


kmharrington:
Thanks, I updated the instructions. I don't think `index_metadata()` ever had a progress bar, it's `index_files()` that has. The first re-run is checking and opening all the tune files, so I expect that one to take some time. Afterwards it should be as bad? 

If it continues to take a long time you can give the update script a minimum ctime to prevent it from running over old data: https://simons1.princeton.edu/~kmharrin/sotodlib/g3tsmurf.html#sotodlib.io.load_smurf.G3tSmurf.index_metadata

docs/g3tsmurf.rst
```diff
@@ -7,16 +7,70 @@ G3tSMuRF
...
+    detdb: '{g3tsmurf_dir}/detdb.db'
+
+    imports:
+    - sotodlib.io.load_smurf
+
+    obs_loader_type: 'g3tsmurf'
+
+    #metadata:
+  
+
```
mhasself, line 71: It looks (in the code) like obs_id corresponds to G3 filenames... please mention that briefly in this documentation (or direct the user to the docstrings that elucidate the mapping).

What other info is made available through ObsDb?  Can I select observations based on the slot_id (for example)?

mhasself (approved):
Great to see this pushed all the way through to a context.yaml!  Couple doc requests in comments.

It looks like the Context-compatible loader produces AxisMan with "channels" axis instead of "dets" axis.  My impression after our discussion a few weeks ago was that I had almost convinced you to use "dets" there.  Anyway, at least mention this difference in the docstring or something.

docs/g3tsmurf.rst
kmharrington: Went ahead and added more docstrings. 

slot_id isn't technically something we'll be able to select on, it's going to line up exactly with tunesets/detsets.

kmharrington:
Yeah, I was more convinced to change it all to dets and then got less convinced when I realized how soon I'm going to be writing the dets/channels table mapping (it will be quite soon).

PR #134, opened 2021-06-10
Allow for different noises between SAT/LAT bands

smsimon:
This requires making different bands for the LAT and SAT.

Wafer type and tube types now: "LAT_UHF", "SAT_UHF", "LAT_MF", "SAT_MF", "LAT_LF", "SAT_LF"

Frequency bands now: "LAT_f030", "LAT_f040", "LAT_f090", "LAT_f150", "LAT_f230", "LAT_f290", "SAT_f030", "SAT_f040", "SAT_f090", "SAT_f150", "SAT_f230", "SAT_f290"


smsimon:
@keskitalo Can you check that I updated the bands in https://github.com/simonsobs/sotodlib/blob/noise_update/sotodlib/toast/sim_hwpss.py correctly? I assumed you only needed the SAT bands since the HWP is only on SATs

@Zhilei and @keskitalo I wasn't sure how to update the bands in https://github.com/simonsobs/sotodlib/blob/noise_update/sotodlib/toast/sim_sso.py on line 280. I didn't know if having LAT/SAT bands would break anything (e.g. if your dictionary gets looped through for a SAT sim and there are LAT-specific bands in it. Bands now have the form "LAT_f090".

smsimon:
@mhasself I wasn't sure how dependent your metadata code files were to the instrument model names so I left your band definitions unchanged in https://github.com/simonsobs/sotodlib/blob/noise_update/sotodlib/core/metadata/detdb.py and https://github.com/simonsobs/sotodlib/blob/noise_update/tests/test_metadata.py

smsimon:
Note: Failure of check is most likely because the checks use tests/test_hardware.py from the main branch, which uses the old band name versus the new one. "f090" will need to be changed to "SAT_f090" in the main branch prior to merging this pull request.

Zhilei:
Thanks for the heads-up! I think it will likely break things in the sim_sso.py file, but I will listen to Reijo since I think Reijo wrote the part around Line 280.

keskitalo:
> @keskitalo Can you check that I updated the bands in https://github.com/simonsobs/sotodlib/blob/noise_update/sotodlib/toast/sim_hwpss.py correctly? I assumed you only needed the SAT bands since the HWP is only on SATs

Yes, that is correct.

> @Zhilei and @keskitalo I wasn't sure how to update the bands in https://github.com/simonsobs/sotodlib/blob/noise_update/sotodlib/toast/sim_sso.py on line 280. I didn't know if having LAT/SAT bands would break anything (e.g. if your dictionary gets looped through for a SAT sim and there are LAT-specific bands in it. Bands now have the form "LAT_f090".

The current implementation in `sim_sso.py` is a little fragile. We should be getting the center frequencies from the detector bandpasses, not from a dictionary. That can be resolved in a separate PR. For now, I just remove the "SAT_" or "LAT_" prefix when looking up the center frequency.

@Zhilei I created a new branch on `pwg-scripts` called `noise_update`. I verified that the branches work together. We just need to merge them at the same time.

keskitalo:
I fixed two unit tests that were broken due to the changes in this branch. The remaining failure seems to be genuine and should be fixed in a separate PR.

```
ERROR: test_10_kernels (tests.test_coords.FootprintTest)
Test creation of sensible WCS kernels for basic cases.
----------------------------------------------------------------------
  File "/Users/reijo/Software/git/sotodlib/tests/test_coords.py", line 63, in test_10_kernels
    fp = coords.get_footprint(tod, wcs_kernel=wcsk, focal_plane=fp, sight=sight)
  File "/Users/reijo/Software/git/sotodlib/sotodlib/coords/helpers.py", line 227, in get_footprint
    proj.get_planar(asm, output=output)
TypeError: get_planar() got an unexpected keyword argument 'output'
```

Zhilei:
> > @keskitalo Can you check that I updated the bands in https://github.com/simonsobs/sotodlib/blob/noise_update/sotodlib/toast/sim_hwpss.py correctly? I assumed you only needed the SAT bands since the HWP is only on SATs
> 
> Yes, that is correct.
> 
> > @Zhilei and @keskitalo I wasn't sure how to update the bands in https://github.com/simonsobs/sotodlib/blob/noise_update/sotodlib/toast/sim_sso.py on line 280. I didn't know if having LAT/SAT bands would break anything (e.g. if your dictionary gets looped through for a SAT sim and there are LAT-specific bands in it. Bands now have the form "LAT_f090".
> 
> The current implementation in `sim_sso.py` is a little fragile. We should be getting the center frequencies from the detector bandpasses, not from a dictionary. That can be resolved in a separate PR. For now, I just remove the "SAT_" or "LAT_" prefix when looking up the center frequency.
> 
> @Zhilei I created a new branch on `pwg-scripts` called `noise_update`. I verified that the branches work together. We just need to merge them at the same time.

Thank you so much for looking into this! 
Sounds good! If you have tested them together I think we are ready to merge anytime, please let me know.

keskitalo (approved):
Looks good!

sotodlib/toast/sim_sso.py
```diff
@@ -276,14 +276,17 @@ def _observe_sso(self, sso_az, sso_el, sso_dist, sso_dia, tod, comm, prefix, foc
 
         if rank == 0:
             log.info("{}Observing the SSO signal".format(prefix))
-        
+
+        # FIXME: we should get the center frequency from the bandpass
         band_dict = {'f030' : 27, 'f040': 39, 'f090': 93,
              '150': 145 , 'f230': 225, 'f290': 285}
```
mhasself, line 282: missing 'f' in '150'?

keskitalo: Nice catch!

mhasself:
> @mhasself I wasn't sure how dependent your metadata code files were to the instrument model names so I left your band definitions unchanged in https://github.com/simonsobs/sotodlib/blob/noise_update/sotodlib/core/metadata/detdb.py and https://github.com/simonsobs/sotodlib/blob/noise_update/tests/test_metadata.py

Ya, I think this is fine, no need for changes here if the tests are passing.

I think the way I intend to use the band tags ("f150") is different from the way you need them for the instrument model.  You're basically using this as a key to lookup parameters for a bunch of detectors; naturally as the model gets more detailed you'll need to be more specific and allow for greater variation throughout the instrument ("SAT1_f150_lower_quadrant").  For data lookup, I will hold on to the very simple "f150" tags -- whatever the gory details are, there are a bunch of detectors centered near 150 GHz and I want to refer to all of them with this tag (and without needing to specify or know additional things).

smsimon:
@mhasself Great, that sounds good to me. Thanks!

PR #136, opened 2021-06-10
More general crosslinking

keskitalo:
This branch fixes an issue in old cross-linking maps and adds support to non-constant elevation scans.


keskitalo:
Merging this branch after communication with @mhasself

PR #141, opened 2021-06-16
Updates to instrument model to match instrumentation documentation

smsimon:
Largely minor instrument updates

### DIFF
--- a/sotodlib/sim_hardware.py
+++ b/sotodlib/sim_hardware.py
@@ -506,7 +506,7 @@ def sim_wafer_detectors(hw, wafer_slot, platescale, fwhm, band=None,
             pol_B[p] = 90.0 + pol_A[p]
         # We are going to remove 2 pixels for mechanical reasons
         kf = dim * (dim - 1) // 2
-        kill = [kf, kf + dim - 2]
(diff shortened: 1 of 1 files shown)


smsimon:
Model has been signed off on by the primary reviewers on the technical side

sotodlib/sim_hardware.py
```diff
@@ -573,8 +582,9 @@ def sim_wafer_detectors(hw, wafer_slot, platescale, fwhm, band=None,
                 dprops["channel"] = doff
                 dprops["AMC"] = doff // chan_per_AMC
                 dprops["bias"] = doff // chan_per_bias
-                dprops["reset_rate_kHz"] = 4.
                 dprops["readout_freq_GHz"] = readout_freq[doff]
+                dprops["bondpad"] = doff-(doff//64)*64
```
keskitalo, line 586: I would like to see the numerical constant `64` assigned to some descriptive variable name rather than used bare. Perhaps declared next to `chan_per_AMC`?

keskitalo (approved):

sotodlib/sim_hardware.py
smsimon: Good call--I added a variable for it

PR #147, opened 2021-07-26
added calc_wn(...) function

RemingtonGerras:
naive white noise calculator - just takes median PSD value at a certain frequency range (defaults to 5-10 Hz).


sotodlib/tod_ops/fft_ops.py
```diff
@@ -205,3 +205,35 @@ def calc_psd(aman, signal=None, timestamps=None, **kwargs):
         
     freqs, Pxx = welch( signal, 1/np.median(np.diff(timestamps)), **kwargs)
     return freqs, Pxx
+
+def calc_wn(aman, pxx, freqs=None, low_f=5, high_f=10):
```
kmharrington, line 209: I would add the option to have pxx inside the AxisManager since you already have freqs as an option. Those will always be paired.

kmharrington (approved):

PR #151, opened 2021-08-12
Run both serial and MPI tests

tskisner:
Run tests both with multiple processes and with MPI explicitly disabled.


tskisner:
Ok, sorry that this PR stagnated for a while.  This includes small changes that run the unit tests both with MPI and also with MPI explicitly disabled for the toast unit tests.

mhasself (approved):
Thanks for fixing this up!

PR #152, opened 2021-08-12
Omnibus 2108

mhasself:
Driver of this work is development of tools for mapmaking of planets (or other bright compact sources)

A few other subsystem fixes along the way.  In summary:

- General improvements to coords code: suppression of low-weight/low-condition when doing “remove_weights”.  Timer context manager to help with profiling.
- Context: help with checking metadata existence.
- tod_ops: Bug-fix polynomial gap filling, change interface (inplace->swap) slightly; now tries to use fast fill code from so3g.  Help with processing Ranges in ways useful for gap-filling bright signal data.  fourier_filter and detrend_data can operate on 1-d arrays, too.  fourier_filter will attempt to use OMP_NUM_THREADS threads.  Bugfix high_pass_sine2.  Test fixtures for detrend and gapfill.
- coords.planets module:  Assistance with ephemeris retrieval, mask&filter useful for compact sources.  Also, a mapmaking function, though it might be better to think of this as a template rather than a one-stop shop swiss knife routine.

@kmharrington Please take a look at the tod_ops changes.

If anyone wants to audit the other stuff, please do so / let me know in the next week or so.

### DIFF
--- a/.github/workflows/install_deps.sh
+++ b/.github/workflows/install_deps.sh
@@ -10,7 +10,8 @@ python3 -m pip install --upgrade pip
 # We upgrade scipy to fix an incompatibility with numpy
 
 python3 -m pip install --upgrade scipy
-python3 -m pip install nose toml mpi4py toast-cmb quaternionarray pyfftw
+python3 -m pip install nose toml mpi4py toast-cmb quaternionarray pyfftw \
+    skyfield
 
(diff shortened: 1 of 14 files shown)


kmharrington (approved):
tod_ops code looks good to me. Thanks for expanding!

PR #154, opened 2021-08-27
Toast3

keskitalo:
TOAST-3 port of `sotodlib`.


tskisner:
Quick update on this work:

- The simulation scripts in the corresponding toast3_bcp branch of pwg-scripts are now working with this branch.  I am checking one more aspect of the sso scripts.
- I am adding a toast.rst file to the docs discussing some details
- This work removes the previous ad-hoc spt3g data format.  Toast HDF5 format is supported and the L3/Book format is implemented in #183, to be merged after this work. 
- I will work on resolving the conflicts next before marking this PR ready for review.

Here are some plots from the `drone_test.sh` script from pwg-scripts.  Drone map, with just spatial common-mode filter and then a binned map, made in source-center Az/El CAR projection:

drone_map_I
drone_hits

mhasself (approved):
Thanks and hooray!  Great that we're now all firmly in the toast3 era.

You probably want to remove tests/toast2/test_time_constant.py?

PR #156, opened 2021-08-31
Make noise elevation scalings consistent

smsimon:
Make the noise elevation scalings consistent with the v3r6 noise estimations


keskitalo (approved):
The numbers seem sensible to me.

PR #161, opened 2021-09-13
Fix typo in smurf status loading

jlashner:
Small typo in the SmurfStatus object. Downsample and Filter disable register should be "Disable" and not "Disabled", which was causing these attributes to always be True.


kmharrington (approved):
Nice find

PR #162, opened 2021-09-23
Tod ops stats

### LINKED ISSUE #144: stats Module for tod_ops
Add in things like these

https://github.com/simonsobs/LATR_Validation/blob/6d89e3eae8e92c32aaa69218587d54982bfbda1d/latrtools/latrtools/__init__.py#L400-L654

skhrg:
Moved most of the important functions form latrtools (resolves #144). More functions may be added later, but the stuff here should be useful for current analysis work.


skhrg:
To be honest I really don't remember where I left off with this. I cleaned up some of the stuff I had but I'm not sure if this has everything we intended on porting over from [LATR_Validation](https://github.com/simonsobs/LATR_Validation/) (or if this has some stuff that we don't need/want to include here). If @kmharrington and/or @jeffiuliano could take a look and see if they remember more than me that would be great. Once this has everything it needs I can throw together the docs and some unit tests.

msilvafe:
@skhrg I recognize that this PR (and issue #144) are pretty out dated. We now have merged in a noise model + fitting function in PR #688 and added in a general binning function in `tod_ops.binning` added in PR #473. So I think what's left in here is basically gaussianity and skewness checks so maybe trim down this PR to that (and add in associated/relevant preprocess processes) so we can merge it or just close this.

skhrg:
Max I can't believe you revived this PR from when I was fresh out of first year, honestly some of this code is pretty embarrassing (I blame @jeffiuliano). I will add a rewrite the gaussian and skewness tests in a new PR (or maybe just a generic nth moment test).

PR #171, opened 2021-10-13
Coord sys revamp

### LINKED ISSUE #157: Detector positions in SSO sim aren't as expected
Extracting detector positions from a Jupiter sim on SAT1 w29 gives this focal plane map:

image

Note the notch in lower right (x=-1.8, y=-13.5).  The instrument model doc and the output of so_hardware_plot produce this image for w29:

image

I believe we all think we're plotting the same thing here -- the detector positions, as they project onto the sky, with the horizontal axis of the plot parallel to lines of constant elevation.

@smsimon @tskisner  I believe this is a simple matter of interchanging "x" and "y" coordinates, when you plot your detector position data.  But as discussed elsewhere it might be good to review the coordinate systems so everything is clear.

mhasself:
Resolves #157.

It was necessary to alter both the plotting code and the layout generation code.  Instead of totally rewriting all the layout code in a right-handed system, this rewrites the 'quat' field in the detector info, once everything is done.  The plotting code was fixed mostly by transposing X and Y in all matplotlib calls.  I tweaked some labeling and coordinate code along the way to address minor issues with the plotting that I noticed.

The plots produced by so_hardware_plot are unchanged, compared to before, except for small changes in the axis ranges and new wafer label sizes.  But the plots now do represent the focal plane as TOAST will use it.

To allow me to use G3 quat code I'm also putting in the "ToastishQuat" wrapper -- I had originally planned to put this into so3g.proj.quat but it is needed here now (and can be moved / duplicated there later).

### DIFF
--- a/sotodlib/coords/helpers.py
+++ b/sotodlib/coords/helpers.py
@@ -433,3 +433,67 @@ def _apply_inverse_weights_map(inverse_weights, target):
         target.shape[1], target.shape[2], target.shape[0], 1)
     m1 = np.matmul(iw, m)
     return m1.transpose(2,3,0,1).reshape(target.shape)
+
+
+class ToastishQuat(np.ndarray):
+    """Wrapper class for numpy arrays carrying quaternions with the ijk1
+    signature.
+
+    The practice in TOAST and quaternionarray is to store quaternions
+    in numpy arrays with shape (..., 4), with the real part of the
+    quaternion at index [..., 3].  In contrast, spt3g_software uses
+    signature 1ijk.
+
+    This class serves two main purposes:
+
+    - It can be used to annotate numpy arrays as carrying quaternions
+      in the ijk1 signature (without disturbing their behavior as
+      numpy arrays).
+    - It provides conversion to and from G3 quaternion containers,
+      including signature correction.
+
+    When instantiating an object of this class with a numpy array as
+    an argument, the object will wrap a view of the array (not a
+    copy).
+
+    When instantiating an object of this class with a G3 quat as an
+    argument, the object will store a copy of the G3 quat, translated
+    to ijk1 signature::
+
+      q_g3 = spt3g.core.quat(1.,2.,3.,4.)
+      q_tq = ToastishQuat(q_g3)
+      q_g3_again = q_tq.to_g3()
+
+      print(q_g3, q_tq, q_g3_again)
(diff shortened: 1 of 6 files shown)


sotodlib/coords/helpers.py
```diff
@@ -433,3 +433,67 @@ def _apply_inverse_weights_map(inverse_weights, target):
         target.shape[1], target.shape[2], target.shape[0], 1)
     m1 = np.matmul(iw, m)
     return m1.transpose(2,3,0,1).reshape(target.shape)
+
+
+class ToastishQuat(np.ndarray):
```
tskisner, line 438: Is there a reason this needs to be a class?  It seems like it would be simpler to just have two small functions to convert back and forth between the formats- especially since it is only being used in one place, to convert back and forth.  If you still want to have a class that inherits from ndarray, can we at least change name?  We are actively working to support both Boost and SciPy format quaternions in toast, so in the very near future this setting will be toggled in `sotodlib.toast.__init__.py` once for the whole package.

For functions maybe `quat_boost_to_scipy()` / `quat_scipy_to_boost()` or `quat_1ijk_to_ijk1()` / `quat_ijk1_to_1ijk()`. Or if we must have a class, can we change it to `ToastQuat` or maybe `ScipyQuat`, `ScalarLastQuat`, etc.

sotodlib/sim_hardware.py
```diff
@@ -686,8 +741,18 @@ def sim_telescope_detectors(hw, tele, tube_slots=None):
...
             tindx += 1
-    return alldets
 
+    # Transform quats from left-handed layout coordinates to
+    # right-handed TOAST-compatible coordintes -- see note at top of
+    # this file.  Note the ToastishQuat is just helping us convert to
+    # and from 3g quats.
+    quats_in = ToastishQuat([d['quat'] for d in alldets.values()])  # shape (n, 4)
+    xi, eta, gamma = so3q.decompose_xieta(quats_in.to_g3())
+    quats_out = ToastishQuat(so3q.rotation_xieta(eta, xi, np.pi/2 - gamma))
```
tskisner, line 751: This is a reasonable stop-gap solution, but keep in mind we are re-writing the contents of this file to do the layout and projection separately.

mhasself: The code is now correct and accurately documented; but I will not object to replacement with a superior thing that is also correct and accurately documented.

PR #172, opened 2021-10-20
G3SuperTimestream support

mhasself:
Preliminary work to support using G3SuperTimestream (in development in so3g) in the place of G3TimestreamMap in our data files.  This should still work fine with old so3g, when processing old data.  But if so3g has G3SuperTimestream, this can extract data from files where G3SuperTimestream has been subbed in.

This has only been tested on "load_smurf" compatible data files... the pipe-s0001 or similar format has additional complexity such that we should probably just write a new schema and extractor for it when we go to G3SuperTimestream.

### DIFF
--- a/sotodlib/io/load_smurf.py
+++ b/sotodlib/io/load_smurf.py
@@ -656,11 +656,22 @@ class G3tSmurf:
                              )
         
             data = frame.get('data')
+            sostream_version = frame.get('sostream_version', 0)
             if data is not None:
-                db_frame.n_samples = data.n_samples
-                db_frame.n_channels = len(data)
-                db_frame.start = dt.datetime.fromtimestamp(data.start.time / spt3g_core.G3Units.s)
-                db_frame.stop = dt.datetime.fromtimestamp(data.stop.time / spt3g_core.G3Units.s)
+                if sostream_version >= 2:  # Using SuperTimestreams
+                    db_frame.n_channels = len(data.names)
+                    db_frame.n_samples = len(data.times)
+                    db_frame.start = dt.datetime.fromtimestamp(
+                        data.times[0].time / spt3g_core.G3Units.s
+                    )
+                    db_frame.stop = dt.datetime.fromtimestamp(
+                        data.times[-1].time / spt3g_core.G3Units.s
+                    )
+                else:
+                    db_frame.n_samples = data.n_samples
+                    db_frame.n_channels = len(data)
+                    db_frame.start = dt.datetime.fromtimestamp(data.start.time / spt3g_core.G3Units.s)
+                    db_frame.stop = dt.datetime.fromtimestamp(data.stop.time / spt3g_core.G3Units.s)
 
                 if file_start is None:
                     file_start = db_frame.start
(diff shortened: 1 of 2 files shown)


jlashner:
I've been testing this with the upcoming version of the smurf-streamer and so far it's been working well. Are you waiting to test on pipe-s0001 files before merging?

mhasself:
> I've been testing this with the upcoming version of the smurf-streamer and so far it's been working well. Are you waiting to test on pipe-s0001 files before merging?

Oh, great.  No, we can proceed if it's been working well on the new smurf-streamer output!

jlashner (commented):
Looks good to me! I've tested using the `load_smurf.load_file` function, but looking closer now I'm pretty sure the indexing function will also have to be updated in order to work with G3tSmurf, which should be fairly straightforward. I can add a commit updating that function but not sure the best way to test it without data on simons1.

kmharrington:
If we have some example files all we need to know is that `G3tSmurf.add_file(path, session)` works. you could even do it with an empty test database so nothing else get's messed up.

mhasself:
Ok, converted back to draft while we sort out that other area.

jlashner:
Alright, tested by running this code from Katie:
```
SMURF = G3tSmurf('', 'test_db.db', '')
session = SMURF.Session()
SMURF.add_file('/home/jlashner/supertimestream.g3', session)
session.commit()

q = session.query(Files)
print(q.count()) ## number of paths you've added
print(q[0].name, q[0].frames, q[0].start, q[0].stop)
```

which returns:
```
Creating FrameType table...
1
supertimestream.g3 [Frame(Observation)<1>, Frame(Wiring)<2>, Frame(Wiring)<3>, Frame(Scan)<4>, Frame(Wiring)<5>, Frame(Wiring)<6>, Frame(Wiring)<7>, Frame(Wiring)<8>, Frame(Scan)<9>] 2021-10-29 17:05:17.024469 2021-10-29 17:05:20.563231
```

kmharrington:
Oh shoot, there's one more place we're loading in files in `load_smurf`. For the SmurfStatus objects. We need to make sure `SmurfStatus.from_file(/file/path/)` and `SmurfStatus.from_time()` work. 

From file just needs the new file to test. To test from time this snippet should work:
```
SMURF = G3tSmurf('', 'test_db.db', '')
session = SMURF.Session()
SMURF.add_file('/home/jlashner/supertimestream.g3', session)
session.commit()

SmurfStatus.from_time(dt.datetime.now().timestamp(), SMURF)
```

sotodlib/io/load_smurf.py
```diff
@@ -656,11 +656,21 @@ def add_file(self, path, session):
...
         
             data = frame.get('data')
+            sostream_version = frame.get('sostream_version', 0)
             if data is not None:
-                db_frame.n_samples = data.n_samples
-                db_frame.n_channels = len(data)
-                db_frame.start = dt.datetime.fromtimestamp(data.start.time / spt3g_core.G3Units.s)
-                db_frame.stop = dt.datetime.fromtimestamp(data.stop.time / spt3g_core.G3Units.s)
+                if sostream_version >= 2:  # Using SuperTimestreams
+                    db_frame.n_channels, db_frame.n_samples = data.data.shape
```
mhasself, line 662: Accessing data.data causes the array to be decompressed, which is otherwise unnecessary.  Better to just use len(data.names), len(data.times).

jlashner:
Alright, addressed matthew's comment and tested `from_file` and `from_time` functions! I think this means we're good to go unless there are any last requests...

jlashner (approved):

kmharrington (approved):

Issue #173, opened 2021-10-23
Multi-slot support for load_status


jlashner:
Currently the G3tSmurf database query for load_status doesn't allow you to specify the stream_Id, which messes things up whenever there are multiple slots.

Issue #170, opened 2021-10-11
Typo in load_smurf


csierra2:
It looks like there is a small typo inside load_smurf.py. I'm assuming `ignore_mission` should really be `ignore_missing`.

```
INFO: sotodlib.io.load_smurf: Axis Manager will have dets and samps axes

---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-93-7acbdf55019d> in <module>
      1 #aman = SMURF.load_data(obs.start, obs.stop, channels=bias_groups[bg4][:,:2] )
----> 2 aman = SMURF.load_data(obs.start, np.max(hk_data['xy_stage_x'][0]), channels=selectChs )
      3 get_trending_flags(aman)
      4 detrend_tod(aman)
      5 #demod.demod_sine(aman, 8)

~/.local/lib/python3.7/site-packages/sotodlib-0.3.0+268.g839fa9e-py3.7.egg/sotodlib/io/load_smurf.py in load_data(self, start, end, stream_id, channels, show_pb, load_biases)
   1198 
   1199         aman = load_file( flist, status=status, channels=channels,
-> 1200                          archive=self, show_pb=show_pb)
   1201 
   1202         msk = np.all([aman.timestamps >= start.timestamp(),

~/.local/lib/python3.7/site-packages/sotodlib-0.3.0+268.g839fa9e-py3.7.egg/sotodlib/io/load_smurf.py in load_file(filename, channels, ignore_missing, load_biases, load_primary, status, archive, obsfiledb, show_pb, det_axis)
   1821         ch_mask = get_channel_mask(channels, status, archive=archive, 
   1822                                    obsfiledb=obsfiledb,
-> 1823                                    ignore_missing=ignore_missing)
   1824     else:
   1825         ch_mask = None

~/.local/lib/python3.7/site-packages/sotodlib-0.3.0+268.g839fa9e-py3.7.egg/sotodlib/io/load_smurf.py in get_channel_mask(ch_list, status, archive, obsfiledb, ignore_missing)
   1565                     channel = session.query(Channels).filter(Channels.name==ch).one_or_none()
   1566                     if channel is None:
-> 1567                         if not ignore_mission:
   1568                             raise ValueError(f"channel {ch} not found in G3tSmurf Archive")
   1569                         continue

NameError: name 'ignore_mission' is not defined
```

Issue #160, opened 2021-09-13
G3tSmurf: Reorganize Metadata update


kmharrington:
Re-write the metadata indexing to use defined file names instead of defined folder names.

Issue #138, opened 2021-06-11
G3tSmurf: Add database update scripts to scripts/ folder


kmharrington:
These are getting long enough it's probably worth standardizing. We could use a "build from scratch" version and a "update the last ~day of data" version. The second needs to be robust to encountering partial observations and/or partial data transfers. Both should include dumping to the DetDb so we can get used to doing that. I'd expect either some form of config.yaml for the 4 required paths or a bunch of passed arguments. 

Here's a starting point: https://github.com/simonsobs/latrt-testing/blob/main/scripts/update_database.py 

This could also include expansions to track sections of the folders that aren't in the correct format / don't contain the right files, so we don't constantly try and fail to add them.

PR #177, opened 2021-10-26
G3tsmurf Misc updates

kmharrington:
This PR fixes a laundry list of issues:  #170 , #160, #137 , #173, #138, #135

Nothing should be breaking existing code. 

highlights:
1. Adds tag tracking to observations, automatically applying tags set at the time of taking the data
2. Adds some new generators that makes it easier to search through the SMuRF folder for things
3. Adds an "official" database update script into pipelines so we can standardize how that's called
4. Many bug fixes.

### DIFF
--- a/pipelines/update_g3tsmurf_database.py
+++ b/pipelines/update_g3tsmurf_database.py
@@ -21,11 +21,11 @@ if __name__ == '__main__':
     
     parser.add_argument('--update-delay', help="Days to subtract from now to set as minimum ctime",
                                default=2, type=float)
-    parser.add_argument('--from_scratch', action="store_true")
+    parser.add_argument('--from-scratch', help="Builds or updates database from scratch",
+                        action="store_true")
     
     args = parser.parse_args()
     
-    print(args.__dict__)
     if args.timestream_folder is None:
         args.timestream_folder = os.path.join( args.data_prefix, 'timestreams/')
        
--- a/pipelines/update_g3tsmurf_database.py
+++ b/pipelines/update_g3tsmurf_database.py
@@ -1,0 +1,57 @@
+import os
+
+import datetime as dt
+import numpy as np
+import argparse
+
+from sotodlib.io.load_smurf import G3tSmurf, Observations, dump_DetDb
+
+
+if __name__ == '__main__':
+    
+    parser = argparse.ArgumentParser()
+    parser.add_argument('data_prefix', help="The prefix to data locations, use individual"
+                                            "locations flags if data is stored in non-standard folders")
+    parser.add_argument('db_path', help="Path to Database to update")
+    
+    parser.add_argument('--timestream-folder', help="Absolute path to folder with .g3 timestreams. Overrides data_prefix")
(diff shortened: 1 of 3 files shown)


sotodlib/io/load_smurf.py
```diff
@@ -1005,10 +1022,189 @@ def update_observation_files(self, obs, session, max_early=5, max_wait=100, forc
...
+                    continue
+                for root, dirs, files in os.walk(os.path.join(path, adir), topdown=False):
+                    for name in files:
+                        try:
+                            try:
+                                ctime = int(name.split('_')[0])
+                            except ValueError:
+                                ctime = actime
+                            fname = '_'.join(name.split('_')[1:])
+                            yield (fname, stream_id, ctime, os.path.join(root, name))
```
jlashner, line 1119: This method of returning ctimes + filenames seems strange to me, and doesn't really work if the first component of the filename is not a ctime (i.e. you will still be throwing away the first word). Maybe this should return the ctime and also the full filename including the first word?

sotodlib/io/load_smurf.py
```diff
@@ -1468,12 +1616,25 @@ def from_time(cls, time, archive, show_pb=False):
...
                 Frames.type_name == 'Observation',
                 Frames.time <= time
-            ).order_by(Frames.time.desc()).first()
+            ).order_by(Frames.time.desc())
 
-        status_frames = session.query(Frames).filter(
+        if stream_id is not None:
+            q = q.join(Files).filter(Files.stream_id==stream_id)
+        else:
+            logger.warning(f"No stream_id given for finding status, not checking for one")
```
jlashner, line 1627: You can use the `_stream_ids_in_range` function to check if there are multiple stream-id options at a given time. I think if no stream_id is passed and there are multiple options this should raise an error.

pipelines/update_g3tsmurf_database.py
```diff
@@ -0,0 +1,57 @@
...
+    parser.add_argument('db_path', help="Path to Database to update")
+    
+    parser.add_argument('--timestream-folder', help="Absolute path to folder with .g3 timestreams. Overrides data_prefix")
+    parser.add_argument('--smurf-folder', help="Absolute path to folder with pysmurf archived data. Overrides data_prefix")
+    
+    parser.add_argument('--detdb-filename', help="File for dumping the context detector database")
+    
+    parser.add_argument('--update-delay', help="Days to subtract from now to set as minimum ctime",
+                               default=2, type=float)
+    parser.add_argument('--from_scratch', action="store_true")
```
jlashner, line 24: `--from-scratch`

jlashner (approved):
Looks good to me! Just a few small comments inline but other than that I think this looks great, and I definitely appreciate the standardized update script.

sotodlib/io/load_smurf.py
kmharrington: I think the big thing here is the fname without the ctime makes for fast/easier pattern matching scripts. Matches similarly with the searching through the folders. Also, all the standard files from smurf all come with the ctime in front. I'd expect that to remain a requirement.

PR #187, opened 2021-11-18
single metadata edge case

kmharrington:
@mhasself  I'm not sure if this is how you'd like to fix it. But I found an edge case in context where if there's only one listed metadata entry to loaders fail. `SuperLoader.load( spec_list, request)` expects as list for `spec_list` and with one metadata entry it's just getting that single dictionary.

### DIFF
--- a/sotodlib/core/context.py
+++ b/sotodlib/core/context.py
@@ -247,8 +247,9 @@ class Context(odict):
         # Load metadata.
         metadata_list = self._get_warn_missing('metadata', [])
         # edge case when only one entry in metadata
-        if type(metadata_list) == dict:
-            metadata_list=[metadata_list]
+        if not isinstance(metadata_list, list):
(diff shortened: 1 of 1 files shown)


mhasself:
So this would happen if your context had a single dictionary:
```
metadata:
  db : '{metadata_lib}/wn_db.gz'
  name: 'wn'
```
instead of a list containing a single dictionary:
```
metadata:
  - db : '{metadata_lib}/wn_db.gz'
    name: 'wn'
```
Yes, YAML can be a bit concise sometimes ...

I think I'd prefer an informative error message instead of just silently accepting misleadingly formatted context structures ... pseudocode:
```
if not isinstance(spec_list, list):
 raise ValueError('helpful message')
```

kmharrington:
FML yamls. Yeah, a better error message is totally fine! Will add that and switch.

kmharrington:
@mhasself  how's this?

mhasself (approved):
Thanks!

PR #190, opened 2021-11-29
G3tSmurf: Fix Frame making and querying

kmharrington:
While trying to track down an error we were seeing I realized that the way I was building the frames table was subtly wrong. I was marking every Status frame as a dump frame instead of only when `frame['dump'] == 1`. This branch fixes that and an indexing error. 

Also edited the `add_file` function so we could use it to fix entries we found to be wrong. 

Lastly, I fixed the actual error we were encountering, where the queries were not returning the correct status objects. This was happening because we were mixing `Frames.Start` for Scan frames and `Frames.Time` for Observation frames and these are not guaranteed to be in order.

### DIFF
--- a/sotodlib/io/load_smurf.py
+++ b/sotodlib/io/load_smurf.py
@@ -673,6 +673,7 @@ class G3tSmurf:
                 dump = False
             else:
                 dump = bool(frame['dump'])
+                
             ## only make Frame once the non-nullable fields are known
             db_frame = Frames(frame_idx=frame_idx, file=db_file,
                              offset = db_frame_offset,
@@ -1396,9 +1397,9 @@ class G3tSmurf:
             stream_id = q[0].stream_id
             
         if status is None:
-            scan_start = session.query(Frames.start).filter(Frames.start > start,
+            scan_start = session.query(Frames.time).filter(Frames.time >= start,
                                                             Frames.type_name=='Scan')
-            scan_start = scan_start.order_by(Frames.start).first()
+            scan_start = scan_start.order_by(Frames.time).first()
                 
 
             try:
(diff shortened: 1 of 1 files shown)


tskisner (commented):
Although I can't test this, the code looks fine.

sotodlib/io/load_smurf.py
```diff
@@ -611,14 +611,29 @@ def add_file(self, path, session):
...
+            db_file = Files(name=path)
+            session.add(db_file)
+        elif not overwrite:
+            logger.info(f"File {path} found in database, use overwrite=True to update")
+            return
+        else:
+            logger.debug(f"File {path} found in database, updating entry and re-making frames")
+            db_frames = db_file.frames 
+            [session.delete( frame ) for frame in db_frames];
+            session.commit()
```
jlashner, line 634: Are we sure we want to commit here? I'd think if the overwrite attempt fails we'd rather leave the db untouched, but could go either way.

jlashner (approved):
Looks good to me! One small question inline, but up to you if you want to address it.

sotodlib/io/load_smurf.py
kmharrington: Yeah, the alternative to committing there felt a lot more complicated and/or more likely to break. 

Instead of deleting, we could go through and update the frames individually, but that would take a much larger re-write of the code or writing an update script that was totally separate from the creation script. The larger re-write was both "work" and I wasn't sure how do to it without adding a whole but of queries to the Frame table that would slow everything down significantly. I didn't like the separate script idea because that's two things we need to update if we change this table.

If we delete and re-create like I'm doing then we have to commit the changes after the delete or the relational mapping gets weird (I'm pretty sure that's true, although I realize now you could convince me I was wrong about that).

Issue #182, opened 2021-11-04
G3tSmurf: load_file breaks for files in the middle of an observation.


kmharrington:
This is true for all cases, loading just files, files+G3tSMuRF and files+obsfiledb. 

The SmurfStatus information is only in the first g3 file in a session. When loading with G3tSMuRF or obsfiledb there are ways to pull out the exact status frame and we should implement those. When loading with just files we should add more helpful error message that points the users to loading the status separately. 

Current work around:
```
status = SmurfStatus.from_file( '/first/file/in/session')
aman = load_file( '/any/file/in/session', status = status)
```

kmharrington:
Best way to tackle this will be working off of what's in [G3tSmurf.load_data](https://github.com/simonsobs/sotodlib/blob/eeda780b8afd346f1b0e983b08135aefeebddcd2/sotodlib/io/load_smurf.py#L1230). Although the obsfiledb version will be a little more challenging.

PR #194, opened 2021-12-16
Adding fix to load_file to load status from start of an observation

jseibert575:
Adds functionality in load_file so if a user tries to load a g3 file from the middle of an observation, it loads the full status information from the g3 file at the start of the observation. If this fails because of reasons (most likely it tries to load a file that doesn't exist), provides an error message and raises an exception. Closes https://github.com/simonsobs/sotodlib/issues/182, but does not address obsfiledb things yet, but can easily add this (just wasn't sure what was necessary right now). Tested this on a variety of observations and it had no problems, but happy to add more if necessary.

### DIFF
--- a/sotodlib/io/load_smurf.py
+++ b/sotodlib/io/load_smurf.py
@@ -2086,7 +2086,16 @@ def load_file(filename, channels=None, ignore_missing=True,
         logger.error("No files provided to load")
     
     if status is None:
-        status = SmurfStatus.from_file(filenames[0])
+        try:    
+            logger.warning('Loading status frame from the file at the start'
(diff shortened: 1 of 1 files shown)


kmharrington:
I have tested this. Thanks a bunch for getting this working! (While we're doing hot fixes, could we add the PR I just made to this branch?)

kmharrington (approved):

jseibert575:
> I have tested this. Thanks a bunch for getting this working! (While we're doing hot fixes, could we add the PR I just made to this branch?)

Added one comment just for clarity but yes works for me! Approved in other PR.

PR #195, opened 2021-12-16
Also check for number of channels

kmharrington:
This is another error I'm seeing show up (loading an incomplete status object). I know why that error is happening but can we add the hot fix here until we've got the longer fix?

### DIFF
--- a/sotodlib/io/load_smurf.py
+++ b/sotodlib/io/load_smurf.py
@@ -2085,10 +2085,13 @@ def load_file(filename, channels=None, ignore_missing=True,
     if len(filenames) == 0:
         logger.error("No files provided to load")
     
-    if status is None:
-        try:    
-            logger.warning('Loading status frame from the file at the start'
+    if status is None or status.num_chans is None:
+        try:
+            if status is None:
+                logger.warning('Loading status frame from the file at the start'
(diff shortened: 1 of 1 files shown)


sotodlib/io/load_smurf.py
```diff
@@ -2085,10 +2085,13 @@ def load_file(filename, channels=None, ignore_missing=True,
...
-    if status is None:
-        try:    
-            logger.warning('Loading status frame from the file at the start'
+    if status is None or status.num_chans is None:
+        try:
+            if status is None:
+                logger.warning('Loading status frame from the file at the start'
                             ' of the corresponding observation.')
+            else:
+                logger.warning("Status information is missing 'num_chans.' Trying to fix.")
```
jseibert575, line 2094: Idea is just to make sure user is aware that this key is missing from the status?

jseibert575 (approved):

sotodlib/io/load_smurf.py
kmharrington: yeah. Idea is to give the actual error instead of one that might point them in the wrong direction. Big thing is that fixing this error _well_ will take me awhile.

Issue #179, opened 2021-10-28
G3tSmurf: Fix Multislot Observations


kmharrington:
G3tSmurf is currently making observations per slot / stream_id. This isn't surprising given how the code is currently written but does not mesh with how context wants observations organized. 

This adds new questions we need to deal with regarding synchronization because each slot will start streaming at slightly different times. Here's an example from the most recent streaming tests:

```
Observation:  1633987456: 2021-10-11 17:24:16 -> 2021-10-11 18:24:29.252010 [1:00:13.252010] (None)
Tuneset/Detset [('1633642729_tune.npy', 'crate1slot2')]
Files:
	 /mnt/so1/data/pton-rd/timestreams/16339/crate1slot2/1633987463_006.g3
	 /mnt/so1/data/pton-rd/timestreams/16339/crate1slot2/1633987463_005.g3
	 /mnt/so1/data/pton-rd/timestreams/16339/crate1slot2/1633987463_004.g3
	 /mnt/so1/data/pton-rd/timestreams/16339/crate1slot2/1633987463_003.g3
	 /mnt/so1/data/pton-rd/timestreams/16339/crate1slot2/1633987463_002.g3
	 /mnt/so1/data/pton-rd/timestreams/16339/crate1slot2/1633987463_001.g3
	 /mnt/so1/data/pton-rd/timestreams/16339/crate1slot2/1633987463_000.g3
Observation:  1633987468: 2021-10-11 17:24:28 -> 2021-10-11 18:24:41.469677 [1:00:13.469677] (None)
Tuneset/Detset [('1633644164_tune.npy', 'crate1slot3')]
Files:
	 /mnt/so1/data/pton-rd/timestreams/16339/crate1slot3/1633987475_006.g3
	 /mnt/so1/data/pton-rd/timestreams/16339/crate1slot3/1633987475_005.g3
	 /mnt/so1/data/pton-rd/timestreams/16339/crate1slot3/1633987475_004.g3
	 /mnt/so1/data/pton-rd/timestreams/16339/crate1slot3/1633987475_003.g3
	 /mnt/so1/data/pton-rd/timestreams/16339/crate1slot3/1633987475_002.g3
	 /mnt/so1/data/pton-rd/timestreams/16339/crate1slot3/1633987475_001.g3
	 /mnt/so1/data/pton-rd/timestreams/16339/crate1slot3/1633987475_000.g3
[...]

PR #199, opened 2021-12-28
G3tSmurf New Years Updates

### LINKED ISSUE #181: G3tSmurf: Start populating sample_start and sample_stop in files
This is needed for working with things through context. 

Should be straight-forward to add in the `update_observation` function.

### LINKED ISSUE #184: G3tSmurf : Add stream_id to readout detector name
Because we can't guarantee that two tune files / channel assignments won't happen at the same time and these names need to be unique.

### LINKED ISSUE #192: G3tSmurf: Use status frames to find observation ends
Right now we're making some un-necessary assumptions about when observations end but that info is in the status frames. So we should use those.

### LINKED ISSUE #198: Fix Datetime Creation in G3tSmurf
We are using `dt.datetime.fromtimestamp(ctime)` all over the place where we should be using `dt.datetime.utcfromtimestamp(ctime)`. 

I'm surprised this hasn't caused more trouble. But I'm working on a larger database update that will also require running scripts on existing databases, so I'll add this there along with the fix.

kmharrington:
Another larger set of G3tSmurf updates that will be matched with some update scripts that need to run on existing tables. These are some bug fixes, some QOL improvements, and some setup to get things ready for the next update of usage with Context. 

#### General
* Creates a site-scripts folder and moves the g3tsmurf database update script into it. This folder is intended for scripts that will be integrated with the workflow database at the site.
*  Split database tables definitions into a new file (because `load_smurf` is still way too large _after_ doing that)
*  Fix docs associated with breaking out tables 
*  Misc. docstring updates
*  Add a `max_ctime` optional argument to all update scripts (I found a use for it once so might as well)
*  Fix all datetime creation to use `datetime.datetime.utcfromtimestamp` (fixes #198 )

#### Observations Table
* Change timestamp and obs_id to be build off the .g3 session instead of the pysmurf action
* Add `stream_id` to `obs_id` and as a column in the database (see discussion in #179 ) 
* Add `action_ctime` and `action_name` for better tracking (seemed necessary and useful while implementing #179  )
* Add `n_samples` to track number of samples in an observation (seemed useful with #181 )
* Use the information in the Wiring frames to find when a data stream is actually finished. This will make for a simpler database update script (fixes #192 )

#### Files Table
* Start populating sample_start and sample_stop for files found in observations. (fixes #181 )

#### Channels Table
* Add stream_id to `name` to make a complete and unique Readout Id. (fixes #184 )

### Updating Existing Databases

The database update script is saved in pwg-scripts here: 
https://github.com/simonsobs/pwg-scripts/tree/master/pwg-duf/g3tsmurf_202112

The script is designed to be run on **this branch** of sotodlib (or master once it's merged)

Usage: 

> source main.sh /path/to/database /path/to/new/database /data/prefix

Data Prefix is where timestream/smurf folders are on your computer. The script will copy the database path to the new path and then run all the updates on the new copy. Keep the old copy around until you're sure the new one works as desired. The entire update takes about 4-6 hours on the current LATRt database, so definitely start it in a screen session.

Notes: 
1. The update script **does not** change the `obs_id` or `timestamp` of existing observations. This is because these are what people are usually sharing with others to find data and I don't want to break all that existing analysis. This does mean that if anyone builds a whole database from scratch the `obs_id` and `timestamp` will be different than ones in existing tables. (Somewhat annoying yes, but more important to get a better setup for the field.) 
2. The update script **does** changes the observation start and stop datetimes because those have been being made with timezone dependent information and we need to fix that. 
3. The update script failed on LATRt's databases for a period around March 2021 because of some file formatting issues then. Watch for failures on updates and let me know if there are periods of data where it fails and you need that information.


mhasself (approved):
I really appreciate the attention to documentation!

Two questions (not even sure if these are part of the new work): 
- Docstring for Observations.stream_id -- isn't the important thing here that this tells you what Smurf card (or UFM, if you like) the data come from?  Not sure that the comment about " Required since level 2 ..." is helpful.  I think that comment would be more relevant in the explanation of obs_id.
- Docstring for Observations.tag -- this is a string... containing comma-delimited tags?

Ignore me if I just don't get it.  Thanks again.

kmharrington:
Added a bit more info on those two. I have the note about L2 data on the `stream_id` because, to a large extent, that's our main departure from the setup of the Context style Observations. (Having one per `stream_id` also technically means we only expect one `detset` per observation).

Issue #176, opened 2021-10-26
AxisManager should accommodate scalars


mhasself:
AxisManager can't handle scalars (not even scalar arrays).  It probably should.

As suggested most recently by @amaurea , for pointing matrix support (weather and site info), but I know this is something many people have wondered about.  There are obviously many use cases in the detector setup realm.

kmharrington:
I've realized this would be super super useful because then we could merge all the details in the `SmurfStatus` object into an AxisManager.

skhrg:
Started taking a whack at this (https://github.com/simonsobs/sotodlib/tree/axisman_scalars)
What's unclear to me is if there is some meaningful equivalent of `restriction` and `intersection` for scalars. 

Comments/suggestions would be appreciated.

kmharrington:
Great to see work on this! The easiest way for us to get comments on the specific code might be to make a draft pull request (it let's us comment on individual lines / have discussions on different parts of the code.

I'm confused what you mean by a scalar array in this instance, could you maybe share an example of your the expected use case for that one? And/or also for just general scalars?

skhrg:
Good point, draft PR is #200 

As far as scalar arrays go, I am not too sure what the use case is, more of a I didn't see a reason not to allow it sort of thing (ie: calling `resolve` with `src` that is an array rather than a single scalar doesn't cause an issue).

mhasself:
Closed via #200

PR #200, opened 2022-01-06
Adding scalars to AxisManager

skhrg:
Adding scalar support to axis manager (#176 )

### DIFF
--- a/docs/axisman.rst
+++ b/docs/axisman.rst
@@ -72,14 +72,22 @@ hold other AxisManagers, and some other stuff too).
 
     dset = core.AxisManager().wrap('tod', tod)
 
+AxisManagers can also hold scalars (ints, floats, strings, etc.), note that ``None`` is considered a scalar as well.
+
+.. code-block:: python
+
+    dset = dset.wrap('scalar', 1.0)
+
 Inspecting::
 
     >>> print(dset)
-    AxisManager(tod[3,10000])
+    AxisManager(tod[3,10000], scalar[0])
     >>> print(dset.tod)
     [[10. 10. 10. ... 10. 10. 10.]
      [11. 11. 11. ... 11. 11. 11.]
      [12. 12. 12. ... 12. 12. 12.]]
+    >>> print(dset.scalar)
+    1.0
 
 
 The value that AxisManager adds is an ability to relate an axis in one
--- a/docs/axisman.rst
+++ b/docs/axisman.rst
@@ -72,7 +72,8 @@ hold other AxisManagers, and some other stuff too).
 
     dset = core.AxisManager().wrap('tod', tod)
 
-AxisManagers can also hold scalars (ints, floats, strings, etc.), note that ``None`` is considered a scalar as well.
+AxisManagers can also hold scalars, a value is considered scalar if
+``'np.isscalar'`` thinks it is a scalar or if it is ``'None'``.
 
 .. code-block:: python
 
@@ -81,7 +82,7 @@ AxisManagers can also hold scalars (ints, floats, strings, etc.), note that ``No
 Inspecting::
 
     >>> print(dset)
-    AxisManager(tod[3,10000], scalar[0])
+    AxisManager(tod[3,10000], scalar)
     >>> print(dset.tod)
     [[10. 10. 10. ... 10. 10. 10.]
      [11. 11. 11. ... 11. 11. 11.]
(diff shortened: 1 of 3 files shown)


kmharrington:
Thanks! 

I think @mhasself  is going to have a much better idea about how this implementation will work out than I will.

mhasself:
Hi @skhrg  --- thanks for working on this.  I haven't tested this but it looks like it's on the right track.

Some notes on what else needs to happen here:
- I think the _assignments[] shouldn't be [None] for the scalar case, but rather [].  Scalars are "0-dimensional".
- Do something sensible for ``__repr__``.  For scalars we might just dump out var=<value of var>.
- Include scalar examples in the documentation.
- This needs to be included in the unit testing...

Here's the unit test I started (in test_core.py), for my attempt at this.
```
    def test_160_scalars(self):
        aman = core.AxisManager(core.LabelAxis('dets', ['a', 'b']),
                                core.OffsetAxis('samps', 100))
        aman.wrap_new('signal', shape=('dets', 'samps'))

        aman.wrap('x', np.array(12))
        aman.wrap('y', np.array(13))
        aman.wrap('z', np.array('hello'))
        aman.wrap('z2', np.array('hello'))

        # Accept trivially promoted scalars                                                             
        aman.wrap('x2', 12)
        aman.wrap('z3', 'hello')

        # Don't just let people wrap any old thing.                                                     
        with self.assertRaises(ValueError):
            aman.wrap('a_dict', {'a': 123})

```
It will a good idea to add some scalar entries in some of the other tests, e.g. test_140, and confirm it behaves as expected.

skhrg:
@mhasself in your unit test was the intention that these lines:
```
        aman.wrap('x', np.array(12))
        aman.wrap('y', np.array(13))
        aman.wrap('z', np.array('hello'))
        aman.wrap('z2', np.array('hello'))
```
are wrapping scalars? In which case would you like me to treat an array with only a single element as a scalar?

mhasself:
> @mhasself in your unit test was the intention that these lines:
> 
> ```
>         aman.wrap('x', np.array(12))
>         aman.wrap('y', np.array(13))
>         aman.wrap('z', np.array('hello'))
>         aman.wrap('z2', np.array('hello'))
> ```
> 
> are wrapping scalars?

Ahh... yes, those are probably not a good idea.  I was playing with the idea of wrapping scalars as 0-dimensional numpy arrays.  But that's perhaps confusing (np.isscalar(np.array(12)) is False).  You can drop those.

> In which case would you like me to treat an array with only a single element as a scalar?

Not totally sure what you mean here, but "no" is a safe answer!

People should be discouraged from using np.array(scalar) ... just use a scalar.

skhrg:
> > In which case would you like me to treat an array with only a single element as a scalar?
> 
> Not totally sure what you mean here, but "no" is a safe answer!
> 
> People should be discouraged from using np.array(scalar) ... just use a scalar.

Yeah this answers my question, by array with a single element i just meant the length 1 array you would get from `np.array(scalar)`

docs/axisman.rst
```diff
@@ -72,14 +72,22 @@ hold other AxisManagers, and some other stuff too).
 
     dset = core.AxisManager().wrap('tod', tod)
 
+AxisManagers can also hold scalars (ints, floats, strings, etc.), note that ``None`` is considered a scalar as well.
```
mhasself, line 75: For docs, please wrap lines at 78 chars or so (like the rest of the file).

What is included in "etc" in this line?  It's probably worth mentioning a scalar is anything that np.isscalar thinks is a scalar.

docs/axisman.rst
```diff
@@ -72,14 +72,22 @@ hold other AxisManagers, and some other stuff too).
...
+
+.. code-block:: python
+
+    dset = dset.wrap('scalar', 1.0)
+
 Inspecting::
 
     >>> print(dset)
-    AxisManager(tod[3,10000])
+    AxisManager(tod[3,10000], scalar[0])
```
mhasself, line 84: The notation scalar[0] implies that scalar is 1-d array with shape (0,).  I think the repr in this case should just be
```
 AxisManager(tod[3,10000], scalar)
```

mhasself (changes requested):
Sorry, just saw that this was marked ready.  Thanks again for doing this.

See a couple of comments inline about formatting.

Also... right now, this will fail:
```
 am.wrap('a', np.array([1]))
 am.wrap('a', np.array([2]))  # ValueError
```
but this succeeds:
```
 am.wrap('a', 1)
 am.wrap('a', 2)  # clobber
```
Although the errors are often annoying, the behavior should be consistent... so please get the scalar case to raise an error, like the array case.  Also please at tests for this behavior (in both scalar and array cases).

mhasself (approved):
Ok, thank you!

PR #202, opened 2022-01-13
add missing condition check

guanyilun:
without the condition check, load_file will fail with an error in accessing the primary field, when the field doesn't exist.


kmharrington (approved):
thanks for the fix!

Issue #105, opened 2021-02-03
Add requirements file


tskisner:
Hosting `sotodlib` on PyPI is a work in progress, pending so3g/spt3g work.  However, when pip installing sotodlib we could already provide a requirements.txt file that eases the installation / verification that dependencies are present.

kmharrington:
Could we revisit this? We need a requirements file for the docker install of sotodlib that goes inside of the sodetlib docker image. 

A couple notes:
* @jlashner  Does the docker image building automatically look for requirements files and install those packages? 
* Is toast ever going to be a requirement for sotodlib? I believe we decided on no but want to make sure we're still good with that. Because we definitely don't want to require toast on the smurf servers.

kmharrington:
In particular this error shows up because our filtering functions need `pyfftw` but that's not installed in the sodetilb docker. But I think adding / using the requirements file is actually the correct way to fix that problem

kmharrington:
Does this part of `setup_opts` effectively work as a requirements file? https://github.com/simonsobs/sotodlib/blob/master/setup.py#L49

tskisner:
The long story about requirements files an setup.py is here:
https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/#requirements-files
https://caremad.io/posts/2013/07/setup-vs-requirement/

But for our purposes (assuming all requirements are listed in setup.py) we could just add a `requirements.txt` file that had:
```
--index-url https://pypi.python.org/simple/

# By default, use our abstract dependencies in setup.py for the
# concrete dependencies from PyPI.
.
```

tskisner:
Closing this issue.  Sotodlib has a proper pyproject.toml and is on pypi.

PR #203, opened 2022-01-14
Update installation dependencies and add requirements to setup.py

BrianJKoopman:
## Description
This PR adds the required packages (that I know of anyway) to the `install_requires` option within `setup.py`. I also updated the intro docs page for installation. 

The thing this doesn't really solve (and I'm not sure it should) is providing any description of installing spt3g/so3g within conda. If a new sotodlib user chooses to use conda, then they might run into difficulty there.

## Motivation
This was motivated by a user installing sotodlib, but finding so3g and PyYAML to be missing. They pointed out that the docs didn't list so3g as a requirement.

I took the approach of adding to setup.py since it should reduce the need for the user to manually install packages, as the dependencies will be checked and installed when running `pip install .` to install sotodlib.

For conda users, I believe they should still install packages available via conda manually first, then pip will make sure remaining packages are installed.

## Testing
I tested this by installing sotdolib with both pip and conda on a clean Ubuntu 20.04 system. The only packages I find missing now when running the tests are spt3g and so3g (so I don't think I'm missing any in the list added to `setup.py`.)

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -34,7 +34,6 @@ Create a new conda environment::
   %> conda create --copy -m -p ${HOME}/software/so
   %> conda activate ~/software/so
   %> conda install pip numpy scipy matplotlib
-  %> pip install quaternionarray
 
 ... Or Use Virtualenv and Pip
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -46,10 +45,8 @@ it::
   %> virtualenv -p python3 ${HOME}/software/so
   %> source ${HOME}/software/so/bin/activate
 
-Now use pip to install the dependencies we need::
-
-    %> pip install numpy scipy matplotlib
-    %> pip install quaternionarray
+pip installable external dependencies will be automatically installed when pip
+installing the sotodlib package.
 
 
 S.O. Affiliated Dependencies
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -56,8 +56,13 @@ Activate / load your python stack from the previous section.  Since you created
 a conda environment or virtualenv directory specifically for S.O. tools, you
 can always delete that directory and make a new one as needed.
 
-Currently the sotodlib package does not require any other S.O. packages.  In
-the future, it will require sotoddb and so3g as dependencies.
+sotodlib requires the following affiliated packages:
+
+- `so3g`_
+- `spt3g_software`_
+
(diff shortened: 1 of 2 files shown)


tskisner (commented):
These changes look good to me.  For reference, in my so3g packaging work, I am using this set of upstream python requirements for so3g:
https://github.com/simonsobs/so3g/pull/108/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R374-R383
Obviously that does not include the compiled dependencies, which are boost, lapack, and qpoint (and spt3g obviously).

BrianJKoopman:
Ah, interesting, thanks Ted. I also noticed after submitting this the [install_deps.sh](https://github.com/simonsobs/sotodlib/blob/master/.github/workflows/install_deps.sh) script used in the test workflow. There are some more package that aren't captured in this PR there, are these also dependencies for sotodlib?

- toml
- mpi4py
- toast-cmb
- pyfftw
- skyfield
- pixell

BrianJKoopman:
> My understanding is that we aren't making sotodlib require toast / it's dependencies. Ex: #105

Gotcha, based on that issue it looks like at least `pyfftw` from this list should be added. Any others? It looks like `mpi4py` is an optional dependency for `toast-cmb`. So I'm still wondering about:
- toml
- skyfield
- pixell

mhasself:
> > My understanding is that we aren't making sotodlib require toast / it's dependencies. Ex: #105
> 
> Gotcha, based on that issue it looks like at least `pyfftw` from this list should be added. Any others? It looks like `mpi4py` is an optional dependency for `toast-cmb`. So I'm still wondering about:
> 
>     * toml
> 
>     * skyfield
> 
>     * pixell

Yes, these are required.  It's possible that toml is only used in modules that also require toast.  But ok to include it.

BrianJKoopman:
Alright, I added those as well. Thanks, Matthew!

tskisner:
@kmharrington , the toast package is only required by the `sotodlib.toast` subpackage as agreed upon earlier.  However, for the record, I would argue that toast is actually easier to install (with pip or conda) than some of the "required" dependencies, and does not depend on MPI or other challenging things.

BrianJKoopman:
I just re-requested review on this. If there are any further changes that I should make, please let me know.

tskisner (approved):
These changes look fine.  If you have a moment, could you bump the require python version to 3.7.0 to match so3g?

PR #207, opened 2022-01-22
Context tuneup

mhasself:
New features:
* samples argument to Context.get_obs
* better chance that get_obs can work in absence of a DetDb
* ObsFileDb.lookup_file enables Context.get_obs(filename=...)

Bug fix / reorganization:
* metadata is looked up relative to the metadata index (the database file)
* obsloader registry lives in context.py

### DIFF
--- a/docs/context.rst
+++ b/docs/context.rst
@@ -1272,53 +1272,15 @@ The purpose of ObsFileDb is to provide a map into a large set of TOD
 files, giving the names of the files and a compressed expression of
 what time indices and detectors are present in each file.
 
-Organization of Files
-=====================
-
-The ObsFileDb is represented on disk by an sqlite database file.  The
-sqlite database contains information about data files, and the
-*partial paths* of the data files.  By *partial paths*, we mean that
-only the part of the filenames relative to some root node of the data
-set should be stored in the database.  In order for the code to find
-the data files easily, it is most natural to place the
-obsfiledb.sqlite file in that same root node of the data set.
-Consider the following file listing as an example::
-
-  /
-   data/
-        planet_obs/
-                   obsfiledb.sqlite     # the database
-                   observation0/        # directory for an obs
-                                data0_0 # one data file
-                                data0_1
-                   observation1/
-                                data1_0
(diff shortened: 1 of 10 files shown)


mhasself:
> One issue I've found is that way context is running the file loading with `resolve_paths=True` isn't going to work for the g3tsmurf databases/file system and I'm not actually sure that's the default behavior we want.

Yes, I see what you mean ... can you try the following work-around?  After loading context, set:
```
context.obsfiledb.prefix = ''
```
Then see if things behave as you might want.

The "prefix" idea was meant to help with moving obsfiledb between systems... but now I think it probably makes sense to use os.join(prefix, filename) instead of prefix+filename, so prefix is ignored if the filenames are absolute.

kmharrington:
Adding that line does fix the query / get me to the "loader doesn't support samples" error. (Which I think I'll wait to implement in a different PR after this one). 

The `%` in this query is wild? 
```
c = self.conn.execute(
                'select name, obs_id, detset, sample_start, sample_stop '
                'from files where name like ?', ('%' + basename, ))
```

Technically that could cause problems if we manage to get two `.g3` sessions started simultaneously (which is totally possible in the field). We'll need the second level with the `stream_id`. Is there a way to just look for and strip off any of the prefix instead of just use the file name?

mhasself:
> Technically that could cause problems if we manage to get two `.g3` sessions started simultaneously (which is totally possible in the field). We'll need the second level with the `stream_id`. Is there a way to just look for and strip off any of the prefix instead of just using the file name?

Yes, that is a wildcard, but it's ok if there are multiple matches, because we go on to check each match in fully resolved path.

kmharrington:
Ok cool, is adding a setup for context so we can input an obsfiledb prefix part of this PR or the next one?

mhasself:
The prefix issue should be fixed now.  We actually changed prefix a while ago so it uses os.path.join.  Latest commits fix the bug with resolve_paths=True and also updates the documentation to not be completely wrong.

kmharrington (approved):

PR #210, opened 2022-01-27
Use static polarization weights, not celestial weights

keskitalo:
This bug effectively made the HWPSS signal time-dependent. Many thanks to @Helbouha for reporting it!


Helbouha:
This looks good! No more time variations. Thanks a lot for the fix!

Helbouha (approved):

PR #211, opened 2022-01-28
AxisManager: save/load to HDF5 (experimental / debugging)

mhasself:
This adds .save() and .load() functions to AxisManager.  This can handle complicated data fields, such as RangesMatrix, other AxisManagers, and FlagManagers.

Note this does effectively provide save/load of RangesMatrix -- something we need for flag archives.

### DIFF
--- a/sotodlib/core/axisman_io.py
+++ b/sotodlib/core/axisman_io.py
@@ -93,7 +93,7 @@ def _safe_scalars(x):
         return tuple([_safe_scalars(_x) for _x in x])
     if isinstance(x, dict):
         return {k: _safe_scalars(v) for k, v in x.items()}
-    if isinstance(x, (np.integer, np.floating, np.str_)):
+    if isinstance(x, (np.integer, np.floating, np.str_, np.bool_)):
         return x.item()
     # Must be fine then!
     return x
--- a/sotodlib/core/axisman_io.py
+++ b/sotodlib/core/axisman_io.py
@@ -85,6 +85,19 @@ def _retype_for_read(data):
 
 # save/load AxisManager to HDF5.
 
+def _safe_scalars(x):
+    # Kill any np.integer, np.floating, or np.str_ vals... they don't serialize.
+    if isinstance(x, list):
+        return [_safe_scalars(_x) for _x in x]
+    if isinstance(x, tuple):
+        return tuple([_safe_scalars(_x) for _x in x])
+    if isinstance(x, dict):
+        return {k: _safe_scalars(v) for k, v in x.items()}
+    if isinstance(x, (np.integer, np.floating, np.str_)):
+        return x.item()
+    # Must be fine then!
+    return x
+
 def _save_axisman(axisman, dest, group=None):
     """
     See AxisManager.save.
@@ -136,6 +149,9 @@ def _save_axisman(axisman, dest, group=None):
                            'args': (v.name, v.count)})
         else:
             raise ValueError(f"No encoder for axis class: {v.__class__}")
+
+    # Sanitize ...
+    schema = _safe_scalars(schema)
 
     if isinstance(dest, str):
         f = h5py.File(dest, 'a')
--- a/sotodlib/core/axisman_io.py
+++ b/sotodlib/core/axisman_io.py
@@ -1,0 +1,234 @@
+import so3g
+import numpy as np
+import json
+import h5py
+
+from .axisman import *
+from .flagman import FlagManager
+
+# Flatten / expand RangesMatrix
+
+def flatten_RangesMatrix(rm):
+    """Given a RangesMatrix, flatten into a small number of arrays that
+    can be written to HDF or npy and then reconstructed later.
+
+    Returns:
+      Dict with arrays called 'shape', 'intervals', and 'ends'.
+    """
+    shape = rm.shape
+    if len(shape) == 2:
+        intervals = [r.ranges().reshape(-1) for r in rm.ranges]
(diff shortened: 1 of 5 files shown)


kmharrington:
So glad to see this! @jlashner and I were just talking about how nice it would be to have an option like this. 

Just tried it on an AxisManager I was already saving and ran into some sort of JSON error with int64. I've checked the AxisManager and the int64s are coming from our channel mapping info but all the other timestream is loading as float64s. I'm not sure if it's just ints that are the issue or if it's all 64 bit data types.

Here's the traceback: 
```
TypeError                                 Traceback (most recent call last)
<ipython-input-8-e652829f00d3> in <module>
----> 1 wn_aman.save('/data/users/kmharrin/ufm_cv4_analysis/dark_202105/wn_100mK.h5')

~/.local/lib/python3.7/site-packages/sotodlib-0.4.0+170.gd212ad8-py3.7.egg/sotodlib/core/axisman.py in save(self, dest, group)
    805         """
    806         from .axisman_io import _save_axisman
--> 807         return _save_axisman(self, dest, group)
    808 
    809     @classmethod

~/.local/lib/python3.7/site-packages/sotodlib-0.4.0+170.gd212ad8-py3.7.egg/sotodlib/core/axisman_io.py in _save_axisman(axisman, dest, group)
    148     dest.attrs['_axisman'] = json.dumps({
    149         'version': 0,
--> 150         'schema': schema,
    151         })
    152     scalars = {}

/opt/anaconda3/lib/python3.7/json/__init__.py in dumps(obj, skipkeys, ensure_ascii, check_circular, allow_nan, cls, indent, separators, default, sort_keys, **kw)
    229         cls is None and indent is None and separators is None and
    230         default is None and not sort_keys and not kw):
--> 231         return _default_encoder.encode(obj)
    232     if cls is None:
    233         cls = JSONEncoder

/opt/anaconda3/lib/python3.7/json/encoder.py in encode(self, o)
    197         # exceptions aren't as detailed.  The list call should be roughly
    198         # equivalent to the PySequence_Fast that ''.join() would do.
--> 199         chunks = self.iterencode(o, _one_shot=True)
    200         if not isinstance(chunks, (list, tuple)):
    201             chunks = list(chunks)

/opt/anaconda3/lib/python3.7/json/encoder.py in iterencode(self, o, _one_shot)
    255                 self.key_separator, self.item_separator, self.sort_keys,
    256                 self.skipkeys, _one_shot)
--> 257         return _iterencode(o, 0)
    258 
    259 def _make_iterencode(markers, _default, _encoder, _indent, _floatstr,

/opt/anaconda3/lib/python3.7/json/encoder.py in default(self, o)
    177 
    178         """
--> 179         raise TypeError(f'Object of type {o.__class__.__name__} '
    180                         f'is not JSON serializable')
    181 

TypeError: Object of type int64 is not JSON serializable
```

skhrg:
> Just tried it on an AxisManager I was already saving and ran into some sort of JSON error with int64.

The `json` library doesn't serialize some numpy data types. The `float64`s aren't a problem because `np.float64` is equivalent to the builtin `float` dtype, but `np.int64` and the builtin `int` dtype are not (I think because in python the `int` dtype doesn't have a fixed size but there may be more to it than that). 

The solution is just to cast the `int64`s to `int`s.

mhasself:
Ah, good find.  Would it bother anyone if we forced scalars to be simple python scalars, at wrap time?  That would limit the space of possibilities early on.  If you really need an int32 for some reason, you can store it in a shape (1,) numpy array.

kmharrington:
This isn't showing up with scalars. It's showing up with `int64` arrays. They definitely don't need to be int64s but I'm also not doing anything special when I make them: https://github.com/simonsobs/sotodlib/blob/master/sotodlib/io/load_smurf.py#L1673 

Since I'm not trying to do anything special here I expect this is an error that will show up for many other cases.

mhasself:
int64 arrays work fine; there's something odder going on... let's figure it out on slack.

mhasself:
Ok ... try latest version.  It cleans up everything (?) in the "schema" to be simple pythonic types.

sotodlib/core/axisman_io.py
```diff
@@ -85,6 +85,19 @@ def _retype_for_read(data):
...
 
+def _safe_scalars(x):
+    # Kill any np.integer, np.floating, or np.str_ vals... they don't serialize.
+    if isinstance(x, list):
+        return [_safe_scalars(_x) for _x in x]
+    if isinstance(x, tuple):
+        return tuple([_safe_scalars(_x) for _x in x])
+    if isinstance(x, dict):
+        return {k: _safe_scalars(v) for k, v in x.items()}
+    if isinstance(x, (np.integer, np.floating, np.str_)):
```
skhrg, line 96: Probably an edge case since off the top of my head I can't think of a use case for it but things that are `np.complex` and `np.bool_` won't get caught by this. May be worth adding them just to cover the full gamut.

You could also just change the condition to `hasattr(x, "item")` which should catch everything.

mhasself: Ah, I will add np.bool_, thanks.

For complex values, I think it's ok if they hit the JSON converter and an error is raised.  We'd need a custom packer/unpacker.  Nothing in schema should be complex (although currently you can do core.LabelAxis('nth_roots', [(-1+0j)**(1/n)**i for i in range(n)]) ... but please don't ...)

skhrg: Yeah, LabelAxis with complex labels is probably a bad enough idea that it throwing an error seems preferable.

PR #213, opened 2022-01-28
Fix RUID load from file and reorder status error handling

kmharrington:
Found an error when loading directly from files (no database) about the RUID handling. If we don't have a database we can't be sure of the readout unique IDs so we shouldn't load anything.

Also re-ordered the trys / messaging on the status load to quiet things down. This order removes a Warning message that was popping up during what should be default behavior when loading without a database.


jlashner (approved):
Looks good to me! Though I haven't tested or anything

Issue #216, opened 2022-02-01
G3tSmurf: update bias line names


kmharrington:
Related to https://github.com/simonsobs/sotodlib/issues/215 

The default g3tsmurf AxisManagers have integer labels for the bias lines, we need to turn them into them strings.

PR #222, opened 2022-02-10
LabelAxis should only accept string labels

### LINKED ISSUE #215: LabelAxis should only accept strings for labels
Be it resolved that LabelAxis should only accept string labels.  The use of integers (and floats, complexes, bools...) as labels is currently supported, to varying degrees, but should instead raise on error on instantiation.

skhrg:
Resolves #215 and #216

### DIFF
--- a/sotodlib/core/axisman.py
+++ b/sotodlib/core/axisman.py
@@ -191,6 +191,8 @@ class LabelAxis(AxisInterface):
     element has been given a unique name.  The vector of names can be
     found in self.vals.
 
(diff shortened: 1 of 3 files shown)


kmharrington (commented):
Thanks for catching the bias line mapping with this. Could we actually do something to make them more obviously strings? (instead of just numbers stored as strings). I'm thinking something like 
```
'b{:02d}.format( bias line number)
```

mhasself (approved):
Thanks!  Please merge unless @kmharrington has lingering concerns.

PR #226, opened 2022-02-14
SmurfStatus Revamp

### LINKED ISSUE #212: G3tSmurf : SmurfStatus Revamp
A couple changes need to be made to SmurfStatus:

(1) we need some sort of integrity checking to say we have found a complete enough status object. Right now we do a bit of that in `load_file` but I think it would be better if it was done at the creation of the status object. 

(2) AxisManagers can accommodate scalars now. This means it's much more straightforward to include the status in loaded smurf AxisManager. (Some things like the readout filter parameters would probably be quite useful to carry around so that later filter functions can deconvolve them). 

(3) Should we just make SmurfStatus an AxisManager?

skhrg:
Resolves #212

### DIFF
--- a/sotodlib/io/load_smurf.py
+++ b/sotodlib/io/load_smurf.py
@@ -15,15 +15,26 @@ from collections import namedtuple
 from enum import Enum
 
 import logging
+
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
 from .. import core
 from . import load as io_load
 
-from sotodlib.io.g3tsmurf_db import (Base, Observations, Tags, Files, Tunes, 
-                                     TuneSets, ChanAssignments, Channels, 
-                                     FrameType, Frames)
+from sotodlib.io.g3tsmurf_db import (
+    Base,
+    Observations,
+    Tags,
+    Files,
+    Tunes,
+    TuneSets,
+    ChanAssignments,
+    Channels,
+    FrameType,
+    Frames,
+)
+
 Session = sessionmaker()
 num_bias_lines = 16
 
@@ -35,24 +46,26 @@ Strings must be unique, in that they must only show up when they should be used
 as observations
 """
 SMURF_ACTIONS = {
-    'observations':[
-        'take_stream_data',
-        'stream_data_on',
-        'take_noise_psd',
-        'take_g3_data',
-        'stream_g3_on',         
+    "observations": [
+        "take_stream_data",
+        "stream_data_on",
+        "take_noise_psd",
+        "take_g3_data",
+        "stream_g3_on",
     ],
 }
- 
+
 
 # Types of Frames we care about indexing
-type_key = ['Observation', 'Wiring', 'Scan']
+type_key = ["Observation", "Wiring", "Scan"]
+
 
 class TimingParadigm(Enum):
     G3Timestream = 1
     SmurfUnixTime = 2
     TimingSystem = 3
     Mixed = 4
+
 
 def get_sample_timestamps(frame):
     """
@@ -72,7 +85,7 @@ def get_sample_timestamps(frame):
             Paradigm used to calculate timestamps.
     """
     logger.warning("get_sample_timestamps is deprecated, how did you get here?")
-    if 'primary' in frame.keys():
+    if "primary" in frame.keys():
         if False:
             # Do high precision timing calculation here when we have real data
             pass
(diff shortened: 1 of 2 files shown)


kmharrington:
I'm running into an error when I try to run `SmurfStatus.from_file` with this. Here's the traceback: 
```ufm_cv4_1645723985: 2022-02-24 17:33:05.169673 -> 2022-02-24 17:33:35.364360 [0:00:30.194687] (None)
  File "test_g3tsmurf_updates.py", line 23, in <module>
    SMURF.update_observation_files(obs,session, force=True)
  File "/home/kmharrin/.local/lib/python3.7/site-packages/sotodlib-0.4.0+213.gb466c07-py3.7.egg/sotodlib/io/load_smurf.py", line 703, in update_observation_files
    status = SmurfStatus.from_file(flist[0].name)
  File "/home/kmharrin/.local/lib/python3.7/site-packages/sotodlib-0.4.0+213.gb466c07-py3.7.egg/sotodlib/io/load_smurf.py", line 1494, in from_file
    return cls(status)
  File "/home/kmharrin/.local/lib/python3.7/site-packages/sotodlib-0.4.0+213.gb466c07-py3.7.egg/sotodlib/io/load_smurf.py", line 1404, in __init__
    self.aman.wrap("filter_a", self.filter_a, [(0, "filter_a")])
  File "/home/kmharrin/.local/lib/python3.7/site-packages/sotodlib-0.4.0+213.gb466c07-py3.7.egg/sotodlib/core/axisman.py", line 567, in wrap
    return self.merge(helper)
  File "/home/kmharrin/.local/lib/python3.7/site-packages/sotodlib-0.4.0+213.gb466c07-py3.7.egg/sotodlib/core/axisman.py", line 771, in merge
    raise ValueError(f'Key: {k} found in {self} and {aman}')
ValueError: Key: filter_a found in AxisManager(dump_frame, filter_a[filter_a], filter_a:IndexAxis(16)) and AxisManager(filter_a[filter_a], filter_a:IndexAxis(16))
```

It looks like we're trying to add an axis and a field with the same name? The file it should be running on: `/mnt/so1/data/chicago-latrt/timestreams/16457/ufm_cv4/1645723985_000.g3`

skhrg:
Ah I thought I fixed that but I guess I didn't commit it. It's a copy paste error where I forgot to change "filter_a" to "filter_b". Having the axis and field have the same name does not give me an error, though I suppose I shouldn't give them the same name.

skhrg:
Someone who knows more about how we store things may know this, but are `filter_a` and `filter_b` always the same length. They seem to be zero padded to the same length in every file I load but I don't know if that's just me getting lucky.

kmharrington:
@msilvafe  @jlashner  Do you happen to know the answer to @skhrg 's question? I've never looked deeply at it.

msilvafe:
> @msilvafe @jlashner Do you happen to know the answer to @skhrg 's question? I've never looked deeply at it.
The length of the a and b parameters are set by the filter order (i.e. 2 pole, 3 pole, etc) we usually use a 4th order filter so a + b should each be length 5 typically but this can change if we change the order. That filter_order should also be returned in the status information though. This is just an implementation of scipy.signal.butter and is described in the pysmurf smurf_processor readme: https://github.com/slaclab/pysmurf/blob/main/README.SmurfProcessor.md

skhrg:
Ah thanks that clears things up.

mhasself:
FYI (not sure it affects this work): there is already some functionality to handle this type of filter in tod_ops.filters.  See https://github.com/simonsobs/sotodlib/blob/32a1b707d0871807925f2a1fcf3894ff092835f0/sotodlib/tod_ops/filters.py#L371 .  

The "iir_params" argument is intended to help folks pass in a filter definition as a single argument (currently a (3,n) array).  Now that we have scalars, that interface can be changed so that iir_params accepts the sort of (sub-)AxisManager that you are constructing here.

skhrg:
I can certainly change things so that `iir_filter` interfaces nicely with this AxisManager. Since according to https://github.com/slaclab/pysmurf/blob/main/README.SmurfProcessor.md the filter is before the downsampling I would assume that here `fscale = downsample_factor`? 

On another note the docstring for `fscale` seems oddly worded
https://github.com/simonsobs/sotodlib/blob/32a1b707d0871807925f2a1fcf3894ff092835f0/sotodlib/tod_ops/filters.py#L380-L382
At first glance it makes `fscale` sound like the original frequency rather than just a scaling factor? Granted my knowledge of filters and z-transforms and whatnot is a bit rusty so maybe I'm off base here and `fscale` is a frequency?

mhasself:
> At first glance it makes `fscale` sound like the original frequency rather than just a scaling factor? Granted my knowledge of filters and z-transforms and whatnot is a bit rusty so maybe I'm off base here and `fscale` is a frequency?

Yes, you're right... it's not worded properly, and should indeed correspond to the downsampling factor!

kmharrington:
I don't think we should change `iir_filter` but I think instead we should add a second filter designed to take an axis manager with this status information and parse it to send to `iir_filter`.

skhrg:
> I don't think we should change `iir_filter` but I think instead we should add a second filter designed to take an axis manager with this status information and parse it to send to `iir_filter`.

`iir_filter` already had the ability to load the filter params from an axis manager. It just assumed all the parameters were in the same field. I just changed it to be able to load them from three separate fields as well (and set the default to load the ones I added in SmurfStatus).

skhrg:
@kmharrington is there anything else you would like me to do for this?

sotodlib/io/load_smurf.py
```diff
@@ -1810,139 +2063,172 @@ def load_file(filename, channels=None, samples=None, ignore_missing=True,
...
 
-    ## Build AxisManager
+    count = sum(map(len, streams["time"]))
+
+    # Build AxisManager
     aman = core.AxisManager(
         ch_info[det_axis].copy(),
-        core.OffsetAxis('samps', count, sample_start)
+        core.OffsetAxis("samps", count, sample_start),
+        status.aman,
```
kmharrington, line 2191: I think instead we want to use wrap to add the status aman as one piece. So after `aman = AxisManager(..)` do something like `aman.wrap("status", status.aman)`. The way this is added right now puts a lot of lower level status things "top level" in the axismanager and I'd rather we not do that.

kmharrington (changes requested):
Ahh sorry, missed that this was ready for another look. 

One requested change, then I'll approve it. Thanks a bunch!

kmharrington (approved):
Thanks a bunch!

PR #227, opened 2022-02-15
Docs workflow

mhasself:
Builds sphinx docs, in picky -W mode, mostly to catch autodoc fails when new dependencies are added.

### DIFF
--- a/.github/workflows/test-docs.yml
+++ b/.github/workflows/test-docs.yml
@@ -1,6 +1,7 @@
 name: "Test docs build"
 on: 
-- push
+  pull_request:
+    branches: [master]
 
 jobs:
   docs:
--- a/.github/workflows/test-docs.yml
+++ b/.github/workflows/test-docs.yml
@@ -7,8 +7,14 @@ jobs:
   docs:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/test-docs.yml
```diff
@@ -0,0 +1,14 @@
...
+
+jobs:
+  docs:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - run: |
+        cd docs/
+        pip install -r requirements.txt
+        sphinx-build -M html . _build -W
```
BrianJKoopman, line 14: This'll definitely work. For slightly nicer formatting when viewing the build logs I'd maybe suggest splitting up the steps and naming them:

```
    - name: clone sotodlib
      uses: actions/checkout@v2

    - name: install dependencies
      working-directory: ./docs
      run: pip3 install -r requirements.txt

    - name: build docs
      working-directory: ./docs
      run: sphinx-build -M html . _build -W
```

BrianJKoopman (approved):
Looks good! I have one small suggestion related to formatting the workflow logs.

.github/workflows/test-docs.yml
mhasself: Ah, working-directory ... I like it ...

PR #228, opened 2022-02-23
Adding first draft of single obs detdb

jseibert575:
Adding first draft of some code to generate a DetDb for a single observation. There's a lot of hard-coded values right now and need to make a variety of improvements to the code for modularity and user-friendly features, but the primary goal of generating a single-obs level DetDb seems to be working. I also don't think this should live in load_smurf but uses some of the functions defined within the module that are useful for now and I'm not sure what the longterm plan is for some of these.

### DIFF
--- a/sotodlib/io/load_smurf.py
+++ b/sotodlib/io/load_smurf.py
@@ -43,7 +43,6 @@ SMURF_ACTIONS = {
         'stream_g3_on',         
     ],
 }
- 
 
 # Types of Frames we care about indexing
 type_key = ['Observation', 'Wiring', 'Scan']
@@ -1176,7 +1175,14 @@ def make_DetDb_single_obs(obsfiledb, obs_id):
     detdb.conn.commit()
     return detdb
 
+def detdb_context_hook(ctx, obs_id):
+    ddb = make_DetDb_single_obs(ctx.obsfiledb, obs_id)
+    ctx.detdb = ddb
+    return ddb
 
+core.Context.hook_sets['obs_detdb_load'] = {
+    'before-load-obs': detdb_context_hook,
+}
 
 
 class SmurfStatus:
--- a/sotodlib/io/load_smurf.py
+++ b/sotodlib/io/load_smurf.py
@@ -1127,6 +1127,56 @@ def dump_DetDb(archive, detdb_file):
     session.close()
     return my_db
 
+def make_DetDb_single_obs(obsfiledb, obs_id, detdb_path):
+
+    # find relevant files to get status
+    c = obsfiledb.conn.execute('select name from files ' 
+                         'where obs_id=?' + 
+                         'order by start', (obs_id,))
+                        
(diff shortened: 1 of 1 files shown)


kmharrington:
This worked for me with some CalDbs except for some tweaks to the naming (which might be a slightly larger question than this, I just started following the first light pipeline ICD with the CalDbs I was testing but have previously called things ruids. tbh I'm indifferent on names but we should get them standardized). 

Apologies for just merging this into a different branch, yes I'm excited to have to untangle the git here later: https://github.com/simonsobs/sotodlib/blob/0d6d52b526004a28c17923d7466b33b9e79e555f/sotodlib/io/load_smurf.py#L1130 

I was able to use it like this: 
```
my_id = obs_list[2]['obs_id']
request = context.get_obs( obs_id=my_id, logic_only=True )
print(request['request'])

detdb = ls.make_DetDb_single_obs(context.obsfiledb, my_id, 'temp_det_db.db')
context.detdb = detdb
context.reload()

meta = context.get_meta(request['request'])
```

A couple notes / questions / paths for more development.

1. It is possible to get to the readout ids without ever opening a file. you can query tunesets from the observations table  (or perhaps its a query to the "association_obs" table) and then use the detsets association table to pull out the readout_ids/channel properties. I'm probably missing something but I don't think the core context obsdb / obsfiledb interfaces have this path (although if they don't I now think they should). But this is specific to g3tsmurf so I think it's legit to depend on that setup.
2. The "saving" step of adding the 880 channels here takes many seconds. This might be more for @mhasself  since you're just calling his functions but if we need to do this every time we want to load an observation it's too much overhead at the moment. There's probably a way to do this as a batch?
3. It would be cool to not have to pass in a file path, as I was changing the column names I learned I had to keep deleting the file in order make it work.

kmharrington:
So these new updates are now fast / nice. Is there any reason we can't merge this now? I am a fan.

jseibert575:
It’s been working fast for me. I’m happy to merge it and keep updating it just as I clean it up and add some more information. Working on PR now to build the loading directly into Context. 

I actually have one commit I’d like to add before merging, in that vein. I will commit that in one second.

jseibert575:
It’s been working fast for me. I’m happy to merge it and keep updating it just as I clean it up and add some more information. Working on PR now to build the loading directly into Context. 

I actually have one commit I’d like to add before merging, in that vein. I will commit that in one second.

jseibert575:
Sorry for the massive wave of commits, this branch was pretty far behind some commits that fixed a bunch of pep8 things that screwed things up i think? But those should be gone now

kmharrington (approved):

mhasself:
Let's clean up the commit history on this before merging.

jseibert575:
There's enough conflicts here that are all reformatting that I'm just going to close this PR and open a new one with a new branch based on the current version of master, with the updates.

PR #229, opened 2022-02-24
Jumpfind

skhrg:
Ported over jumpfinder from [LATR_Validation](https://github.com/simonsobs/LATR_Validation/) repo


kmharrington:
This one also just instantly threw an error for me. Here's what I ran:

```
context = Context('/mnt/so1/users/kmharrin/smurf_context/smurf_context.yaml')
obs_list = context.obsdb.query(f'timestamp > {dt.datetime(2021,11,23).timestamp()} and timestamp < {dt.datetime(2021,11,29).timestamp()}')

my_id = obs_list[2]['obs_id']
aman = context.get_obs(obs_id=my_id)

jumps = tod_ops.jumpfind(aman)
print(jumps)
```

Some sort of array shape error: 
```
ValueError                                Traceback (most recent call last)
<ipython-input-7-8d680d47d4bf> in <module>
----> 1 jumps = tod_ops.jumpfind(aman)
      2 print(jumps)

~/.local/lib/python3.7/site-packages/sotodlib-0.4.0+222.g8029563-py3.7.egg/sotodlib/tod_ops/jumpfind.py in jumpfind(tod, signal_name, axis)
    276     signal = getattr(tod, signal_name)
    277 
--> 278     return np.apply_along_axis(jumpfind_default_pars_recursive, axis, signal)

<__array_function__ internals> in apply_along_axis(*args, **kwargs)

/mnt/so1/shared/software/anaconda3/200627/lib/python3.7/site-packages/numpy/lib/shape_base.py in apply_along_axis(func1d, axis, arr, *args, **kwargs)
    400     buff[ind0] = res
    401     for ind in inds:
--> 402         buff[ind] = asanyarray(func1d(inarr_view[ind], *args, **kwargs))
    403 
    404     if not isinstance(res, matrix):

ValueError: could not broadcast input array from shape (22,) into shape (7,)
```

That list of observations I like to up there is a good set of data to test these one.

skhrg:
This is what I get for testing the wrapper on fake data. The problem here is that `np.apply_along_axis` is trying to make a 2d array, but if two detectors have a different number of jumps then it gets this error. I'll just swap it out for a for loop.

skhrg:
I also didn't realize just how slow this was once it scales from our data to yours. I'll see if I can throw some optimizations in there at some point.
Also, thanks for this chunk of data. When I tried testing on LATRt data before I could only find stuff without jumps which wasn't super useful.

kmharrington (changes requested):
One inline comment. 

Overall: 

1. Messing around with this, it does seem to do a pretty decent job finding jumps when there are jumps and distinguishing them from glitches. When the data is overall behaving and there are just a subset of channels doing wonky things, it looks like it works well and I could see this going into a jump finder.

2. I do see some examples where this scales pretty badly. Ex: three minutes of particularly bad data where it took almost 10 minutes to run on it. (obs_id 'ufm_cv4_1637698295' connected to the context above). 

3. Related to 2, I think we need a way to pass all the possible parameters into the highest level function  (and include some sort of escape "stop if really bad" option). Our glitch flag function does follow the sotodlib default way of passing in lots of parameters: https://github.com/simonsobs/sotodlib/blob/b740669f84a0838fea9c1e16597d96942fffd6f4/sotodlib/flags.py#L43

4. The List of indices are basically ranges/intervals. I think I'd rather this return a Ranges matrix like our other flag type operations. You can copy that and the flag merging / over-writing directly out of glitch flags function.

5. I'm basically 100% sure this sort of algorithm is going to be the lowest hanging fruit for speedups. Lots of array manipulations / I'm already commenting on it being slow for our data. I don't think that should necessarily stop us from merging this version but it is making me hesitate. @mhasself  I'm curious about your thoughts here? It's definitely running a lot in parallel (at least from watching it take over all of simons1's cores while it was running).

skhrg:
> * do see some examples where this scales pretty badly. Ex: three minutes of particularly bad data where it took almost 10 minutes to run on it. (obs_id 'ufm_cv4_1637698295' connected to the context above).
> 
> * Related to 2, I think we need a way to pass all the possible parameters into the highest level function  (and include some sort of escape "stop if really bad" option)

Yeah the easiest "optimization" to make is to set a limit on the depth of the recursion. Since it keeps going until the chunks are either too small or it can't find anymore jumps, crappy data ends up going to really high recursion depths. It may be worth setting a limit on this (probably as a function of tod length) at which point it gives up and you should probably just throw that stretch of time out (or do this based on total jumps found instead of recursion depth).

As far as other speedups go, this is trivially parallelizable since I can split the tod up by detectors (or in time domain and just have the peices overlap so that I don't miss jumps on the edges). I'm not sure if there if sotodlib already has some speicifc way that I should MPI things but if its just mpi4py then its easy enough for me to throw in.

Once we're happy with the jumpfinding abilities I can also throw together a C implementation pretty easily if need be, but since most of the heavy lifting is currently being done by scipy (which I think is mostly making calls to C libraries under the hood anyways) I'm not sure how much of an improvement that gets us.

The other changes you requested are easy enough, I'll get them out in the next few days.

kmharrington (approved):
This data doesn't have the largest 1/f but I am impressed with how well it works at finding jumps. Thanks for getting the formatting matching. 

image

I'm good with merging this current setup but would like to hear from @mhasself about git organization / maybe squashing before you actually hit merge.

I do have a future desire / question about adding this into larger pipelines. Do the heights of the jumps as calculated by the peak finder match the raw timestreams? Could those be fed into a jump fixer? If so, I'd like request future work to track them through into a sparse array like I recently set up with the glitch finder. If it gets into the format it will be quite easy to get them saved into CalDbs (we've been working on that infrastructure).

kmharrington:
Thanks! If we can't pass them directly then let's have separate functions for it. (and definitely a future PR). We want to have options that will work with and without the HWPSS and jumps are definitely more difficult to fix with the HWPSS in there.

skhrg:
2pi jumps makes sense. I can certainly make a matched filter for them and check to see if it finds ones this one misses (or if we can get away with just this one). This also means that when jumpfixing if the calculated jump size is close to a factor of 2pi it is probably safe to assume that it is that factor of 2pi which should make things easier.

sotodlib/tod_ops/jumpfind.py
```diff
@@ -0,0 +1,406 @@
...
+    jumps = np.array([], dtype=int)
+    for i in range(len(x) // (window_size - overlap)):
+        start = i * (window_size - overlap)
+        end = np.min((start + window_size, len(x)))
+        _jumps = jumpfinder(x[start:end], **kwargs) + start
+        jumps = np.hstack((jumps, _jumps))
+    return np.unique(jumps).astype(int)
+
+
+def jumpfind(tod, signal=None, buff_size=10, jumpfinder=jumpfind_tv, **kwargs):
```
mhasself, line 369: It looks like this is supposed to be the main user interface function.  But using it is awkward, because the user needs to specify a particular jumpfinder to use, and then also needs to look up the specific kwargs that that jumpfinder accepts, and tune the parameters for their case of interest.  The interface for jumpfinder functions is totally free right now (it's not even true that they all accept the "min_size" parameter), so it's not possible to do simple experiments where you individually adjust parameters in the jumpfind call (such as jumpfinder=...) to compare results from different jumpfinders.

This means there's not much value added by this version jumpfind function -- in fact it's almost more work to use jumpfind than to use a jumpfinder directly.

I think you should either remove this main entry point, and have the 2-D support & RangesMatrix conversion supported in each individual jumpfinder OR else make jumpfind a little more powerful and general:

* Allow user to pass jumpfinder function by string name (e.g. "tv")
* Have the function accept some basic parameters that control the jumpfinders.  For example, "min_size" is always something the user should be able to specify.  I think you should also have a "min_sigma", which will lead to a min_size computed based on the stdev (appropriately estimated).  Another basic parameter would be the time scale for a jump (how many samples does a particular jump take to occur -- this can used to tease out true glitchy jumps from bright signals that are a little slower, such as planet signals; in code the time scale would be converted to a sample count and used to set windowing and filtering parameters).
* In the case that jumpfinder is specified by name, compute reasonable default parameters to pass to that jumpfinder, _that are based on the data_ and those general basic parameters.  "Based on the data" just means that it can figure out stdev on its own, it doesn't need the user to pass in stdev or min_size all the time.  The user can override the jumpfinder arguments using kwargs ... but jumpfind should provide sensible defaults.

In simple cases, I should get more or less the same answer.  E.g. with just large steps in white noise, these two should both work and give me the ~same answers:
```
  j1 = jumpfind(tod, jumpfinder='tv')
  j2 = jumpfind(tod, jumpfinder='gaussian')
```

sotodlib/tod_ops/jumpfind.py
```diff
@@ -0,0 +1,406 @@
...
+    Find jumps in tod.signal_name.
+    Expects tod.signal_name to be 1D of 2D
+
+    Arguments:
+
+        tod: axis manager
+
+        signal: Signal to jumpfind on. If None than tod.signal is used.
+
+        buff_size: How many samples to flag around each jump in RangesMatrix
```
mhasself, line 380: The choice of 10 for the default seems very arbitrary.  I think this should be off by default, so the user just gets single-sample jump locations, or else the default should be to have it computed based on other function parameters (such as the aforementioned "time_scale" for filtering/windowing).

tests/test_tod_ops.py
```diff
@@ -140,5 +140,28 @@ def test_basic(self):
...
+        jump_locs = np.array([200, 400, 700])
+        sig_jumps[jump_locs[0]:] += 5
+        sig_jumps[jump_locs[1]:] -= 10
+        sig_jumps[jump_locs[2]:] -= 5
+
+        tod.wrap('sig_jumps', sig_jumps, [(0, 'samps')])
+
+        # Find jumps, using a conservative min_size here because the jumps
+        # are large and we want to avoid false positives
+        jumps = tod_ops.jumpfind(tod, signal=tod.sig_jumps, min_size=2)
```
mhasself, line 157: Calling this "conservative" is a but confusing.  I think it's not surprising that min_size should be a few times the white noise level (which is 1 here) to avoid false positives.

(This is another reason a min_sigma argument will be a little more powerful / less confusing.)

tests/test_tod_ops.py
```diff
@@ -140,5 +140,28 @@ def test_basic(self):
                                        detrend='linear')
         self.assertEqual(sig1f.shape, tod['sig1d'].shape)
 
+class JumpfindTest(unittest.TestCase):
+    def test_jumpfinder(self):
```
mhasself, line 144: I think this only tests the tv jumpfinder... please test the gaussian one too.

sotodlib/tod_ops/jumpfind.py
```diff
@@ -0,0 +1,406 @@
...
+
+    # Mean subtract the data
+    _x = x - x.mean()
+
+    # Scale data to have std of order 1
+    if _x.std() > 10:
+        _x = _x / (10 ** (int(np.log10(_x.std()))))
+
+    # Take cumulative sum
+    x_step = np.cumsum(_x)
```
mhasself, line 55: Add to comment that this is equivalent to convolution with a step function, as one might do in a matched filter (or something to that effect).  I think you should also move the mean removal closer, since it's essential to this method's success.

(Thanks for the off-line explanation of this!)

mhasself (changes requested):
General notes:
- There's lots of good stuff in the docstrings, and they mostly render properly in sphinx.  But please try to get them into either numpy or google format, so they also look good (and consistent other sotodlib files) in raw text.
- Some recurring spelling issues: gaussain, negetive, smalled... please track down all and fix.
- In docstrings, if you start a phrase with a capital letter then it is a sentence and it should end in a period.
- In docstrings, a line break does not cause a line break in the sphinx output; use a period and fill the paragraph with appropriate indentation.

My main request is for a bit more power in the jumpfind interface; see long inline comment.

Soft suggestion on nomenclature: "jumpfind" is a new compound verb; consider using find_jumps(...) instead of jumpfind(...) along with "jumpfinder_XYZ" instead of "jumpfind_XYZ" for the support functions?  I don't want to stifle your artistic license too much... but verb_noun is a popular function name recommendation in style guides.

sotodlib/tod_ops/jumpfind.py
```diff
@@ -0,0 +1,406 @@
...
+           Note that x will by modified in place with jumps removed
+           pass x.copy() to preserve x (say to pass to a better jump fixer)
+
+        min_chunk: The smallest chunk of data to look for jumps in
+
+        min_size: The smalled jump size counted as a jump
+
+        win_size: Number of samples to average over when checking jump size
+
+        max_depth: The maximum recursion depth, set negetive to not use
```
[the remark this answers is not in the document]
skhrg: I meant "to not use" as in to not use a max depth. I'll clear that up.

sotodlib/tod_ops/jumpfind.py
```diff
@@ -0,0 +1,406 @@
...
+    # If std is basically 0 no need to check for jumps
+    if np.isclose(x.std(), 0.0):
+        return np.array([], dtype=int)
+
+    # Mean subtract the data
+    _x = x - x.mean()
+
+    # Scale data to have std of order 1
+    if _x.std() > 10:
+        _x = _x / (10 ** (int(np.log10(_x.std()))))
```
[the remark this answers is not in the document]
skhrg: It's because having several large jumps can skew the std high. There are tod's where the std in the not so jumpy regions are <1 but the tod as a whole is something closer to 5 or 6. I actually don't know if this is needed. When I did some testing where I rescaled things to have higher noise level (ie: white noise with std of 50 instead of 1) this helped find some jumps that were below the noise level. For real data I don't know if we need this since we shouldn't have anything with stds that high (I hope).

sotodlib/tod_ops/jumpfind.py
```diff
@@ -0,0 +1,406 @@
...
+    """
+    Apply gaussain filter to data and then search for jumps
+
+    Arguments:
+
+        x: Data to jumpfind on, expects 1D
+
+        sigma: Sigma of gaussain kernal
+
+        order: Order of gaussain filter
```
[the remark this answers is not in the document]
skhrg: I agree, the higher order stuff is vestigial from when I was playing around with them.

sotodlib/tod_ops/jumpfind.py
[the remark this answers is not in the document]
mhasself: Rescaling by the stdev makes sense.  My point was that snapping to powers of 10 is not doing anything other than potentially discretizing the behavior.  Why not just:
```
if _x.std() != 0:
  _x /= _x.std()
```

sotodlib/tod_ops/jumpfind.py
skhrg: 10 was just because typically the slop in the position of the jump is less than 10 samples. But this uncertainty is mostly only an issue in super jumpy regions (where jumps right next to each other aren't always distinguished). It's also less of an issue with the TV filtering, the gaussian adds some width to the jumps when it smooths so you can get some additional error.

sotodlib/tod_ops/jumpfind.py
skhrg: Yeah. The main reason why I resisted having each function support 2d and ranges previously was because when testing its was more convenient to avoid all of that (and axis managers) so that I could quickly make some noise + jumps and check how well the output array of jumps matched the input. So maybe the correct thing would be to do make them all support it. I do like having some sort of wrapper for the other reason of it avoiding having to copy the same sort of code at the start and end of every function. But I do agree it is confusing to use if you don't already know all the kwargs for the jumpfinders (which for obvious reasons was not an issue for me).

Making it more powerful and general sounds good but I'll have to think a bit on how to deal with the differing kwargs in a way thats not just as annoying. Maybe I want to add more power to the wrapper and then just add an additional function for each jumpfinder that calls the wrapper but with the jwargs specified? It adds function bloat to the file but then at least the arguments will show up when people `?` the functions in an interactive shell/notebook and show up in autocomplete for people whose editors are configured for that.

> Allow user to pass jumpfinder function by string name (e.g. "tv")

I have mixed feelings on this. I understand the convenience from a user perspective, but these sorts of "select the mode with a string" parameters always add the need for a set of if statements at the start of the function that needs to be changed whenever a mode is added/deleted which feels clunky to me vs just passing a function.

> Have the function accept some basic parameters that control the jumpfinders. For example, "min_size" is always something the user should be able to specify. 

This makes sense.

> I think you should also have a "min_sigma", which will lead to a min_size computed based on the stdev (appropriately estimated). 

The problem here is that std is effected by jumps. So if you do this and have big jumps it can skew the std enough to drop even mid sized jumps below your threshold. One way around this (which one of the jumpfinder tries to do), is to recompute the std and the stuff based on it at each recursion, the downside of this is it tends to find less jumps per recursion to it takes more iterations to complete. 

> Another basic parameter would be the time scale for a jump (how many samples does a particular jump take to occur -- this can used to tease out true glitchy jumps from bright signals that are a little slower, such as planet signals; in code the time scale would be converted to a sample count and used to set windowing and filtering parameters).

Yeah, this is something I have toyed around with in the past. I didn't find a great way to determine the time scale of the jump. Part of what complicates this is that when you zoom in on them a good chunk of jumps are actually multiple jumps happening in quick succession. But its also not something I spent a ton of time on so I can certainly revisit it. 

> In the case that jumpfinder is specified by name, compute reasonable default parameters to pass to that jumpfinder, that are based on the data and those general basic parameters. "Based on the data" just means that it can figure out stdev on its own, it doesn't need the user to pass in stdev or min_size all the time. The user can override the jumpfinder arguments using kwargs ... but jumpfind should provide sensible defaults.

This also seems reasonable. I have a function that tries to do this for the gaussian jumpfinder that typically works better than me specifying params by default. But the issue with this (as I stated above) is the way that jumps skew the std.

>  In simple cases, I should get more or less the same answer. E.g. with just large steps in white noise, these two should both work and give me the ~same answers:

Yeah, as far as I can tell this is true. You typically only see the difference in their ability to (not) flag sources and in more jumpy regions. I actually don't know if there is a case where the gaussian one outperforms the TV one, I just left it in just in case we do find such a case down the line.

mhasself: > > I think you should also have a "min_sigma", which will lead to a min_size computed based on the stdev (appropriately estimated).
> 
> The problem here is that std is effected by jumps. So if you do this and have big jumps it can skew the std enough to drop even mid sized jumps below your threshold. One way around this (which one of the jumpfinder tries to do), is to recompute the std and the stuff based on it at each recursion, the downside of this is it tends to find less jumps per recursion to it takes more iterations to complete.

Yes, "appropriately estimated" was meant to get at that.  Here's a trick I often use to get a white noise stdev estimate, in the presence of 1/f and jumps/glitches:
```
# Get difference vector.
delta = np.diff(signal)
# Find ~1 sigma limits
lims = np.quantiles(delta, [0.159, 0.841])
# Convert to equivalent signal stdev
sigma = (lims[1] - lims[0]) / 8**.5
```
That's robust to outliers.

skhrg: Ah, that's clever. I'll give it a shot. Seems like something that should maybe exist as a function somewhere in sotodlib, this is likely not the only place where a robust estimate of the noise level (but not a full calculation of the white noise level) would be useful.

skhrg: @mhasself would you mind taking a look at some of the changes I've made to this in terms of making it more general. I've only done a bit but I want to make sure that I'm moving in a reasonable direction (I always have a hard time judging how bad my interfaces are since having written them using them is easy for me).

I'll throw in some stuff with determining parmeters based on stdev and min_sigma and whatnot in a bit. The initial testing I have done with them have gone well so I don't foresee any problems there.

mhasself: Yes, I like the changes!

Couple of notes as you continue ... :  I think it's considered good practice to pass parameters in key=value form, especially when there's a chance that the interface here will be changed or extended.  So instead of:
```
           jumps = jumpfinder(
                _signal, min_chunk, min_size, win_size, max_depth, **kwargs
            )
```
you should probably use
```
           jumps = jumpfinder(
                _signal, min_chunk=min_chunk, min_size=min_size, win_size=win_size, max_depth=max_depth, **kwargs
            )
```

Also, you say in a comment that kwargs have to be passed in a certain order ... they do not, because they are kwargs.  The function will pull out any keys it recognizes, and leave the rest in its own kwargs.

skhrg: > Couple of notes as you continue ... : I think it's considered good practice to pass parameters in key=value form, especially when there's a chance that the interface here will be changed or extended. So instead of:

Sure. 

> Also, you say in a comment that kwargs have to be passed in a certain order ... they do not, because they are kwargs. The function will pull out any keys it recognizes, and leave the rest in its own kwargs.

Ah that's right. For some reason I was under the impression that if you pass one that it doesn't recognize first it will assume the ones that come after are also part of the kwargs. But I guess its a dict so the ordering doesn't matter.

skhrg:
@mhasself do you mind taking another look at this? I think I got most of the stuff you requested. The one thing that I haven't got figured out is determining parameters based on an input time scale, but I think it might be better to just save that for a future PR.

Also, sorry about the messy commit history. I'll do a round of squashing before I merge.

mhasself (approved):
I like the new tests.

Nice bit of work!

Issue #224, opened 2022-02-11
G3tSMuRF : Observation updating strategy 


kmharrington:
We just recently hit two different issues with trying to find the end of an Observation. 

Issue one is more straightforward, I've been using updated to the `'AMCc.SmurfProcessor.FileWriter.IsOpen'` key to decide if an observation ends but it could be that one or `'AMCc.SmurfProcessor.SOStream.open_g3stream'`. So we need to also watch for that one. 

Issue two might be more complicated I always just look at the last file available in a g3session to check for the stream ending information. But we hit an edge case (in `/mnt/so1/data/chicago-latrt/timestreams/16445/ufm_cv4/1644597953_00X.g3`) where there is one single scan frame in the last file and all the stream ending frames are in the second to last file. I'm assuming this is something that just "can happen" with how our g3 files are made so we're going to need to be able to deal with it.

jlashner:
Yeah, stream ending in general is a bit tricky. After the end_stream flag is set using either of the two registers mentioned above, the stream will be held open until scan frames stop (and are replaced with idle frames) so that we are sure we've saved any queued / bufferred data.

I think technically the safest way to find the end of an observation is by filename, all files with a given session-id will always correspond to the same observation. If you'd like we can modify the g3 file-writer to add an EndObservation frame or something when a file is closed due to the end of an observation.

kmharrington:
oh that last option sounds the best. I moved away from just looking at files because it's tricky to do correctly when the database is updating while data is still coming in. So, I've been leaving `obs.stop = None` until I get an indication that the observation is complete. I think explicitly ending the whole file is my favorite option.

jlashner:
Cool, I feel like this should be an `Observation` frame but I think that would mess up some of the other G3tSmurf archiving functions.

kmharrington:
I think it might mean we need to change how we handle `Observation` frames in two places but I don't think that means this shouldn't be an `Observation` frame. If this update increases the version number of the .g3 files types then it should be pretty straightforward to work it out?

kmharrington:
I've just noticed that at least some of our dump frames have: 
```
'AMCc.SmurfProcessor.FileWriter.IsOpen': False
'AMCc.SmurfProcessor.SOStream.open_g3stream': 1
```

What's the difference between these two?

PR #230, opened 2022-02-26
add more logic and options to find end of observations

kmharrington:
Adding more logic to finding the end of observations. Related to issue https://github.com/simonsobs/sotodlib/issues/224 

This should fix all the inconsistencies/issues we've been seeing the the LATRt data.

### DIFF
--- a/sotodlib/io/load_smurf.py
+++ b/sotodlib/io/load_smurf.py
@@ -96,6 +96,35 @@ def get_sample_timestamps(frame):
                           for t in frame['data'].times()])
         return times, TimingParadigm.G3Timestream
 
+def _file_has_end_frames(filename):
+    ended = False
+    
+    ## Search through file looking for stream closeout
+    reader = so3g.G3IndexedReader(filename)
+    logger.debug(f"Reading {filename} to look for observation end.")
+    while True:
+        frames = reader.Process(None)
+        if not frames:
+            break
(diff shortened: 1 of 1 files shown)


sotodlib/io/load_smurf.py
```diff
@@ -96,6 +96,35 @@ def get_sample_timestamps(frame):
                           for t in frame['data'].times()])
         return times, TimingParadigm.G3Timestream
 
+def _file_has_end_frames(filename):
+    ended = False
+    
+    ## Search through file looking for stream closeout
+    reader = so3g.G3IndexedReader(filename)
+    logger.debug(f"Reading {filename} to look for observation end.")
+    while True:
```
jseibert575, line 105: You know significantly more about the file structure than me, so you probably know this loop will always terminate, but just making sure these conditionals will catch every possible outcome.

jseibert575 (approved):
Looks good! Did not go line by line, but understand the general process. Only comment is making sure that loop will definitely terminate every time the function is called, but think it's fine to merge and disentangle any bugs that show up in actual use instead of thinking about hypotheticals.

sotodlib/io/load_smurf.py
kmharrington: yeah, it will terminate at the end of the file no matter what

Issue #231, opened 2022-03-03
Add ManifestDb uniqueness constraints


kmharrington:
(I'll note that I'm not totally sure about the in between steps on how the `ManifestDb` tables are made but I can see the issue from inspecting the tables directly in sqlite)

Right now we can accidentally run `ManifestDb.add_entry` over and over with the exact same parameters. And when there are two identical entries in the database the assertions in the metadata loading fail. 

Can we figure out a way to add unique constraints on the schema definition or could we add some integrity checking to the `add_entry` function? It's not immediately obvious to me how to do it with the code, I think because I'm only partially following how the code builds the tables.

PR #232, opened 2022-03-04
metadata: ManifestDb throws or replaces duplicate index entries

mhasself:
Addresses #231 .

Note it can't protect you from other kinds of invalid data, such as overlapping ranges.


kmharrington (approved):
Works for me, thanks!

PR #237, opened 2022-03-14
Handle scalars properly when making copy of AxisManager

### LINKED ISSUE #236: When loading data with load_smurf, cannot run `aman.restrict` on returned axis manager
When loading with sotodlib.io.load_smurf, either using `load_data` or `load_file`, the returned axis manager cannot successfully run `aman.restrict`. This causes a crash when using `load_data`, [here](https://github.com/simonsobs/sotodlib/blob/d33e5224e0721bdc69c4af9f1fd22961bff21550/sotodlib/io/load_smurf.py#L1256), but the same error persists when data is successfully loaded from a g3 file, and then I try to run the `aman.restrict` line manually. 

The bug seems to be directly related to the `aman['status']` sub axis manager, when the code in `aman.restrict` tries to run [this line](https://github.com/simonsobs/sotodlib/blob/d33e5224e0721bdc69c4af9f1fd22961bff21550/sotodlib/core/axisman.py#L721). Because the thing stored in `aman['status']` is now an axis manager, it tries to run `aman['status'].copy()` which gets to [this line](https://github.com/simonsobs/sotodlib/blob/d33e5224e0721bdc69c4af9f1fd22961bff21550/sotodlib/core/axisman.py#L288), and fails because the first key/value pair in `aman['status'] is `dump_frame` and then either a `bool` or a `None`, which has no method `.copy()`. 

I'm not sure if this is a bug in the way the new SmurfStatus object/axis manager is created (@skhrg ) or something that axis managers are more generally expecting (@mhasself ), i.e. that the logic in the way axis managers are constructed is incorrectly assuming something about the structure of any axis managers passed to `aman.restrict`, so I'm not sure what the correct path to fix this is, but I think it should definitely be fixed somewhere. 

The error can be reproduced by running the following code, if one is so inclined:
```
import sotodlib.io.load_smurf as ls
aman = ls.load_file('/mnt/so1/data/ucsd-k2so/timestreams/16456/ufm_mv11/1645644498_004.g3')
aman['status'].copy()
```

OR:

```
import sotodlib.io.load_smurf as ls
aman = ls.load_file('/mnt/so1/data/ucsd-k2so/timestreams/16456/ufm_mv11/1645644498_004.g3')
aman.restrict("samps", (0, 120014))
```

Let me know if I can give any more details, or answer any questions.

skhrg:
Fixes #236

### DIFF
--- a/sotodlib/core/axisman.py
+++ b/sotodlib/core/axisman.py
@@ -285,7 +285,7 @@ class AxisManager:
         if axes_only:
             return out
         for k, v in self._fields.items():
-            if np.isscalar(v):
+            if np.isscalar(v) or v is None:
                 out._fields[k] = v
             else:
                 out._fields[k] = v.copy()
--- a/sotodlib/core/axisman.py
+++ b/sotodlib/core/axisman.py
@@ -285,7 +285,10 @@ class AxisManager:
         if axes_only:
             return out
         for k, v in self._fields.items():
-            out._fields[k] = v.copy()
+            if np.isscalar(v):
+                out._fields[k] = v
+            else:
+                out._fields[k] = v.copy()
         for k, v in self._assignments.items():
             out._assignments[k] = v.copy()
         return out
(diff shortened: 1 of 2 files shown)


mhasself (changes requested):
Thanks for the investigation!

Can you add something to the tests for this?

mhasself (approved):
Perfect, thanks!

Issue #234, opened 2022-03-09
ChanAssignments in G3tSmurf Table only including band 7


jseibert575:
For some reason, the most recent channel assignment in the tuneset we're currently using is only including the band 7 channel assignments. I think I've figured out it comes from these lines (https://github.com/simonsobs/sotodlib/blob/ac3545d5f3d73e1d5f665e039aaf9dad25c265b5/sotodlib/io/load_smurf.py#L843) in the updating lines to add to the ChanAssignments table, but I'm not sure if this has been a problem for anyone else with other indexes, nor why we haven't seen this before. I'm going to write a workaround but don't want to implement any bugfix until I figure out if this is just an issue with the way we're updating our index now, since this is the first (and only) tuneset copied over since the SupRSync bug was solved that wasn't copying tunesets, so it might be related.

 @kmharrington probably would have seen this the most, so tagging here.

kmharrington:
Oh I know why that line of code is in there but don't know why it's now causing you issues. I put it in there because pysmurf started making two sets of channel assignments every time we ran setup notches. It had something to do with trying to speed up the tuning. But, I didn't want to make 2x2000 channels when it was a pretty good bet that 2000 of them would never be used. 

I'm guessing the folder setup has changed to put more than just one band's channels assignments in a folder? 

A quick bug fix is to just delete the line and accept lots of extra channels. A longer elegant fix might be to combing the tune+channel assignment indexing so we only index channels assignments that are called out in a tune file.

jseibert575:
Can we check later ctime per band? Like check the band name in the file name, which should generate two files per band, and then just take the later of the two?

kmharrington:
yes that's also an option. Will have to add more logic to that function though because right now it assumes that there is only one meaningful channel assignment per folder.

PR #238, opened 2022-03-15
Bugfix for multiple channel assignments in the same folder

jseibert575:
This closes https://github.com/simonsobs/sotodlib/issues/234, by making sure that if there are channel assignments for multiple bands in a given action folder, it will make sure to add one for each band to the ChanAssignments Table in the archive. I checked that this works by using:

```
SMURF = ls.G3tSmurf('/mnt/so1/data/ucsd-k2so/timestreams',
               meta_path='/mnt/so1/data/ucsd-k2so/smurf',
               db_path='/home/jseibert/indexes/test_fix_k2so.db')

am = ls.load_file('/mnt/so1/data/ucsd-k2so/timestreams/16456/ufm_mv11/1645644498_004.g3',archive=SMURF)
```

The returned axis manager has all the band, channel info necessary in `am.ch_info`, which was not the case before, and there are no errors where information for different channels are failing to load.

### DIFF
--- a/sotodlib/io/load_smurf.py
+++ b/sotodlib/io/load_smurf.py
@@ -958,12 +958,12 @@ class G3tSmurf:
                     # decide if this is the last channel assignment in the directory
                     # needed because we often get multiple channel assignments in the same folder
                     root = os.path.join("/", *path.split("/")[:-1])
-                    cha_times = [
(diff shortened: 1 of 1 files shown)


kmharrington (approved):
This is making me want to totally change around the updating but this is definitely the most complete short-term fix.

PR #239, opened 2022-03-18
Adding functions to generate a detdb from an obs_id and load via context

jseibert575:
Because the branch was far enough behind, and the source file was edited (mostly for pep8 compliance) the rebasing was causing difficulty and making the commit history very dirty. The discussion about this PR can be seen here https://github.com/simonsobs/sotodlib/pull/228, but otherwise this is ready to be merged.

### DIFF
--- a/sotodlib/io/load_smurf.py
+++ b/sotodlib/io/load_smurf.py
@@ -1394,9 +1394,9 @@ def make_DetDb_single_obs(obsfiledb, obs_id):
     detdb.conn.commit()
     return detdb
 
-def detdb_context_hook(ctx, obs_id):
+def obs_detdb_context_hook(ctx, obs_id):
     ddb = make_DetDb_single_obs(ctx.obsfiledb, obs_id)
-    ctx.detdb = ddb
+    ctx.obs_detdb = ddb
     return ddb
 
 core.Context.hook_sets['obs_detdb_load'] = {
--- a/sotodlib/io/load_smurf.py
+++ b/sotodlib/io/load_smurf.py
@@ -1394,15 +1394,6 @@ def make_DetDb_single_obs(obsfiledb, obs_id):
     detdb.conn.commit()
     return detdb
 
-def obs_detdb_context_hook(ctx, obs_id):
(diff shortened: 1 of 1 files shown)


skhrg:
PEP8 stuff was my fault sorry about that. My linter was complaining enough when I was doing SmurfStatus stuff that I autoformatted it to get rid of them. I probably shouldn't have done that to a file with so much active development.

sotodlib/io/load_smurf.py
```diff
@@ -1309,6 +1309,100 @@ def dump_DetDb(archive, detdb_file):
...
+        detdb.get_id(ruids[i])
+        detdb.add_props('base',ruids[i],readout_id=ruids[i],smurf_band=ch['band'].item(),
+                        smurf_channel=ch['channel'].item(),res_frequency=ch['freqs'].item(),commit=False)
+
+    detdb.conn.commit()
+    return detdb
+
+def obs_detdb_context_hook(ctx, obs_id):
+    ddb = make_DetDb_single_obs(ctx.obsfiledb, obs_id)
+    ctx.obs_detdb = ddb
```
kmharrington, line 1399: have you tested this one? I had needed to run ctx.reload after setting the detdb to get it to work when I was testing this out.

kmharrington (approved):

sotodlib/io/load_smurf.py
```diff
@@ -1309,6 +1309,100 @@ def dump_DetDb(archive, detdb_file):
...
+    return ddb
+
+core.Context.hook_sets['obs_detdb_load'] = {
+    'before-load-obs': detdb_context_hook,
```
kmharrington, line 1403: also the docs are failing here (obs_detdb_context_hook vs detdb_context_hook)

sotodlib/io/load_smurf.py
jseibert575: I think I'm going to pull these lines for now from the PR, and include them in the PR that updates Context directly to handle loading this straight in to a context object, and let this PR focus on just building the detdb.

PR #240, opened 2022-03-21
Add IO interfaces for sparse arrays

kmharrington:
This is part of a larger project I'm working on for creating descriptive caldbs off of observations but I'm breaking this out for easier review. Sparse arrays look like they're going to be super useful for saving the values behind calculated flags without need to create ndets x nsamps boolean masks. They go directly into AxisManagers and expanding the IO for saving them is straightforward. 

I'm not very excited about the try-catch import statement I've put in everywhere, but the scipy docs say they're actively moving away from the old matrix versions of the sparse arrays. All the functionalities I've been using are interchangeable between the two versions so I figured I'd make it so later down the line we can just change the import statement.

### DIFF
--- a/sotodlib/core/axisman_io.py
+++ b/sotodlib/core/axisman_io.py
@@ -68,21 +68,21 @@ def expand_RangesMatrix(flat_rm):
 
 ## Flatten and Expand sparse arrays
 def flatten_csr_array(arr):
-    """Extract information from scipy.sparse.csr_array for saving in 
+    """Extract information from scipy.sparse.csr_array for saving in
     hdf5 files"""
     return {
-        'data' : arr.data,
-        'indices' : arr.indices,
-        'indptr' : arr.indptr,
-        'shape' : arr.shape,
+        'data': arr.data,
+        'indices': arr.indices,
+        'indptr': arr.indptr,
+        'shape': arr.shape,
     }
 
 
 def expand_csr_array(flat_arr):
-    """Reconstruct csr arrays from flattened versions
+    """Reconstruct csr_array from flattened versions
     """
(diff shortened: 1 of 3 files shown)


sotodlib/core/axisman_io.py
```diff
@@ -60,6 +66,24 @@ def expand_RangesMatrix(flat_rm):
...
 
+## Flatten and Expand sparse arrays
+def flatten_csr_array(arr):
+    """Extract information from scipy.sparse.csr_array for saving in 
+    hdf5 files"""
+    return {
+        'data' : arr.data,
+        'indices' : arr.indices,
+        'indptr' : arr.indptr,
+        'shape' : arr.shape,
```
mhasself, line 77: remove spaces before :

sotodlib/core/axisman_io.py
```diff
@@ -60,6 +66,24 @@ def expand_RangesMatrix(flat_rm):
...
+    }
+
+
+def expand_csr_array(flat_arr):
+    """Reconstruct csr arrays from flattened versions
+    """
+    return csr_array( (flat_arr['data'], flat_arr['indices'], 
+                       flat_arr['indptr']), 
+                     shape=flat_arr['shape'])
```
mhasself, line 86: "csr_arrays" in docstring.  Remove space between open brackets.

mhasself (changes requested):
Cool!  Glad this worked out smoothly.  Thanks for the test.

Some formatting nitpicking, especially axisman_io.py, which has to this point been fairly PEP8 compliant.  Please try to clean up trailing whitespace, too, though I know some editors make that easier than others.

PR #242, opened 2022-03-23
Context changes to support on-the-fly detdb

jseibert575:
This PR adds functionality to Context to build the obs_detdb, a DetDb-like object that is constructed upon calling either get_obs or get_meta (or check_meta) after providing a specific obs_id. It populates the attribute `ctx.obs_detdb` and leaves `ctx.detdb` alone to make clear that the object is not the same as a traditional DetDb, though it does still belong to that class. 

Adds a hook, living in the load_smurf module, that is called with key `before-load-obs` after the obs_id is initialized. Here is an example Context file:

```
tags:
  g3tsmurf_dir: '/mnt/so1/users/kmharrin/smurf_context'

obsfiledb: '{g3tsmurf_dir}/latrt_db_v4.db'
obsdb: '{g3tsmurf_dir}/latrt_db_v4.db'

imports:
  - sotodlib.io.load_smurf

obs_loader_type: 'g3tsmurf'

context_hooks: obs_detdb_load
```
Notice that the detdb key isn't included here.

I've tested this using a few obs_ids and haven't had any issues yet. I've also checked the metadata loading which does work, but am not including that in the example Context file because there are some bugs that are related to the construction of the CalDb I'm loading, having nothing to do with the changes addressed in this PR. I made some changes to the metadata loading functions to make sure the correct DetDb is passed through to the right places, and I checked that it is working properly, so those shouldn't be an issue. (The problem emerges only when trying to merge that resultant axis manager with the loaded timestream, which is related to the way I've named detectors in the CalDb, and shouldn't require a PR to address).

### DIFF
--- a/sotodlib/core/context.py
+++ b/sotodlib/core/context.py
@@ -71,6 +71,7 @@ class Context(odict):
         self.obsdb = None
         self.detdb = None
         self.obsfiledb = None
+        self.obs_detdb = None
 
         for to_import in self.get('imports', []):
             importlib.import_module(to_import)
@@ -200,6 +201,18 @@ class Context(odict):
         if isinstance(obs_id, dict):
             obs_id = obs_id['obs_id']  # You passed in a dict.
 
+        # Call a hook after preparing obs_id but before loading obs
+        self._call_hook('before-load-obs', obs_id=obs_id,
+                        detsets=detsets, samples=samples)
+
+        # Identify whether we should use a detdb or an obs_detdb
+        # If there is an obs_detdb, use that.
+        # Otherwise, use whatever is in self.detdb, even if that is None.
+        if self.obs_detdb is not None:
+            detdb = self.obs_detdb
+        else:
+            detdb = self.detdb
+
         # If the obs_id is colon-coded, decode them.
         if ':' in obs_id:
             tokens = obs_id.split(':')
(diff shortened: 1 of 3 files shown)


sotodlib/core/context.py
```diff
@@ -307,17 +320,43 @@ def get_meta(self, request):
         """
         if isinstance(request, str):
             request = {'obs:obs_id': request}
+
+        # Call a hook after preparing obs_id but before loading obs
+        obs_id = request['obs:obs_id']
+        self._call_hook('before-load-obs', obs_id=obs_id,
```
kmharrington, line 326: should probably include dets as something passed to this hook

sotodlib/io/load_smurf.py
```diff
@@ -2350,7 +2358,7 @@ def load_g3tsmurf_obs(db, obs_id, dets=None, samples=None, **kwargs):
         "select name from files " "where obs_id=?" + "order by start", (obs_id,)
     )
     flist = [row[0] for row in c]
-    return load_file(flist, dets, samples=samples, obsfiledb=db)
+    return load_file(flist, dets, samples=samples, obsfiledb=db, short_labels=False)
```
kmharrington, line 2361: sorry I messed that up!

mhasself (changes requested):
Sorry for the delay on the review.  This looks good but I have one request:

You are calling the same hook, "before-load-obs", in 3 different places.  Only one of those might legitimately be thought of as "before the observation is loaded".  But clearly you want to do the same thing in each case: provide an observation-specific detdb.  So how about we change the name of this hook to "before-use-detdb"?

mhasself (approved):
Great!

PR #243, opened 2022-03-25
docs: mock modules manually

mhasself:
For sphinxarg.ext.


mhasself:
@BrianJKoopman I have learned some new things about mocks in sphinx...

BrianJKoopman (approved):
Neat! That's useful to know. socs could really use that.

PR #244, opened 2022-03-29
change import order and add pyfftw to requirements

kmharrington:
Small fixes that I found as I updated our python and was installing things from scratch.


tskisner (approved):
Looks fine to me after tests pass.

PR #245, opened 2022-03-31
so-metadata script

mhasself:
ManifestDb problems can be difficult to debug ... this command line tool helps with that, a little, by presenting the DB structure and content summary in human readable form.

Also you can re-write filenames, by prefix, which is useful for moving things between systems (as I am about to do...).


jseibert575 (approved):
This looks great! Gave it a quick once-over, and am looking forward to using it.

PR #246, opened 2022-03-31
tod_ops: cleanup iir_filter

mhasself:
I would like to revisit the interface for iir_params -- see amendments in this PR.

The good news is that the options and logic are greatly simplified.  The bad news is that this doesn't defualt to looking in "status" anymore.

@skhrg @kmharrington  This will break the integration with 'status' in g3tsmurf.  Can you have the g3tsmurf loader instead populate "iir_params" directly?  I think that's more general.  Also you will need to compute the original sampling frequency and use that to set fscale (it is not simply the decimation factor, it turns out).  Feel free to add that to this PR if convenient.

In the proposed system here, parameters "a", "b", and "fscale" are loaded from an AxisManager called "iir_params".  I am open to other names for these things (especially "iir" and "fscale"), but I want them to remain somewhat general rather than being tied to Smurf and Smurf status.

### DIFF
--- a/sotodlib/io/load_smurf.py
+++ b/sotodlib/io/load_smurf.py
@@ -2298,6 +2298,14 @@ def load_file(
     aman.wrap("timestamps", _get_timestamps(streams), ([(0, "samps")]))
     aman.wrap("status", status.aman)
 
+    # If readout filter in enabled build iir_params AxisManager
+    if status.filter_enabled:
+        iir_params = core.AxisManager()
+        iir_params.wrap("a", status.filter_a)
+        iir_params.wrap("b", status.filter_b)
+        iir_params.wrap("fscale", 1/status.flux_ramp_rate_hz)
+        aman.wrap("iir_params", iir_params)
+
     # Conversion from DAC counts to squid phase
     aman.wrap("signal", np.zeros((aman[det_axis].count, aman["samps"].count), "float32"), [(0, det_axis), (1, "samps")])
     for idx in range(aman[det_axis].count):
--- a/sotodlib/tod_ops/filters.py
+++ b/sotodlib/tod_ops/filters.py
@@ -368,7 +368,8 @@ def high_pass_sine2(freqs, tod, cutoff, width=None):
     return 0.5 + 0.5 * np.sin(phase)
 
 @fft_filter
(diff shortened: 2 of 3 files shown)


skhrg:
I can have it populate iir_params directly but now that axismanagers support scalars do we want fscale to just be a separate scalar?

mhasself:
> I can have it populate iir_params directly but now that axismanagers support scalars do we want fscale to just be a separate scalar?

Yes ... sorry if it's not clear, see the test for how iir_params should look.  Two vectors and a scalar.

mhasself:
> In that case do I even need to change anything other than what the params in SmurfStatus are called? Since I can just set the `iir_params` argument to be the smurfstatus axismanager anyways.

If you load these things into the special iir_params key, then it is easier to write generic code.  One can just say "filters.iir_filter(invert=True)" instead of "filters.iir_filter(invert=True, iir_params='status')".  It's nice to have sensible defaults.  So I think it would be better to *not* change the names of things in SmurfStatus, and instead to simply add a new thing, iir_params, with new key names to match this interface. 

> 
> > Also you will need to compute the original sampling frequency and use that to set fscale (it is not simply the decimation factor, it turns out).
> 
> Sorry I'm a bit confused by this. Isn't the original frequency just the decimation factor * current frequency?

Not sure what is confusing... but to say it differently, iir_params should have an element called "fscale" that is computed as
```
 fscale = 1/flux_ramp_rate_hz
```

skhrg:
> Not sure what is confusing... but to say it differently, iir_params should have an element called "fscale" that is computed as

Sorry I guess I worded that poorly. I totally forgot that `flux_ramp_rate_hz` was a thing that we had on hand vs something we needed to calculate (hence the confusion as the whether it was just decimation factor * current frequency or something more complicated). 

> I think it would be better to not change the names of things in SmurfStatus, and instead to simply add a new thing, iir_params, with new key names to match this interface.

Yeah this seems like the reasonable thing to do.

kmharrington (approved):
I supposed this is fine with me. 

If we start doing more things like this we'll end up with a lot of possible top-level arguments that specific functions are looking for. Ex: I could see us doing it with time constants as well. I guess this is fine?  I don't have necessarily have a better idea.

PR #247, opened 2022-04-01
AxisManager: concatenate can handle fields ...

mhasself:
... that don't share the concatenation axis.  If you let it.


kmharrington (approved):
Thanks for setting all this up, going to be super useful!

PR #248, opened 2022-04-05
Toast3 source

gabrielecoppi:
- Added an operator to simulate artificial source
- Added functions to deal with local coordinates transformation

### DIFF
--- a/sotodlib/coords/__init__.py
+++ b/sotodlib/coords/__init__.py
@@ -4,4 +4,4 @@ from .helpers import (
     Timer, DEG,
 )
 from . import planets
-from .local import * 
+from . import local
--- a/sotodlib/coords/__init__.py
+++ b/sotodlib/coords/__init__.py
@@ -4,3 +4,4 @@ from .helpers import (
     Timer, DEG,
 )
 from . import planets
+from .local import * 
--- a/sotodlib/coords/local.py
+++ b/sotodlib/coords/local.py
@@ -1,0 +1,458 @@
+import numpy as np
+from astropy import units as u
+
+
+'''
+Most of the functions are adapted from pymap3d, to include error calulations 
+'''
+
+def ellipsoid(model='WGS84'):
+
+    '''
+    Return the major and minor semiaxis given an ellipsoid model.
+    Returns:
+    - a: astropy.Quantity, semiaxis major in meter
+    - b: astropy.Quantity, semiaxis minor in meter
+    '''
+
+    if model == 'WGS84':
+        a = 6378137.0*u.meter         
+        b = 6356752.31424518*u.meter 
+
+        return a, b
+
+def check_quantity(val, unit):
+
+    if isinstance(val, u.Quantity):
+        return val.to(unit)
+    else:
+        return val * unit
+
+def ecef2lonlat(x, y, z, ell='WGS84'):
+    '''
+    convert ECEF to geodetic coordinates
+    Parameters
+    ----------
+    x : float, array (numpy or Quantity)
+        target x ECEF coordinate. If numpy float or array, it is considered in meters
+    y : float, array (numpy or Quantity)
+        target y ECEF coordinate. If numpy float or array, it is considered in meters
+    z : float, array (numpy or Quantity)
+        target z ECEF coordinate. If numpy float or array, it is considered in meters
+    ell : string, optional
+          reference ellipsoid
+    Returns
+    -------
+    lat : float, array (Quantity)
+          target geodetic latitude 
+    lon : float, array (Quantity)
+          target geodetic longitude
+    alt : float, array (Quantity)
+          target altitude above geodetic ellipsoid
+    based on:
+    You, Rey-Jer. (2000). Transformation of Cartesian to Geodetic Coordinates without Iterations.
+    Journal of Surveying Engineering. doi: 10.1061/(ASCE)0733-9453
+    '''
+
+    x = check_quantity(x, u.meter)
+    y = check_quantity(y, u.meter)
+    z = check_quantity(z, u.meter)
+
+    a, b = ellipsoid(ell)
+
+    r = np.sqrt(x**2 + y**2 + z**2)
+
+    E = np.sqrt(a**2 - b**2)
+
+    # eqn. 4a
+    U = np.sqrt(0.5*(r**2 - E**2) + 0.5*np.sqrt((r**2 - E**2)**2 + 4*E**2*z**2))
+
+    Q = np.sqrt(x**2 + y**2)
+    huE = np.sqrt(U**2 + E**2)
+
+    Beta = np.arctan2(huE*z, U*Q)
+
+    # eqn. 13
+    eps = ((b*U - a*huE + E**2)*np.sin(Beta)) / (
+        a*huE*1 / np.cos(Beta) - E**2*np.cos(Beta)
+    )
+    Beta += eps
+
+    lat = np.arctan2(a*np.tan(Beta), b)
+
+    lon = np.arctan2(y, x)
+
+    # eqn. 7
+    alt = np.hypot(z - b * np.sin(Beta), Q - a * np.cos(Beta))
+
+    # inside ellipsoid?
+    inside = (x**2/a**2 + y**2/a**2+ z**2/b**2 < 1)
+
+    try:
+        if inside.any():  # type: ignore
+            # avoid all false assignment bug
+            alt[inside] = -alt[inside]
+    except (TypeError, AttributeError):
+        if inside:
+            alt = -alt
+
+    return lat, lon, alt
+
+def hor2enu(az, el, srange, deg=True):
+    """
+    Azimuth, Elevation, Slant range to target to East, North, Up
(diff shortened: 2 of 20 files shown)


sotodlib/coords/__init__.py
```diff
@@ -4,3 +4,4 @@
     Timer, DEG,
 )
 from . import planets
+from .local import *
```
mhasself, line 7: These function are quite specialized and I'd prefer to keep them in an isolated sub-sub-module.  So:
```
 from . import local
```

sotodlib/coords/local.py
```diff
@@ -0,0 +1,458 @@
+import numpy as np
+from astropy import units as u
+
+
+'''
+Most of the functions are adapted from pymap3d, to include error calulations 
+'''
```
mhasself, line 7: Please use double quotes for comment blocks.  """/"""

Typo "calculations"

Explain purpose of this submodule.  For example, the fact that these functions are important for drone observations / simulations! 

Maybe mention enu/geodetic/ecef coord systems with a brief line about each one?  We can have all these docs appear on the readthedocs page automagically.

sotodlib/coords/local.py
```diff
@@ -0,0 +1,458 @@
...
+
+def ellipsoid(model='WGS84'):
+
+    '''
```
mhasself, line 11: Remove blank line between signature and docstring; use """ for docstring.  (Please do this for all functions.)

Should this function raise an error if called with invalid model?  Or just silently return None?

sotodlib/coords/local.py
```diff
@@ -0,0 +1,458 @@
...
+        return a, b
+
+def check_quantity(val, unit):
+
+    if isinstance(val, u.Quantity):
+        return val.to(unit)
+    else:
+        return val * unit
+
+def ecef2lonlat(x, y, z, ell='WGS84'):
```
mhasself, line 31: Throughout the module you use the "lonlat" abbreviation (great) but you often return those fields in the order (lat, lon, alt).  How would you feel returning/accepting (lon, lat, alt) everywhere?  (Note that's a naturally right-handed coordinate system, which we like.)

sotodlib/toast/ops/sim_source.py
```diff
@@ -0,0 +1,694 @@
...
+        s_nu: same dimension as tb
+            spectral radiance s_nu, unit: W*sr−1*m−2*Hz−1
+    """
+    h = constants.h.value
+    c = constants.c.value
+    k_b = constants.k_B.value
+
+    x = h * nu / (k_b * tb)
+
+    return 2 * h * nu ** 3 / c ** 2 / (np.exp(x) - 1)
```
mhasself, line 53: We should probably make this a little more unit-correct... extracting a bunch of .value and then computing with them shows a hilarious disrespect for astropy.units.

mhasself (changes requested):
A few comments -- because I care a lot about what happens in .coords!  (I have paid less attention to the toast part.)

Also please consider adding this to the docs (even as just a basic stub) -- at the end of docs/coords.rst, for example, you could put:
```
Local coordinate systems
------------------------

.. automodule:: sotodlib.coords.local
   :members:
```

sotodlib/toast/ops/sim_source.py
```diff
@@ -0,0 +1,694 @@
...
+from toast.traits import trait_docs, Int, Unicode, Float, Instance, List, Bool, Quantity
+
+from toast.ops.operator import Operator
+from toast.instrument import Focalplane
+
+from toast.utils import Logger, Timer
+
+from toast.observation import default_values as defaults
+
+from ...coords.local import *
```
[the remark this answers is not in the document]
tskisner: I'll chime in on this separate from an individual review...  Please let's use relative imports for code in a package that calls other code in the same package.

workflows/toast_so_sim.py
```diff
@@ -402,6 +415,18 @@ def simulate_data(job, args, toast_comm, telescope, schedule):
         timer=timer,
     )
 
+    # Optionally write out the data
+    if args.out_dir is not None:
+        hdf5_path = os.path.join(args.out_dir, "data")
+        if toast_comm.world_rank == 0:
```
tskisner, line 421: Thanks, this is a good idea to put it in a subdirectory and create that on one process.

workflows/toast_so_sim.py
```diff
@@ -623,6 +648,11 @@ def dump_spt3g(job, args, data):
             (ops.elevation_model.out_model, ops.elevation_model.out_model),
         ]
     )
+
+    print([
```
tskisner, line 652: Was this a stale debug print statement?

workflows/toast_so_sim.py
```diff
@@ -789,7 +821,7 @@ def main():
             name="filterbin",
             enabled=False,
         ),
-        so_ops.MLMapmaker(name="mlmapmaker", enabled=False, comps="TQU"),
+        #so_ops.MLMapmaker(name="mlmapmaker", enabled=False, comps="TQU"),
```
tskisner, line 824: The MLMapmaker is disabled by default here (`enabled=False`), so there is no need to comment it out.

sotodlib/toast/ops/sim_source.py
```diff
@@ -0,0 +1,694 @@
...
+
+
+
+def tb2s(tb, nu):
```
tskisner, line 35: This function looks the same as the one defined in `toast.ops.sim_sso.py`.  We should consolidate that to one function and just import it from the other source file.  In toast3, we support `astropy.units` everywhere (in toast2 we were very loose about that).  I think we can modify these functions (in `sim_sso.py`) to propagate the units until we have to convert to a value.  This would address the point that @mhasself was getting at.

tskisner (changes requested):
Thanks for this work!  I think we can remove some duplicate code and propagate units in a couple places before merging.

sotodlib/toast/ops/sim_source.py
```diff
@@ -0,0 +1,694 @@
...
+
+    x = h * nu / (k_b * T_cmb)
+
+    slope = 2 * k_b * nu ** 2 / c ** 2 * ((x / 2) / np.sinh(x / 2)) ** 2
+
+    return s_nu / slope
+
+a = 6378137.0 #semiaxis major
+b = 6356752.31424518 #semiaxis minor
+
```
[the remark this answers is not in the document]
gabrielecoppi: Yes you are right. Old leftovers removed

sotodlib/toast/ops/sim_source.py
gabrielecoppi: Ok I set them to just keep the units

sotodlib/coords/local.py
```diff
@@ -0,0 +1,458 @@
...
+    
+    else:
+        delta_x, delta_y, delta_z = np.zeros_like(x)*u.meter, \
+            np.zeros_like(y)*u.meter, \
+            np.zeros_like(z)*u.meter      
+    
+    return x, y, z, delta_x, delta_y, delta_z
+
+def lonlat2ecef_error(cos_lon, cos_lat, sin_lon, sin_lat, alt, a, b, \
+                      delta_lon, delta_lat, delta_alt):
```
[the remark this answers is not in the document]
gabrielecoppi: Done

sotodlib/coords/local.py
gabrielecoppi: Done. Also every input now follows the convention Lon,Lat,Alt

sotodlib/coords/local.py
```diff
@@ -0,0 +1,458 @@
...
+    - b: astropy.Quantity, semiaxis minor in meter
+    '''
+
+    if model == 'WGS84':
+        a = 6378137.0*u.meter         
+        b = 6356752.31424518*u.meter 
+
+        return a, b
+
+def check_quantity(val, unit):
```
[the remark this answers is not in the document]
gabrielecoppi: Done

sotodlib/coords/local.py
gabrielecoppi: Done

sotodlib/coords/__init__.py
gabrielecoppi: Done

workflows/toast_so_sim.py
gabrielecoppi: I think is some leftover from the first time that I looked at how to export data and then completely forgot about it since I saved always in HDF5. Removed

sotodlib/toast/ops/sim_source.py
gabrielecoppi: I moved the functions to a utils file and kept the units. There is no check on the output because the input units are set so that the output should be the expected one

sotodlib/coords/local.py
mhasself: That's very helpful, thanks.

workflows/toast_so_sim.py
gabrielecoppi: Done

PR #250, opened 2022-04-06
Set timezone to UTC

### LINKED ISSUE #249: timezone dependency
load_data function in sotodlib.io.load_smurf.G3tSmurf class require start time and stop time.
However, these timestamps appear to be dependent on the time zone settings of the analysis machine and the daq machine.
Could you modify the code to be timezone independent?

skhrg:
Should close #249 
Note that here all its doing is setting the timezone to UTC without doing any conversions. I can instead have it convert the input datetimes to UTC if that is preferred.

### DIFF
--- a/sotodlib/io/load_smurf.py
+++ b/sotodlib/io/load_smurf.py
@@ -185,7 +185,8 @@ class G3tSmurf:
         """
         Takes an input (either a timestamp or datetime), and returns a datetime.
         Intended to allow flexibility in inputs for various other functions
-
+        Note that x will be assumed to be in UTC and the datetime will reflect that
+        
         Args
         ----
             x: input datetime of timestamp
@@ -197,9 +198,9 @@ class G3tSmurf:
         if np.issubdtype(type(x), np.floating) or np.issubdtype(type(x), np.integer):
             return dt.datetime.utcfromtimestamp(x)
         elif isinstance(x, np.datetime64):
-            return x.astype(dt.datetime)
+            return x.astype(dt.datetime).replace(tzinfo=dt.timezone.utc)
         elif isinstance(x, dt.datetime) or isinstance(x, dt.date):
-            return x
+            return x.replace(tzinfo=dt.timezone.utc)
         raise (Exception("Input not a datetime or timestamp"))
 
     def add_file(self, path, session, overwrite=False):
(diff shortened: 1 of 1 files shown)


mhasself:
But now if someone constructs a datetime and explicitly sets some timezone, that will be silently force-shifted to UTC?

Is there a way to "replace" the timezone in the datetime only if the user has not explicitly set it themselves (i.e. is datetime.timezone == None unless the user says "actually I really want timezone=EDT").  I.e. can you change the _default_ timezone without either forcing the datetime's timezone to UTC or forcing a conversion from local time to UTC?

I wonder if this actually is not an sotodlib problem (that probably contradicts what I said 2 hours ago ...).  datetime has rules for timezones, and if we accept datetimes as arguments, then the user needs to pass in well-constructed datetimes that mean the right thing.

I have always avoided datetimes, because of these issues... it's been easier to just work with strings "2022-04-06 12:00" and be really clear that they only ever mean UTC.

mhasself:
Ok, there is indeed a "naive" datetime, where the tz isn't specified, and we can decide to interpret as UTC.  I didn't read the whole page... you got this @skhrg ! :)

https://docs.python.org/3/library/datetime.html#aware-and-naive-objects

mhasself:
> Ah so you want naive datetimes to be UTC, but if a timezone is specified don't convert it? Easy enough.

Yes, I think that's a fair compromise that will work for people who don't know what they're doing as well as people who do...

skhrg:
I did assume here that if you pass a `np.datietime64` its in UTC, the [numpy docs](https://numpy.org/doc/stable/reference/arrays.datetime.html#basic-datetimes) seem to indicate that that is the intended usecase anyways so I think that's fair enough.
The existing code also assumed that if you pass a timestamp its in UTC which seems reasonable enough as well.

skhrg:
Can someone review this? It's not high priority but given how simple it is I'd like to get it cleared so the linked issue can be closed.
I assume @mhasself is the person to review this since he commented on it previously.

mhasself:
> Can someone review this? It's not high priority but given how simple it is I'd like to get it cleared so the linked issue can be closed. I assume @mhasself is the person to review this since he commented on it previously.

I'm happy with it but @kmharrington should approve since it's in the smurf stuff.

PR #254, opened 2022-04-14
Make Glitch Flagging and White Noise CalDbs

kmharrington:
Here's my unambitious first pass at making the pipeline scripts to calculate the glitch flags and white noise (really it's "white" noise) for different observations. 

Considering the shear amount of things I just straight copied from the "make-source-flags" scripts there's a lot of low-hanging fruit for normalization and generalization between these different pipeline objects. But I expect that will be the next PR.


mhasself (changes requested):
Sorry for the slow review -- as discussed offline I think these need to be in a single pipeline element, "preprocess-tod".

Glitch patching before evaluating white noise can be accomplished with the gapfill module.

PR #258, opened 2022-04-28
Initial Imprinter

kmharrington:
This is an initial setup for the "imprinter" (the thing that decides what should get bookbound). 

On simons1 you can test with something like: 

```
python3 imprinter.py /mnt/so1/data/pton-rd/ /mnt/so1/users/kmharrin/smurf_context/pton-rd_v1.db --min-ctime=1647117381 --stream-ids=crate1slot3,crate1slot5 --ignore-singles
```

Right now I have set the default overlap time to zero because we don't actually have many examples of data with significant overlap, but I would assume that default would be expanded in the field (or we'd change the logic to be percent of observation). All those would require minor alterations to this script.


guanyilun (approved):
Thanks for this and aplogize for the slow turnaround! I have tested it with success. The codes are easily understandable, so I have no objection to merging this.

Issue #261, opened 2022-05-06
Improve DetDb+Context behavior for metadata loading


mhasself:
(From discussion with @kmharrington and @jseibert575 -- thanks!)

DetDb was designed without fully accepting how incrementally we would build up our per-detector metadata (first with only a readout_id and detset/wafer, then with some readout info, then bias_line info, then a full mapping to physical detector), and the extent to which this info would be tailored to a specific obs_id.  We need to build in better support for that, and make the data more useful along the way.  The metadata system can support a lot of the things that the DetDb time range x table functionality was intended to handle; it now makes more sense to lean on the metadata system. 

We want the following behaviors, when loading metadata & TOD through the Context system:
- A way for some metadata (as loaded by context.get_meta) to be merged into the active (per-obs) DetDb info.  (This would be available for subsequent metadata loaders and the TOD loader to use).
- The DetDb info to be included in the returned AxisManager (let's call that "tod"), under the key ``det_info``.  This will probably have a hierarchical structure with at least 2 levels; equivalent to DetDb table names (readout_id, det_id, wafer, sky.x, sky.y, sky.angle ...)
- Support for usage pattern "context.get_metadata -> trim metadata to select detectors -> context.get_tod(trimmed metadata)".
- ResultSet interface on tod.det_info. The DetDb structure is simple enough that it need not be a full sqlite database ... it's enough to interact with it as a table.

Also, along the way:
- ``tod.obs_info`` field that holds the observation's ObsDb entry
- obsfiledb detector names will be mandated to be the same as readout_id
- #233

mhasself:
Closed via #265 .

PR #265, opened 2022-05-14
Detdb transformation

mhasself:
Addresses everything listed in #261 .  (Supersedes #264 as the comments below reflect major interface changes.)

I am still interested in feedback on the context.get_obs/get_meta interface, but I think it's much better now.  You can do any of the following for get_obs:
```
   ctx.get_obs(obs_id)
   ctx.get_obs(obs_id, dets=['det_01', 'det_02'])
   ctx.get_obs(obs_id, detsets=['w29'])
   ctx.get_obs(obs_id, dets={'readout_id': ['det_01', 'det_02']})
   ctx.get_obs(obs_id, dets={'detset': ['w29']})
   ctx.get_obs(obs_id, dets={'band': 'f090'})
```

Also, filename= still works. And you can throw in samples=(start, stop). And you can still use "obs_colon_tags" (internal nomenclature is now "free tags"). Another example, with 'band' as a free tag:
```
   ctx.get_obs(obs_id+':f090', samples=(0, 1000))
```
For get_meta, the interface is basically the same:
```
   meta = ctx.get_meta(obs_id)  # a string
   meta = ctx.get_meta(obs_id, dets={'band': 'f090'})
```
With that meta result you can do:
```
   tod = ctx.get_obs(meta)
```
and it will magically work, even if you manipulate the meta .dets and/or .samps axes. So that's cool.

Couple other notes:

- The way to add new det_info fields is through a special metadata entry. The loader in that case should return a ResultSet with columns dets:readout_id and dets:new stuff. We can make this more general but it would be nice to limit the options here. This is highly compatible with simple HDF5 tables.
- The AxisManager .dets axis always just contains the readout_id. We can add a promotion step that flips it over to some other field, at the end. If we're loading other 2-d metadata, along the way, that happen to be indexed by det_id (this seems unlikely...) then we can add a metadata setting that indicates what det_info field to reconcile against (i.e. to translate .dets from dets:whatever to dets:readout_id before trying to combine the AxisManagers)
- The DetDb is not used except to turn it into the first iteration of det_info. The hook Joe added might still work... it will probably only need to be called from one place. We don't need to store obs_detdb.  We should consider redefining the hook to act on det_info (the ResultSet) instead of a DetDb.

### DIFF
--- a/sotodlib/core/metadata/loader.py
+++ b/sotodlib/core/metadata/loader.py
@@ -123,7 +123,8 @@ class SuperLoader:
           expresses that the result should be limited to only a subset
           of those detectors.  This is notated in practice by
           including dets:* fields in the index data in the ManifestDb,
-          or in the request dict.
+          or in the request dict.  Only fields already present in
+          det_info may be included in the request dict.
 
         Returns:
           A list of tuples (unpacker, item), corresponding to each
@@ -274,19 +275,20 @@ class SuperLoader:
 
         """
         # Augmented request.
+        aug_request = _filter_items('obs:', request, False)
         if self.obsdb is not None and 'obs:obs_id' in request:
             obs_info = self.obsdb.get(request['obs:obs_id'], add_prefix='obs:')
             if obs_info is not None:
-                obs_info.update(request)
-                request = obs_info
+                obs_info.update(aug_request)
+                aug_request.update(obs_info)
             if dest is None:
                 dest = core.AxisManager()
             obs_man = core.AxisManager()
-            for k, v in obs_info.items():
-                obs_man.wrap(k[len('obs:'):], v)
+            for k, v in _filter_items('obs:', obs_info).items():
+                obs_man.wrap(k, v)
             dest.wrap('obs_info', obs_man)
             
-        def reraise(e, spec):
+        def reraise(spec, e):
             e.args = e.args + (
                 "\n\nThe above exception arose while processing "
                 "the following metadata spec:\n"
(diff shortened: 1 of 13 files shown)


sotodlib/core/metadata/loader.py
```diff
@@ -71,229 +85,408 @@ def load_raw(self, spec_list, request, detdb=None):
...
+                mi2.merge(a)
+
+            elif isinstance(mi1, core.AxisManager):
+                # For AxisManager results, the dets axis *must*
+                # reconcile 1-to-1 with some field in det_info, and
+                # that may be used to toss things out based on
+                # index_line.
+                det_restricts = _filter_items('dets:', index_line, remove=True)
+                dets_key = 'name'
+                new_dets, i_new, i_info = core.util.get_coindices(mi1.dets.vals, det_info['name'])
```
kmharrington, line 218: I think there's a bug around here but don't know the right fix. Trying to run this with some previously working cal_dbs, I get an error about looking up "name." Printing the det_info ResultSet gets me
```
ResultSet<[readout_id,smurf_band,smurf_channel,res_frequency,detset], 878 rows>
```
If I change `det_info['name']` to `det_info['readout_id']` everything works and I get the correct calibration data loaded.

mhasself: Ah ... yes, this was a stray.  It should be readout_id now, as you suggest.  Fixed.

kmharrington (changes requested):
I think I've gotten through all these changes (although am continuing to build off them in the readout-->detector id matching) so might find other things still. 

I think the only change left (which we'd talked about on slack) is what to do with the `ResultSet.axismanager` function. I think just renaming it so something less general than `axismanager` is good for me but could move it / whatever. I've got a `to_axismanager` function I'll be adding from another branch that's more general but less powerful.

sotodlib/core/metadata/loader.py
```diff
@@ -71,229 +85,526 @@ def load_raw(self, spec_list, request, detdb=None):
...
+    Notes:
+      The input arguments det_info and new_info may be modified by
+      this function.  Passing in None for det_info is convenient to
+      initialize a det_info from a new_info where the columns are
+      named dets:... .
+
+    """
+    new_keys = _filter_items('dets:', new_info.keys)
+    if (len(new_keys) != len(new_info.keys)):
+        reraise(spec, RuntimeError(
```
kmharrington, line 444: because I'm good at hitting every one of these errors. I'm getting a "NameError: name 'reraise' is not defined" error here.

kmharrington (approved):
Very excited for all these changes

PR #267, opened 2022-05-27
make-uncal-beam-map

mhasself:
site_pipeline.make_uncal_beam_map

This runs successfully on ACT data and SAT sims.  There are a couple key features missing (demod + downsampling), but this is still useful.  Here's the plot it made for an ACT TOD:

Some other fixes / changes / supports made along the way:
- `coords.planets` has ways to auto-populate detector splits, and map them simultaneously.
- `coords.BeamMapAxes` provides a matplotlib "projection" that is good for beam maps.  Axes for enmaps get auto-labeled with xi and eta and will auto-rescale to deg/arcmin/arcsec depending on the viewport.
- minor/rare bugs fixed in `coords.P` (when threading not in use), `metadata.loader` (when applying a result to only a subset of the detectors that it has data for)
- `make_source_flags` is updated for new metadata practices.
- `site_pipeline.util` has a few new things -- most useful is probably `lookup_conditional`, which helps a config file to provide different values depending on what data the user is working with (e.g. to set a different beam size or map resolution for 'f090' vs. 'f030').
- `tod_ops` has an `IdentityFilter` that can be used as the starting point for a filter chain, and `fourier_filter` will skip the FFT/iFFT entirely if all you're doing is an `IdentityFilter`.
- `tod_ops` has an `fft_trim` function, to restrict .samps to a nice FFT length.  (Having that functionality inside fourier_filter isn't actually practical in realistic cases.)

### DIFF
--- a/sotodlib/coords/__init__.py
+++ b/sotodlib/coords/__init__.py
@@ -4,3 +4,4 @@ from .helpers import (
     Timer, DEG, ScalarLastQuat
 )
 from . import planets
+from .plotting import *
--- a/sotodlib/site_pipeline/util.py
+++ b/sotodlib/site_pipeline/util.py
@@ -3,6 +3,7 @@ import os
 import inspect
 import logging
 import time
+from astropy import units as u
 
 from .. import core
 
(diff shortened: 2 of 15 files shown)


sotodlib/coords/__init__.py
```diff
@@ -4,3 +4,4 @@
     Timer, DEG, ScalarLastQuat
 )
 from . import planets
+from .plotting import *
```
kmharrington, line 7: I tend to worry about any code that automatically imports anything from matplotlib (yes, I'm a HUGE fan of pysmurf, /s). There's nothing in there that will crash things if you don't have X11 forwarding set up or anything?

sotodlib/site_pipeline/util.py
```diff
@@ -9,9 +17,223 @@ class ArchivePolicy:
...
+        self.pattern = kwargs['pattern']
+
+    def get_dest(self, **kw):
+        """Returns full path to destination directory.
+
+        """
+        return os.path.join(self.root_dir, self.pattern.format(**kw))
+
+
+def parse_angle(angle, default_units='deg', desired_units='deg'):
```
kmharrington, line 49: I'm not the biggest fan of just adding units in here. It seems like the type of functionality we'd want in core if we're building units into the `sotodlib` at all. Here makes me expect it to end up a bit haphazard.

Is there a reason we can't just do astropy units?

kmharrington (changes requested):
I will admit I have not looked at any of the mapping math/setup. I'm leaving that for @Zhilei 

I like the changes to `tod_ops`, those all look great. 

A couple in-line comments. 

I'm not a big fan of units showing up in `util` for `site_pipeline`. I think we might want a more comprehensive thing within sotodlib or at least a place where something could become more comprehensive. If we aren't interested in more comprehensive I'd think we should default to using astropy units.

mhasself:
Thanks for the comments @kmharrington  -- I believe I've addressed them.  Though we should talk about the best way to deal with matplotlib (I always have 'agg' in my rcparams, so I forget how annoying the X issue can be).

Please note I've added a demonstration of this code (though for a single ACT Uranus obs), in
https://github.com/simonsobs/pwg-scripts/pull/34

kmharrington (approved):
Thanks for the updates on the units. 

I've checked that the general imports to sotodlib / sotodlib.coords work when X11 isn't set up / with no background adjustments for matplotlib. So I'm happy with that.

PR #268, opened 2022-06-01
linearize timestamps. close #253

### LINKED ISSUE #253: Add `timestamps_linearized` to g3tsmurf.load_file
Because of the latency of the smurf processor the unix timestamps have nonuniform spacing as discussed in #127, the smurf header has counter information that is written at the FPGA level and so does not suffer from delays in the smurf processor. The counter can be used to linearize the timestamps. We should implement in the g3tsmurf loader an added axis that corresponds to this linearized timestamps, this could be implemented as:
```
fsamp, offset = np.polyfit(am.primary["FrameCounter"], am.timestamps, 1)
timestamps_linearized = offset + fsamp*am.primary ["FrameCounter"]

am.wrap("timestamps_linearized", timestamps_linearized, [(0,"samps")])
```

kmharrington:
Add functionality to automatically linearize unix timestamps.

### DIFF
--- a/sotodlib/io/load_smurf.py
+++ b/sotodlib/io/load_smurf.py
@@ -2103,7 +2103,7 @@ def _get_sample_info(filenames):
     return out
 
 
-def _get_timestamps(streams, load_type=None):
+def _get_timestamps(streams, load_type=None, linearize_timestamps=True):
     """Calculate the timestamp field for loaded data
 
     Args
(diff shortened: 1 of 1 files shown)


msilvafe (approved):
Approve given that we're pushing loading timing system timestamps to a separate issue.

kmharrington:
Sounds good thanks! If you all want to make those changes go for it. Only difference between this and doing it with loading AxisManagers is using that `io_load.hstack_into` function to pull fields out of the streams (which should have all primary fields available).

PR #270, opened 2022-06-10
AxisManager utility functions

jlashner:
A couple of useful axismanager functions implemented in sodetlib usage. Katie said a few may be more generally applicable so I'm copying them here.

If these are merged in I'll remove them from sodetlib.

### DIFF
--- a/sotodlib/core/axisman_util.py
+++ b/sotodlib/core/axisman_util.py
@@ -1,0 +1,48 @@
+"""
+Various AxisManager utility functions
+"""
+
+import numpy as np
+from sotodlib.core import AxisManager
+
+class RestrictionException(Exception):
+    """Exception for when cannot restrict AxisManger properly"""
+
+def restrict_to_times(am, t0, t1, in_place=False):
+    """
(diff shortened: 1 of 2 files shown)


sotodlib/core/axisman_util.py
```diff
@@ -0,0 +1,48 @@
+"""
+Various AxisManager utility functions
+"""
+
+import numpy as np
+from sotodlib.core import AxisManager
+
+class RestrictionException(Exception):
+    """Exception for when cannot restrict AxisManger properly"""
```
mhasself, line 9: AxisManager.

mhasself (changes requested):
One could see either of these getting into the class itself (restrict_to_times would need to be a bit more general; restrict_by_value or something).  But good to have them in here.

Please add basic smoke testing for these to tests/test_core.py

jlashner:
Thanks for the review! Comments implemented and tests added

mhasself (approved):
Sorry for the slow follow-up.  Thanks for the changes.

PR #272, opened 2022-06-15
Wiregrid and stimulator simulation operators

keskitalo:
This PR contains both the wiregrid and stimulator simulation operators and their unit tests. There are matching projects in `pwg-scripts` that have been used to integration-test each operator. Further development is likely needed but both operators are functional.

### DIFF
--- a/tests/test_sim_stimulator.py
+++ b/tests/test_sim_stimulator.py
@@ -25,7 +25,7 @@ class SimStimulatorTest(unittest.TestCase):
         focalplane = sotoast.SOFocalplane(
             hwfile=None,
             telescope="LAT",
-            sample_rate=100 * u.Hz,
+            sample_rate=1000 * u.Hz,
             bands="LAT_f090",
             wafer_slots=None,
             tube_slots=None,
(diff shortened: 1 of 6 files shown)


keskitalo:
@tskisner ok to merge? This PR is pretty separate from the rest of the code base.

tests/test_sim_stimulator.py
```diff
@@ -0,0 +1,77 @@
...
+import toast
+
+import sotodlib.toast as sotoast
+import sotodlib.toast.ops as so_ops
+
+
+class SimStimulatorTest(unittest.TestCase):
+    def test_instantiate(self):
+        """Test instantiating a simulation operator."""
+        comm, procs, rank = toast.get_world()
```
tskisner, line 21: For all of the unit test classes using toast, we should guard against import errors in case toast is not installed.  This is the technique I used in the toast / data PR (note the debugging `raise` statement in there which should be replaced by `toast_available = False`):  https://github.com/simonsobs/sotodlib/blob/4e4fb294884d5e7d859260631b5676da0e866832/tests/test_toast_books.py#L24-L46

tests/test_sim_wiregrid.py
```diff
@@ -0,0 +1,95 @@
...
+import toast
+
+import sotodlib.toast as sotoast
+import sotodlib.toast.ops as so_ops
+
+
+class SimWireGridTest(unittest.TestCase):
+    def test_instantiate(self):
+        """Test instantiating a simulation operator."""
+        comm, procs, rank = toast.get_world()
```
tskisner, line 21: Need a toast import guard here.

tskisner (changes requested):
Just a couple changes needed in the unit tests to handle case if toast not installed.

tests/test_sim_stimulator.py
keskitalo: Added import guard.

tests/test_sim_wiregrid.py
keskitalo: Added import guard.

PR #274, opened 2022-06-16
Update smurf caldbs

jseibert575:
site_pipeline.update_smurf_caldbs

This module will generate CalDbs from various detector outputs from sodetlib and smurf. At the time of posting as draft, has the ability to add IVs, bias group maps, and bias steps. Depending on command line arguments at the time of running the script, one can update any subset of those three, in an implementation that makes it easy to add more detector metadata if things come up. 

I generated and tested everything on Katie's dataset described here https://github.com/simonsobs/pwg-scripts/tree/master/pwg-tds/real-data-coadd/noise-data, and with a config file that can be found on simons1 at `/home/jseibert/pipeline-dev/update-smurf-caldbs/config.yaml` if anyone wants to try running it themselves with a similar setup. 

A few main questions for discussion, though I'm happy to get any comments and can change things as necessary, just wanted to put a first version up for people to glance over. Lot of words below, though most of it is a brain dump so feel free to pick and choose what to read over...

1. The way this functions is that for a given obs_id, we will locate the most recent calibration and associate it with that particular obs. If a more elegant way of doing that is required, I'm happy to switch it up but this seemed easy to put in place. At some point, I had convinced myself that a time range was the more valid way to implement the CalDbs for different products (especially bias group maps) instead of just tying them to an individual obs_id, but I talked myself out of that. Happy to be reconvinced if anyone has strong feelings...
2. The IV in particular contains pieces of information that could be included for a particular obs that are a function of chosen bias point. The most obvious of these things is the responsivity estimate from the IV curve for a chosen bias point. I didn't include that here, though again I'm happy to change that if we want. The main reason for not including in this first pass is that the bias point for a given observation is currently only made available by loading the entire TOD, which is time consuming when updating the records. I built and tested this with data that was taken before a change was made in sodetlib to register the action whenever detectors are biased using those functions, but I'm not sure that action actually reports usable information. Happy to expand on this if people want to discuss it further, but for now I left it out in the interest of efficiency.
3. Should we include the Level 2 (I think this is the right level?) filename produced by sodetlib that is used to generate the CalDb somehow in metadata to make it clear what file was used in case someone wants to backtrack?

### DIFF
--- a/docs/site_pipeline.rst
+++ b/docs/site_pipeline.rst
@@ -135,19 +135,31 @@ Here's an annotated example:
 Inputs
 ``````
 
-The formal inputs should all be made available through the Context.
-They are:
+The Context should cause the TOD to be loaded with all supporting
+metadata loaded into the AxisManager.  Here are key members that will
+be processed:
 
-- Obs Book
-- HWP Angle
-- Chanmap
-- Basecal - store in 'cal_base'
-- RelCal model - store in 'cal_flat'
-- Timeconst model
-- Focal plane
-- Glitch flags - store in 'glitch_flags'
-- Noise cuts - store in 'noise_cuts'
-- HWP Spinsync
+- Deconvolution step:
+
+  - ``'timeconst'``
+  - ``'iir_params'``
+
+- Calibration:
+
+  - Whatever is listed in preprocessing.cal_keys
+
+- Pointing correction:
+
+  - ``'boresight_offset'``
+
+- Demodulation and downsampling:
+
+  - not implemented
+
+- Planet mapmaking:
+
+  - ``'source_flags'``
+  - ``'glitch_flags'`` - optional
 
 
 Support
--- a/docs/site_pipeline.rst
+++ b/docs/site_pipeline.rst
@@ -69,6 +69,87 @@ Here's an annotated example:
     default: {'xyr': [0., 0., 0.1]}
 
 
+make-uncal-beam-map
+-------------------
+
+.. automodule:: sotodlib.site_pipeline.make_uncal_beam_map
+   :members:
+   :undoc-members:
+
+
+Command line arguments
+``````````````````````
+
+.. argparse::
+   :module: sotodlib.site_pipeline.make_uncal_beam_map
+   :func: _get_parser
+   :prog: make-uncal-beam-map
+
+Config file format
+``````````````````
+
+Here's an annotated example:
+
+.. code-block:: yaml
+
+  # Data source
+  context_file: ./act_uranus/context.yaml
+
+  # Sub-observation data grouping
+  subobs:
+    use: detset
+    label: wafer_slot
+
+  # Database of results
+  archive:
+    index: 'archive.sqlite'
+    policy:
+      type: 'directory'
+      root_dir: './'
+      pattern: 'maps/{product_id}'
+
+  # Output selection and naming
+  output:
+    map_codes: ['solved', 'weights']
+    pattern: '{product_id}_{split}_{map_code}.fits'
+
+  # Plot generation
+  plotting:
+    zoom:
+      f090: [10, arcmin]
+      f150: [10, arcmin]
+
+  # Preprocessing
+  preprocessing:
+    cal_keys: ['abscal', 'relcal']
+    pointing_keys: ['boresight_offset']
+
+  # Mapmaking parameters
+  mapmaking:
+    force_source: Uranus
+    res:
+      f090: [15, arcsec]
+      f150: [15, arcsec]
+
+
+Inputs
+``````
+
+The formal inputs should all be made available through the Context.
+They are:
+
+- Obs Book
+- HWP Angle
+- Chanmap
+- Basecal - store in 'cal_base'
+- RelCal model - store in 'cal_flat'
+- Timeconst model
+- Focal plane
+- Glitch flags - store in 'glitch_flags'
+- Noise cuts - store in 'noise_cuts'
+- HWP Spinsync
+
+
 Support
 =======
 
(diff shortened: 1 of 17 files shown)


jseibert575:
All the additional commits are from rebasing off of Matthew's open PR in make-uncal-beam-map. I'm happy to try and clean those up and fix that, if desired.

kmharrington:
A couple quick thoughts without looking too closely at the code (sorry if I miss anything obvious):

Are you using tools from the beam map PR? If not, it'd be *much* easier to read this one without all of that intermixed.

For your point (2), 99% of people who interact with the calibration databases will ONLY care about the responsivity calibration. I know it's a function of bias point for each observation and that needs to be found and calculated for IVs, but that's what everyone will want. (Yes there are things to think about if we actually *want* people using IV calibrations, but still.)

For the bias step databases, are you double checking that the bias point in each observation matches the bias point in the for the steps? 

All the SMuRF metadata files have stream ID and tune file saved in them, cross checking that with each observation would probably be good.

jseibert575:
> A couple quick thoughts without looking too closely at the code (sorry if I miss anything obvious):
> 
> Are you using tools from the beam map PR? If not, it'd be _much_ easier to read this one without all of that intermixed.

I'm only using one line, just to establish a logger. I'll pull it out for now since this is just in draft anyway.
> 
> For your point (2), 99% of people who interact with the calibration databases will ONLY care about the responsivity calibration. I know it's a function of bias point for each observation and that needs to be found and calculated for IVs, but that's what everyone will want. (Yes there are things to think about if we actually _want_ people using IV calibrations, but still.)

That's totally fine with me. I might add a command line arg to add this to the caldb though, so that people can avoid loading the TOD if they don't want to do so, since that would add a bunch of run-time to this.
> 
> For the bias step databases, are you double checking that the bias point in each observation matches the bias point in the for the steps?

I'm not, but I can! Particularly if I load the TOD before. Is it a safe assumption to say that for any observation loaded here the bias point doesn't change throughout the observation (at least in terms of set bias)? In that case, I could just load the first couple samples from the observation to speed things up.
> 
> All the SMuRF metadata files have stream ID and tune file saved in them, cross checking that with each observation would probably be good.

Could you clarify what you mean by cross checking that with each observation?

mhasself:
Thanks for posting this.  (Note that to avoid seeing all the make-uncal-beam-map stuff, you could initiate a draft PR from your branch into the make-uncal-beam-map branch.)

I am looking through the update-smurf-caldb script and have a few questions ...

For the IV curves, it looks like this doesn't have the responsivity?  That's understandable since the IV curve is not the same thing as the bias selection / bias application.  So how to we get the responsivity in there (or into a similar but separate place)?

For the bias_step is "Si" the responsivity?

All of these seem to make use of obs_id, passed in on the command line.  Since this is an "update" script, I think what you'll want to do is scan for *all* obs_id (but see below...), starting from some min timestamp, and add anything to the database that you haven't already processed.

The extraction and summary of information looks good -- but ultimately what we are going to want is a way to find the right metadata (e.g. calibration factors and timestamps from a bias_step) for a level 3 Obs Book.  So I think the logic in here (or in somewhere) has to be smart enough to figure out what time ranges a given result applies to.

For example, the results of a bias_step might be deemed to apply to all data from that wafer, going both directions in time, up to any rebiasing events, or another bias_step being closer.  This isn't straightforward to figure out.  But we have to figure it out somehow, at this stage.  If there are metadata missing from the g3tsmurf / G3 archive that would make this easier, we need to identify that and get it into the system.  And we don't really want to guess at this stuff.  We want to have enough record keeping that we _know_ what bias_step (if any) is applicable to every point in time.

kmharrington:
(this is me totally coming back to this while making slides for the DAQ review). But, @mhasself, I think I disagree with making this sort of functionality at this point. I don't think it's the FLP's job to figure out how and when we should use what calibration method. 

> For example, the results of a bias_step might be deemed to apply to all data from that wafer, going both directions in time, up to any rebiasing events, or another bias_step being closer. This isn't straightforward to figure out. But we have to figure it out somehow, at this stage. If there are metadata missing from the g3tsmurf / G3 archive that would make this easier, we need to identify that and get it into the system. And we don't really want to guess at this stuff. We want to have enough record keeping that we know what bias_step (if any) is applicable to every point in time.

Our record keeping, through operation books now, will have to be good enough that we'll know when all operations where called and of what time. We'll be able to reconstruct and test out calibration methods for science results as time goes on and we understand the systems better. 

For the first light pipeline I think we need something strictly defined and automatable: Ex: "the bias step CalDb will return the last bias step information taken before an observation as long as the bias points are within 0.05V of the observation bias points." Maybe we add in configurable options for "time since operation," etc. This has made me realize we need to carefully define a lot of these (with help from detectors and readout) but I don't think we should consider ourselves responsible for "the best cal method" from the start.

mhasself:
> (this is me totally coming back to this while making slides for the DAQ review). But, @mhasself, I think I disagree with making this sort of functionality at this point. I don't think it's the FLP's job to figure out how and when we should use what calibration method.

In case there is a misunderstanding, I didn't mean to imply that we'd construct some CalDb that jumps between recommending IV calibrations, then bias_step calibrations, then back to IV.  Each calibration method should definitely be kept separate.

I was just proposing a slightly more complicated version of "the bias step CalDb will return the last bias step information taken before an observation".  Because bias_steps apply to time before them as well as to the time after them.

The transcription that this script already does, and then the association-with-observations part, could be done separately.  The latter could be left to "update-level3-dbs" element, for example.  There's some advantage to having this script do it, since it is perhaps more readily aware of the chronology since it's closer to the data ... but update-level3-dbs has the advantage of approaching the problem from the perspective of "I need each observation to have a bias_step calibration".

A topic for data flow telecon.

kmharrington:
Ok yeah, this is a data flow topic but also a readout and detector topic TBH. Because I can already tell you we're never going to use IV calibrations in science data.

PR #276, opened 2022-07-09
add demodulation tod module (exported from latrt_testing)

tdalford1:
Adds in demodulation module originally used for LATR tester timestream data analysis.

### DIFF
--- a/sotodlib/tod_ops/demodulation.py
+++ b/sotodlib/tod_ops/demodulation.py
@@ -1,0 +1,264 @@
+import numpy as np
+from sotodlib.tod_ops import filters
+from scipy.optimize import curve_fit
+import matplotlib.pyplot as plt
+
+
+def _remove_demod_placeholders(aman):
+    '''Remove some temporary data from aman that was used for demodulation'''
+    if 'highband_sine' in aman:
+        aman.move('highband_sine', None)
+    if 'highband_cos' in aman:
+        aman.move('highband_cos', None)
+    if 'mod_sine' in aman:
+        aman.move('mod_sine', None)
+    if 'mod_cos' in aman:
+        aman.move('mod_cos', None)
+    return
+
+
+def demod_sine(aman, freq=8.0, lp_fc=1, signal_name='signal',
+               demod_name='demod_signal', dets_axis='dets'):
+    '''Demodulate a sinusoidal function from AxisManager.signal_name and then
+    place it into AxisManager.demod_name, using a sqrt(sin + cos) wave.
+
+    args:
+
+        aman                 - axis manager to use
+        freq                 - demodulation frequency (Hz)
+        lp_fc                - low pass filter cutoff frequency to use (Hz)
+        signal_name          - name of axis manager signal to demodulate
+        demod_name           - name of resulting demod signal to put in aman
+        det_axis             - name of detector axis manager name to use
+    '''
+    try:
+        sinewave = np.sin(2*np.pi*freq*(aman.timestamps-aman.timestamps[0]))
+        coswave = np.cos(2*np.pi*freq*(aman.timestamps-aman.timestamps[0]))
+
+        aman.wrap('highband_sine', aman[signal_name] *
+                  sinewave[None, :], [(0, dets_axis), (1, 'samps')])
+        aman.wrap('highband_cos', aman[signal_name] *
+                  coswave[None, :], [(0, dets_axis), (1, 'samps')])
+
+        aman.wrap('mod_sine', sinewave**2, [(0, 'samps')])
+        aman.wrap('mod_cos', coswave**2, [(0, 'samps')])
+
+        div_mod_sin = filters.fourier_filter(aman,  filters.low_pass_butter4(
+            fc=lp_fc), detrend=None, signal_name='mod_sine')
+        div_mod_cos = filters.fourier_filter(aman,  filters.low_pass_butter4(
+            fc=lp_fc), detrend=None, signal_name='mod_cos')
+
+        if demod_name in aman:
+            aman.move(demod_name, None)
+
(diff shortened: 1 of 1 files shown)


sotodlib/tod_ops/demodulation.py
```diff
@@ -0,0 +1,264 @@
+import numpy as np
+from sotodlib.tod_ops import filters
+from scipy.optimize import curve_fit
+import matplotlib.pyplot as plt
```
kmharrington, line 4: matplotlib imports like this cause issues in modules because people use/connect to them in so many different ways. Can we move this one to inside the plotting if statements?

sotodlib/tod_ops/demodulation.py
```diff
@@ -0,0 +1,264 @@
...
+    phases = np.array([attrs['phase'] for attrs in fit_attrs])
+    # convert this phase to a global phase
+    global_phases = phases - (2 * np.pi * freq) * (
+        times[0] - aman.timestamps[0])
+    # convert to usual range
+    global_phases = np.unwrap(np.sort(np.mod(global_phases, 2 * np.pi)))
+    mean, median, std = (np.mean(global_phases), np.median(global_phases),
+                         np.std(global_phases))
+
+    print('mean, median, std of fitted phases: %.2f, %.2f, %.2f' % (
```
kmharrington, line 232: use logging for print statements instead of just going to sysout (you can copy what's in `sotodlib.io.load_smurf`)

sotodlib/tod_ops/demodulation.py
```diff
@@ -0,0 +1,264 @@
...
+        plt.plot(tt, sinfunc(tt, *popt), label='fit, phase = %.2f' % phi)
+        plt.legend()
+        plt.show()
+
+    return {"amplitude": A, "phase": phi, "offset": c, "freq": freq}
+
+
+def fit_phase(aman, middle_relative_time, index_limit=180, threshold=.8,
+              freq=8.0, plot=False):
+    '''Fit a sinewave to the data to find the chopper phase.
```
kmharrington, line 200: Make this comment a bit more general? This can be used in more than chopper data.

Maybe "fit sinewave to modulated data, only useful in regions where modulation signal has sufficiency signal-to-noise".

sotodlib/tod_ops/demodulation.py
```diff
@@ -0,0 +1,264 @@
+import numpy as np
```
kmharrington, line 1: Could you add a comment string up here about what this module is for. 

"""
Functions for demodulating general sinusoidal modulation in timestreams. Often used in lab testing or with optical choppers. 
"""

I don't want this confused with any HWP demodulators because those will be different.

sotodlib/tod_ops/demodulation.py
```diff
@@ -0,0 +1,264 @@
...
+    if 'highband_cos' in aman:
+        aman.move('highband_cos', None)
+    if 'mod_sine' in aman:
+        aman.move('mod_sine', None)
+    if 'mod_cos' in aman:
+        aman.move('mod_cos', None)
+    return
+
+
+def demod_sine(aman, freq=8.0, lp_fc=1, signal_name='signal',
```
kmharrington, line 20: We started using a different way of passing in arguments in sotodlib than I originally used in latrt-testing. We also standardized the detector axis name to dets. 

For the two demodulation functions could you:

1. remove the `dets_axis` keyword arg and replace `aman['dets_axis']` with just `aman.dets` and `'det_axis'` with `'dets'`.

2.  replace `signal_name='signal'` with `signal=None`, replace `aman[signal_name]` with `signal` and have 
```
if signal is None:
    signal = aman.signal
```
at the beginning of the functions.

[Example of how we've started doing this](https://github.com/simonsobs/sotodlib/blob/master/sotodlib/tod_ops/jumps.py#L350).

kmharrington (changes requested):
Thanks a bunch for making the start of this!

There are formatting / standardizations to change for how we've started doing things in sotodlib (which changed from how I was writing things in latrt-testing).

sotodlib/tod_ops/demodulation.py
tdalford1: agreed, done

sotodlib/tod_ops/demodulation.py
tdalford1: makes sense, done

sotodlib/tod_ops/demodulation.py
tdalford1: done. Is logger.debug the right choice here?

sotodlib/tod_ops/demodulation.py
tdalford1: done

sotodlib/tod_ops/demodulation.py
tdalford1: Sounds good. I think I changed these to what you mentioned. Is the 'samps' axis name used here also universal?

sotodlib/tod_ops/demodulation.py
```diff
@@ -0,0 +1,277 @@
+"""Functions used for demodulating general sinusoidal modulation in
+timestreams. Often used in lab testing or with optical choppers."""
+
+import numpy as np
+from sotodlib.tod_ops import filters
+from scipy.optimize import curve_fit
+
+import logging
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
```
kmharrington, line 10: I realize we probably shouldn't be setting the logging level on input (did I do that in `load_smurf`? I should probably fix that...) Other two lines are good.

kmharrington (approved):
Looks good to me! One small change then lets go ahead and merge it.

Issue #278, opened 2022-07-14
Fix Channel Loading by readout index instead of readout channel


kmharrington:
Katie knows what this means and will do it.

PR #279, opened 2022-07-15
Change how channel indexing works on load.

kmharrington:
Recently we realized that the indexing of how to build a channel mask was pretty not-useful if you just wanted to blindly load sections of channels. This updates `load_file` and `get_channel_mask` to just accept `channels= range(x,y)` and return the resonator channels at those indexes. Will be useful for batching and just limiting detector loads in general.

Solves this issue: https://github.com/simonsobs/sotodlib/issues/278

### DIFF
--- a/sotodlib/io/load_smurf.py
+++ b/sotodlib/io/load_smurf.py
@@ -1759,8 +1759,10 @@ def get_channel_mask(
         if np.isscalar(ch):
             if np.issubdtype(type(ch), np.integer):
                 # this is an absolute readout INDEX (not band*512+ch)
-                if ch >= status.num_chans:
+                if not ignore_missing and ch >= status.num_chans:
                     raise ValueError(f"Requested Index {ch} > {status.num_chans}")
+                if ch >= status.num_chans:
+                    continue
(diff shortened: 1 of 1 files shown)


jseibert575:
Trying to test this out, doing the following:

```
aman = ls.load_file('/mnt/so1/shared/todsims/real-data-coadd/noise-data/timestreams/16500/ufm_cv4/1650060983_000.g3')
```

compared to 

```
aman = ls.load_file('/mnt/so1/shared/todsims/real-data-coadd/noise-data/timestreams/16500/ufm_cv4/1650060983_000.g3', 
                    channels=range(3,8))
```

And unfortunately it seems like the second actually takes **more** time somehow. Does that mean something is broken?

Also, checking with `sys.getsizeof` both axis managers take up the same amount of size, somehow? (64 bytes). Not sure if this is a valid metric, but curious nonetheless...

jseibert575:
This seems to work as intended, so approving!

jseibert575 (approved):
Looks great! Seems to work as intended, so approved.

PR #280, opened 2022-07-15
Adding calibration observations to the G3tSmurf Observations table

jseibert575:
This PR adds calibration observations to the G3tSmurf observations table. In particular, it adds `take_iv`, `take_bias_steps`, `take_bgmap`, and `take_noise`, gives any observations an obs_id that looks something like `cal_<stream_id>_<session_id>`. It also adds a `calibration` column to the Observations table, which will require a database update script before we can merge this into master, which @kmharrington said she would help provide at least an outline for.

### DIFF
--- a/sotodlib/io/load_smurf.py
+++ b/sotodlib/io/load_smurf.py
@@ -697,9 +697,9 @@ class G3tSmurf:
                     break
 
             if calibration:
-                obs_id=f"cal_{stream_id}_{session_id}",
+                obs_id=f"cal_{stream_id}_{session_id}"
             else:
-                obs_id=f"{stream_id}_{session_id}",
+                obs_id=f"{stream_id}_{session_id}"
             
             # Build Observation
             obs = Observations(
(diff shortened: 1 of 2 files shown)


sotodlib/io/load_smurf.py
```diff
@@ -692,14 +700,26 @@ def add_new_observation(
...
             # Build Observation
-            obs = Observations(
-                obs_id=f"{stream_id}_{session_id}",
-                timestamp=session_id,
-                action_ctime=action_ctime,
-                action_name=action_name,
-                stream_id=stream_id,
-                start=start,
-            )
+            if calibration:
```
kmharrington, line 703: could you change this to something like: 
```
if calibration:
    obs_id = f"cal_{stream_id}_{session_id}"
else:
    obs_id=f"{stream_id}_{session_id}"
    
obs = Observations(
    obs_id=obs_id,
    timestamp=session_id,
    action_ctime=action_ctime,
     action_name=action_name,
    stream_id=stream_id,
    start=start,
    calibration=calibration,
)
```

This makes it much easier to keep track of how different variables are set. And easier to change with time if we need to edit how observations are made in the future.

kmharrington (changes requested):
A couple in-line comments to reduce the amount of copy-pasted lines. 

For altering existing tables, follow these examples as a guide: https://github.com/simonsobs/pwg-scripts/tree/master/pwg-duf/g3tsmurf_202112 

This doesn't require any table updates like what's in that python script. But the sql script shows adding columns and the `main.sh` shows how to "just run it."

kmharrington:
Update for folks. This changes requires a table alteration to existing databases: 

```
prompt> sqlite3 database_name.db
sqlite> ALTER TABLE OBS ADD COLUMN CALIBRATION;
sqlite> .exit 
```

kmharrington (approved):
Looks good!

PR #281, opened 2022-08-01
fit planet maps, make statistics, calculate profiles and harmonic transforms

nad-dch:
Fit beam maps and compute the beam transfer function


nad-dch:
Thank you for the comments! I will start addressing them.

PR #286, opened 2022-08-03
Mlmapmaker fix

amaurea:
This fixes a nasty bug in the MLMapmaker class that would prevent the CG chain from converging. I think this was introduced when I generalized the class to handle joint solution of arbitrary signals. There was then a choice between a) Every signal having to handle the cuts, b) Having SignalCut zero out-the cut samples so the rest of the signals can operate on all samples obliviously.

I chose b) since it makes it easier to add new signals, but the consequence is that SignalCut *MUST* be the first signal in the list of signals to solve for if cuts are enabled. However, I made a mistake where the all the code that used the MLMapmaker class expected a different ordering than the one MLMapmaker actually implemented (which was the reverse of what I intended to implement - MLMapmaker mistakenly required SignalCut to be the *last* signal).

I've fixed this now and confirmed that the CG chain properly converges now. However, the fragility of this signal-ordering is worrying. Maybe there is a better solution that is better than both a) and b)...

This pull request depends on the master_toast3_merge pull request, so don't pull it before that.


amaurea:
Maybe this should be rebased, but I think it might fast-forward correctly once master_toast3_merge is merged.

PR #287, opened 2022-08-04
Master toast3 merge

amaurea:
Merge of master and toast3, which I've used for my recent maximum likelihood mapmaker work.


amaurea:
Looks like master was already merged into toast3, so this pull request ends up having hardly any changes in it.

PR #290, opened 2022-08-09
Fix Tune file missing edge case

kmharrington:
Edge case for when trying to load with a g3tsmurf instance but when we're missing tune files in the database. This makes things fail nicely instead of badly.


jlashner (approved):
Looks good!

PR #291, opened 2022-08-12
G3tSmurf Database Updates

### LINKED ISSUE #206: G3tSmurf: Change Observation Error handling
Right now these assertions are working to usefully clear out observations that should not exist: https://github.com/simonsobs/sotodlib/blob/master/sotodlib/io/load_smurf.py#L561 

However it shows up in any sort of log as "unexplained error at Observations XYZ" which is not the most useful or real error message.

### LINKED ISSUE #241: G3tSMuRF : Actually populate the Tunes to Files links
Apparently this got ignored / replaced with just adding TuneSet to File links. Do both.

kmharrington:
Many changes related to databasing based on errors found with LATR testing

* `Observations.start` is now the time of the first data point in a file instead of the time the Observation frame is created
* Automatically order `Observation.files` by start time (new thing I didn't sqlalchemy could do, makes those lists much nicer)
* Use `Observations.session_id` to know what files should be part of an observation. No more fuzzy logic there.
* Close #241 
* Close #206  (still could be nicer but I'm seeing this as a massive improvement for little work)
* More filtering by `stream_id`

### DIFF
--- a/site-scripts/update_g3tsmurf_database.py
+++ b/site-scripts/update_g3tsmurf_database.py
@@ -55,7 +55,6 @@ if __name__ == '__main__':
             obs, 
             session, 
             force=True,
-            max_early=90,
         )
 
     if args.detdb_filename is not None:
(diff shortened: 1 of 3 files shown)


jlashner (approved):
Seems good to me! I haven't been keeping up with G3tSmurf enough to fully understand it, but I don't see any obvious problems so if you've tested, I approve!

PR #292, opened 2022-08-16
Updates to get things working for End-to-end Testing

### LINKED ISSUE #224: G3tSMuRF : Observation updating strategy

### LINKED ISSUE #269: Add no_signal=True load option for Smurf loader
Context now implements this and we also have examples where this would be useful in `sodetlib` as well.

kmharrington:
Various things we needed to update just in general or for end-to-end tests.

* use observation frames to determine if an observation. Technically completely closes #224 even though I apparently decided to close the issue before doing this. 
* add stream_id to the SmurfStatus object because we need a way to pass that around and that makes the most sense
* Add the `no_signal=True/False` option to the g3tsmurf loader. closes #269  even though it's not actually needed for this I just felt like finally doing that.

### DIFF
--- a/sotodlib/io/load_smurf.py
+++ b/sotodlib/io/load_smurf.py
@@ -2159,6 +2159,7 @@ def load_file(
     channels=None,
     samples=None,
     ignore_missing=True,
+    no_signal=False,
     load_biases=True,
     load_primary=True,
     status=None,
@@ -2184,6 +2185,8 @@ def load_file(
           being loaded from the list.
       ignore_missing : bool
           If true, will not raise errors if a requested channel is not found
+      no_signal : bool
+          If true, will not load the detector signal from files  
       load_biases : bool
           If true, will load the bias lines for each detector
       load_primary : bool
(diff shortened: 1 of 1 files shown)


jlashner (approved):
Oops I reviewed these out of order, but looks great :)

kmharrington:
haha, thanks!

PR #293, opened 2022-08-16
Clean Up for the g3tsmurf update script

kmharrington:
Several things I've been meaning to do for awhile for the database update script that's part of the first light pipeline. Moved to the correct place, print statements are now entirely replaced with logger, control over logging level in the script, and some documentation updates.


mhasself (approved):
I like where this going!  Thanks.

PR #294, opened 2022-08-18
Batched Loading of level 2 AxisManagers

kmharrington:
The main additional function here came from discussions at the face-to-face where we wanted to make easier ways to get sections of observations instead of just loading complete observations. `get_batch` is a generator that will yield AxisManagers that are subsets of complete observations. 

@mhasself @jlashner @jseibert575  Interested in what you all think of this?

### DIFF
--- a/docs/g3tsmurf.rst
+++ b/docs/g3tsmurf.rst
@@ -239,15 +239,15 @@ Batched load of Observations
 ============================
 
 Loading long large observations into memory at once can cause issues with memory
-usage, especially on smaller computing facities. This function is a generator
+usage, especially on smaller computing facilities. This function is a generator
 than can be called to automatically split observations into smaller sections.
 
 .. automodule:: sotodlib.io.g3tsmurf_utils
     :special-members: get_batch
     :noindex:
 
-Obervations Files
-=================
+Observation Files
+==================
 
 There are many instances where we might want to load the SMuRF metadata
 associated with actions that have made it into the database. These function take
--- a/docs/g3tsmurf.rst
+++ b/docs/g3tsmurf.rst
@@ -119,7 +119,7 @@ AxisManagers loaded with G3tSmurf will all have the form::
             'TESRelaySetting', 'UnixTime'
         biases (optional): (bias_lines, samps)
             Bias values during the data
-        ch_info : AxisManager (dets,)
+        det_info : AxisManager (dets,)
             Information about channels, including SMuRF band, channel,
              frequency.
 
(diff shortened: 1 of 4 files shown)


sotodlib/io/g3tsmurf_utils.py
```diff
@@ -0,0 +1,195 @@
...
+                    yield load_file( 
+                        filenames, 
+                        channels=det_chunk, 
+                        samples=samp_chunk, 
+                        archive=archive,
+                        **load_file_args,
+                    )
+    except GeneratorExit:
+        pass
+    session.close()
```
jlashner, line 194: I think the session should be closed immediately after you're done using it instead of waiting til after the data's processed

sotodlib/io/g3tsmurf_utils.py
```diff
@@ -0,0 +1,195 @@
...
+    n_samps=None, 
+    det_chunks=None, 
+    samp_chunks=None,
+    test = False,
+    load_file_args={},
+):
+    """A Generator to loop through and load AxisManagers of sections of
+    Observations. When run with none of the optional arguments it will default
+    to returning the full observations. Some arguments over-write others as
+    described in the docstrings below.
```
jlashner, line 78: Though ideally analysis scripts will pull this data from the axis-manager, I think the loop-order of the `dets` and `samps` axes should be included here

sotodlib/io/g3tsmurf_utils.py
```diff
@@ -0,0 +1,195 @@
...
+    logger.debug(f"Loading data with det_chunks: {det_chunks}.")
+    logger.debug(f"Loading data in samp_chunks: {samp_chunks}.")
+    
+    try:
+        for det_chunk in det_chunks:
+            for samp_chunk in samp_chunks:
+                if test:
+                    yield (det_chunk, samp_chunk)
+                else:
+                    yield load_file(
```
jlashner, line 185: How long does it take for the SmurfStatus to be loaded? Maybe that should be loaded once and passed in to each batch.

docs/g3tsmurf.rst
```diff
@@ -232,8 +234,35 @@ files. For Example::
     
         return os.path.join(base_dir, 'outputs',info)
 
+ 
+Batched load of Observations
+============================
+
+Loading long large observations into memory at once can cause issues with memory
+usage, especially on smaller computing facities. This function is a generator
```
jlashner, line 242: facities --> facilities

docs/g3tsmurf.rst
```diff
@@ -232,8 +234,35 @@ files. For Example::
...
+than can be called to automatically split observations into smaller sections.
+
+.. automodule:: sotodlib.io.g3tsmurf_utils
+    :special-members: get_batch
+    :noindex:
+
+Obervations Files
```
jlashner, line 249: Obervations --> Observations

jlashner (approved):
Hi Katie, looks great and I can see this being really useful. A few small nitpicky comments and then I think its good to go

sotodlib/io/g3tsmurf_utils.py
kmharrington: not long but it is additional IO, I'll add it this.

sotodlib/io/g3tsmurf_utils.py
kmharrington: I'm pretty sure that you technically want to leave it open until you're done with any objects you got from the database. But that's still well before this point so I'll move it up.

PR #295, opened 2022-08-23
metadata.ManifestDb: add inspection and modification API

mhasself:
New functions in ManifestDb to help with cases where one needs to inspect existing db before trying to modify it:
- .inspect() -- load records, matching only fields provided (which can be Index or Endpoint)
- .update_entry() -- starting from records returned by .inspect(), change the data.
- .remove_entry() -- based on records returned by .inspect(), delete entries.

@jseibert575 and @kmharrington please let me know what else might be helpful.

(In looking at this (first time in a while) I am seeing that having a separate table for files was probably not a good idea.  There's an advantage to having a special *type* for endpoint data that contains filenames (because you can perform path resolution), but it would be much easier to treat it like other endpoint data.  I think we should change the interface to not treat filename= in a special way, and then later change the internal implementation.  Existing archives can be translated on load in a pretty straightforward way.)

### DIFF
--- a/docs/context.rst
+++ b/docs/context.rst
@@ -385,12 +385,12 @@ following:
   See examples in :ref:`metadata-indexes`.
 - The entries in the Archive do not need to be per-detector.  You can
   specify results for a whole group of detectors, if that group is
-  enumerated in the DetDb.  For example, if DetDb contains a column
-  ``band``, the dataset could contain columns ``dets:band`` and
-  ``cal`` and simply report one calibration number for each frequency
-  band.  (On load, the Context Metadata system will automatically
-  broadcast the ``cal`` number so that it has shape ``(n_dets,)`` in
-  the fully populated AxisManager.)
+  enumerated in the DetDb (or added to det_info using other metadata).
+  For example, if DetDb contains a column ``band``, the dataset could
+  contain columns ``dets:band`` and ``cal`` and simply report one
+  calibration number for each frequency band.  (On load, the Context
+  Metadata system will automatically broadcast the ``cal`` number so
+  that it has shape ``(n_dets,)`` in the fully populated AxisManager.)
 - You can store all the results (i.e., results for multiple
   ``obs_id``) in a single HDF5 dataset.  This is not usually a good
   idea if your results are per-detector, per-observation... the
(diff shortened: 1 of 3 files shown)


sotodlib/core/metadata/manifest.py
```diff
@@ -160,22 +180,49 @@ def from_database(cls, conn, table_name='input_scheme'):
...
+          partial: if True, then operate in "inspection" mode (see notes).
+          strict: if True, then reject any requests that include
+            entries in params that are not known to the schema.
 
         Returns:
           (where_string, values_tuple, ret_cols)
+
+        Notes:
+          The normal mode (partial=False) requires that every "in"
+          column in the scheme has a key=value pair in the params
```
jseibert575, line 197: key-value

jseibert575 (approved):
This looks great to me! Looking forward to using this functions. Only one minor (and unnecessary) docstring typo, leaving aside the changes to filenames that you mentioned in the write-up.

jseibert575:
> (In looking at this (first time in a while) I am seeing that having a separate table for files was probably not a good idea. There's an advantage to having a special _type_ for endpoint data that contains filenames (because you can perform path resolution), but it would be much easier to treat it like other endpoint data. I think we should change the interface to not treat filename= in a special way, and then later change the internal implementation. Existing archives can be translated on load in a pretty straightforward way.)

I think this change makes sense. More general code regarding different datafields in the ManifestDbs is a good thing, I think.

docs/context.rst
```diff
@@ -385,12 +385,12 @@ following:
...
 - The entries in the Archive do not need to be per-detector.  You can
   specify results for a whole group of detectors, if that group is
-  enumerated in the DetDb.  For example, if DetDb contains a column
-  ``band``, the dataset could contain columns ``dets:band`` and
-  ``cal`` and simply report one calibration number for each frequency
-  band.  (On load, the Context Metadata system will automatically
-  broadcast the ``cal`` number so that it has shape ``(n_dets,)`` in
-  the fully populated AxisManager.)
+  enumerated in the DetDb (or added to det_info using other metadata).
+  For example, if DetDb contains a column ``band``, the dataset could
```
kmharrington, line 389: Band is a highly overloaded term in SO world. Could we make a habit of using something like bandpass, fcode, freq_code, etc? I use entries with `bandpass = "f090"` in the detector wafer info.

kmharrington (approved):
Yeah, looks like some useful tools, one comment on the docs which has nothing to do with the actual code.

PR #298, opened 2022-08-27
HWPSS simulation fix

Helbouha:
This is a fix to a bug in the HWPSS simulation which was causing spurious peaks to appear in the Fourrier spectrum. This implementation avoids this by first interpolating the HWPSS template to fast sampled HWP angles, and then downsampling to the original sample rate using [`scipy.signal.decimate`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.decimate.html#scipy.signal.decimate)

### DIFF
--- a/sotodlib/toast/sim_hwpss.py
+++ b/sotodlib/toast/sim_hwpss.py
@@ -16,7 +16,11 @@ XAXIS, YAXIS, ZAXIS = np.eye(3)
 
 
 class OpSimHWPSS(toast.Operator):
-    """ Simulate HWP synchronous signal """
+    """ Simulate HWP synchronous signal 
+        N.B: The HWPSS template is interpolated (with a 5th order spline interpolation).
+            Interpolation errors produce spurious peaks in the frequency domain which are 7 to 8 orders of magnitude below
+            the amplitude of the original signal.
+    """
 
     def __init__(self, name, fname_hwpss, mc=0):
         """
--- a/sotodlib/toast/sim_hwpss.py
+++ b/sotodlib/toast/sim_hwpss.py
@@ -6,7 +6,7 @@ import pickle
 import sys
 
 import numpy as np
-import scipy.signal
+import scipy.interpolate as scp
 
 import toast
 import toast.qarray as qa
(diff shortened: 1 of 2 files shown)


sotodlib/toast/sim_hwpss.py
```diff
@@ -43,8 +44,19 @@ def exec(self, data):
...
+            
             # Get HWP angle
             chi = tod.local_hwp_angle()
+            
+            # Get times
+            times = tod.local_times()
+            
+            # Fast sample HWP angle with a factor 100
+            fast_time = np.linspace(times[0], times[-1], 1+(len(times)-1)*100)
+            hwp_rate = (chi[1] - chi[0]) / (times[1] - times[0]) / (2*np.pi) # Assuming constant HWP rotation
```
keskitalo, line 56: I don't think we can assume constant rotation rate. We know that is not the case in real experiments.

sotodlib/toast/sim_hwpss.py
```diff
@@ -43,8 +44,19 @@ def exec(self, data):
...
             focalplane = obs["focalplane"]
+            
             # Get HWP angle
             chi = tod.local_hwp_angle()
+            
+            # Get times
+            times = tod.local_times()
+            
+            # Fast sample HWP angle with a factor 100
+            fast_time = np.linspace(times[0], times[-1], 1+(len(times)-1)*100)
```
keskitalo, line 55: This can be a significant increase in memory consumption. Do we absolutely need to oversample 100X or would, say, 10X do the job as well?

sotodlib/toast/sim_hwpss.py
```diff
@@ -97,30 +109,35 @@ def exec(self, data):
...
+                    qweights * np.interp(fast_chi, self.chis, iquv[1]) +
+                    uweights * np.interp(fast_chi, self.chis, iquv[2])
                 )
 
-                iquss -= np.median(iquss)
+                fast_iquss -= np.median(fast_iquss)
 
-                # Co-add with the cached signal
+                # Downsample iquss to original sampling rate
+                iquss = scipy.signal.decimate(fast_iquss,10)
```
keskitalo, line 137: Have you checked that the supersampling and decimation do not introduce a phase shift in the signal? You can do this by applying the same operations to the time stamps and confirming that they match the input time stamps.

keskitalo (changes requested):
This is awesome! Please see the few questions I added. We'll need to address the constant rotation rate requirement.

sotodlib/toast/sim_hwpss.py
Helbouha: This is no longer a requirement in the latest fix.

sotodlib/toast/sim_hwpss.py
Helbouha: This is also no longer required in the latest fix.

sotodlib/toast/sim_hwpss.py
Helbouha: Decimation no longer required.

keskitalo (approved):
Excellent work! Thank you so much for identifying this issue and resolving it.

PR #299, opened 2022-08-29
Import quick fix from PR 298

keskitalo:
Simple fix to HWPSS simulations from @Helbouha . Also included in #298 .


keskitalo:
Will merge this trivial fix that only affects the TOAST simulation pipeline as soon as the unit tests pass.

PR #301, opened 2022-08-30
Add so3g as requirement

zonca:
it is imported in sim_hardware

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,6 +54,7 @@ setup_opts["install_requires"] = [
     'PyYAML',
     'toml',
     'skyfield',
+    'so3g',
     'pixell',
     'scikit-image',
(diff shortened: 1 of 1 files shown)


tskisner:
If someone is running against the so3g wheel from pypi, this will work fine.  What happens if someone is running against a git checkout of so3g which is installed somewhere to `$PYTHONPATH`?  Would doing `pip install .` on sotodlib then try to pip install so3g from PyPI?  Would users running against a dev copy just need to do `pip install --no-deps .` when installing sotodlib?

zonca:
No, unless you are running pip install --upgrade

kmharrington (approved):
so3g is definitely imported a lot more places than just sim_hardware.

PR #303, opened 2022-09-13
Check if end of file is before requested sample range

### LINKED ISSUE #302: Batched loading of TOD loads more files than needed
When using the new batched TOD generator both @jeffiuliano and I ran across the issue where when loading the n'th tod it seeems to be loading n files. The signal in the returned TOD is still correct but this makes the data loading take increasingly longer at each iteration.

Adding a print statement to `load_file` shows `flist` grow each iteration with the first element being correct and the rest being other files with some odd sample bounds.

Here is an example:

```
In [148]: amans = gu.get_batch(obs.obs_id, SMURF, n_samps=10*60*200)

In [149]: for aman in amans:
     ...:     print(aman.signal.size)
     ...: 
     ...: 
[('/mnt/so1/data/penn-latr/timestreams/16628/ufm_mv15/1662875703_000.g3', 0, 120000)]
100%|████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:16<00:00, 16.99s/it]
189600000
[('/mnt/so1/data/penn-latr/timestreams/16628/ufm_mv15/1662875703_000.g3', 120000, 240000), ('/mnt/so1/data/penn-latr/timestreams/16628/ufm_mv15/1662875703_001.g3', 0, 119966)]
100%|████████████████████████████████████████████████████████████████████████████████████████████| 2/2 [00:33<00:00, 16.63s/it]
189600000
[('/mnt/so1/data/penn-latr/timestreams/16628/ufm_mv15/1662875703_000.g3', 240000, 360000), ('/mnt/so1/data/penn-latr/timestreams/16628/ufm_mv15/1662875703_001.g3', 119966, 239966), ('/mnt/so1/data/penn-latr/timestreams/16628/ufm_mv15/1662875703_002.g3', 0, 119930)]
100%|████████████████████████████████████████████████████████████████████████████████████████████| 3/3 [00:58<00:00, 19.39s/it]
189600000
[('/mnt/so1/data/penn-latr/timestreams/16628/ufm_mv15/1662875703_000.g3', 360000, 480000), ('/mnt/so1/data/penn-latr/timestreams/16628/ufm_mv15/1662875703_001.g3', 239966, 359966), ('/mnt/so1/data/penn-latr/timestreams/16628/ufm_mv15/1662875703_002.g3', 119930, 239930), ('/mnt/so1/data/penn-latr/timestreams/16628/ufm_mv15/1662875703_003.g3', 0, 119892)]
 50%|██████████████████████████████████████████████                                              | 2/4 [00:32<00:32, 16.07s/it]
```

Where the `obs` is `ufm_mv15_1662875703: 2022-09-11 05:55:01.989792 -> 2022-09-11 13:37:19.964351 [7:42:17.974559] (None)`

skhrg:
Fixes #302

### DIFF
--- a/sotodlib/io/load_smurf.py
+++ b/sotodlib/io/load_smurf.py
@@ -2290,6 +2290,8 @@ def load_file(
             if stop is not None:
                 if file_start > sample_stop:
                     continue
+                if file_stop < sample_start:
+                    continue
                 stop = sample_stop - file_start
 
             start = max(0, sample_start - file_start)
--- a/sotodlib/io/load_smurf.py
+++ b/sotodlib/io/load_smurf.py
@@ -2287,10 +2287,10 @@ def load_file(
         stop = sample_stop
         for filename, out in zip(filenames, outs):
             file_start, file_stop = out["sample_range"]
+            if file_stop <= sample_start:
+                continue
             if stop is not None:
                 if file_start >= sample_stop:
-                    continue
-                if file_stop <= sample_start:
                     continue
                 stop = sample_stop - file_start
 
--- a/sotodlib/io/load_smurf.py
+++ b/sotodlib/io/load_smurf.py
@@ -2288,9 +2288,9 @@ def load_file(
         for filename, out in zip(filenames, outs):
             file_start, file_stop = out["sample_range"]
             if stop is not None:
-                if file_start > sample_stop:
+                if file_start >= sample_stop:
                     continue
-                if file_stop < sample_start:
+                if file_stop <= sample_start:
                     continue
                 stop = sample_stop - file_start
 


mhasself:
My radar is telling me that your test should be outside the "if stop is not None".

Also ... do you agree that your new test, as well as the existing one, should include equality? ... i.e.  `if file_start >= sample_stop` and `if file_stop <= sample_start`?  (Generally with python-style indexing if start == stop, that corresponds to 0 samples, so skip.)

skhrg:
Agreed on both points.

kmharrington:
I agree with `if file_stop < sample_start:` being outside the `if stop is not None` but disagree with the equalities. 

Samples in files are returned as [start, stop). So if the files samples are `[0, 500)` and `[500,1000)` and the sample request is `[500, 700)` then we don't actually want to open the first file. Similarly, if the sample request is `[300,500)` then we don't need to open the second file.

kmharrington:
Also thanks so much for digging into this and figuring out what was missing!

mhasself:
> So if the files samples are `[0, 500)` and `[500,1000)` and the sample request is `[500, 700)` then we don't actually want to open the first file.

I think that the logic that implements your example (for the first file) is:
```
   if file_stop <= sample_start:   # if 500 <= 500:
      continue                     # yes, skip that first file
```
I.e., this includes equality _as a condition for skipping the file_.  That's what I meant.  I think https://github.com/simonsobs/sotodlib/pull/303/commits/6bb597fa94a2062b678c2c38d646e5faf9aab53f gets it right.

kmharrington (approved):
_Right_. I swear I can read code.

Looks great! Thanks for the fix!

PR #305, opened 2022-09-14
Loading batcher extension

jeffiuliano:
It will be useful to be able to batch subsets of observation, so this extends get_batch to take start, end times and loop only through that piece of the observation.

Other changes:
- get_batch can take an `Observation` object and not just an `obs_id` string (for convenience)
- Since I'm now using the `make_datetime` tool in get_batch too, it seemed better to make it it's own function that can be imported rather than a static method of `G3tSmurf`.

### DIFF
--- a/sotodlib/io/g3tsmurf_utils.py
+++ b/sotodlib/io/g3tsmurf_utils.py
@@ -12,6 +12,31 @@ from sotodlib.io.g3tsmurf_db import Observations, Files
 logger = logging.getLogger(__name__)
 
 
+
+def make_datetime(x):
+    """
+    Takes an input (either a timestamp or datetime), and returns a datetime.
+    Intended to allow flexibility in inputs for various other functions
+    Note that x will be assumed to be in UTC if timezone is not specified
+
+    Args
+    ----
+        x: input datetime of timestamp
+
+    Returns
+    ----
+        datetime: datetime of x if x is a timestamp
+    """
+    if np.issubdtype(type(x), np.floating) or np.issubdtype(type(x), np.integer):
+        return dt.datetime.utcfromtimestamp(x)
+    elif isinstance(x, np.datetime64):
+        return x.astype(dt.datetime).replace(tzinfo=dt.timezone.utc)
+    elif isinstance(x, dt.datetime) or isinstance(x, dt.date):
+        if x.tzinfo == None:
+            return x.replace(tzinfo=dt.timezone.utc)
+        return x
+    raise (Exception("Input not a datetime or timestamp"))
+
 def get_obs_folder(obs_id, archive):
     """
     Get the folder associated with the observation action. Assumes
@@ -60,7 +85,7 @@ def get_obs_plots(obs_id, archive):
     return [os.path.join(path,f) for f in os.listdir(path)]
 
 def get_batch( 
-    obs_id, 
+    specifier, 
     archive,
     ram_limit=None, 
     n_det_chunks=None,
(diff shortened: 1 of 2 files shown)


sotodlib/io/g3tsmurf_utils.py
```diff
@@ -129,13 +138,41 @@ def get_batch(
...
+    if "status" in load_file_args:
+        status = load_file_args.pop("status")
+    else:
+        status = SmurfStatus.from_file(filenames[0])
+
+    if start == None:
+        start = obs.start
+    if end == None:
+        end = obe.stop
+    samprate = 4e3 / (status.downsample_enabled * status.downsample_factor)
```
skhrg, line 167: Doesn't this divide by 0 if `status.downsample_enabled` is `False`?

jeffiuliano: Yeah, that's definitely what it does.  Will fix

skhrg: I expect `status.downsample_factor` is 1 when it isn't downsampled. But I may be wrong.

sotodlib/io/g3tsmurf_utils.py
```diff
@@ -117,6 +122,10 @@ def get_batch(
     samp_chunks: None or list of tuples
         if specified, each entry in the list is successively passed to load the
         AxisManagers as  `load_file(... samples = list[i] ... )`
+    start: Datetime or timestamp. Begin batching at this time.
+    end: Datetime or timestamp.  End betching at this time.
```
kmharrington, line 126: spelling

sotodlib/io/g3tsmurf_utils.py
```diff
@@ -90,7 +95,7 @@ def get_batch(
     
     Arguments
     ----------
-    obs_id : string
+    obs_id : string, or Observatin object
```
kmharrington, line 98: spelling

sotodlib/io/g3tsmurf_utils.py
```diff
@@ -129,13 +138,41 @@ def get_batch(
...
+        
+    if "status" in load_file_args:
+        status = load_file_args.pop("status")
+    else:
+        status = SmurfStatus.from_file(filenames[0])
+
+    if start == None:
+        start = obs.start
+    if end == None:
+        end = obe.stop
```
kmharrington, line 166: typo. did you test this?

sotodlib/io/g3tsmurf_utils.py
```diff
@@ -129,13 +138,41 @@ def get_batch(
...
+    if "archive" in load_file_args:
+        archive = load_file_args.pop("archive")
+        
+    if "status" in load_file_args:
+        status = load_file_args.pop("status")
+    else:
+        status = SmurfStatus.from_file(filenames[0])
+
+    if start == None:
+        start = obs.start
```
kmharrington, line 164: `if start is None:` (I don't think == is standard for getting Nones in python)

if start is None can we just set `samp_s` to 0 and skip the math? that will avoid rounding errors and missed edge cases by default. Same `if stop is None`, just set `samp_e` to `obs_samps`.

kmharrington (changes requested):
I started reviewing this piecemeal but then I realized I don't like the strategy of guessing the samples vs time with all those down sample assumptions. We know that information already we don't have to guess. 

The Files table has `start`, `stop`, `sample_start`, and `sample_stop` (I see the docstring says the sample ones aren't implemented, they are actually implemented I missed that in an update). 

The Frames table (files have a lists of frames) has `time`, and `n_samples`. Each frame is only a couple seconds long so you should be able to use frame edges very effectively to determine what sample ranges you want to load for a specific time range.

jeffiuliano:
Oh, yeah, I like that better too.  I had started by trying to just load all the timestamps and no data, which I thought (and still do) _should_ be fast, but was actually not at all fast.  But using tables is even better, good idea.  I'll update to that strategy when I get a chance

skhrg:
@jeffiuliano what is status of finishing this up? Also the option to only give you one in every n batches would be useful for when we want to take a quick look at a long timestream (ie: grab the first 10 mins of each hour in a overnight TOD).

mhasself:
Hi folks -- can we retreat this to a "Draft PR", or is anyone passionate about whipping it into shape for final review in the near term?

PR #307, opened 2022-09-21
Update hardware sim and toast workflows.

tskisner:
* Use new xi/eta/gamma support in toast when simulating detector layout.

* Fix projected LAT rotation on the sky, and also the tube positions.

* Add LAT co-rotation options when making focalplane plots.

* Support focalplane plots in both Xi / Eta and X / Y coordinates.

* Cache detector beams in SimSSO and SimSource, in case we have the same beam for many detectors.

* Small tweaks to get_wafer_offsets arguments

* Add a new toast workflow that just loads data and reduces it.

### DIFF
--- a/docs/toast.rst
+++ b/docs/toast.rst
@@ -1,5 +1,5 @@
 ===========
-TOAST 
+TOAST
 ===========
 
 The ``sotodlib.toast`` package contains toast operators for use in
@@ -45,7 +45,7 @@ easiest way forward is to use a conda environment.  This is also the
 path to take if you want to use the latest upstream code version.
 
 To begin, make sure you already have a recent conda "base" environment.
-If you are starting from scratch, I recommend running the "miniforge" `installer from conda-forge <https://github.com/conda-forge/miniforge>`_.  
+If you are starting from scratch, I recommend running the "miniforge" `installer from conda-forge <https://github.com/conda-forge/miniforge>`_.
 Make sure your base conda environment is activated::
 
     conda env list
@@ -73,8 +73,8 @@ Now we can install the toast package itself after activating our environment::
     conda activate simons
     ./platforms/conda.sh
 
-Whenever you pull new changes to your git checkout, you can re-run the 
-``platforms/conda.sh`` script to install an updated version of the toast 
+Whenever you pull new changes to your git checkout, you can re-run the
+``platforms/conda.sh`` script to install an updated version of the toast
 package.  Now install ``so3g``::
 
     pip install so3g
@@ -101,12 +101,19 @@ Batch Use
 -------------------
 
 When running workflows in batch mode, we typically use a slightly different
-structure from a notebook.  There are helper functions in toast to parse
-operator options from the commandline and / or a config file, and to automatically
-create named instances of those operators for use in workflow.  There is
-a large-scale workflow in ``sotodlib/workflows/toast_so_sim.py``.  This
-script has nearly every possible operator in its config options, and they
-can be selectively enabled or disabled from the commandline or config file.
+structure from a notebook. There are helper functions in toast to parse
+operator options from the commandline and / or a config file, and to
+automatically create named instances of those operators for use in workflow.
+There is a large-scale simulation and reduction workflow in
+``sotodlib/workflows/toast_so_sim.py``. This script has nearly every possible
+operator in its config options, and they can be selectively enabled or disabled
+from the commandline or config file.
+
+For processing data that already exists on disk, you can use
+``sotodlib/workflows/toast_so_map.py`` as an example workflow. This loads data
+in the native toast HDF5 format and / or L3 books.
+
+.. note:: Loading data from books will be added in PR 183.
 
 
 Reference
@@ -124,6 +131,8 @@ instance usually contains detectors from one readout card / wafer.
 
 .. autoclass:: sotodlib.toast.SOFocalplane
    :members:
+
+.. note::  There are more extensive updates to the S.O. toast instrument classes coming in PR 183.
 
 
 Operators
--- a/sotodlib/toast/ops/corotator.py
+++ b/sotodlib/toast/ops/corotator.py
@@ -32,11 +32,22 @@ XAXIS, YAXIS, ZAXIS = np.eye(3)
 
 @trait_docs
 class CoRotator(Operator):
-    """The LAT focalplane projected on the sky rotates as the cryostat
-    (co-rotator) tilts.  Usually the tilt is the same as the observing
-    elevation to maintain constant angle between the mirror and the cryostat.
+    """Compute the LAT focalplane rotation.
 
-    This operator be applied *before* expanding the detector pointing
+    The optical design configuration of the LAT projected on the sky has the "O6" cryo
+    tube in the positive azimuth direction when the telescope is at 60 degrees
+    elevation.
+
+    If corotate_lat is False, then the projected focalplane will rotate on the sky.
+    The saved corotation angle will be zero.  If corotate_lat is True, then the
+    corotation angle is set to compensate for the rotation caused by the observing
+    elevation.
+
+    The rotation of the projected focalplane is given by:
+
+        R = elevation - offset - corotator angle
+
+    This operator should be applied *before* expanding the detector pointing
     from boresight.
 
     """
@@ -60,7 +71,7 @@ class CoRotator(Operator):
     corotator_angle = Unicode(
         "corotator_angle",
         allow_none=True,
-        help="Observation shared key for corotator_angle",
+        help="Observation shared key for corotation angle",
     )
 
     elevation = Unicode(
@@ -70,7 +81,8 @@ class CoRotator(Operator):
     )
 
     corotate_lat = Bool(
-        True, help="Rotate LAT receiver to maintain focalplane orientation"
+        True,
+        help="If True, rotate LAT receiver to maintain projected focalplane orientation"
     )
 
     def __init__(self, **kwargs):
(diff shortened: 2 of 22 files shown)


tskisner:
This PR against the toast3 branch adds fixes for the detector focalplane layout needed to produce consistent LAT tube geometry and rotation.  One change is that simulating the hardware model is now done with a `sim_nominal()` function, rather than the poorly named `get_example()`.  The returned model has no detectors (only the telescope / tube / wafer / band information).  Then one of several functions can be called to simulate detector locations.  The default one uses the toast geometry layout helper functions, but there is a stub for a "better" one using optics data.

One consequence of this is that to do the initial "full" instrument detector sim and generate a TOML / JSON file, toast is required.  After that file exists, it can be modified, trimmed, plotted, etc without the toast sub-module.  So I would advocate that we automate the sotodlib releases to append one of these instrument files to the release page (or some other place).

I am running a SAT test now.

workflows/toast_so_sim.py
```diff
@@ -221,44 +219,77 @@ def job_create(config, jobargs, telescope, schedule, comm):
...
+    ops.weights_azel.hwp_angle = ops.sim_ground.hwp_angle
+
+    ops.weights_radec.detector_pointing = ops.det_pointing_radec
+    ops.weights_radec.hwp_angle = ops.sim_ground.hwp_angle
+
+    # Select Pixelization and weights for solve and final binning
+
+    if ops.pixels_wcs_azel.enabled:
+        args.pixels_solve = ops.pixels_wcs_azel
+        args.weights_solve = ops.weights_azel
```
keskitalo, line 274: I'm confused. Where is `weights_solve` added to the `args` namespace?

tests/test_hardware.py
```diff
@@ -7,19 +7,24 @@
...
+from sotodlib.vis_hardware import plot_detectors
 
-from sotodlib.sim_hardware import get_example
+from ._helpers import create_outdir, mpi_multi
 
-from sotodlib.sim_hardware import (sim_wafer_detectors,
-                                   sim_telescope_detectors)
 
-from sotodlib.vis_hardware import plot_detectors
+toast_available = None
```
keskitalo, line 19: I think we have discussed this import guard before. Since you set `toast_available` to `None` and immediately check its value, the test is redundant and can be omitted. Furthermore, because of Python scoping rules, you don't even need to declare `toast_available` before the `try-except` clause.

docs/toast.rst
```diff
@@ -104,10 +104,14 @@ When running workflows in batch mode, we typically use a slightly different
 structure from a notebook.  There are helper functions in toast to parse
 operator options from the commandline and / or a config file, and to automatically
 create named instances of those operators for use in workflow.  There is
-a large-scale workflow in ``sotodlib/workflows/toast_so_sim.py``.  This
+a large-scale workflow in ``sotodlib/workflows/toast_so_map.py``.  This
```
keskitalo, line 107: I think you mixed `toast_so_map.py` and `toast_so_sim.py` on line 111.

sotodlib/toast/ops/sim_sso.py
```diff
@@ -131,6 +121,11 @@ def _check_detector_pointing(self, proposal):
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
+        # Store of per-detector beam properties.  Eventually we could modify the
```
keskitalo, line 124: I really appreciate these helpful comments.

sotodlib/toast/ops/sim_sso.py
```diff
@@ -278,62 +292,96 @@ def _get_sso_position(self, data, sso, times, observer):
...
             x = (az - sso_az.to_value(u.rad)) * np.cos(el)
             y = el - sso_el.to_value(u.rad)
-            r = np.sqrt(x ** 2 + y ** 2)
+            r = np.sqrt(x**2 + y**2)
+
             good = r < radius
             sig = beam(x[good], y[good], grid=False)
             signal[good] += sig
 
+            # if idet == 0:
```
keskitalo, line 363: Did you intend to drop these debug statements?

sotodlib/toast/ops/sim_sso.py
```diff
@@ -363,3 +411,137 @@ def _provides(self):
 
     def _accelerators(self):
         return list()
+
```
keskitalo, line 414: And these lines?

keskitalo (approved):
Thanks a ton on your thorough work on the focalplane simulation and the co-rotator!

I only have minor comments. Please see if they merit any action.

smsimon (approved):
There's been a desire from the instrument side to skip a detector number on the pin/slot (i.e. killed pixels) so that if pin/slot locations change (e.g. in Advanced SO) or we kill pixels that are dead in the model, the detector numbers don't change. Now might be a good time to implement that in sim_focalplane.py. I _think_ you just have to change doff to px in line 354 (or something similar) though we should check if it has any other implications.

sotodlib/toast/ops/corotator.py
```diff
@@ -86,35 +98,46 @@ def _exec(self, data, detectors=None, **kwargs):
             if obs.telescope.name != "LAT":
                 continue
 
+            el_rad = obs.shared[self.elevation]
+
+            if self.corotator_angle not in obs.shared:
```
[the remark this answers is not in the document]
tskisner: The `ensure()` method is for detector data (https://github.com/hpc4cmb/toast/blob/ee47b337eccf57f5911893e1b1b7021219ff5fe0/src/toast/observation_data.py#L608) where it can re-use the existing detector data buffer by re-assigning it to different detector sets.

workflows/toast_so_sim.py
tskisner: This is done in the new `select_pointing()` [function here](https://github.com/simonsobs/sotodlib/blob/d052fb851f1ccd565be3c1ae1e93deb4b0d8fed4/workflows/toast_so_sim.py#L272-L281)  Many of these workflow helper functions (which deal with runtime options configuring multiple operators) can move to within the toast submodule.  However, I wanted to do that in a future, smaller PR.

tests/test_hardware.py
tskisner: Good point, I removed that.  Also, in the `toast3_data` PR, I moved this check into the `_helpers.py` file for use in multiple tests.

docs/toast.rst
tskisner: Yes, clarified and fixed.

sotodlib/toast/ops/sim_sso.py
tskisner: I removed the commented out debugging statements from the operator itself, but left the commented out plotting function at the bottom of the source.  We could delete that later or move to some other source if it is useful.

workflows/toast_so_sim.py
keskitalo: I see. Sort of like a `workflow_tools` package? I'm very much in favor of that!

I was thrown off because symbols in the `args` namespace are usually declared via `argparse` but now we are expanding it to include symbols generated as the workflow progresses. That is fine with me but it is a slippery slope into using it as a vehicle for all information flow in the workflow. Perhaps we want to call it something else than `args` if we expand its meaning?

tskisner: Ok, I see your point on abusing the `args` namespace.  I propose storing these kinds of workflow variables in the `job` namespace that we create when loading configs / options and instantiating the operators and templates.  For the high-level set of functions in the toast submodule, these should be helper functions that work on configuring many operators together.  The old "`pipeline_tools`" code was basically trying to externally configure individual operators.  Now that operators have all their configuration built-in, the kinds of things we would put in `sotodlib.toast.workflow` would be things like:

- Return a list of all nominal simulation operators and their starting defaults
- Return a list of all nominal reduction operators and their starting defaults
- Select the pointing / weights being used depending on which operators are enabled at runtime
- Function to "save" or "load" data using whichever operator combination has been enabled at runtime

The `simulate_data()` and `reduce_data()` functions in the workflows are kind of specific to the workflow, but I guess we could have a `simulate_nominal_data()` or similar in the workflow submodule.  There is a balance between reducing boilerplate / duplication in the script and hiding what the script is doing.  If a workflow has a main() that just calls 3 functions, it is really hard to follow what it is actually doing.  On the other hand, perhaps "following what is going on" is better left to interactive notebooks, and these workflows should just be about running things while reducing code duplication across scripts.

I'll fix the use of the args namespace here, and we can discuss further the workflow submodule on the next PR.

sotodlib/toast/ops/corotator.py
```diff
@@ -86,35 +98,46 @@ def _exec(self, data, detectors=None, **kwargs):
...
+                    self.corotator_angle,
+                    shape=(obs.n_local_samples,),
+                    dtype=np.float64,
+                )
+
             if self.corotate_lat:
-                obs["corotator_angle"] = -obs["scan_el"]
+                # We are corotating the receiver.  We set this to a fixed angle
+                # based on the nominal elevation for the scan.
+                corot = obs["scan_el"].to_value(u.radian) - LAT_COROTATOR_OFFSET.to_value(u.radian)
```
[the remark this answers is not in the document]
tskisner: This is now changed to use the mean boresight elevation if the target / intended "scan_el" is not defined.

workflows/toast_so_sim.py
tskisner: This is now fixed

tskisner:
The changes in this PR are now rebased against the toast3 branch and have changes to make the code compatible with the latest toast-3 pre-release.  The `toast3_bcp` branch of pwg-scripts is now also compatible with this code, and the drone sim, Jupiter SAT, and Jupiter LAT examples are working:
drone_10_08
out_f150_i1_Jupiter
out_f030_w42_Jupiter

I will merge this PR after the upstream toast-3.0.0a5 wheels are available.

tskisner:
Upstream wheels of toast that support this are now on PyPI.  Merging.

PR #308, opened 2022-10-05
Add function for plotting flag statistics.

msilvafe:
This function uses the native stats function built into the `RangesMatrices` to plot the cut fractions for a given tod.
Example_Plot_Glitch_Stats

### DIFF
--- a/sotodlib/flag_utils.py
+++ b/sotodlib/flag_utils.py
@@ -5,7 +5,7 @@ import time
 from scipy import stats
 
 def plot_glitch_stats(tod, flags='flags', glitches='glitches', N_bins=30, save_path='./',
-                      save_name='glitch_flag_stats.png'):
+                      save_name='glitch_flag_stats.png', save_plot=True):
     """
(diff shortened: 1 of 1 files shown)


kmharrington (changes requested):
Two small changes, could you make saving the plot optional and returning the figure possible? 

When running this in a notebook I expect we'd just want it returned and not saved. Also returning the figure will let people make exits before saving if they'd like to. 

Thanks for setting this up!

msilvafe:
> Two small changes, could you make saving the plot optional and returning the figure possible?
> 
> When running this in a notebook I expect we'd just want it returned and not saved. Also returning the figure will let people make exits before saving if they'd like to.
> 
> Thanks for setting this up!

Ok I think I addressed both in [13e4ea3](https://github.com/simonsobs/sotodlib/pull/308/commits/13e4ea3035508f306788430cc5dcd3256356fe18). Let me know if this is ok to merge then.

kmharrington (approved):
works for me, thanks a bunch!

PR #310, opened 2022-10-12
Add smurf timestamps

msilvafe:
Added in functionality to process the high quality timing words from the Smurf timing system. I tested this on some data we took on when we were testing the timing system at 4kHz as below:
```
aman = load_smurf.load_file('/mnt/so1/data/ucsd-sat1/timestreams/16505/crate1slot3/1650583410_000.g3')
plt.plot(aman.timestamps, aman.primary['UnixTime'])
plt.xlabel('SMuRF EPICS Timestamps', fontsize = 16)
plt.ylabel('UNIX Timestamps', fontsize = 16)
```
Test_SMuRF_Timestamps

This data had the PPS level triggered not edge triggered (clearly visible in the plot) and the data wasn't downsampled so this should be tested on the data from SAT1 where these two issues are fixed

### DIFF
--- a/sotodlib/io/load_smurf.py
+++ b/sotodlib/io/load_smurf.py
@@ -117,7 +117,7 @@ def get_sample_timestamps(frame):
 
 def _file_has_end_frames(filename):
     ended = False
-    
+
     ## Search through file looking for stream closeout
     reader = so3g.G3IndexedReader(filename)
     logger.debug(f"Reading {filename} to look for observation end.")
@@ -344,7 +344,7 @@ class G3tSmurf:
             ## happens if there are no scan frames in file
             file_start = db_file.frames[0].time
             file_stop = db_file.frames[-1].time
-            
+
         db_file.start = file_start
         db_file.stop = file_stop
         db_file.n_channels = total_channels
@@ -701,7 +701,7 @@ class G3tSmurf:
                 obs_id=f"cal_{stream_id}_{session_id}"
             else:
                 obs_id=f"{stream_id}_{session_id}"
-            
+
             # Build Observation
             obs = Observations(
                 obs_id=obs_id,
@@ -752,13 +752,13 @@ class G3tSmurf:
             Files.name.like(prefix + "/"+ str(obs.timestamp)+"%")
         ).order_by(Files.start).all()
 
-        
+
         logger.debug(f"Found {len(flist)} files in {obs.obs_id}")
         if len(flist)==0:
             ## we don't have .g3 files for some reason, shouldn't be possible?
             logger.debug(f"Found no files associated with {obs.obs_id}")
             return
-        
+
         ## set start to be the first scan frame in file/observation
         obs.start = flist[0].start
 
@@ -1069,10 +1069,10 @@ class G3tSmurf:
                         f"Add new Observation: {stream_id}, {ctime}, {obs_path}"
                     )
                     self.add_new_observation(
-                        stream_id, 
-                        action, 
-                        ctime, 
-                        session, 
+                        stream_id,
+                        action,
+                        ctime,
+                        session,
                         calibration=(action in SMURF_ACTIONS["calibrations"])
                     )
                 except Exception as e:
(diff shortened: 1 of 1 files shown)


sotodlib/io/load_smurf.py
```diff
@@ -2125,20 +2133,32 @@ def _get_timestamps(streams, load_type=None, linearize_timestamps=True):
...
                 load_type = TimingParadigm.G3Timestream
         else:
             load_type = TimingParadigm.G3Timestream
 
+    if load_type == TimingParadigm.TimingSystem:
+        s, ns = split_ts_bits(io_load.hstack_into(None,
+                                streams["primary"]["Counter2"]))
+        # Add 20 years in seconds (accounting for leap years) to handle
+        # offset between EPICS time referenced to 1990 relative to UNIX time.
+        counter2 = s + ns*1e-9 + 5*(4*365 - 1)*24*60*60
```
kmharrington, line 2158: could we just hard code this number since it seems like it's not changing? 

You have double checked that the difference between the unix time and the timing system data is small right? (Making sure this isn't off by some weird amount).

sotodlib/io/load_smurf.py
```diff
@@ -2125,20 +2133,32 @@ def _get_timestamps(streams, load_type=None, linearize_timestamps=True):
...
-          if true and using unix timing, linearize the timing based on the 
+          if true and using unix timing, linearize the timing based on the
           frame counter
     """
     if load_type is None:
         # determine the desired loading type. Expand as logic as
         # data fields develop
         if "primary" in streams:
-            if "UnixTime" in streams["primary"]:
+            if np.diff(io_load.hstack_into(None,
```
kmharrington, line 2143: I'm concerned that this is the only way we know we're using the timing system? Is there not some sort of header information or status information that just says "we're using the timing system"? If it's something that we need to load from the status frame or something then we can add it into `load_file`, but if there's nothing.... what can we add to SMuRF to get us that info?

Also, you're sure this math is backwards compatible? It's not going to trip up data where there isn't a timing system?

kmharrington (approved):
I'm going to approve this PR because I know we need to get working with timing synced data at the SAT, but if you're not able to make the changes / questions in-line, please submit sotodlib issues for them before merging. Thanks!

sotodlib/io/load_smurf.py
jlashner: I think this is the best and easiest way to check for the time being. There is technically the "timing link up" variable, but I don't love relying on a status variable since that could potentially change mid-stream (though you could make the argument that it invalidates the data anyway...). Once we get more data and understand the failure modes better I'm totally fine changing this though.

That being said, the mean is probably not the right thing to use here since there are large negative diffs when the counter resets. `np.mean(np.abs(np.diff(Counter0)))` seems more reliable.

sotodlib/io/load_smurf.py
mhasself: That offset evaluates to 630288000.  But I would have expected
```date -u -d '1990-01-01' +%s
631152000
```
The equation should have + 1, not - 1, for the leap years.

msilvafe: Whoops, totally found this -1 error before when we were trying this out and you corrected me then but I propagated the error through in my haste. Fixed in [0666539](https://github.com/simonsobs/sotodlib/pull/310/commits/06665399a060ae4f1a6bfc97eb3e4739a3a21911) 

For Katie's comment I'm just going to leave in the math so that it's somewhat human readable how I got that number. We should inspect more when the timing system is all setup with the new software but it appears that the two timestamps (EPICS vs UNIX) are the same up to ~0.2s. Here's some plots showing things:

This plot shows the offset between the EPICS + UNIX timestamps: 
inspect_offset_zoom

This plot shows a histogram of the diff of the timestamps for EPICS (excluding the issues around the PPS reset) and UNIX:
ts_hist

This is the fixed version of the plot I put above but now the timestamps on the 2 axes match.
fixed_leap_year

PR #314, opened 2022-10-20
Add v0 of HWP angle loading

kmharrington:
Expand the g3tsmurf loading from configs a bit and add a utility function for loading up HWP angles. This assumes the HWP angles are saved in the format generated by the `update-hwp-angles` site pipeline script. 

The loading for the HWP data is _slow_. We're going to have to fix that soon-ish. But this does work and does get the HWP angles loading into the AxisManager. We can upgrade this function when we get faster methods.

### DIFF
--- a/sotodlib/io/g3tsmurf_utils.py
+++ b/sotodlib/io/g3tsmurf_utils.py
@@ -4,13 +4,15 @@
 import os
 import numpy as np
 import logging
+import yaml
+
+from so3g.hk import load_range
 
 from sotodlib.io.load_smurf import load_file, SmurfStatus
 from sotodlib.io.g3tsmurf_db import Observations, Files
-
+from scipy.interpolate import interp1d
 
 logger = logging.getLogger(__name__)
-
 
 def get_obs_folder(obs_id, archive):
     """
@@ -201,4 +203,42 @@ def get_batch(
                     )
     except GeneratorExit:
         pass
+    
+def load_hwp_data(aman, configs=None, data_dir=None):
+    """Load HWP data that will match with aman timestamps and 
+    interpolate angles into hwp_angle
+    
+    Arguments
+    ----------
+    aman : AxisManager
+        AxisManager of loaded SMuRF data
+    configs : (optional) config filename or dictionary containing "hwp_prefix"
+    data_dir : (optional) the "hwp_prefix" data directory. overrides configs
+    
+    Wraps HWP data into aman.hwp_angle
+    """
+    if "hwp_angle" in aman:
+        raise ValueError("hwp_angle already exists in aman")
+        
+    if configs is None and data_dir is None:
+        raise ValueError("Must pass config or data_dir")
+    if data_dir is None:
+        if type(configs)==str:
+            configs = yaml.safe_load( open(configs, "r"))
+        data_dir = configs["hwp_prefix"]
+        
+    data = load_range(
+        float(aman.timestamps[0]-5), float(aman.timestamps[-1]+5), 
+        fields=["hwp.hwp_angle"], data_dir=data_dir
+    )
+    time = data["hwp.hwp_angle"][0]
+    if len(time) == 0:
+        raise ValueError("No HWP data found that overlaps aman")
+    if time[0] > aman.timestamps[0] or time[-1] < aman.timestamps[-1]:
+        logger.warning(f"HWP data does not cover all of AxisManager, extrapolations" /
+                      "may be unstable")
+        
+    hwp = interp1d(data["hwp.hwp_angle"][0],data["hwp.hwp_angle"][1], bounds_error=False,
+                 fill_value="extrapolate" )
+    aman.wrap( "hwp_angle", hwp(aman.timestamps), [(0,"samps")])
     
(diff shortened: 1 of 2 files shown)


sotodlib/io/g3tsmurf_utils.py
```diff
@@ -202,3 +204,41 @@ def get_batch(
...
+    )
+    time = data["hwp.hwp_angle"][0]
+    if len(time) == 0:
+        raise ValueError("No HWP data found that overlaps aman")
+    if time[0] > aman.timestamps[0] or time[-1] < aman.timestamps[-1]:
+        logger.warning(f"HWP data does not cover all of AxisManager, extrapolations" /
+                      "may be unstable")
+        
+    hwp = interp1d(data["hwp.hwp_angle"][0],data["hwp.hwp_angle"][1], bounds_error=False,
+                 fill_value="extrapolate" )
```
mhasself, line 242: This sort of thing doesn't work on data with a branch cut, right?

A brute force solution is to "unwrap" the HWP angle and resample that, then put the result back in the [0, 2pi) branch.  float64 should have enough precision to handle that without adding noise, at least for ~days of data at a time.

sotodlib/io/g3tsmurf_utils.py
```diff
@@ -202,3 +204,41 @@ def get_batch(
...
+def load_hwp_data(aman, configs=None, data_dir=None):
+    """Load HWP data that will match with aman timestamps and 
+    interpolate angles into hwp_angle
+    
+    Arguments
+    ----------
+    aman : AxisManager
+        AxisManager of loaded SMuRF data
+    configs : (optional) config filename or dictionary containing "hwp_prefix"
+    data_dir : (optional) the "hwp_prefix" data directory. overrides configs
```
msilvafe, line 216: Can you point to an example `hwp_prefix` file or dictionary in the doc-string or readthedocs?

msilvafe (changes requested):
The main actual comment is just addressing what Matthew brought up. I think if you address that and have tested that it works this can be merged. 

Otherwise I think it would be useful if you included in the PR comment, doc string, or readthedocs  an example of what the loading in the presence of HWP data looks like (like what's the function call) and what is the structure of the config file.

sotodlib/io/g3tsmurf_utils.py
kmharrington: @msilvafe  any chance you have the code for that you could just paste into this and merge into this branch? I'm more interested in connecting all the IO dots and would rather leave the math to the SAT folks.

sotodlib/io/load_smurf.py
```diff
@@ -198,6 +198,8 @@ def from_configs(cls, configs):
         -----
         configs - dictionary containing `data_prefix` and `g3tsmurf_db` keys
         """
+        if type(configs)==str:
+            configs = yaml.safe_load( open(configs, "r"))
```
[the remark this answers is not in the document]
kmharrington: https://simons1.princeton.edu/confluence/display/CHILAT/SO%3A+Data+Packaging+FLP#SO:DataPackagingFLP-P10R1DataLoadingExample

sotodlib/io/g3tsmurf_utils.py
kmharrington: Actually, I guess you also haven't seen these changes? https://sotodlib.readthedocs.io/en/latest/hwp.html

sotodlib/io/g3tsmurf_utils.py
```diff
@@ -30,20 +32,22 @@ def get_obs_folder(obs_id, archive):
...
-        get_obs_folder(obs_id,archive),
+        get_obs_folder(obs_id, archive),
         "outputs"
     )
 
     if not os.path.exists(path):
-        logger.error(f"Path {path} does not exist, how does {obs.obs_id} exist?")
-    return [os.path.join(path,f) for f in os.listdir(path)]
+        logger.error(f"Path {path} does not exist, " +
+                     f"how does {obs.obs_id} exist?")
```
msilvafe, line 49: Also this logger is trying to return `obs.obs_id` but `obs` isn't defined.

sotodlib/io/g3tsmurf_utils.py
```diff
@@ -53,13 +57,15 @@ def get_obs_plots(obs_id, archive):
...
-        get_obs_folder(obs_id,archive),
+        get_obs_folder(obs_id, archive),
         "plots"
     )
 
     if not os.path.exists(path):
-        logger.error(f"Path {path} does not exist, how does {obs.obs_id} exist?")
-    return [os.path.join(path,f) for f in os.listdir(path)]
+        logger.error(f"Path {path} does not exist, " +
+                     f"how does {obs.obs_id} exist?")
```
msilvafe, line 66: Same here as line 49.

msilvafe (approved):
I'm going to go ahead and approve so that I'm not blocking merging but I would like to see this tested with the commits that I added in before it gets merged.

kmharrington:
looks good, and I've testing it again. Thanks Max

PR #315, opened 2022-10-21
update-hwp-angles writes 60s frames

mhasself:
Instead of a single frame for each file it processes.

Also added a "--file" option to force reprocess something.

This makes it possible to load short segments of the output ~efficiently.


mhasself:
> Approving but could you update the instructions of the file tag

Done...

@yukisakurai let me know if you intend to review this.

yukisakurai (approved):
I tried this branch locally and both the file argument and frame part worked as expected.
This is more efficient, thank you!

PR #317, opened 2022-10-25
HWP angle calculator with quad enable option

yukisakurai:
There is a hardware issue that makes frequently flipping quad value (rotation direction) in P10R1. This PR adds an option to enable quad or not in config file. This option should be basically on if hwp readout is working correctly, but it is useful for lab. testing and debug.
You need to include `enable_quad: False` to config yaml file if you want to turn off it.
Also I modified reference slit finding module to fix hwp_rate and updated docs.  
(This PR has been overwritten from RP#315.)

**Angle solution**
image
image
**hwp_rate**
image
image

### DIFF
--- a/docs/axisman.rst
+++ b/docs/axisman.rst
@@ -227,12 +227,13 @@ Pointing information required for :mod:`sotodlib.coords`.
 * | ``boresight_equ`` `[samps]` - AxisManager with boresight in equitorial
   |  coordinates.
 
+
 HWP information 
 
 * | ``hwp_angle`` `[samps]` - AxisManager with hwp rotation angle required for
   | :mod:`sotodlib.io.g3tsmurf_utils.load_hwp_data`
 * | ``hwpss_ext`` `[dets, samps]` - the fields of extracted HWP synchronous signal
-  | required for :mod:`sotodlib.hwp.extract_hwpss`
+  | derived from :mod:`sotodlib.hwp.extract_hwpss`
 
 ---------
 Reference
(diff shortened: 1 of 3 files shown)


kmharrington (changes requested):
I like the change with letting quad be fixed from the config file.

My change request is the commit history, it looks like you copied the lines of code from PR#315 into this and re-commited? A better way to do it is to merge that branch into yours. I believe that would be `git merge frame-ify-hwp hwp`. Then we'd be 100% sure everything from PR#315 was copied over and git would understand how to make the final merge. As is we're going to get merge conflicts when we try and merge both of these PRs. 

Could you remove your commit where you added in the #315 changes and merge the branch instead? Thanks!

kmharrington:
Just an update, since we've merged #315  you should be able to rebase this branch off of master, then we'll be good to go.

yukisakurai:
I rebased this branch from master and added the updates.

mhasself (approved):
The rebase against master looks like it was using a stale version of that branch.  I think you need to "git fetch", then "git rebase origin/master" and "git push --force".  Approving but please fix this first.

In cases where quad is disabled, the direction of spin is assumed to be positive ... but perhaps the direction to assume should be a configuration parameter?  Or at least some way to control it?  I defer to your experience on this though.

yukisakurai:
Thank you guys for your suggestions.
I processed "git fetch" → "git rebase origin/master" → "git push --force". There is no conflict now.
I have made it possible to force the rotation direction from the config.
`enable_quad` → `force_quad` (0 or 1 or -1)
ex) `force_quad: -1` if you want to force negative rotation direction

kmharrington (approved):
looks good, let's try it!

PR #318, opened 2022-10-26
site_pipeline.util: redirect StreamHandler to stdout

guanyilun:
Is there any reason we don't want the log stream to go to stdout by default? Going to stdout allows prefect to pick it up.

### DIFF
--- a/sotodlib/site_pipeline/util.py
+++ b/sotodlib/site_pipeline/util.py
@@ -3,6 +3,7 @@ import os
 import inspect
 import logging
 import time
+import sys
 from astropy import units as u
 
 from .. import core
@@ -251,7 +252,7 @@ def init_logger(name, announce=''):
     logger.propagate = False
     logger.setLevel(logging.DEBUG)
 
-    ch = logging.StreamHandler()
+    ch = logging.StreamHandler(sys.stdout)
     formatter = _ReltimeFormatter('%(asctime)s: %(message)s (%(levelname)s)')
 
     ch.setLevel(logging.INFO)


kmharrington (approved):
I definitely tend to prefer stdout.

mhasself:
Warnings / errors usually go to stderr so that they don't interfere with parseable output (for one) ... though this isn't really a pipeline where that would matter much.  Are you sure prefect can't pick things up from stderr?  That is hard to believe... many errors (such as standard python exceptions) end up on stderr.

kmharrington:
can we have it pick up both? being able to look at the info logs would have already been nice. Info really shouldn't go to stderr though.

mhasself:
Ya, I suppose logging in site_pipeline could go to stdout ... but other things will go to stderr.  Maybe add a "2>&1" to the end of your commands when you call them through prefect?  (The default wrapper shell seems to be bash.)

guanyilun:
> Are you sure prefect can't pick things up from stderr?

Yes it can pick things up from stderr, which is how Katie was able to paste link to tracebacks in prefect flow runs. Currently it's not getting the logging output because normal logs are not redirected to stdout.

mhasself:
> I notice the default logger in sotodlib is also not going to stdout https://github.com/simonsobs/sotodlib/blob/master/sotodlib/__init__.py
> 
> Which is why the hwp and g3tsmurf loggers aren't printing to prefect either.

I think generically it is appropriate and standard practice for warning and error messages to go to stderr.  The site_pipeline logger is being used for chatter / progress, so stdout is acceptable.

Yilun and I spoke about this and he is going to confirm that it's possible to capture stderr, too, when calling external scripts.  That will make sure we have everything in the prefect log, no matter if the script succeeded and no matter where the writes to stderr originated.

kmharrington:
I think I've just learned that I really dislike loggings defaults. I would like this:

```
from sotodlib.io.load_smurf import logger
logger.info("testing")
```

to just automatically print to the screen. (it doesn't even if I used logging outside of sotodlib).

guanyilun:
I checked that adding 2>&1 after commands allows prefect to capture logs properly, but I agree with Matthew that changing the default to stdout is more appropriate for what we are using it for. Merging it now!

mhasself:
> I think I've just learned that I really dislike loggings defaults. I would like this:
> 
> ```
> from sotodlib.io.load_smurf import logger
> logger.info("testing")
> ```
> 
> to just automatically print to the screen. (it doesn't even if I used logging outside of sotodlib).

Ya ... I don't totally understand how the hierarchy of loggers in sotodlib are related to each other.  But libraries are not supposed to be chatty so not printing "info" is a good default.  If you have an app using it where you think printing load_smurf's info is useful, you can do logger.setLevel('INFO'), just for that one submodule.  That is handy.  I think I only want to be forced to see warnings or worse from everyone... but I want to be free to use info and debug within any submodule.

mhasself:
Related to the above ... even if you do `logger.setLevel('DEBUG')` for load_smurf, you won't see debug messages printed unless you *also* change the log level of the parent logger's main handler: `logging.getLogger('sotodlib').handlers[0].setLevel('DEBUG')`.  (This because the handler is set to "info" in `sotodlib/__init__.py`.)

PR #320, opened 2022-10-26
Change default concatenation behavior for AxisManagers

kmharrington:
I would really like to change the default behavior of AxisManager concatenation because the current behavior is causing issues for what I'd like to save in my todo pre-processing scripts. Effectively I have a script that saves an AxisManager with `dets` and `dets,samps` fields along with an additional field aligned just along `samps` that I'd like to have available later through the metadata loading. 

When I try and load these the [loader is calling concatenate](https://github.com/simonsobs/sotodlib/blob/master/sotodlib/core/metadata/loader.py#L285) on them and the presence of the `samps` only field is causing an error. I don't think we can change the default behavior of the loader because it's able to call concatenate on AxisManagers, ResultSets, and numpy arrays. So instead I'd like to change AxisManager.concatenate. 

This PR adds a fourth option for what to do with fields we aren't concatenating along called `'exact'`. The idea is if the field is identical across all of the items we'll just add it to the new AxisManager. I added a bunch of test cases to make sure it was working for as many of the cases I think of, and it seems to all work. It deals with scalars and anything that has a `.shape` property.

### DIFF
--- a/sotodlib/core/axisman.py
+++ b/sotodlib/core/axisman.py
@@ -501,19 +501,27 @@ class AxisManager:
                 output.wrap(k, new_data[k], axis_map)
             else:
                 if other_fields == "exact":
-                    if not np.all([i[k].shape==items[0][k].shape for i in items]):
-                        raise ValueError(
-                            f"The field '{k}' does not share axis '{axis}'; "
-                            f"{k} is not identical across items "
-                            f"pass other_fields='drop' or 'first' or else "
-                            f"remove this field from the targets.")
-                    if not np.all([i[k]==items[0][k] for i in items]):
-                        raise ValueError(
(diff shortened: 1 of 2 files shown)


sotodlib/core/axisman.py
```diff
@@ -427,7 +430,7 @@ def concatenate(items, axis=0, other_fields='fail'):
           target axis).
 
         """
-        assert other_fields in ['fail', 'first', 'drop']
+        assert other_fields in ['exact','fail', 'first', 'drop']
```
mhasself, line 433: space after comma

mhasself (approved):
Makes sense!  Thanks for implementing this superior default behavior.

PR #324, opened 2022-11-03
Jump test fix

skhrg:
Fixes issue where jumpfinder test will randomly fail sometimes (see [slack discussion](https://simonsobs.slack.com/archives/C60J078P7/p1667416485276259) for details).

Main problem seems to be that I set the min jump size to just 2 sigma, that has been upped to 5.
The jumps themselves have also been made more extreme and the seed is set to 0 for consistency.


kmharrington (approved):
Thanks for the checks / updates!

PR #328, opened 2022-11-04
Use fftw backend for welch

skhrg:
Speeds up PSD computation by switching `scipy`'s backend to use `fftw`.

As far as I can tell this backend switch only matters for the `calc_wn` function so I could have set it just within that scope, but I set it globally just in case we add more stuff that uses `scipy` FFT stuff down the line.

Timing it (code at bottom) yields a 73.95/54.11 = ~1.37x speedup.
Not a tremendous amount but given that the example speedup in the [pyfftw docs](https://pyfftw.readthedocs.io/en/latest/source/tutorial.html#integration-with-3rd-party-libraries) is ~1.6 it seems about right.

I can keep looking into this but for now this is a pretty easy boost for free.

```
import timeit
setup = """
from sotodlib.tod_ops.fft_ops import calc_psd
from tests.test_tod_ops import get_tod
aman = get_tod("white")
"""
print(timeit.timeit(setup=setup, stmt="calc_psd(aman)", number=100000)
```

### DIFF
--- a/sotodlib/tod_ops/fft_ops.py
+++ b/sotodlib/tod_ops/fft_ops.py
@@ -15,7 +15,7 @@ logger = logging.getLogger(__name__)
 
 def _get_num_threads():
     # Guess how many threads we should be using in FFT ops...
-    return so3g.useful_info().get('omp_num_threads', 4)
+    return so3g.useful_info().get("omp_num_threads", 4)
 
 
 # Setup scipy operations to use fftw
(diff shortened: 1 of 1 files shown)


skhrg:
MPI tests ran for over an hour without finishing so I cancelled them. Will do some digging as to how I broke things.

BrianJKoopman:
> MPI tests ran for over an hour without finishing so I cancelled them. Will do some digging as to how I broke things.

Probably not related to your PR, they broke for everyone. @tskisner fixed them in #334. I'd suggest rebasing or merging master into here.

kmharrington (approved):
Every little bit helps. Thanks for looking into this!

skhrg:
> Probably not related to your PR, they broke for everyone. @tskisner fixed them in #334. I'd suggest rebasing or merging master into here.
> 
> Edit: And now I noticed the force-push.

Yeah rebasing fixed the stalling out problem but now it's segfaulting. Still in the midst of debugging.

It passes all tests when I call `mpirun -np 2 python setup.py test` on simons1 so it seems to be a GitHub actions thing.
If anyone has more experience debugging this sort of stuff I welcome the help.

sotodlib/tod_ops/fft_ops.py
```diff
@@ -1,17 +1,32 @@
...
     return so3g.useful_info().get('omp_num_threads', 4)
 
+
+# Setup scipy operations to use fftw
+try:
+    pyfftw.config.NUM_THREADS = _get_num_threads()
+    set_global_backend(pyfftw.interfaces.scipy_fft)
+    pyfftw.interfaces.cache.enable()
+except ValueError:
+    logger.debug("Unable to set pyffw backend")
```
mhasself, line 27: Please change the message to "Failed to configure pyfftw as scipy fft backend" (fix spelling, mention scipy).

skhrg:
Doing another round of testing with python 3.10, the difference between the pyfftw backend and vanilla scipy is negligible. I suspect this is related to the performance increases in recent versions of python (3.11 is even faster!).

So unless there is any opposition in the next day or so I will just close this without merging.

I am happy to keep looking into this. Implementing welch in C/C++ would be fairly straightforward. There is also [this](https://github.com/xmikos/simplespectral/tree/master), where someone essentially re-implemented the scipy code line by line with pyfftw (but I don't see how this is different than changing the backend).

PR #329, opened 2022-11-04
Drone scans

gabrielecoppi:
Update on the drone operator to include several drone movements (Attached a scan of the drone in elevation)

Other changes

- More accurate power output calculation (Other parameters are required now)
- Set the drone to keep constant distance from the telescope (based on field test this is feasible and definetly more convient)

[mapmaker_map.fits_I.pdf](https://github.com/simonsobs/sotodlib/files/9940452/mapmaker_map.fits_I.pdf)


tskisner:
The test failures in this PR should be fixed if you rebase against master:
```
git fetch
git checkout master
git rebase origin/master
git checkout drone_scans
git rebase origin/drone_scans
git rebase master
# Fix any conflicts (there should not be any in this case)
git push -f origin drone_scans
```
Or if you are using a merge-style workflow:
```
git fetch
git checkout master
git rebase origin/master
git checkout drone_scans
git rebase origin/drone_scans
git merge master
# Fix any conflicts (there should not be any in this case)
git push origin drone_scans
```

gabrielecoppi:
Ok re-submitted the PR with another small improvement (including noise for the source emission).  I was also looking at this [lines](https://github.com/simonsobs/sotodlib/blob/4616f73fb9aba9ffae2e59fa7ec436fe3ea73f95/sotodlib/toast/ops/sim_source.py#L992), can we just remove those? It seems to me that we are writing twice the source azimuth and elevation

tskisner:
Hi @gabrielecoppi , thanks for revisiting this PR.  I agree with your previous comment [about the lines of code here](https://github.com/simonsobs/sotodlib/blob/4616f73fb9aba9ffae2e59fa7ec436fe3ea73f95/sotodlib/toast/ops/sim_source.py#L992-L1004).  The source az/el coordinates are being stored in a toast "shared" data object, which means that only one copy will exist in memory on each node.  Lines 992 and 993 are storing a copy of this same information and on every process.  Not sure any downstream code expects those names, but you could comment them out and see if anything breaks.  The shared object containing the source position can be used by the toast WCS pixelization operator to make maps in source-centered coordinates, which is why it is nice to have it.

tskisner:
@gabrielecoppi , I am happy to help rebase this against master and fix any conflicts.  However, we should not do that until just before merging.  I was waiting to fix the wcs axis swap upstream in toast and then testing this PR before merging, but I think this has diverged so much from master that we should get it merged first and then make a new branch for further testing (and for the bug fix that @chervias wanted to make).

gabrielecoppi:
I am happy to merge it in to main. I agree that we should probably merge it and then add other fixes

PR #334, opened 2022-11-05
When running unit tests, cleanup MPI communicators.

tskisner:
This ensures that all communicators are cleaned up at the end of each unit test.  This seems to be needed on github CI services to avoid a deadlock when running MPI-enabled tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,6 +51,7 @@ jobs:
           python3 -m pip install --upgrade wheel
           python3 -m pip install mpi4py
           python3 -m pip install so3g
+          python3 -m pip install pshmem
           python3 -m pip install --pre toast
 
       - name: Install sotodlib
--- a/tests/_helpers.py
+++ b/tests/_helpers.py
@@ -78,6 +78,34 @@ def create_outdir(subdir=None, comm=None):
     return retdir
 
 
+def close_data_and_comm(data):
+    """Delete a toast data object AND the input Comm.
+
+    Multiple toast.Data objects can be created from a single toast.Comm,
+    and so the communicators are not freed when deleting the data object.
+    However for unit tests we frequently use a helper function to produce
+    a simulated dataset and then want to fully clean up that along with
+    the communicators that were used.  This is especially true on CI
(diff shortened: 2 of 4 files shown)


BrianJKoopman (dismissed):
Great, thanks for the fix!

BrianJKoopman (commented):
Actually, sorry, just noticed something I thought I'd ask about. The Actions logs are now filled with errors like:
```
mpi4py.MPI.Exception: Invalid communicator, error stack:
PMPI_Barrier(289): MPI_Barrier(comm=MPI_COMM_NULL) failed
PMPI_Barrier(248): Null communicator
Exception ignored in: <function MPIShared.__del__ at 0x7f6bfc8d61f0>
  File "/opt/hostedtoolcache/Python/3.8.14/x64/lib/python3.8/site-packages/pshmem/shmem.py", line 227, in __del__
    self.close()
  File "/opt/hostedtoolcache/Python/3.8.14/x64/lib/python3.8/site-packages/pshmem/shmem.py", line 360, in close
    self._nodecomm.barrier()
  File "mpi4py/MPI/Comm.pyx", line 1560, in mpi4py.MPI.Comm.barrier
  File "mpi4py/MPI/msgpickle.pxi", line 690, in mpi4py.MPI.PyMPI_barrier
```

Seems to be due to the clean up of the MPI communicator, but is this innocuous? Before tests were breaking these weren't showing up.

tskisner:
Hmmm, strange, those line numbers in the error don't match up with the source in the latest release of `pshmem`, where there is already a check in the destructor for a null communicator.  I don't see how the tests could be using an old version, will check the logs.

tskisner:
OK- apparently installing the pre-release of toast allows pip to install the pre-release of any dependencies.  There is indeed a pre-release of pshmem available that we are testing at NERSC, and apparently it does have a bug that was triggered here.  I'll update the scripts to work around that.  Somewhat of a corner case of a corner case!

tskisner:
Ok, errors are gone now.

BrianJKoopman (approved):
Awesome, thanks for the fix on the fix!

Interesting to know about the pre-release installation.

PR #335, opened 2022-11-07
Fixing exception handling and others

yukisakurai:
I was pointed out that update_hwp_angle.py has stopped in the middle of the process.
I modified some exception handlings, and also I fixed rotation direction handling.

### DIFF
--- a/sotodlib/hwp/g3thwp.py
+++ b/sotodlib/hwp/g3thwp.py
@@ -90,8 +90,9 @@ class G3tHWP():
         # 1: positive rotation direction, -1: negative rotation direction
         self._force_quad = int(self.configs.get('force_quad', 0))
         if np.abs(self._force_quad) > 1:
-            logger.error("force_quad in config file must be 0 or 1 or -1")
-            sys.exit(1)
+            logger.warning("force_quad in config file must be 0 or 1 or -1")
+            if self._force_quad > 1: self._force_quad = 1
+            else: self._force_quad = -1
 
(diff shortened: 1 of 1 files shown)


sotodlib/hwp/g3thwp.py
```diff
@@ -90,8 +90,9 @@ def __init__(self, config_file=None):
         # 1: positive rotation direction, -1: negative rotation direction
         self._force_quad = int(self.configs.get('force_quad', 0))
         if np.abs(self._force_quad) > 1:
-            logger.error("force_quad in config file must be 0 or 1 or -1")
-            sys.exit(1)
+            logger.warning("force_quad in config file must be 0 or 1 or -1")
+            if self._force_quad > 1: self._force_quad = 1
```
kmharrington, line 94: PEP8 standards say no single line ifs / elses like this. 
```
if conditional:
    do_things()
else:
    do_other_things()
```

kmharrington (approved):
Sorry I missed the added `sys.exits` reviewing the previous PR. Let's never use those as a rule. 

Going to approve this because I know we need these angle calculators running at UCSD. But please fix the if statement formatting. Single line ifs/elses aren't allowed in PEP8 and make the code harder to read.

sotodlib/hwp/g3thwp.py
yukisakurai: Applying autopep8 fixes this.

PR #326, opened 2022-11-03
import bookbinder from a separate repo

guanyilun:
bookbinder has been managed in a separate repo: https://github.com/simonsobs/bookbinder. As discussed in data flow call, the plan is to manage it under sotodlib.

PR #337, opened 2022-11-11
import bookbinder from a separate repo (with commit history)

guanyilun:
Similar to #326 but preserves commit history.


mhasself:
The test failure is because when you're using the pip-installable so3g, there is no independent spt3g package.  It can be fixed by always importing so3g before you import spt3g; i.e.
```
import so3g  # noqa: F401
from spt3g import core
```

algadoreo (approved):

mhasself:
The FLP name for bookbinder is "make-obs-book".  Can you rename the site_pipeline file to "make_obs_book.py"?

If this same script is going to make all kinds of books (I think we expect it will), then just "make_book.py".

Also please put the code in a main(args=None) function, and call that from the `if __name__ ...` block.

We will, eventually, hide all of these behind a single command line interface (so it can be invoked as "so-site-pipeline make_book <args>" or similar.

mhasself (changes requested):
(See comment in PR)

guanyilun:
fixed!

sotodlib/site_pipeline/make_book.py
```diff
@@ -0,0 +1,52 @@
+import os, os.path as op
+import argparse
+from sotodlib.iobookbinder import Bookbinder
```
mhasself, line 3: This line has a typo.  Has this been tested?

mhasself (changes requested):
Thanks for this -- see comments.

sotodlib/site_pipeline/make_book.py
guanyilun: ah sorry I didn't test the executable script because I'm intending to rewrite it in a new PR, but I've fixed the typo and import now.

mhasself (approved):

PR #339, opened 2022-11-14
Make focal plane

skhrg:
First pass at make focal plane.

One thing to note (with more details in comments in the code) are that I made some assumptions about the format of my inputs that aren't currently true. Feedback on whether or not it makes sense to assume these things (or if I should load in this information in a different way) is much appreciated.

There are also a couple of features that I have kinda implement but not fully: outlier detection, more complex priors, and applying the optics model.

Outlier detection is partially tested but my main pain point there is what to do with the outliers. Do I add another field to the AxisManager that contains a mask for them? Or is better to just forgo it at this step and instead just include the likelihood for each `readout_id -> det_id` match and let the stages down the line decide how they want to deal with low likelihood matches.

More complex priors aren't too hard to add in, I just need to figure out what sort of priors actually help. I am in the process of testing this but it may take some time.

Applying the optics model offered a minor but non-zero improvement in my previous testing and I already have the needed data and code for the LAT (and have spoken to SAT people about getting the equivalent from them). But it is more of a low priority extra since it only really matters if we have relatively high error on the measured pointing and polarization angles. The main thing I need to figure out here is how to avoid having a clunky interface.

### DIFF
--- a/sotodlib/site_pipeline/make_focal_plane.py
+++ b/sotodlib/site_pipeline/make_focal_plane.py
@@ -173,12 +173,16 @@ def main():
         "--instrument",
         help="Instrument mapping from sky to focal plane, feature not currently implemented",
     )
+    # Should this point to a config file?
+    parser.add_argument(
+        "-p", "--priors", help="Amount to set prior to from detmap", default=1
+    )
     args = parser.parse_args()
 
     # Load data
     aman = AxisManager.load(args.pointing_data)
     g3u.add_detmap_info(aman, args.detmap)
-    bg_map = np.load(bg_path, allow_pickle=True).item()
+    bg_map = np.load(args.bias_map, allow_pickle=True).item()
 
     # TODO: apply instrument to pointing if availible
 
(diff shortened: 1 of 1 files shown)


sotodlib/site_pipeline/make_focal_plane.py
```diff
@@ -0,0 +1,264 @@
...
+    xy_rs[:, 0] -= xy_rs[:, 0].min()
+    xy_rs[:, 1] /= xy[:, 1].max() - xy[:, 1].min()
+    xy_rs[:, 1] -= xy_rs[:, 1].min()
+    return xy_rs
+
+
+def gen_priors(aman, prior, method="flat", width=1, basis=None):
+    """
+    Generate priors from detmap.
+    Currently priors will be 1 everywhere but the location of the detmap's match.
```
Zhilei, line 61: I guess the location of the detmap's match is infinite? Meaning that we trust the detmap result 100% for now?

sotodlib/site_pipeline/make_focal_plane.py
```diff
@@ -0,0 +1,264 @@
...
+
+def match_template(
+    focal_plane, template, out_thresh=0, avoid_collision=True, priors=None
+):
+    """
+    Match fit focal plane againts a template.
+
+    Arguments:
+
+        focal_plane: Focal plane after either having optics model applied or being fit with channel map.
```
Zhilei, line 108: I assume this focal_plane argument will contain the detmap information?

Zhilei (commented):
Thank you so much for pushing this through! It contains the scope of make-focal-plane and beyond.
I am wondering whether the interface with the time-domain results file and the detmap results can be more explicit. Please point it out if I have missed them!

sotodlib/site_pipeline/make_focal_plane.py
skhrg: It's whatever gets passed in as `prior` right now. In the case where we want to trust the detmap 100% there is a flag that can be passed that will skip all the fitting stuff and just copy the `det_id` from the detmap.

sotodlib/site_pipeline/make_focal_plane.py
skhrg: Sorry these argument names haven't changed since a very early version and may be unclear. Docstrings are also out of date, I'll update them soon.
`focal_plane` is the pointing information that you produce.
`template` is the designed `det_x`, `det_y`, and `angle` of the detectors. I currently pull these from the values in the detmap. There are probably other sources that I could pull this from if we want things to work even if there is no detmap provided.

sotodlib/site_pipeline/make_focal_plane.py
```diff
@@ -0,0 +1,264 @@
...
+def main():
+    # Read in input pars
+    parser = ap.ArgumentParser()
+
+    # NOTE: Eventually all of this should just be metadata I can load from a single context?
+
+    # Making some assumtions about pointing data that aren't currently true:
+    # 1. I am assuming that the HDF5 file is a saved aman not a pandas results set
+    # 2. I am assuming that it contains the measured detector polangs
+    # 3. I am assuming it comtains aman.det_info
```
skhrg, line 161: @Zhilei here are the interface assumptions I am making that are currently incorrect. I imagine we will land on something between what currently exists and what I am assuming here but I wanted to make the assumptions that required the least extra code here for the first pass.

Zhilei (commented):
Looks good!

sotodlib/site_pipeline/make_focal_plane.py
Zhilei: This makes sense, thanks for taking the first step with what we have!

Zhilei (approved):
OK, thanks again for the great work @skhrg 
Since this adds a new file, I think we can move this to merge (unless there are any objections).
From here, I think it will be useful if _one example_ of using these functions can be shared, probably under the pwg-scripts repo.

skhrg:
I have an old example here https://github.com/simonsobs/pwg-scripts/tree/master/pwg-bcp/make-focal-plane
But that calls the functions directly, I will add an updated example soon.

skhrg:
Just noticed that @Zhilei marked this as approved. But I'm not sure if merging right now is the right move given the discussion on loading metadeta like pointing from the call 2 weeks ago. 

Will defer the decision to @Zhilei and @kmharrington

PR #341, opened 2022-11-16
modified demod code to support reverse rotation

tterasaki:
The demodulation code was only valid for forward rotation. That was because in the calculation of hwp speed, diff of hwp_angle was moded by 2 pi. If diff of hwp_angle is negative, that became close to 2 pi. The calculated speed is used for center frequency of bandpass filter. So that resulted in too large center frequency of bandpass filter.

I modified this by adding pi before mod and subtracting pi after mod. Besides, add bpf_center to arguments as manual input of center frequency of bandpass filter.

### DIFF
--- a/sotodlib/hwp/hwp.py
+++ b/sotodlib/hwp/hwp.py
@@ -117,8 +117,8 @@ def demod(aman, signal='signal', name='signal_demod',
         aman.move(name, None)
     
     if bpf_center is None:
-        speed = (np.sum(np.abs((np.diff(aman.hwp_angle) + np.pi)%(2*np.pi) - np.pi)) / 
-                 (aman.timestamps[-1] - aman.timestamps[0]))/ (2 * np.pi)
+        speed = (np.sum(np.abs(np.diff(np.unwrap(aman.hwp_angle)))) /
+                 (aman.timestamps[-1] - aman.timestamps[0])) / (2 * np.pi)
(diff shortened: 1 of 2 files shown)


sotodlib/hwp/hwp.py
```diff
@@ -112,16 +115,19 @@ def demod(aman, signal='signal', name='signal_demod',
         aman.move(name + '_wo_lfilt', None)
     if name in aman.keys():
         aman.move(name, None)
-
-    speed = (np.sum(np.diff(aman.hwp_angle) % (2 * np.pi)) /
-             (aman.timestamps[-1] - aman.timestamps[0])) / (2 * np.pi)
-    prelfilt = tod_ops.filters.low_pass_butter4(fc=4 * speed + bpf_width)
+    
+    if bpf_center is None:
+        speed = (np.sum(np.abs((np.diff(aman.hwp_angle) + np.pi)%(2*np.pi) - np.pi)) /
```
yukisakurai, line 120: np.unwarp can be used to this process
speed = (np.sum(np.abs(np.diff(np.unwrap(aman.hwp_angle)))) /
               (aman.timestamps[-1] - aman.timestamps[0])) / (2 * np.pi)

yukisakurai (approved):
Approved. I modified one point by myself, so please check it before merging.

PR #342, opened 2022-11-17
Add mod2pi option and fix handling of encoder instance name 

ykyohei:
1
Fixed the default instance of `load_data` from `HAB` to None.

The instance name should be initialized by config.yaml, default values, or explicitly specified in load_data function.
Therefore the following error looks unnecessary.
> else:
>             logger.error("Can not find field instance")
>             return {}

2
Added mod2pi option to the `g3thwp.analyze`
I added this option because the monotonically increasing hwp angle is useful for evaluating hwp angle jitter.

### DIFF
--- a/sotodlib/hwp/g3thwp.py
+++ b/sotodlib/hwp/g3thwp.py
@@ -101,7 +101,7 @@ class G3tHWP():
         self._output = self.configs.get('output', None)
 
     def load_data(self, start=None, end=None,
-                  data_dir=None, instance='HBA'):
+                  data_dir=None, instance=None):
         """
         Loads house keeping data for a given time range and
         returns HWP parameters in L2 HK .g3 file
@@ -145,9 +145,6 @@ class G3tHWP():
                 self._field_instance = instance
             else:
                 self._field_instance = 'observatory.' + instance + '.feeds.HWPEncoder'
-        else:
-            logger.error("Can not find field instance")
-            return {}
 
         # load housekeeping data with hwp keys
         logger.info('Loading HK data files ')
@@ -281,7 +278,7 @@ class G3tHWP():
 
         return data
 
-    def analyze(self, data, ratio=0.25, fast=True):
+    def analyze(self, data, ratio=0.25, mod2pi=True, fast=True):
         """
         Analyze HWP angle solution
         to be checked by hardware that 0 is CW and 1 is CCW from (sky side) consistently for all SAT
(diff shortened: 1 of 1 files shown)


sotodlib/hwp/g3thwp.py
```diff
@@ -790,21 +790,20 @@ def _calc_angle_linear(self):
...
+        angle_last_revolution = (self._encd_cnt_split[-1] - self._ref_cnt[-1]) * \
+            (2 * np.pi / self._num_edges) % (2 * np.pi) + len(self._ref_cnt) * 2 * np.pi
+        self._angle = np.array([((self._encd_cnt_split[i] - self._ref_cnt[i]) * \
+                          (2 * np.pi / np.diff(self._ref_indexes)[i - 1]) \
+                              % (2 * np.pi) + i * 2 * np.pi ).flatten() \
+                                  for i in range(1, len(self._encd_cnt_split) - 1)])
+        self._angle = np.concatenate([angle_first_revolution, self._angle.flatten(), angle_last_revolution])
+        self._angle = direction * self._angle
+        if mod2pi:
+            self._angle = self._angle % (2 * np.pi)
```
yukisakurai, line 804: % (2*np.pi) is already applied at previous process.
Original motivation of this line is to deal in case of direction = -1.
So we should do like below if we want to implement mod2pi option:

self._angle = direction * self._angle % (2 * np.pi)
if not mod2pi: self._angle = np.unwrap(self._angle)

I think it's ok to add this option, but user can easily do this outside of this script.
What do you think?

sotodlib/hwp/g3thwp.py
```diff
@@ -790,21 +790,20 @@ def _calc_angle_linear(self):
...
+
+        ii = np.where(np.diff(self._ref_indexes) != 1140)[0]
```
yukisakurai, line 806: This line was deleted in current master branch. I just forget to delete in debug process...

ykyohei: Okay this is my fault I'll delete again.

sotodlib/hwp/g3thwp.py
ykyohei: I added this option not to do the mod2pi.
My opinion is that "Not doing mod2pi on angle at all" and "np.unwrap(angle%2pi)" is not an equivalent operation and former better for detecting some kinds of error.

Actually, I didn't know about the `np.unwrap` so I think it's fine to delete this optio. 
I agree that this is minor point.

yukisakurai: Ah I agree that "not doing" and "unwrap" is not completely same for debug point of view.
OK, let's add this option.

PR #343, opened 2022-11-17
fix hwp ref index

ykyohei:
I found a bug in the hwp angle calculation and fixed it.
The ecoder ref_index is shifted by 1 and this is making an angle glitch once in one hwp revolution.
Please see the following notebook for the difference after debug.

https://gist.github.com/ykyohei/54f91be27dc9bf180dc7d6d196014b5f


yukisakurai (approved):
I left offset=0 in debug process. As you can see, I defined offset to intend to set to 1... Thanks for pointing that out.

PR #344, opened 2022-11-23
Update g3thwp

ykyohei:
I added `_key_formatting`, `_data_formatting` and `_slowdata_process` for maintainability.
This modification includes small bugfixes.
I have tested in several conditions, and it seems working same as before.
I hope these modifications doesn't affect the data reduction and packing.

I added a function (`eval_angle`) to evaluate and subtract the non-uniformity of hwp angle.
this function is assumed to be called after `analyze` before `write_solution`.
I would like to hear your opinion about where to implement this function.

I think this subtraction is important to satisfy the requirement of angle jitter especially when the uniformity is close to worst.
The attached plot shows the comparison of hwp angle jitter.
Figure 71 (1)

### DIFF
--- a/sotodlib/hwp/g3thwp.py
+++ b/sotodlib/hwp/g3thwp.py
@@ -284,14 +284,15 @@ class G3tHWP():
 
         Args
         -----
-            data : dict
-                HWP HK data from load_data
-            suffix: Specify whether to use 1st or 2nd encoder, '' or '_2'
-                '' for 1st encoder, '_2' for 2nd encoder
+        data : dict
+            HWP HK data from load_data
+        suffix: Specify whether to use 1st or 2nd encoder, '' or '_2'
+            '' for 1st encoder, '_2' for 2nd encoder
+
         Returns
         --------
-            dict
-                {'rising_edge_count', 'irig_time', 'counter', 'counter_index'}
+        dict
+            {'rising_edge_count', 'irig_time', 'counter', 'counter_index'}
         """
         keys = ['rising_edge_count', 'irig_time', 'counter', 'counter_index']
         if 'irig_time'+suffix not in data.keys():
@@ -322,15 +323,15 @@ class G3tHWP():
 
         Returns
         --------
-            dict
-                {stable, locked, hwp_rate, slow_time}
+        dict
+            {stable, locked, hwp_rate, slow_time}
 
         Notes
         ------
-            - Time definition -
-            if fast_time exists: slow_time = fast_time
-            elif: irig_time exists but no fast_time, slow_time = irig_time
-            else: slow_time is per 10 sec array
+        - Time definition -
+        if fast_time exists: slow_time = fast_time
+        elif: irig_time exists but no fast_time, slow_time = irig_time
+        else: slow_time is per 10 sec array
         """
         # hwp speed calc. (approximate using ref)
         hwp_rate_ref = 1 / np.diff(fast_time[self._ref_indexes])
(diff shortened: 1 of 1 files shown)


sotodlib/hwp/g3thwp.py
```diff
@@ -322,145 +424,73 @@ def analyze(self, data, ratio=0.25, mod2pi=True, fast=True):
...
             logger.warning('There is no correct IRIG timing. Stop analyze.')
             return {}
-        if len(irig_time) < len(irig_time_2):
-            logger.info('Use 2nd encoder IRIG timing.')
-            irig_time = irig_time_2
-            rising_edge = rising_edge_2
-            counter = data['counter_2']
-            counter_index = data['counter_index_2']
-            quad_time = data['quad_2'][0]
-            quad = self._quad_form(data['quad_2'][1])
```
ykyohei, line 375: comment about small bug I found.

Previously there was a small bug of using 2nd encoder as follows
`data['counter_2'] =>data['counter_2'][1]` etc

sotodlib/hwp/g3thwp.py
```diff
@@ -322,145 +424,73 @@ def analyze(self, data, ratio=0.25, mod2pi=True, fast=True):
...
-            hwp_rate = np.zeros(len(fast_irig_time), dtype=float)
-        else:
-            fast_irig_time = []
-            locked = []
-            stable = []
-            hwp_rate = []
+        # hwp angle calc.
+        if ratio is not None:
+            logger.info(f"Overwriting reference slit threshold by {ratio}.")
+            self._ref_range = ratio
```
ykyohei, line 441: comment about small bug I found.

Previously the `ratio` parameter of `analyze` was not doing anything.
I modified to be able to overwrite the ratio. 
This option is useful in debug process.

sotodlib/hwp/g3thwp.py
```diff
@@ -850,7 +880,7 @@ def _quad_form(self, quad):
         quad[(quad < 0.5)] = 0
         offset = 0
         logger_bit = True
-        for quad_split in np.array_split(quad, len(quad) / 100):
+        for quad_split in np.array_split(quad, 1 + np.floor(len(quad) / 100)):
```
ykyohei, line 883: Comment about small bug I found

Previously when len(quad) < 100, np.array_split tries to split array into zero splits and raise error.
I added 1 to avoid error.

yukisakurai (changes requested):
Thank you for your effort. The code readability is greatly improved.
I think you simplified multiple lines to single lines, but some of them are too long line.
We should modify them or run `autopep8` to follow pep8 formatting.

We should test this branch by running `update_hwp_angle.py ` with all of P10R1 and TSAT data for not to stop data packing.
As we did local discussion, I will test P10R1 and @ykyohei will test TSAT data.
After that, let's merge this version into master.

Regarding `eval_angle`, it is on my to do list, so thanks for bringing this on a table.
I think your "moving_average per one rotation" procedure would be  fine to eliminate angle non-uniformity due to encoder, but this is assuming that there is no rotation wobble due to rotator. 
Since this is related to systematics, it is good to include the function in the class, but wait to add it to `analyze` after a thorough discussion of the function's internals.
The name of `eval_angle` might be good to rename.

ykyohei:
We have tested to run `update_hwp_angle.py` with P10R1 and recent TSAT data, and the data packing passed without errors except for the existing known ones.
I think at least the data processing & packing part of this PR is ready to be merged.
Is the pep8 format the only one left task?

ykyohei:
I have implemented the template subtraction of the non-uniformity of encoder slits.
Also `g3thwp.py` follows pep8 format now.

Here is the link to the demonstration of template subtraction.
https://gist.github.com/ykyohei/0dcb65606f6e96897c18d49d152209f9

yukisakurai (approved):
Perfect, thank you. Please merge this version.
Let's keep touching about aliasing treatment in downsampling.

PR #347, opened 2022-12-01
Add operation searches for standard operations

kmharrington:
I had  been debating adding then but worth it so folks don't need to copy paste this around their own code. Adds three functions to automatically find the last bias group map, bias step analysis, and iv analysis relative to a give observation. This uses searches / functionality that is being defined as the "standard" way to index and find these files at level 2, but that means you need a relatively recent copy of sodetlib (currently p10r1 branch) for your files to be made with the "correct" information for these searches. But will be useful for p10r1 data and later cooldowns.

### DIFF
--- a/sotodlib/io/g3tsmurf_utils.py
+++ b/sotodlib/io/g3tsmurf_utils.py
@@ -6,6 +6,9 @@ import numpy as np
 import logging
 import yaml
 
+from sqlalchemy import desc
+
+
 from so3g.hk import load_range
 
 import sotodlib.core as core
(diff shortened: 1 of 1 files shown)


tterasaki (approved):
Thank you Katie! I checked it with P10R1 data and it worked well.

sotodlib/io/g3tsmurf_utils.py
```diff
@@ -68,6 +71,89 @@ def get_obs_plots(obs_id, archive):
...
+        "oper,bias_steps%",
+        session
+    )
+    if oper is None:
+        logger.error(f"Unable to find Bias Step associated with {my_obs_id}")
+    
+    files = get_obs_outputs(oper.obs_id, SMURF) 
+    file = [f for f in files if "bias_step_analysis.npy" in f]
+    if len(file) != 1:
+        logger.error("fUnable to find analysis file in {get_obs_folder(oper.obs_id, SMURF)}")
```
msilvafe, line 128: Same typo @guanyilun pointed out here and in iv one too.

msilvafe (approved):
Looks good, just fix f-string typo then happy to have merged.

Issue #346, opened 2022-11-30
make_det_info_wafer and make_read_det_match failing to match up with add_det_info() in g3tsmurf_utils


davidvng:
Attempting to create databases using `make_det_info_wafer.py` and `make_read_det_match.py` and comparing the resulting wafer info to the output of `add_det_info()` in `g3tsmurf_utils`. In an example case, I'm finding that the first method is cutting channels due to not finding matching readout IDs to detector IDs. (1698 IDs as opposed to 1808 IDs using `add_det_info()`). I tried to add a "NO_MATCH" det_id entry to `make_det_info_wafer.py` but that only adds a single ID after loading in using context.

Corresponding PR to all of this: #345 

@kmharrington and I are hoping @mhasself could comment on this?

kmharrington:
A bit more context here (pun totally intended). We found this comparing the AxisManagers loaded straight from files vs the ones loaded through Context. The context load pattern follows the one in the docs here: https://sotodlib.readthedocs.io/en/latest/site_pipeline.html#context-file-format 

Right now, the `readout_2_detector_manifest.db` assigns a `det_id` of `"NO_MATCH"` to any readout_id where DetMap doesn't find a corresponding detector.  If we just use the the first manifest db, 
```
metadata:
    - db: "/path/to/readout_2_detector_manifest.db"
      det_info: true
      det_key: "dets:readout_id"
```
this gets us an AxisManager with all 1808 channels present. But 110 of those channels have `aman.det_info.det_id == "NO_MATCH"`. 

When we add the wafer info back:
```
metadata:
    - db: "/path/to/readout_2_detector_manifest.db"
      det_info: true
      det_key: "dets:readout_id"
    - db: "/path/to/det_info/wafer/manifest.db"
      det_info: true
      det_key: "dets:det_id"
```

The loader quietly enforces the constraint that each `det_id` in the AxisManager has an entry in the wafer manifest _and_ that that entry is unique. So, if the wafer manifest doesn't have any entry for "NO_MATCH" then the AxisManager is just returned with 1698 channels and there's no notice that we're cutting some just because of lack of entries. If we have a single "NO_MATCH" entry in the wafer manifest, we get 1699 channels back where one of those "NO_MATCH"s were kept. 

At a minimum, I don't think we want the databases quietly doing data selection for us, there should be some feedback about what's happening. But there's an additional question of if it's possible and/or reasonable to remove that uniqueness constraint as well.

mhasself:
[...]

PR #350, opened 2022-12-09
Fix det info multicast

mhasself:
As mentioned in #346, det_info indices were enforced to be unique, which prevented broadcasting place-holder data using det_id='NOT_FOUND'.

This PR fixes that, tests it, and fixes a couple other minor things.

Check the docs for new context syntax... you want to set multi: true on the detector wafer info table, not the det_id-readout_id mapping input.

### DIFF
--- a/docs/context.rst
+++ b/docs/context.rst
@@ -849,32 +849,24 @@ contain only the ``obs_id``.  For each item in the context.yaml
 Changing det_info
 -----------------
 
-Metadata entries can be used to change the active ``det_info`` table
-that is used to screen and match metadata results.  This is
-accomplished through special metadata entries with the following
-syntax::
+Metadata entries can be used to augment the  ``det_info`` table
+that is used to screen and match metadata results.  For example::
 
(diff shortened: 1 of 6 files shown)


kmharrington (approved):
This looks really good to me, excited to test it out. 

@davidvng  could you merge this into your read-det-match branch, update the context, and see if this fixes the multicasting issues we were seeing?

I notice Matthew uses `"NOT_FOUND"` and I had started with `"NO_MATCH"`, do we care about standardizing them? The actually words don't matter to me, I think the convey the same message.

mhasself:
> I notice Matthew uses `"NOT_FOUND"` and I had started with `"NO_MATCH"`, do we care about standardizing them? The actually words don't matter to me, I think the convey the same message.

Arg, sorry I didn't mean to start a new thing.  We can go with NOT_FOUND if that's already in master anywhere.

kmharrington:
`NO_MATCH` is what's already in master in `site-pipeline`

PR #351, opened 2022-12-09
HK Utils: Turn HK (cosampled) .g3 data into AxisManagers

sanahabhimani:
This is a first iteration of turning HK .g3 data and outputting a slew of axismanagers. Arguments are the same as `load_range()` (start, stop, config file).  The functions:

1. Group the fields from a config file by device/feed name, 
2. Test whether data per device/feed is cosampled, and 
3. Output either (a) an axismanager per device for cosampled data, or (b) an axismanager per field/channel for a device that doesn't have cosampled data (i.e., 372 channels)

Couple Things to Note:
1. Testing whether the data is cosampled doesn't assert that all cosampled time stamps are the same because I noticed very slight changes in timestamps for devices that we know output cosampled data. Instead, it checks that the length of the time arrays of channels in a device are the same. We should discuss whether we want to make this assertion more robust. For example, maybe you take the `diff` of the times for channels in a device, do some more math, decide on some threshold value that if the timestamps are not far off within a certain limit, consider it as cosampled.
2. The final product of this hk tool is to output an axismanager of axismanagers. Right now, it just outputs a list of axismanagers. 
3. The reconfiguring I've done in order to group data by device may be worth a discussion on what `load_range()` could further output? `load_range()` outputs a dictionary of alias keys and data; to properly group the field names by device, I make a new dictionary of field names, aliases, and data. 
4. At the moment, this code requires a config file but it shouldn't be that much more effort to add arguments for a few field names and aliases


sotodlib/io/hk_utils.py
```diff
@@ -0,0 +1,133 @@
...
+
+            device_data = group[field][2]
+
+            info = {alias: device_data}
+            data.update(info)
+
+        try:
+            # check whether hk device is cosampled
+            assert len(times[0]) == len(times[-1])
+            # TODO: diff of times, if less than say 10%??, treat as cosampled
```
kmharrington, line 99: times is a list of arrays of times? It can have more than length 2? If so, I think I'd rewrite this to 
```
assert np.all( [len(t)==len(times[0]) for t in times])
```

I do think it's a good idea to also check the samples too with some level of precision. I would make the cosampled precision a kwarg of the function and set it at that 10% guess you started with. That will let folks mess with it later if they want

sotodlib/io/hk_utils.py
```diff
@@ -0,0 +1,133 @@
...
+        for field in group:
+            field_dict = {field: fields_data[field]}
+            device_data.update(field_dict)
+
+        grouped_data.append(device_data)
+
+    return grouped_data
+
+
+def make_hkaman(grouped_data):
```
kmharrington, line 73: I would have `make_hkaman` call `get_grouped_hkdata` (have it take in those parameters and call it inside the function). Users will pretty much always want to just use `make_hkaman` (or in that case call it `get_hkaman`?) I don't think there's a reason to make them call two functions each time. 

Or, if you want to keep those separate functions (because I can see how that's helpful) then make a `get_hkaman` the wraps the two. I just want the main entry point for users to be just one function.

sotodlib/io/hk_utils.py
```diff
@@ -0,0 +1,133 @@
...
+            device_axis = 'hkdata_' + device_name
+            samps_axis = 'hksamps_' + device_name
+
+            hkaman_cos = core.AxisManager(core.LabelAxis(device_axis, aliases),
+                                          core.OffsetAxis(samps_axis, len(time)))
+            hkaman_cos.wrap('times', time, [(0, samps_axis)])
+            hkaman_cos.wrap(device_name, np.array([data[alias] for alias in aliases]), [(0, device_axis), (1, samps_axis)])
+            amans.append(hkaman_cos)
+
+        # TODO: make an aman of amans
```
kmharrington, line 129: It looks like this should just be a series of `merge` or `wrap` commands. Have you tried it? I think it's worth just adding that in before merging.

kmharrington (changes requested):
Sorry for the delay getting back to this. Looking pretty good!

A bunch of inline comments/requests but I don't think any of them are too too complicated.  I do think it's worth finishing aman of amans before merging this because it looks like you're really really close to that.

sotodlib/io/hk_utils.py
```diff
@@ -0,0 +1,133 @@
...
+        field = online_fields[alias]
+
+        info = {field: [alias, time, data]}
+        fields_data.update(info)
+
+    # now group field names by feed/device
+    sorted_fields = sorted(_get_swap_dict(online_fields))
+
+    grouped_feeds = []
+    key_func = lambda field: field.split('.')[0:4]
```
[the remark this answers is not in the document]
sanahabhimani: with the way the .g3 files output field information, yes. but we should confirm whether things will change at the site. for example, right now, our hk files says: 

`observatory.CloudCity.feeds.temperatures.setpoint`,
`observatory.CloudCity.feeds.temperatures.supply_temp`,
`observatory.HAN.feeds.temperatures.Channel_1_T`

I can't remember whether `observatory` will change to `sat1` or if another identity will be attached to this string.

sotodlib/io/hk_utils.py
sanahabhimani: half-resolved! Still need to add the 10% guess; do we need this portion as a first pass for merging? Probably we want UCSD to test this hk tool soon? Wondering if I should make checking the samples a GitHub issue to address as UCSD concurrently tests this tool. What do you think?

sotodlib/io/hk_utils.py
```diff
@@ -0,0 +1,133 @@
...
+            # if not cosampled, make aman for each field in device/feed
+            print("{} is not cosampled".format('.'.join(field.split('.')[0:3])))
+            for field in group:
+                alias = group[field][0]
+                time = group[field][1]
+                data = group[field][2]
+
+                device_name = field.split('.')[1]
+                device_axis = 'hkdata_' + device_name
+                samps_axis = 'hksamps_' + device_name
```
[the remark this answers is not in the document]
sanahabhimani: Yes, `device_name`'s in my working example look like this:

```
Falcon
HAN
LEIA
LOBOT
LSA22HG
LSA22HG
LSA22HG
LSA22HG
LSA22HG
LSA22HG
LSA22HG
LSA22HG
LUKE
YODA
```

So added labels to make them more unique per axis.

kmharrington (changes requested):
I've already seen how this works and like this as the first step. Next step will be the option to co-sample the fields to reduce the number of total axes that are loaded.

My one request for changes is documentation. Especially for the "entry points" (functions you expect people to use) we need docstrings that tell folks what each function argument means and an explanation for how the returned data is formatted: (Ex from g3tsmruf which is a slightly bad example because the return info is out of date https://github.com/simonsobs/sotodlib/blob/master/sotodlib/io/load_smurf.py#L2189 ). And example config file with 1-2 entries in the docstring would also be good for this setup.

sanahabhimani:
Just added more docstrings, it's redundant since I wrote a wrapper function but let me know if you're looking for something else/more.

sanahabhimani:
Oh wait.. I forgot to add the returns output in the docstrings. Standby

sanahabhimani:
Now added returns output to docstrings, ready for re-review!

kmharrington (approved):
Looks good. Thanks!

PR #368, opened 2023-01-10
Focal plane coords

skhrg:
Module to generate focal plane from physical optics, this is just the LAT for now. I need to touch base with SAT optics folks about SAT optics again (but it should be simpler than the LAT).

One question I did have was if the corotator angle would be something that will be nominally loaded into the axis manager for LAT data (I assume yes)? And if we had a standardized name for it so I can have this code pull it from the aman by default.

### DIFF
--- a/sotodlib/coords/focal_plane.py
+++ b/sotodlib/coords/focal_plane.py
@@ -12,42 +12,59 @@ from scipy.interpolate import interp2d
 
 logger = logging.getLogger(__name__)
 
-LAT_FOV=3.7 # LAT FOV in degrees
-'''
+"""
 Dictionary of zemax tube layout.
 +ve x is to the right as seen from back of the cryostat *need to check*
+          11
+   5    3    4    6
+      1    0    2
+   9    7    8    10
           12
-   6    4    5    7
-      2    1    3
-  10    8    9    11
-          13
 Below config assumes a 30 degree rotation
-'''
-LAT_TUBES={'c':1,'i1':4,'i2':2,'i3':8,'i4':9,'i5':3,'i6':5, 'o1':6,'o2':10,'o3':13,'o4':11,'o5':7,'o6':12}
+"""
+LAT_TUBES = {
+    "c": 0,
+    "i1": 3,
+    "i2": 1,
+    "i3": 7,
+    "i4": 8,
+    "i5": 2,
+    "i6": 4,
+    "o1": 5,
+    "o2": 9,
+    "o3": 12,
+    "o4": 10,
+    "o5": 6,
+    "o6": 11,
+}
 
-def pix2sky(x,y,tube,rot=0,opt2cryo=0.):
-    '''Routine to map pixels from arrays to sky
+
+def pix2sky(x, y, tube, rot=0, opt2cryo=0.0):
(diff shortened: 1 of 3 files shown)


Zhilei:
I am wondering whether @laurensaunders has any inputs in the corotator angle. She is the most experienced with the ACU work.

sotodlib/coords/focal_plane.py
```diff
@@ -0,0 +1,280 @@
...
+import logging
+import numpy as np
+from scipy.interpolate import interp2d
+from scipy.spatial.transform import Rotation as R
+from sotodlib import core
+
+logger = logging.getLogger(__name__)
+
+# Dictionary of zemax tube layout.
+# +ve x is to the right as seen from back of the cryostat *need to check*
```
Zhilei, line 16: I am not sure about this layout, here is what I keep the record when I looked into this
https://docs.google.com/spreadsheets/d/1od2IfJo6Czt9ExH7wDADbaRPL77qi270XpjnF0f5c4U/edit#gid=0

Zhilei (changes requested):
Thanks for the nice PR! I would suggest we check the layout of the optics tubes before we proceed to merge.

sotodlib/coords/focal_plane.py
skhrg: Interesting, there is an off by one error that I introduced for indexing purposes but even with that many of these are inconsistent. I will check with @sdicker to see what he thinks.

skhrg:
Added in SAT optics and some performance improvements. I'd like to get this merged soonish since I want this done for `finalize-focal-plane`. @Zhilei what do you think about merging for now and when I get back to NA I will double check the OT indexing and if there is a problem put in another PR with the fix.

tskisner:
Just a note that we should call these functions in the [stub that already exists for it](https://github.com/simonsobs/sotodlib/blob/2cf83c7b06a1e175ebd3fd75bf11dd26c4540b5b/sotodlib/sim_hardware.py#L47-L66) when simulating focalplane layout.

Also, I previously gathered my notes on [LAT optics tube configuration](https://docs.google.com/document/d/1RjcPj8zEsid-Q2USwBT_g2ww5LmmTkQUwoZDIhniAKg/edit?usp=sharing) here if it helps.

skhrg:
Thanks Ted, I think your notes agree with what I got from Simon. If we did get the mapping wrong it should (I think) be obvious the first time we compare this against measured pointing so I am not super worried about getting it 100% correct right now.
I will fill out that stub when I get back to NA, may put that is a second PR since I don't like how old this branch is getting.

skhrg:
Also, I think we should gather all these notes on tube layout and the optics somewhere on the confluence so its in a central location. Any oppositions/ideas where that belongs?

sdicker:
For that matter all the optical design files are on the old wiki - I claim
nicely organized - maybe a link from such a confluence page could be in
order.

skhrg:
All the pages from the old wiki are converted to confluence pages. I can migrate it to the main space and then we can add in any information from Ted and Zhilei's notes that is missing. Maybe we put this in a LAT_Optics page under the LAT tree?

Zhilei (approved):
Let's move forward!

PR #372, opened 2023-01-14
bookbinder update

### LINKED ISSUE #373: bookbinder should not add G3Timestream to G3TimesampleMap
In recent spt3g, '0.3-175-g5d56a18', bookbinder tests fail with this:
```
>           output[c] = core.G3Timestream(np.array(self.data[c][:len(output.times)]))
E           ValueError: Cannot add member (Azimuth_Corrected): not a supported vector type.

sotodlib/io/bookbinder.py:82: ValueError
```

`output` is a G3TimesampleMap.  Those were only intended to house simple vectors; G3VectorInt, G3VectorDouble, G3VectorBool ...  The fact that it accepted G3Timestream was an accident, I guess because G3Timestream used to inherit from G3VectorDouble?

In any case, it doesn't work now (I guess because of recent interesting change to G3Timestream) ... probably just change it to G3VectorDouble.

guanyilun:
This PR includes some bugfixes and improvements to imprinter+bookbinder stack. A brief summary of the changes:
- `imprinter`: 
     - minor bugfixes, add `get_books` function
     - database migration to include more fields
     - improve test coverage
     - change configurations such that one instance watches over multiple `g3tsmurf` dbs. This also suggests that there will only be a single imprinter db for all data sources (SAT1, TSAT, etc.)
     - update `update_book_plan` script to reflect imprinter change
- `load_smurf`: 
     - add `db_args` option to allow more configurable db connection
- `bookbinder`: 
     - fix #373
     - update `make_book` script to reflect imprinter change

### DIFF
--- a/sotodlib/io/imprinter.py
+++ b/sotodlib/io/imprinter.py
@@ -269,7 +269,7 @@ class Imprinter:
         try:
             readout_ids = self.get_readout_ids_for_book(book)
         except ValueError:
-            pass
+            readout_ids = None
         hkfiles = []  # fixme: add housekeeping files support
 
         start_t = int(book.start.timestamp()*1e8)
--- a/sotodlib/io/imprinter.py
+++ b/sotodlib/io/imprinter.py
@@ -322,6 +322,21 @@ class Imprinter:
         if session is None: session = self.get_session()
         return session.query(Books).filter(Books.bid == bid).first()
 
+    def get_books(self):
+        """Get all books from database.
+
+        Parameters
+        ----------
+        session: BookDB session
+
+        Returns
+        -------
+        books: list of book objects
+
+        """
+        if session is None: session = self.get_session()
+        return session.query(Books).all()
+
     def get_unbound_books(self, session=None):
         """Get all unbound books from database.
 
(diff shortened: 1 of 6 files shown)


sotodlib/io/imprinter.py
```diff
@@ -269,7 +288,7 @@ def bind_book(self, book, session=None, output_root="out"):
         try:
             readout_ids = self.get_readout_ids_for_book(book)
         except ValueError:
-            pass
+            readout_ids = None
```
kmharrington, line 291: will this still make books just with no readout ids? Asking because I don't think we want to make books without readout ids. Instead we should skip over them and then trigger an alert of some type.

sotodlib/io/imprinter.py
```diff
@@ -322,6 +341,21 @@ def get_book(self, bid, session=None):
         if session is None: session = self.get_session()
         return session.query(Books).filter(Books.bid == bid).first()
```
kmharrington, line 342: I would use the function `.one()` instead of `.first()` for queries that are supposed to be unique. Since book id is the primary key you should never get more than one in the query, but this is another check.

`.one_or_none()` is also useful if there's a chance the book doesn't exist.

kmharrington (changes requested):
One suggestion but then one concern about readout ids. I don't think we want an error fallback if the IDs aren't getting parsed correctly.

sotodlib/io/imprinter.py
guanyilun: yes that's what's it currently doing, but I've now remove the error handling here

kmharrington (approved):
Great thanks!

PR #377, opened 2023-01-23
toast/ops: fix some None checks on traitlets

mhasself:
When allow_none=True, and with a @validate function, the None case needs to be specifically addressed.


tskisner (approved):
Thanks for this fix!

PR #378, opened 2023-01-25
Full working version

keskitalo:
This PR solves a long-outstanding request to account for the planet angular diameter in simulating SSO signal. It is enabled with a simple argument: `--sim_sso.finite_sso_radius`. If simulation is run without this option _and_ the planet solid angle is over 10% of the beam solid angle, a warning is produced but the execution proceeds.

If the beam modelers agree, we can make this option the default.


Zhilei (approved):
Looks great!
Since this is not a very involved PR, I suggest we merge first and start using it.

PR #379, opened 2023-01-25
io: context-compatible loader for toast3 HDF5 format TOD

mhasself:
To test this requires a context.yaml ... will place script in pwg-scripts in a moment.


tskisner (approved):
This seems fine as a first draft.  In later revisions we might extract more information like the simulated readout_ids.

PR #380, opened 2023-01-25
Random minor fixes

mhasself:
Happy for comments but these are pretty minor.

Bug fix:
- planets: handle pixell's new non-tuple Geometry object

Tidying:
- metadata loader: clean up how errors are presented
- AxisManager: pre-check for conflicting keys when merging
- tod_ops.filters: add some logger.infos


keskitalo (approved):
Only minor modifications. Everything seems to be in order.

PR #381, opened 2023-01-26
Hwpss fit fix

msilvafe:
This PR includes 3 main changes:
1. A fix to the offset in the nonlinear (scipy.curve_fit) fitter that broke the fitting when I integrated it in with the preprocess module.
2. A new hwpss fitting function `hwpss_linreg` that speeds up the fitting ~100x and now returns a reduced $\chi^2$ of the fit in the processed axis manager which can be used for data quality monitoring and selection.
3. I modified the demodulation function `demod_tod` to now include an optional argument for bandpass filtering the timestream before demodulation.

I tested all of this with some simulated data (see plots below). I think that these are all working well now and fully capture all of the functionality of the other functions in the hwp module. To avoid confusion I will remove all of the duplicate functions for hwpss fitting and demodulation (and their supporting functions) in the module but will save that for a separate pull request.

HWPSS Fit         |  Red $\chi^2$ of Fit   |  HWPSS Removal and Demod
:-------------------------:|:-------------------------:|:-------------------------:
  |   |  

Testing code + these plots can be replicated w/ jupyter notebook on simons1 at: `/mnt/so1/users/msilvafe/notebooks/flp_dev/20230125_Test_new_HWPSS_fns.ipynb`


tterasaki:
Thank you, Max!
I'll try and check the functions later, but let me comment about the definition of the chi2.
In the code the chi2 is defined as the sum of deviation between data and fitting.
But I think it is not correct. Chi2 should be divided by sigma of each data point.

So I would like to change the name.
Or to get better than that, estimate sigma from white noise level and divide sum of squares by that.

msilvafe:
> Thank you, Max! I'll try and check the functions later, but let me comment about the definition of the chi2. In the code the chi2 is defined as the sum of deviation between data and fitting. But I think it is not correct. Chi2 should be divided by sigma of each data point.
> 
> So I would like to change the name. Or to get better than that, estimate sigma from white noise level and divide sum of squares by that.

Hi Tomoki, thanks for looking this over. When you test the functions feel free to fix the red $\chi^2$ definition and commit that.

sotodlib/hwp/hwp.py
```diff
@@ -115,41 +115,105 @@ def demod(aman, signal='signal', name='signal_demod',
...
     aman.move(name + '_prebfilt', None)
     aman.move(name + '_wo_lfilt', None)
 
 
-# Functions to test with preprocessing pipeline added by Max S-F below 
+# Functions to test with preprocessing pipeline added by Max S-F below
+def hwpss_linreg(aman, signal='signal', modes=[2, 4, 6, 8], mask_flags=True,
+                 add_to_aman=True, name='hwpss_extract'):
+    """
+    Fir out half-waveplate synchronous signal (harmonics of rotation frequency),
```
kmharrington, line 156: typo

sotodlib/hwp/hwp.py
```diff
@@ -115,41 +115,105 @@ def demod(aman, signal='signal', name='signal_demod',
...
+    The harmonics to fit out are defined by ``modes`` and the operation is
+    performed on the ``signal`` axis of the ``aman`` axis manager.
+
+    Args
+    ----
+    aman (AxisManager): Axis manager to perform fit.
+    signal (string): Axis within aman to perform fit on.
+    modes (list): List of hwp rotation frequency harmonics to fit.
```
kmharrington, line 164: do these need to be ints?

sotodlib/hwp/hwp.py
```diff
@@ -220,18 +285,19 @@ def fit_func(x, A0, As, Bs, N_mode, N, sins, coss):
 
     for i in range(aman.dets.count):
         params_0 = list(As[:, i]) + list(Bs[:, i]) + list(modes)
-        params_0.append(np.mean(aman[signal][i, m[i]]))
+        params_0.append(np.median(aman[signal][i, m[i]]))
```
kmharrington, line 288: any reason not to make this `nanmedian` instead of `median`? (I often use `nanmedian` when I just want to blindly not have to worry about edge cases).

kmharrington (approved):
Looks good to me! A couple small in-line comments you can take or leave.

msilvafe:
@kmharrington I think I addressed all of your inline comments

@tterasaki I updated the red $\chi^2$ definition. I'm still not sure that this is really catching the bad fits properly so I'm not sure I'm calculating this properly. Below is the histogram of the red $\chi^2$ which does peak around 1 as it should but if I plot the fits >>1 or <<1 it looks by eye like they're fitting ok.
Red_chi2_updated

tterasaki:
I found a bug in the demod_tod.
An argument 'detrend' in the fourier_filter  defaults to 'linear'. That must not be applied.
That caused the strange behavior in the analysis of pol measurement with spinning HWP. I'll report in DM.

msilvafe:
> I found a bug in the demod_tod.
> 
> An argument 'detrend' in the fourier_filter  defaults to 'linear'. That must not be applied.
> 
> That caused the strange behavior in the analysis of pol measurement with spinning HWP. I'll report in DM.

Thanks for the report, I fixed this in PR #402 (already merged) and I think the mods to g3tsmurf_utils (commit d9d2412) should go in a different PR probably.

kmharrington:
Something I'm seeing while debugging other things. Could we put the fit equations in the docstrings? Something to help someone interpret what the coefficients mean without having to look through the code?

msilvafe:
@kmharrington I added in the description of what is being fit. There was an error before where it was properly fitting and returning the correct fitted curve into the input axis manager but the output fit coefficients array in the proc_aman was (n-1)/2 as long as it should have been. That's now been fixed.

@tterasaki I have added the some fit statistics. This is just an ordinary least squares regression of the form:

$$\hat{\beta} = (X^TX)^{-1}X^Ty$$

In this case, $\hat{\beta}$ is the estimate of the HWPSS coefficient amplitudes (matrix size $N_{dets} \times 2N_{harmonics}+1$), so for a single detector, the functional form is:

$$y = \beta_0 + \beta_1\sin(\theta_{HWP}) + \beta_2\cos(\theta_{HWP}) + \beta_3\sin(2\theta_{HWP}) + \beta_4\cos(2\theta_{HWP}) + \ldots + \beta_{2n+1}\sin(n\theta_{HWP}) + \beta_{2n+2}\cos(n\theta_{HWP})$$

The $X$ matrices are arrays of the trig functions size ($N_{samples} \times N_{harmonics}$) just that the estimate of the HWPSS is $\hat{y} = \hat{\beta}X$. For an OLS regression like this the (assuming [homoscedaticity](https://en.wikipedia.org/wiki/Homoscedasticity_and_heteroscedasticity), which I'm assuming here for simplicity though not sure it's the correct thing), then the covariance of the fitted coefficients can be calculated as:

$$Var{\hat{\beta}} = (X^TX)^{-1}\hat{\sigma}^2$$

The variance here is the variance of the regression (which is also the reduced $\chi^2$ of the fit [see Wikipedia article](https://en.wikipedia.org/wiki/Reduced_chi-squared_statistic)) defined as:

$${\sigma}^2 = \frac{\sum_{N_{samples}} (y_i - \hat{y_i})^2}{N_{samples} - p - 1}$$

where $p$ is the number of degrees of freedom of the fit (or the number of $2N_{harmonics} + 1$ in this case). So I am now returning this fit covariance ($Var{\hat{\beta}}$), the reduced $\chi^2$, and the $R^2$ of the fit. Hopefully, this is everything that you were hoping for from this function. I would really like to merge this PR but of course let me know if you think it still needs modifications.

I did test this functions as is on some simulation data and lab data from p10r2 below are some fit plots and results of the fit statistics described above for one of the observations ids for mv19:

HWPSS Fit         |  Fit Statistics
:-------------------------:|:-------------------------:
download (15) | download (14)

tterasaki:
The codes looks good! But I found the fitting takes about 1minutes for 1 hour timestream.

So I want to implement hybrid version of binning and linarg, I wrote a example in simons1.
`/home/tterasaki/workspace/for_share/hwpss_fit_fix_review/230302_hwp_fit_discussion.ipynb`
I would like to implement like this.

But, maybe it takes time. Let's merge this PR not to be too behind the master.

tterasaki (approved):

msilvafe:
> The codes looks good! But I found the fitting takes about 1minutes for 1 hour timestream.
> 
> So I want to implement hybrid version of binning and linarg, I wrote a example in simons1. `/home/tterasaki/workspace/for_share/hwpss_fit_fix_review/230302_hwp_fit_discussion.ipynb` I would like to implement like this.
> 
> But, maybe it takes time. Let's merge this PR not to be too behind the master.

@tterasaki, thanks for taking a look. We can't implement the binned fitting as in your jupyter notebook. The glitch flag mask will break. To use the binned fitting, we need to create a util function (maybe in the flags module) that bins ranges_matrices once we can properly bin the flag info (boolean arrays), then we can implement a binned version of this fitting. But currently, 1 hour of a full wafers' worth of data taking ~45 sec (that's what is in your notebook) seems pretty good, and the improvement from binning only cut this time in half (to ~22 sec), whereas before the `scipy.curve_fit` function was taking ~10 min so I'm happy as is. We can open another issue for binning the ranges matrices properly to update to binned fitting in the future. Going ahead and merging this now. Thanks!

msilvafe:
@tterasaki I take it back, let's do the binning. I'll open a new PR.

PR #382, opened 2023-01-26
Support for Moon observations

keskitalo:
This PR extends the SimSSO operator to work on sources that are larger or comparable in size to the beam map.

It explicitly adds support for Moon observations. There is presently no Lunar emission model included in the data file `data/planet_data.h5`. This may be desirable as the surface temperature of the Moon greatly depends on latitude and surface illumination. It is not obvious to me how refined we want to make the emission model.

Here is an illustration of the Moon observed at 150GHz using the `i1` optics tube:
image (3)

### DIFF
--- a/sotodlib/toast/ops/sim_sso.py
+++ b/sotodlib/toast/ops/sim_sso.py
@@ -224,14 +224,22 @@ class SimSSO(Operator):
 
         return
 
-    def _get_planet_temp(self, sso_name):
+    @function_timer
+    def _get_sso_temperature(self, sso_name):
         """
-        Get the thermodynamic planet temperature given
+        Get the thermodynamic SSO temperature given
         the frequency
         """
+        log = Logger.get()
         dir_path = os.path.dirname(os.path.realpath(__file__))
         hf = h5py.File(os.path.join(dir_path, "data/planet_data.h5"), "r")
-        if sso_name in hf.keys():
+        if sso_name == "Moon":
+            freq = np.linspace(0, 1000, 1001)[1:] * u.GHz
+            T = 300 # Kelvin
+            emissivity = 1.0
+            tb = 1 / (k / (h * freq) * np.log(1 + (np.exp(h * freq / (k * T)) - 1) / emissivity))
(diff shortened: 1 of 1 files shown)


Zhilei (approved):
Looks good, although I have to admit that I haven't checked that pixel counting in the added lines.

We do need a better thermal model of the Moon. I wrote a moon emission simulator considering its varying thermal properties and also its orientation on the sky within CLASS (Fig.2 of this paper https://ui.adsabs.harvard.edu/abs/2020ApJ...891..134X/abstract). 

Let me look into how I can properly share that code with SO.

Zhilei:
BTW, I am curious about the dark band across the map. Is that artifact from filtering the bright source?

sotodlib/toast/ops/sim_sso.py
```diff
@@ -224,21 +224,30 @@ def _exec(self, data, detectors=None, **kwargs):
...
         """
+        log = Logger.get()
         dir_path = os.path.dirname(os.path.realpath(__file__))
         hf = h5py.File(os.path.join(dir_path, "data/planet_data.h5"), "r")
-        if sso_name in hf.keys():
+        if sso_name == "Moon":
+            freq = np.linspace(0, 1000, 1001)[1:] * u.GHz
+            T = 300 # Kelvin
+            emissivity = 1.0
+            tb = 1 / (k / (h * freq) * np.log(1 + (np.exp(h * freq / (k * T)) - 1) / emissivity))
```
mhasself, line 240: What is happening with this line?  Are k & h defined in this scope?  Looks like a stray...

keskitalo: That is an incomplete attempt to add a simple emission model of the Moon. I just pushed the completed version. Here is the same simulation as above but now using the emission model:

image

keskitalo: @Zhilei that cooler band going through the source is indeed from filtering.

mhasself (approved):
Cool!

PR #386, opened 2023-01-28
Missing import

skhrg:
Missing import I came across when working on something else.


kmharrington:
I fixed this 2 weeks ago in https://github.com/simonsobs/sotodlib/pull/371 

If someone will review that one I would rather merge it there.

skhrg:
Ah, I hadn't seen that. I will just close this PR since I can just keep the fix locally until that gets merged.

Issue #357, opened 2022-12-20
Demonstrate multi-wafer books and smart fall backs


kmharrington:
I think this is a combination of `make-book-plans` and `make-obs-books` and I'm not sure where the line is between them. But with P10R1 (and P10R2) data we have examples of SMuRF data that is co-sampled across different UFMs. 

* We need to demonstrate the we can find when books need to be made with multiple wafers in a book and then make them. 
* We also need to have a fall-back structure where if the data are not co-sampled we can bind them into single-wafer books.  (we could test that with LATRt data where we're reading multiple UFMs but they aren't co-sampled).

guanyilun:
On the second point, how should the fall-back mechanism work? it doesn't sound very practical to check the timestamps in each frames across multiple observations. Is there any flag that says the data is cosampled so I could just refer to that?

kmharrington:
@jlashner  @msilvafe  Is there anything within the .g3 files we can use to know if the timing system is operating correctly / we can assume the timestreams are co-sampled between UFMs? 

It's not insane to check some of the frames in the files to be sure it's working. Right now it definitely isn't with TSAT, they don't have a timing system, ex:
```
book = "/mnt/so1/shared/site-pipeline/data_pkg/books/pton-tsat/obs/16735/obs_1673517465_tsat_1101111/"

for frame2 in core.G3File(book+"D_obs_crate_so04_slot2_000.g3"):
    if frame2.type == core.G3FrameType.Scan:
        break
        
for frame3 in core.G3File(book+"D_obs_crate_so04_slot3_000.g3"):
    if frame3.type == core.G3FrameType.Scan:
        break
        
f2_times = np.array([t.time / core.G3Units.s for t in frame2["signal"].times ])
f3_times = np.array([t.time / core.G3Units.s for t in frame3["signal"].times ])

print( np.median(f2_times - f3_times), np.std(f2_times - f3_times))
```

kmharrington:
Oh, other thought: If we can get some flag in the files that tell us it's running off the timing system then we can add a column to the level2 tables to indicate it's timing level. That way imprinter could work off that table.

kmharrington:
Ok, I am going to update my previous comment and say we HAVE to have something in the files to tell us if there is a timing system running. Comparing books from TSAT (no timing) and UCSD (with timing) I'm seeing that we aren't going to be able to tell by the timestamps alone. Writing some fast math to guess timing difference here is probably doomed to failure:

[...]

PR #387, opened 2023-01-30
Bookbinder: Check for timing system

algadoreo:
As pointed out in #357, if the timing system isn’t on, the timestamps between wafers don't line up and the system should fall back to making individual books for each wafer. Although the Imprinter will check for a timing system from the database and instruct the Bookbinder, the Bookbinder should still double-check for the presence of timing counters.

* If the counters are missing where they are expected (i.e., the database indicates the timing system was on), an error should be thrown to be caught by Imprinter, which will rerun Bookbinder in single-wafer mode.

* If the counters are present, calculate the correct timestamps and write them into every `.times` field in the output Book. Timestamps are calculated using the logic in load_smurf.py (ver. Jan 23, 2023):

https://github.com/simonsobs/sotodlib/blob/0f3c7b7a0ff882d39cd3e3b8c2258ede0c7ba308/sotodlib/io/load_smurf.py#L2160-L2166

### DIFF
--- a/sotodlib/io/bookbinder.py
+++ b/sotodlib/io/bookbinder.py
@@ -619,13 +619,15 @@ class FrameProcessor(object):
             List containing frames to be entered into output Book
         """
 
-        def generate_missing_smurf_samples(t, dtype=np.int64):
+        def generate_missing_smurf_samples(f, t):
             """
             Produce a frame filled with the FLAGGED_SAMPLE_VALUE at the timestamps given in t,
             to be used as fill-in values for missing data.
 
             Parameters
             ----------
+            f : G3Frame
+                A frame from the current timestream, from which to extract metadata
             t : numpy.ndarray
                 Vector of timestamps
 
@@ -635,12 +637,13 @@ class FrameProcessor(object):
                 Frame containing a G3SuperTimestream with FLAGGED_SAMPLE_VALUE at each input timestamp
             """
             frame = core.G3Frame(core.G3FrameType.Scan)
-            data = so3g.G3SuperTimestream(f['data'].names, core.G3VectorTime([core.G3Time(_t) for _t in t]))
-            if dtype in [np.float32, np.float64]:
-                assert f['data'].quanta is not None
-                data.quanta = f['data'].quanta
-            data.data = np.full((len(data.names),len(t)), self.FLAGGED_SAMPLE_VALUE, dtype=dtype)
-            frame['data'] = data
+            for k in set(f.keys()).intersection(['data', 'tes_biases', 'primary']):
+                data = so3g.G3SuperTimestream(f[k].names, core.G3VectorTime([core.G3Time(_t) for _t in t]))
+                if f[k].data.dtype in [np.float32, np.float64]:
+                    assert f[k].quanta is not None
+                    data.quanta = f[k].quanta
+                data.data = np.full((len(data.names),len(t)), self.FLAGGED_SAMPLE_VALUE, dtype=f[k].data.dtype)
+                frame[k] = data
             return frame
 
         if f.type != core.G3FrameType.Housekeeping and f.type != core.G3FrameType.Scan:
@@ -699,7 +702,7 @@ class FrameProcessor(object):
                     self._smurf_gaps.append((self._prev_smurf_frame_last_sample, current_timestamp))
                     # Insert dummy samples into self.smbundle
                     gap_times = self._smurf_timestamps[self._next_expected_smurf_sample_index : current_smurf_sample_index]
-                    self.smbundle.add(generate_missing_smurf_samples(gap_times, dtype=f['data'].dtype))
+                    self.smbundle.add(generate_missing_smurf_samples(f, gap_times))
                     # account for missing samples
                     self._next_expected_smurf_sample_index += gap_nsamples
                 # update values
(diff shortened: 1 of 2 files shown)


kmharrington:
@algadoreo  any chance you have some books created with this branch?

algadoreo:
I just made a book in `/mnt/so1/users/jleung/books/16733/obs_1673307576_sat1_1111101` on simons1. It’d be great if you could re-run your tests on it to see if it fixes the issue or not!

kmharrington:
Yeah, looks like I'm getting zero time differences between the different UFMs!

mhasself:
When I look at that book, I see mv9 has two fewer samples than all the other slots.

algadoreo:
There are 2 dropped samples in the original mv9_004.g3 file: t=167330998373897728 (between sample # 1327 and 1328) and t=167331048574397696 (between sample # 101727 and 101728). Otherwise, all other samples line up, so the code is doing what we expect.

I think dropped samples is a whole other problem not within the scope of this PR.

sotodlib/io/bookbinder.py
```diff
@@ -153,19 +210,22 @@ def add(self, f):
...
         if 'tes_biases' in f.keys():
-            if self.bias is None:
-                self.bias = self.create_new_supertimestream(f['tes_biases'])
-            self.bias.times.extend(f['tes_biases'].times)
-            self.bias.data = np.hstack((self.bias.data, f['tes_biases'].data)).astype(np.int32)
+            if self.biases is None:
+                self.biases = so3g.G3SuperTimestream(f['tes_biases'].names, core.G3VectorTime([]),
+                                    np.empty((len(f['tes_biases'].names), 0), dtype=np.int32))
+            self.biases.times.extend(f['tes_biases'].times)
+            self.biases.data = np.hstack((self.biases.data, f['tes_biases'].data)).astype(np.int32)
```
guanyilun, line 223: I guess the performance fix will go to another PR, but I'm a little confused about the mixed use of int32 and int64 across different fields. Could you clarify that for me?

sotodlib/io/bookbinder.py
```diff
@@ -183,54 +243,35 @@ def rebundle(self, flush_time):
...
 
-            Parameters
-            ----------
-            flush_time : G3Time
-                Output will contain buffered data up to (but not including) this time
-            use_rids : bool
-                Whether the `.names` field should be replaced with readout_ids
+        # Since G3SuperTimestreams cannot be shortened, the data must be copied to
+        # a new instance for rebundling of the buffer
+        signalout = so3g.G3SuperTimestream(self.signal.names, tout, self.signal.data[:,:len(tout)])
```
guanyilun, line 261: I think it's always good to state our assumptions in terms of assertion statements. For example, are we assuming `len(tout) > 0` here?

sotodlib/io/bookbinder.py
```diff
@@ -417,6 +479,61 @@ def determine_state(self, v_az, v_el, dy=0.001):
 
         self.current_state = state
 
+    def check_times(self, f):
```
guanyilun, line 482: maybe we should consider a better name for this function because `check_times` makes it sound like a "pure" function but it actually replaces the `.times` field in place in the case that `t[0]>BOOK_START_TIME && t[-1] < BOOK_END_TIME`. Maybe something like `correct_times` is a better name? This in place modification could also be stated clearer in the docstr. Also, I think timestamps from your `get_timestamps` function should be more accurate than the `BOOK_{START,END}_TIME` from imprinter, so I don't think we want to trim frames based on these inaccurate metadata.

guanyilun (approved):
I like the changes here, so I'm approving with some minor comments for consideration / clarification.

sotodlib/io/bookbinder.py
algadoreo: Those are the types in the original SMuRF files (int32 for `'data'` and `'tes_biases'` and int64 for `'primary'`). I did ask @mhasself at one point about the discrepancy with the TOD format doc, and then we forgot to update it. Nonetheless, I plan to add type-checking assertions in a subsequent PR where I optimize the array type-casting.

sotodlib/io/bookbinder.py
algadoreo: I don’t think so (but not 100% certain). G3SuperTimestream can be initialized with 0-length vectors, though I don’t know if it breaks anything down the line (in `FrameProcessor.flush`). Even more reason to have comprehensive tests!

sotodlib/io/bookbinder.py
algadoreo: I struggled a lot with this one; originally it was `trim_frames`, but then became `adjust_times` when the main purpose shifted to the timestamp calculation. I’m not super happy with `check_times` either, so I’m open to suggestions for a more descriptive name.

You also bring up a good point about the operations being done in-place—do you think it makes more sense to make it a subroutine (no return statements) instead?

sotodlib/io/bookbinder.py
kmharrington: @jlashner  will know what's actually coming out of the smurf system.

sotodlib/io/bookbinder.py
```diff
@@ -508,13 +619,15 @@ def __call__(self, f):
             List containing frames to be entered into output Book
         """
 
-        def generate_missing_smurf_samples(t, dtype=np.int64):
+        def generate_missing_smurf_samples(f, t):
```
kmharrington, line 622: If you're working on this function then wouldn't this be the place to fill in missing samples / make the number of samples per UFM all line up? (https://github.com/simonsobs/sotodlib/issues/389) 

It's very inter-related to this PR and is equally going to prevent us from having a working loader function / run analysis off books. So if it's not this PR it needs to be the priority for the next one.

sotodlib/io/bookbinder.py
jlashner: This is correct, the TES signal and bias data are both 32 bit ints, while the "primary" data which contains timing counters and timestamps are 64 bit.

sotodlib/io/bookbinder.py
algadoreo: The current gap checker (this function) only checks for gaps only between—not within—frames. I have a good idea of how I can implement the latter, but I think it would be a large enough operation that it should be its own PR.

PR #390, opened 2023-01-30
Migrate QDS Monitor into site_pipeline

BrianJKoopman:
This migrates over the qds package from [pwg-scripts/pwg-qds/](https://github.com/simonsobs/pwg-scripts/tree/master/pwg-qds), along with its documentation.

I updated the docs a bit, generalizing connection details and including a full script example that was otherwise committed to the `pwg-scripts` repo.

This adds `influxdb` as a dependency.

And I wasn't sure how you want to handle the package structure here, so I didn't import `Monitor` into any of the above levels like in `site_pipeline`. Happy to adjust as suggested though.

Also, I did pull over all the commits from pwg-scripts. I'd expect this all to be squashed on merge though, but for completeness they're there.


kmharrington (approved):
Thanks for setting up everything here!

I'm not sure we want influxdb as a requirement for all of sotodlib. We have at least one other dependency (DetMap) that is only used inside of site-pipeline. Influx is similar to that one because it's very likely to only be used there (it's dissimilar because it's pip installable so it could go either way). 

I think I would go ahead and import Monitor like we import utils in `__init__.py` (but I'm also going to instantly start using this PR and can just do this).

guanyilun:
Not directly relevant, but maybe we should be reconsidering the option to manage some site pipeline script in the prefect / workflow system, because that would allow us to decompose them in terms of prefect tasks and flow, and have a lot finer control over their executions than simply wrapping an executable script. I had to do this for the make_obs_book flow in order to keep track of the status of each book. Having a separate place for these scripts also move one-script dependencies like this out of sotodlib to keep things clean.

kmharrington:
One note, the reason to bring this is in that this isn't a one-script dependency. As soon as it works for checking if the timing outputs we're also going to want to use it for things like reporting on if the HWP angle solutions are failing for some reason or that tod_processing is getting a lot of unlocked detectors. So I think "ability to push to our alert system" is a more general setup. 

I know enough about prefect to understand that using more detailed task/flow things could be useful but not enough to see a whole vision about what this would look like for data packaging ( ex: instead of having it "all out" of sotodlib why not have it "all in"? ) This is something we could discuss on Thursday. If it's something you really want to do then I'd say to put together a sketch of how the whole system would end up working, what would need to change, etc. Without new info I think we'll end up having the same conversation we've had in the past.

guanyilun:
Yeah, I'm not suggesting to move every site_pipeline scripts to prefect, only those that we want finer control over its executions such as make_obs_books. On the other hand, wrapping them all in prefect will make it harder to run as executable script in command line which is not good either, so maybe a better solution is just to keep two implementations coexist. In light of this, we should also try to make site_pipeline scripts more prefect friendly. I've set an example here: https://github.com/simonsobs/sotodlib/compare/master...dataflow-update

BrianJKoopman:
> I'm not sure we want influxdb as a requirement for all of sotodlib. We have at least one other dependency (DetMap) that is only used inside of site-pipeline. Influx is similar to that one because it's very likely to only be used there (it's dissimilar because it's pip installable so it could go either way).

It looks like there are others besides `detmap`, `pycpd` for instance (though also not listed in the `setup.py` file.) How do you want to handle dependencies for the `site_pipeline` subpackage?

> I think I would go ahead and import Monitor like we import utils in `__init__.py` (but I'm also going to instantly start using this PR and can just do this).

Sounds good, I'll let you do that then.

BrianJKoopman:
Following our discussion on Slack, `influxdb` has been moved to a dependency group called `site_pipeline`, installable via `pip install .[site_pipeline]` in [dbacd64](https://github.com/simonsobs/sotodlib/pull/390/commits/dbacd64187a3e42ce31b2228890cd2f34947b338). Other `site_pipeline` dependencies can be added to the list here:
```
setup_opts["extras_require"] = {
    "site_pipeline": [
        "influxdb",
    ],
}
```

kmharrington (approved):
awesome, thanks a bunch!

Issue #119, opened 2021-04-16
AxisManagers -- Non-aligned Information?


kmharrington:
Looking for thoughts on how we might want to encode information in Axis managers that aren't aligned along axes. 

The main example I have here are units. We'd like a way to track units of signal. But I could also imagine we might want to encode units of other arrays pointing/encoders in particular.

My first two thoughts on options:

1. make `aman['units'] = 'string'`
2. Have some sort of `aman._units = {}` that keeps track of `{'field_name' : unit}`

jeffiuliano:
The second option there makes sense to me.  We could then set up each other object to have an eg,  `aman.signal.units`, which is like a property, that just does  `aman._units.get(name)` or `aman._units[name]=unit` under the hood (would need to look at how aman is built more closely to know if something like that is straightforward to do)

tskisner:
Consider using astropy units for labels / attributes / traits of your data.  They can be used to express even complicated compound units:

https://github.com/hpc4cmb/toast/blob/e247526822a14b9a0c7634bc9c7e02c0fcbdf2d0/src/toast/tests/instrument.py#L53-L62

Note that I'm **not** suggesting that you use `Quantity` objects for the full data (i.e. wrapped numpy arrays with a unit associated to the array), since I'm not sure of the performance of manipulating those or passing the buffer to compiled code.  However, if you have a scalar "units" label to your object, you could just use astropy units for that variable.  Or maybe a string representation of the unit that is compatible with astropy.  For example:
```
In [1]: import numpy as np                                                                          

In [2]: from astropy import units as u                                                              

In [3]: unit = u.K * np.sqrt(1 * u.second)                                                          

In [4]: unit                                                                                        
Out[4]: <Quantity 1. K s(1/2)>

In [5]: unit_str = "{}".format(unit)                                                                

In [6]: unit_str                                                                                    
Out[6]: '1.0 K s(1/2)'

In [7]: roundtrip = u.Quantity(unit_str)                                                            

In [8]: roundtrip                                                                                   
Out[8]: <Quantity 1. K s(1/2)>

# Convert to micro K
In [9]: unit.to(u.uK * np.sqrt(1 * u.second))                                                       
Out[9]: <Quantity 1000000. s(1/2) uK>

# Get the actual (unit-less) conversion factor
In [10]: unit.to(u.uK * np.sqrt(1 * u.second)).to_value()                                           
Out[10]: 1000000.0
```

jeffiuliano:
[...]

PR #391, opened 2023-01-31
Added ability to serialize quantities

skhrg:
Following some discussion from #119 this adds the ability to save/load AxisManagers with Quantities wrapped (which seems to be the only thing that wrapping Quantities breaks). I don't think this closes that issue since the task of adding in SO specific units that was started in [this branch](https://github.com/simonsobs/sotodlib/tree/add_units) may be incomplete.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -379,6 +379,7 @@ class TestAxisManager(unittest.TestCase):
                                       shape=(aman.dets.count, aman.samps.count)))
 
         aman.wrap('quantity', np.ones(5) << u.m)
+        aman.wrap('quantity2', (np.ones(1) << u.m)[0])
 
         # Make sure the saving / clobbering / readback logic works
         # equally for simple group name, root group, None->root group.
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -4,6 +4,7 @@ import os
 import shutil
 
 import numpy as np
+import astropy.units as u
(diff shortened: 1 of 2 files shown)


kmharrington (changes requested):
I like this as a starting point for adding units to the system. 

I'm pretty sure axisman_io has complete test coverage at this point, so could you add tests here to verify the units don't break anything? https://github.com/simonsobs/sotodlib/blob/master/tests/test_core.py#L356

tests/test_core.py
```diff
@@ -377,6 +378,8 @@ def test_500_io(self):
         aman.wrap('sparse', csr_array( ((8,3), ([0,1], [1,54])), 
                                       shape=(aman.dets.count, aman.samps.count)))
 
+        aman.wrap('quantity', np.ones(5) << u.m)
```
mhasself, line 381: Looking good but something is up with scalars ... If I add this:
```
        aman.wrap('quantity2', (np.ones(5) << u.m)[0])
```
it causes the test to fail.

skhrg: Looks like it is because scalar Quantities are size-0 arrays which seems to  mess with some stuff that h5py does when it is loading it back. I can add a separate schema for scalar quants to get around this.

kmharrington (approved):
Looks good to me now. Thanks!

PR #392, opened 2023-02-01
update_g3tsmurf_database: use config and rewrite it to be more prefect friendly

guanyilun:


kmharrington (approved):
thanks for reformatting

PR #393, opened 2023-02-01
G3tSmurf Timing Updates

kmharrington:
This has been rebased off of #392  and #390  so we need to wait for these two to merge here. I would also like to wait until [this PR](https://github.com/simonsobs/smurf-streamer/pull/38) has started running at UCSD because I'd prefer to not need to repair the UCSD databases more than once (and this should technically be tested on "good" files). 

This PR adds a timing tracking column to the Files and Observation tables in G3tSmurf. It also adds our first example of the QDS monitor into the update-g3tsmurf-db script which will let us set up a grafana alter if the timing information dies at some point. 

Merging this PR will come with the need to run some table updates on the existing databases. I have scripts for these but @guanyilun  we'll need to coordinate running those scripts on the prefect managed databases. 

* Scripts to Update the databases / repair timing columns are here: https://github.com/simonsobs/pwg-scripts/tree/master/pwg-duf/g3tsmurf_timing 
* Monitor config files for this example are here: https://github.com/simonsobs/site-pipeline-configs

### DIFF
--- a/sotodlib/io/g3tsmurf_db.py
+++ b/sotodlib/io/g3tsmurf_db.py
@@ -79,7 +79,8 @@ class Observations(Base):
     action_name = db.Column(db.String)
     
     stream_id = db.Column(db.String)
-    
+    timing = db.Column(db.Boolean)
+
     # in seconds
     duration = db.Column(db.Float)
     n_samples = db.Column(db.Integer)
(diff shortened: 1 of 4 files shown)


guanyilun (approved):
Looking good! You may consider also adding the `Monitor.from_configs` to the nice documentation here before we forget about this.

kmharrington:
fixed the rebasing after merging the monitor PR (and checked that `from_configs` is caught by the autodoc). @guanyilun  go ahead and merge this when you have time to simultaneously do the database updates.

PR #395, opened 2023-02-02
Bookbinder: Optimizing array allocations

algadoreo:
@jlashner did some digging and found that the `ndarray.astype()` casting during data buffering is taking up ~20% of the runtime because it is rebuilding the arrays from scratch. We can remove these calls, instead initializing the arrays with the types described in [TOD format doc](https://github.com/simonsobs/tod2maps_docs) and adding assertions to ensure that subsequent input data has the right type.

### DIFF
--- a/sotodlib/io/bookbinder.py
+++ b/sotodlib/io/bookbinder.py
@@ -193,6 +193,7 @@ class _SmurfBundle():
         self.signal = None
         self.biases = None
         self.primary = None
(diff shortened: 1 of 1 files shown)


kmharrington (approved):
Nice find!

jlashner (approved):

guanyilun (approved):

PR #396, opened 2023-02-03
Bookbinder: Optimize data buffer I/O

algadoreo:
@mhasself suggested that I can minimize the number of numpy array allocations by changing how the buffered data is stored. Instead of creating a new ndarray every time new data is added, we append the new data to a running list (as a new item in the list) that only gets converted into a contiguous array at the output call.

A simple test where I add a 1000-by-1000 array to the buffer 10 times show a notable speedup between the two methods.

### DIFF
--- a/sotodlib/io/bookbinder.py
+++ b/sotodlib/io/bookbinder.py
@@ -214,7 +214,7 @@ class _SmurfBundle():
 
         if self.signal is None:
             self.signal = {'names': f['data'].names, 'times': [], 'data': []}
-        assert np.asarray(f['data'].names) == np.asarray(self.signal['names']) # there's a bug where G3VectorString(x)==G3VectorString(x) returns False
+        assert (np.asarray(f['data'].names) == np.asarray(self.signal['names'])).all() # there's a bug where G3VectorString(x)==G3VectorString(x) returns False
         assert f['data'].data.dtype == self.dtypes['data']
         self.signal['times'].append(f['data'].times)
         self.signal['data'].append(f['data'].data)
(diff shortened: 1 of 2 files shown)


sotodlib/io/bookbinder.py
```diff
@@ -212,28 +212,29 @@ def add(self, f):
...
         self.times.extend(f['data'].times)
 
-        assert f['data'].data.dtype == self.dtypes['data']
         if self.signal is None:
-            self.signal = so3g.G3SuperTimestream(f['data'].names, core.G3VectorTime([]),
-                                    np.empty((len(f['data'].names), 0), dtype=self.dtypes['data']))
-        self.signal.times.extend(f['data'].times)
-        self.signal.data = np.hstack((self.signal.data, f['data'].data))
+            self.signal = {'names': f['data'].names, 'times': [], 'data': []}
+        assert np.asarray(f['data'].names) == np.asarray(self.signal['names']) # there's a bug where G3VectorString(x)==G3VectorString(x) returns False
```
mhasself, line 217: Does this work?  Don't you need an all(...)?  I get "ValueError: The truth value of an array with more than one element is ambiguous." for an equivalent test.

algadoreo: Oh now I see why my unit test didn’t catch that: it uses a single channel, so it only had one element (and so didn’t have that error). Let me rectify that.

mhasself (approved):
Thanks, glad to see this.

guanyilun (approved):

jlashner:
I believe this PR broke the bookbinder in python3.7, since `dtype` being an argument to hstack was only introduced in a new version of numpy that requires python >= 3.8. Is python3.7 still supported?

algadoreo:
I think the version of tod_stack (_unstable) that uses Python 3.7 just breaks Bookbinder in general, so there was discussion of moving up to 3.10—though I don’t know the timeline for it. I’ve been developing using tod_stack_unstable/py310.

mhasself:
Python 3.7 isn't on our official support list anymore, as 3 years have now passed since 3.8 was released.  But one shouldn't hurry to break things ...

Isn't the hstack(dtype=...) unnecessary as long as all the contributors have the same, right dtype?  We should never end up with inconsistency there, anyway -- so perhaps there is a superior solution that doesn't require hstack(dtype=...).

algadoreo:
Sure, I can remove them. In principle, the stacked arrays should never be different types since they are checked before being added to the buffer, but I was paranoid (about the output of hstack) and wasn’t aware of the problem with 3.7.

PR #397, opened 2023-02-06
Point source simulation

keskitalo:
This PR adds a TOAST operator that simulates point sources based on a catalog and beam map. The sources can be static, variable or even transient and their SED and polarization properties can evolve over time.

The PR has a sibling on TOAST side: https://github.com/hpc4cmb/toast/pull/632.

Example usage here: https://github.com/simonsobs/pwg-scripts/tree/master/pwg-tds/point-source-sim


keskitalo:
The CI test failure seems to come from issues reaching an ubuntu server. Not related to the proposed changes.

keskitalo:
The catalog units were in fact Jy, not Jy/sr. That is now reflected in the code.

keskitalo:
@mhasself @Zhilei @amaurea Are there any unaddressed concerns regarding this PR? Nominally I need at least one approving review but if there are no concerns I can go ahead and merge.

mhasself (changes requested):
Neat.  Can you please add documentation or example of the catalog file format?

I know it's in the pwg-script you linked, but it is then only this PR that connects them together.

keskitalo:
@mhasself Thanks for the reminder. I added the operator (along with the stimulator and wiregrid simulators) into the docs.

mhasself (approved):
Great!

Issue #388, opened 2023-01-30
Book format requests


mhasself:
I have some requests for the obs/oper Book format.  I'm looking at, for example, /mnt/so1/shared/site-pipeline/data_pkg/books/ucsd-sat1/oper/16748/oper_1674860955_sat1_0010000 and trying to write Context indexing and loading tools.

1. Info in M_index.yaml:

- "telescope" field (e.g. "sat1")
- "book_type" (e.g. "oper")
- include a list of the included "stream_ids" (e.g. ["ufm_mv14", "obs_ufm_mv6"]).  (These should be in standard order, i.e. in the same order that is used to generate slot-flags.)
- If possible, include the g3tsmurf "detset" name for each stream_id ({"ufm_mv9": "sch_ufm_mv9_1673656362", ...}).
- include a dump of the sample range for each file_index; e.g. if there are files 000 and 001 then "samples_ranges" might be = [[0, 10000], [10000, 12543]].

2. When there are no interesting housekeeping data, the "ancil" entry should still be populated.  It would be simply a G3TimesampleMap with .times populated (it's the .times I want).

3. The "sample_range" (in Scan frames) should store index ranges in Python range notation, where the second index is not included.  For example, the first 10000 samples would be denoted as [0, 10000], not [0, 9999].

4. The filename for D_ prefix files is supposed to be `D_{stream_id}_{file_index}.g3`.  But the books I'm looking at include the book type in there, e.g. `D_`**oper**`_ufm_mv19_000.g3`.  I don't think we need that.

I might tack on a few more things as I keep working on this.

mhasself:
5. The "sample_range" should be with respect to an increasing "sample index" that spans all files in an observation.  For example, the file listed below has a first Scan frame with sample range (0, 9999).  But I expected to see (55000, 55999).  (And as per 3 above it should ultimately be (55000, 56000)).

`/mnt/so1/shared/site-pipeline/data_pkg/books/ucsd-sat1/obs/16749/obs_1674963154_sat1_1111101/A_ancil_001.g3`

kmharrington:
Just noting that issue overlaps with this issue: https://github.com/simonsobs/sotodlib/issues/359 

But this is a much more complete list than I wrote up.

mhasself:
Looking at recent books, I confirm that items 2 and 4 are fixed.  Thanks!

1, 3, 5 seem unchanged.

Issue #365, opened 2022-12-21
Write a Context loader for level 3 SMuRF Books


kmharrington:
We need a loader for our level 3 books before we can do anything related to site automated analysis / data quality checking. 

This can technically start now but it will not be complete until #356 and #357 are finished.

PR #398, opened 2023-02-07
Book loader

mhasself:
This contribution has 2 parts:
* a context-compatible loader function for "obs" and "oper" Books.
* a Book tester and obsfiledb writer for validating and indexing obs/oper Books.

(This is a draft PR for now in order to keep iterating with bookbinder output as it evolves.)

People who liked this also liked: #388, #365.


sotodlib/io/load_book.py
```diff
@@ -0,0 +1,223 @@
...
+                if delta > start:
+                    dest_offset += (delta - start)
+                    start = 0
+                else:
+                    start -= delta
+
+                if dest_offset >= samples[1] - samples[0]:
+                    break
+
+    det_info = core.metadata.ResultSet(
```
kmharrington, line 219: I'd like to build the `smurf` entry into the det_info here as well. We don't need to follow to functional path in `load_smurf` but it's something used in analysis fairly often. All these get built off the status frame that's in the first file on an g3 session and so should also be at the beginning of every book frame. 

https://github.com/simonsobs/sotodlib/blob/master/sotodlib/io/load_smurf.py#L2381 

The `status` axismanager is also something we'll want to track through, it has things like what filters are applied to the data as we're taking it, etc. 

At level 2 we've been using the `SmurfStatus.from_file( filename)` function to easily make a status object that can use used for all these. I've already seen this work for books.

sotodlib/io/load.py
```diff
@@ -587,6 +588,7 @@ def hstack_into(dest, src_arrays):
     {
         'pipe-s0001': load_observation,
         'toast3-hdf': load_toast_h5_obs,
+        'obs-book': load_obs_book,
```
kmharrington, line 591: could also just stick a line in `load_book` like this:

https://github.com/simonsobs/sotodlib/blob/29c83f97d1928dba15ff7c29453bfaca16ec74be/sotodlib/io/load_smurf.py#L2479 

It's not clear to me why we need all these here and/or if I should have the level 2 one here.

sotodlib/io/load_book.py
```diff
@@ -0,0 +1,223 @@
...
+import os
+
+import sotodlib
+from sotodlib import core
+
+#from . import load
+
+import numpy as np
+
+def _extract_1d(src, src_offset, dest, dest_offset):
```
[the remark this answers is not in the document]
mhasself: Ya, you caught me on this ... I will see if there are more things I could take advantage of.

The existing io.load stuff was written in the G3TimestreamMap days and contains many intrinsic inefficiencies because of that.  But I should be more careful about that.

sotodlib/io/load_book.py
```diff
@@ -0,0 +1,223 @@
...
+    samples[1] = min(max(samples), sample_range[1])
+
+    file_map = db.get_files(obs_id)
+
+    # Prepare the output structure.
+    aman = core.AxisManager(
+        core.LabelAxis('dets', [p[1] for p in pairs_req]),
+        core.OffsetAxis('samps', samples[1] - samples[0], samples[0], obs_id))
+    aman.wrap('primary', core.AxisManager(aman.samps))
+    aman.wrap('tes_bias', core.AxisManager(aman.samps))
```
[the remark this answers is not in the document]
mhasself: The docs do not generalize unambiguously to the multi-UFM case.  I'm happy to talk about a sensible generalization.

sotodlib/io/load.py
mhasself: I think if one wants to draw a line between what could be here and what should instead only be added on submodule import, it is whether the submodule has additional non-standard / weird dependencies (e.g. moby2, sqlalchemy (?)).  I don't mind moving this and the toast_h5 one out ... I think I mostly was just being quick and forgot to clean up.

sotodlib/io/load_book.py
mhasself: Cool, will work this in.

mhasself:
> A bunch of in-line comments / requests. Big picture is this is already diverging from the level 2 loading/AxisManager formats and I'd like that not to be the case without good reason.

Yes, great ... let's make it a more natural extension.

> I am confused about a lot of what's in `check_book` but haven't gone through it line by line. Isn't this the sort of thing we'd want to build in to a test system and/or just get bookbinder and imprinter to actually make these correctly? The one exception to this is the "find all the files for the `obsfiledb` function, that is useful but could also go in the `update-obs-db` script.

Yes, this (or its descendant) should be built into a test system / run regularly after a book is bound.  It turns out that the parsing you do for obsfiledb incorporation is also a good time to check that the book follows the schema.  So it's dual purpose, but not artificially so.  The essence of this can be moved to library level later (and used by bookbinder/imprinter/update-stuff/etc as needed).  But I think it is worth committing rather than having it floating around my dev area.

sotodlib/io/load_book.py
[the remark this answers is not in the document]
kmharrington: Ahh gotcha. I'm not against updating what's in `io.load` to be more efficient and having that flow down into updates for g3tsmurf loading, etc. I think we're all using SuperTimestreams at this point.

Mostly interested in keeping the same functions/structure.

sotodlib/io/load_book.py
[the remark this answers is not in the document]
kmharrington: Right, the generalization is a thing to work out in more detail....

What you have per detset isn't what we had per lvl2 axismanager for biases. Those are bias_line x samps arrays instead of single line arrays. At some points in the past I've been building `biases x dets` arrays for maps and things. So I have liked having that as a label axis and not a lot of single dim arrays. 

If we're going to want any arrays that are `biases x dets` in level 3 data then I we're going to want to extend those label axes to have names per detset or something. We also _definitely_ need to start building the caldbs for smurf things off of books / for use at level three because the standard formats we've had for bg_maps need updates. This is all probably work a discussion on one of the Thursday calls.

sotodlib/io/load_book.py
kmharrington: noting that the per-detset organization will be a thing to track here as well.

sotodlib/site_pipeline/check_book.py
```diff
@@ -0,0 +1,422 @@
...
+The config file for this should look something like this:
+
+.. code-block:: yaml
+
+    # Database setup
+    obsfile_db: './obsfiledb.sqlite'
+    root_path: '/'
+
+    # Work-arounds
+    stream_file_pattern: 'D_obs_{stream_id}_{index:03d}.g3'
```
guanyilun, line 16: a heads-up that I'm fixing the D-prefixed file naming in #401 and will propagate the changes to previous books once that's merged.

mhasself:
As agreed on our call, I am marking this as ready for review.  I've created issue #442 to capture a couple of things I didn't get to fixing.

On where to register the obsloader ... I think it's ok to do this in load.py, since this doesn't bring in any new weird dependencies and will probably end up being the loader that is used the most, overall, in the life of SO.  (The load_smurf one can be put in there too... just add sqlchemy and whatever else as a package dependency.)

sotodlib/io/load_book.py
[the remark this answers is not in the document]
mhasself: The major difference is that this loader assumes it knows exactly how many samples it is supposed to load, and how many detectors.  That means it needs a well-constructed obsfiledb... but with those in hand it is very easy to preallocate the data, and not have to use hstack anywhere.

mhasself:
@jlashner  Here's how I test this; first the config file for check-book:
```
{
    # Database setup
    obsfiledb: './obsfiledb3.sqlite',
    root_path: '/',

    # Work-arounds
    'extra_files': ['M_index.yaml'],
    'extra_extra_files': ['Z_bookbinder_log.txt'],
    #'extra_extra_files': ['frame_splits.txt'],
}
```

And a test script for using a loader without a Context:
```
from sotodlib import core
from sotodlib.io import load

book_loader = load.OBSLOADER_REGISTRY['obs-book']

db = core.metadata.ObsFileDb.from_file('obsfiledb.sqlite')

all_obs = db.get_obs()
obs_id = all_obs[0]
print(obs_id)

det_table = db.get_det_table(obs_id)

dets = det_table['dets:readout_id'][10:20]

print('No signal ...')
tod0 = book_loader(db, obs_id, no_signal=True)
print('  -- ', tod0.shape)

print('Whole thing ...')
tod0 = book_loader(db, obs_id)
print('  -- ', tod0.shape)

print('A few dets ...')
tod = book_loader(db, obs_id, dets=dets)
print('  -- ', tod.shape)

print('A few samples ...')
tod = book_loader(db, obs_id, samples=[1,1001])
print('  -- ', tod.shape)
```

sotodlib/site_pipeline/check_book.py
```diff
@@ -0,0 +1,143 @@
...
+    print(' -- adding %i detsets and %i file refs' % (len(detset_rows), len(file_rows)))
+    for name, dets in detset_rows:
+        if len(db.get_dets(name)) == 0:
+            db.add_detset(name, dets)
+    for row in file_rows:
+        db.add_obsfile(**row)
+
+
+if __name__ == '__main__':
+    util.main_launcher(main, get_parser)
```
kmharrington, line 143: missing a util import here.

kmharrington (approved):
Modulo the two new in-line fixes, I vote we merge this and move on to the next set.

mhasself:
> Modulo the two new in-line fixes, I vote we merge this and move on to the next set.

Fixed -- will merge shortly.

PR #399, opened 2023-02-07
Bookbinder: Backward compatibility with Python 3.7

algadoreo:
This reverts commit 75dafd8. The Python 3.7 version of Numpy doesn’t include dtype as an option for `hstack`. Since it was only put in to satisfy my paranoia, we remove it here to ensure continued compatibility.

@jlashner can you check that this works?

### DIFF
--- a/sotodlib/io/bookbinder.py
+++ b/sotodlib/io/bookbinder.py
@@ -265,19 +265,19 @@ class _SmurfBundle():
 
         # Since G3SuperTimestreams cannot be shortened, the data must be copied to
         # a new instance for rebundling of the buffer
-        signalstack = np.hstack(self.signal['data'], dtype=self.dtypes['data'])
+        signalstack = np.hstack(self.signal['data'])
         signalout = so3g.G3SuperTimestream(self.signal['names'], tout, signalstack[:,:len(tout)])
         self.signal = {'names': self.signal['names'], 'times': [self.times], 'data': [signalstack[:,len(tout):]]}
 
         if self.biases is not None:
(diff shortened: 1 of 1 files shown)


jlashner:
Huh, now when I switch back to python3.7 and re-run I'm getting this error:

```
  File "test_hwp_binding.py", line 53, in <module>
    frameproc_config={"readout_ids": rids})()
  File "/home/jlashner/.local/lib/python3.7/site-packages/sotodlib/io/bookbinder.py", line 1100, in __call__
    output += self.frameproc(f)  # FrameProcessor returns a list of frames (can be empty)
  File "/home/jlashner/.local/lib/python3.7/site-packages/sotodlib/io/bookbinder.py", line 693, in __call__
    f = self.replace_times_and_trim_frame(f)
  File "/home/jlashner/.local/lib/python3.7/site-packages/sotodlib/io/bookbinder.py", line 527, in replace_times_and_trim_frame
    f[k].times = t
ValueError: Cannot set .times because it conflicts with the established number of samples (400).
```
Seems concerning that this is failing error is being raised in 3.7 but not 3.10... I'll have to look a bit more into what's going on here

guanyilun:
@jlashner this is a known problem with the spt3g built in the tod_stack_unstable/py37. The recommended solution was to switch to py310...

jlashner:
Ah I see, that makes sense then. I was using a local build of spt3g but removed it when I upgraded to py310. But if this is also a numpy versioning issue we should merge this anyway. @guanyilun have you tested to see if this PR fixes the issue?

jlashner (approved):

guanyilun:
Yes, this fixes the issue for me. Good to merge!

PR #400, opened 2023-02-07
Bookbinder HWP binding

jlashner:
This PR adds functionality to the bookbinder which adds processed HWP data to books.

If the `hwp_root` is passed, this will load in the processed HWP g3 files, and use it to fill out the ancillary data files in books.

This tries to replicate the format described in the tod_format doc as closely as possible. For the negative integer values that report HWP conditions, I use a bitmask with flags for when the hwp is stationary, stable, etc. based on what's in the processed data file. I assume this is something people have opinions about so we should probably discuss the correct flags for this.

This makes a few other small changes in the bookbinder that I came across when testing:
- the `ancil` field is now added to ancillary frames even if there's no data (Matthew also noticed this in #388)
- G3Files are now properly closed with EndProcessing frames

I've tested this on a few books on simons1, but probably needs to be tested with the whole p10r2 archive. What is the best way to do this?

### DIFF
--- a/sotodlib/io/bookbinder.py
+++ b/sotodlib/io/bookbinder.py
@@ -6,6 +6,15 @@ from spt3g import core
 import numpy as np
 import itertools
 import yaml
+from scipy.interpolate import interp1d
+# 
+class HWPFlags:
+    STATIONARY = 2**0
+    UNLOCKED = 2**1
+    UNSTABLE = 2**2
+
+# Minum spin rate before we declare stationary
+MIN_HWP_RATE = 0.2  # Hz
 
 
 def pos2vel(p):
@@ -313,6 +322,7 @@ class FrameProcessor(object):
                  timing_system=False, **config):
         self.hkbundle = None
         self.smbundle = None
+        self.hwp_loader = None
         self.flush_time = None
         self.maxlength = config.get("maxlength", 10000)
         self.FLAGGED_SAMPLE_VALUE = config.get("flagged_sample_value", -1)
@@ -571,6 +581,33 @@ class FrameProcessor(object):
         if smurf_primary is not None:
             f['primary'] = smurf_primary
 
+        # Ancillary data object to be written to disk (co-sampled HK encoder data)
+        anc_data = core.G3TimesampleMap()
+        anc_data.times = smurf_data.times
+
+        ## Adds HWP data if present
+        if self.hwp_loader is not None:
+            ts = np.array(smurf_data.times) / core.G3Units.s
+            hwp_angle = self.hwp_loader.hwp_angle_interp(ts)
+            hwp_angle = np.mod(hwp_angle, 2*np.pi)
+
+            hwp_locked = self.hwp_loader.locked_interp(ts)
+            hwp_stable = self.hwp_loader.stable_interp(ts)
+            hwp_rate = self.hwp_loader.stable_interp(ts)
+
+            flags = np.zeros_like(hwp_angle, dtype=np.int32)
+            m = hwp_rate < MIN_HWP_RATE
+            flags[m] = np.bitwise_or(flags[m], HWPFlags.STATIONARY)
+            m = hwp_locked != 1
+            flags[m] = np.bitwise_or(flags[m], HWPFlags.UNLOCKED)
+            m = hwp_stable != 1
+            flags[m] = np.bitwise_or(flags[m], HWPFlags.UNSTABLE)
+
+            hwp_angle[flags != 0] = -flags[flags != 0]
+
+            anc_data['hwp_enc'] = core.G3VectorDouble(hwp_angle)
+
+
         try:
             hk_data = self.hkbundle.rebundle(flush_time)
         except AttributeError:
@@ -596,18 +633,14 @@ class FrameProcessor(object):
                 flag_smurfgap = np.logical_or(flag_smurfgap, np.logical_and(t > gap_start, t < gap_end))
             f['flag_smurfgap'] = core.G3VectorBool(flag_smurfgap)
 
-            # Ancillary data (co-sampled HK encoder data)
-            anc_data = core.G3TimesampleMap()
-            anc_data.times = smurf_data.times
             anc_data['az_enc'] = core.G3VectorDouble(cosamp_az)
             anc_data['el_enc'] = core.G3VectorDouble(cosamp_el)
-
-            # Write ancillary data to frame
-            f['ancil'] = anc_data
-
             self.determine_state(hk_data['Azimuth_Velocity'], hk_data['Elevation_Velocity'])
             f['state'] = self.current_state
+
         finally:
+            # Write ancillary data to frame
+            f['ancil'] = anc_data
             output += [f]
 
         return output
@@ -770,9 +803,11 @@ class Bookbinder(object):
     """
     def __init__(self, smurf_files, hk_files=None, out_root='.', book_id=None,
                  session_id=None, stream_id=None, start_time=None, end_time=None,
-                 smurf_timestamps=None, timing_system=False, verbose=True, **config):
+                 smurf_timestamps=None, timing_system=False, verbose=True, 
+                 hwp_root=None, **config):
         self._smurf_files = smurf_files
         self._hk_files = hk_files
+        self._hwp_root = hwp_root
         self._book_id = book_id
         self._out_root = out_root
         self._session_id = session_id
@@ -793,6 +828,9 @@ class Bookbinder(object):
         self.MAX_SAMPLES_PER_CHANNEL = self.MAX_SAMPLES_TOTAL // self.max_nchannels
         self.DEFAULT_TIME = core.G3Time(1e18)  # 1e18 = 2286-11-20T17:46:40.000000000 (in the distant future)
         self.OVERWRITE_ANCIL_FILE = config.get("overwrite_afile", False)
+        self.hwp_loader = None
+        self.ancil_writer = None
+        self.writer = None
 
         if isinstance(self._hk_files, str):
             self._hk_files = [self._hk_files]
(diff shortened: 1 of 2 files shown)


sotodlib/io/bookbinder.py
```diff
@@ -571,6 +582,37 @@ def flush(self, flush_time=None):
...
+                flags[m] = np.bitwise_or(flags[m], HWPFlags.UNLOCKED)
+                m = hwp_stable != 1
+                flags[m] = np.bitwise_or(flags[m], HWPFlags.UNSTABLE)
+                m = np.isnan(hwp_angle)
+                flags[m] = np.bitwise_or(flags[m], HWPFlags.NO_DATA)
+
+                hwp_angle[flags != 0] = -flags[flags != 0]
+            else:
+                hwp_angle = np.full(len(smurf_data.times), -HWPFlags.NO_DATA)
+
```
algadoreo, line 613: Perhaps this should be in its own method (e.g., `get_hwp_angles`), to avoid having too much in `flush`.

sotodlib/io/bookbinder.py
```diff
@@ -1034,6 +1090,14 @@ def compile_mfile_dict(self):
              'end_time':   self._end_time.time/core.G3Units.s,
              'n_frames':   self.frame_num,
              'n_samples':  self.sample_num}
+
+        if self.hwp_loader is not None:
+            if self.hwp_loader.hwp_angle_interp is not None:  # If there's data
+                hwp_rate = float(np.mean(self.hwp_loader.data['slow']['hwp_rate']))
```
algadoreo, line 1096: What’s the reason for the `float()` here?

sotodlib/io/bookbinder.py
```diff
@@ -1125,3 +1196,105 @@ def __call__(self):
...
+                times['slow'], data['slow']['locked'], bounds_error=False)
+            self.stable_interp = interp1d(
+                times['slow'], data['slow']['stable'], bounds_error=False)
+            self.hwp_rate_interp = interp1d(
+                times['slow'], data['slow']['hwp_rate'], bounds_error=False)
+        else:
+            self.hwp_angle_interp = None
+            self.locked_interp = None
+            self.stable_interp = None
+            self.hwp_rate_interp = None
```
algadoreo, line 1300: Do we anticipate needing this elsewhere? If so, might be good as a separate module/file in the sotodlib/io/ directory.

I usually put a newline at the end of my files out of habit, but I guess that’s not strictly necessary.

algadoreo (approved):
This looks good to me on a read-through. Since I don’t have anything to test it with, do you plan on adding a unit test to test_bookbinder.py? (Doesn’t have to be now.) Otherwise, my only comments are on the organizational aspects.

Also, thanks for adding the `EndProcessing` frame—I wasn’t aware that was a thing.

sotodlib/io/bookbinder.py
```diff
@@ -571,6 +582,37 @@ def flush(self, flush_time=None):
...
+        ## Adds HWP data if present
+        if self.hwp_loader is not None:
+            if self.hwp_loader.hwp_angle_interp is not None:
+                ts = np.array(smurf_data.times) / core.G3Units.s
+                hwp_angle = self.hwp_loader.hwp_angle_interp(ts)
+                hwp_angle = np.mod(hwp_angle, 2*np.pi)
+
+                hwp_locked = self.hwp_loader.locked_interp(ts)
+                hwp_stable = self.hwp_loader.stable_interp(ts)
+                hwp_rate = self.hwp_loader.stable_interp(ts)
```
guanyilun, line 598: typo? hwp_rate_interp?

sotodlib/io/imprinter.py
```diff
@@ -257,7 +257,7 @@ def register_books(self, bids, obs_lists, commit=True, session=None):
             self.register_book(session, bid, obs_list, commit=False)
         if commit: session.commit()
 
-    def bind_book(self, book, session=None, output_root="out", message=""):
+    def bind_book(self, book, session=None, output_root="out", message="", hwp_root=None):
```
guanyilun, line 260: I think I might prefer to no change to the function signature here because they need to be propagate downstream. Shouldn't hwp_root be already specified in the g3tsmurf config file which is stored in `self.sources[book.tel_tube]['g3tsmurf']`? This will be easier for downstream tasks.

guanyilun (changes requested):

sotodlib/io/bookbinder.py
```diff
@@ -1125,3 +1196,105 @@ def __call__(self):
...
+        Gets HWP files that may have data between start and stop time
+        """
+        files = []
+        for subdir in os.listdir(self.root_dir):
+            try:
+                tcode = int(subdir)
+            except:
+                continue
+
+            if not (((start//1e5) - 1) <= tcode <= stop//1e5):
```
guanyilun, line 1227: Why not stop//1e5 + 1?

sotodlib/io/bookbinder.py
```diff
@@ -1125,3 +1196,105 @@ def __call__(self):
...
+
+        files = np.array(sorted(files))
+        file_times = np.array([int(os.path.basename(f).split('.')[0]) for f in files])
+
+        m = (start <= file_times) & (file_times < stop)
+        if not np.any(m):
+            self.files = []
+            return self.files
+
+        # Add file before for good measure
```
guanyilun, line 1241: What about the file afterward?

sotodlib/io/bookbinder.py
```diff
@@ -1125,3 +1196,105 @@ def __call__(self):
...
+                if bname not in data:
+                    data[bname] = {
+                        k: [] for k in block.keys()
+                    }
+                    times[bname] = []
+
+                ts = np.array(block.times) / core.G3Units.s
+
+                # Includes a sec before and after for interpolation
+                m = ((start - 1) < ts) & (ts < (stop + 1))
```
guanyilun, line 1273: We should be careful about the assumption here. The start and stop times in bookbinder come from g3tsmurf db. @kmharrington 
could clarify how that timestamp is obtained. These timestamps often do not line up with the actual .times field, probably more so when timing system is on. Here it's assuming they can't be off by more than 1s. Is that a safe assumption?

sotodlib/io/bookbinder.py
jlashner: Sounds good, I can add this in a method of the HWPLoader

sotodlib/io/bookbinder.py
jlashner: I had to add this because mean returns a `np.float`, which isn't formatted well in yaml

sotodlib/io/bookbinder.py
jlashner: Ah I'm too spoiled by the autoformatter that Brian added in ocs/socs haha. I'll add it. I'm not sure how useful it is elsewhere in sotodlib, so I kind of think we should just keep it here for now since we can easily move it to a separate module later if that ends up being helpful

sotodlib/io/bookbinder.py
jlashner: Yep, typo, good catch

sotodlib/io/bookbinder.py
jlashner: My reasoning for these start time buffers is just because the sample with timestamp `t` will fall in a file timestamped before the time `t`. So we have to check the file before the start time, but the file after the end time shouldn't contain any samples that we care about.

However as you mention below, the timestamp in the filename may not actually be a lower bound for the samples in file, so you are right that we should probably include a buffer on the `stop` time also.

sotodlib/io/bookbinder.py
jlashner: Ah this is a really good point. Since there is this uncertainty, I think we may as well just use a much larger buffer, of like `30 sec` or so on each side.

sotodlib/io/bookbinder.py
algadoreo: In that case, could you add a brief comment for future viewers?

jlashner:
Great thanks for your comments! Just added a commit that addresses both of your feedback, so if you could re-review that would be awesome.

As far as testing goes, I could add unit tests for the HWPLoader class. I think one thing we're lacking is a end-to-end test of the bookbinder. One that generates a small fake data archive and attempts to run the `bind_book` function, which should cover a lot of the bookbinder class. Maybe we can add that in a separate PR?

guanyilun (approved):
looking good, thanks for the changes!

guanyilun:
> I think one thing we're lacking is a end-to-end test of the bookbinder. One that generates a small fake data archive and attempts to run the `bind_book` function, which should cover a lot of the bookbinder class. Maybe we can add that in a separate PR?

+1 for setting end-to-end testing for bookbinder

algadoreo:
I’m good with merging this. End-to-end testing is definitely outside the scope of this PR. If we’re going to make one though, we should include the other HK stuff since that has not been tested at all.

kmharrington (changes requested):
Two in-line comments then another question:

I see HWP rate in the `M_index.yaml`. Can we add an overall flag there as well. Something that we could build a database off of later for indexing what the HWP is doing. I think that it's basically going to be some level of all or any calls to the timestream flag you built. (What those flags mean also need to go into the tod2maps docs).

sotodlib/io/bookbinder.py
```diff
@@ -6,6 +6,16 @@
 import numpy as np
 import itertools
 import yaml
+from scipy.interpolate import interp1d
+# 
+class HWPFlags:
```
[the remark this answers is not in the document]
jlashner: This doesn't add an additional field for flags. In the tod_format docs it says that the `hwp_enc` field should be a float between [0, 2pi), and a negative integer to report any other conditions. So these flags are just used to determine what that negative integer should be.

I actually wasn't sure if these are actually good flags to use and wanted to open it up to a larger discussion, since I'm not so familiar with the HWP data and don't really know what goes into constructing the angle (to be honest I don't really know how "locked" and "stable" should be interpreted). Do you have any suggestions for what flags would be useful here?

jlashner:
Hi all, thanks for the great feedback! I was finally able to finish implementing changes.

I changed the HWPFlags based on my new understanding of what's in the level 2.5 HWP files, so we only are flagging if "stationary" (i.e. stable and not locked) and "no data" (i.e. not stable).

I also added the HWP rate taken once every 10 minutes to the meta file, along with a description to tell us whether the HWP is stationary, changing, or constant.

Let me know if there are any other issues!

kmharrington:
looks good to me. Sorry for the delay on the review

kmharrington (approved):
Looks good to me, sorry for the detail on the review.

jlashner:
Should we bother merging this since we decided to not include HWP data in the ancillary data files? I'm fine with just closing it.

kmharrington:
If the changes aren't breaking changes then I think I'd like to "keep this in our back pocket" so that if the plan changes again no one has to re-do all you've done.

PR #401, opened 2023-02-07
Imprinter: add support for timing system and its fallback

guanyilun:
This PR includes various small changes to imprinter to work with the timing updates. This completes the last missing piece of #357. Also fixes point 4 in #388.

### DIFF
--- a/sotodlib/io/imprinter.py
+++ b/sotodlib/io/imprinter.py
@@ -669,7 +669,7 @@ def stream_timestamp(obs_id):
         timestamp of the observation
 
     """
-    return "_".join(obs_id.split("_")[:-1]), obs_id.split("_")[-1]
+    return "_".join(obs_id.split("_")[1:-1]), obs_id.split("_")[-1]
 
 def get_obs_type(obs: G3tObservations):
     """Get the type of observation based on the observation id
--- a/sotodlib/io/imprinter.py
+++ b/sotodlib/io/imprinter.py
@@ -3,6 +3,7 @@ import numpy as np
 from collections import OrderedDict
 from typing import List
 import yaml, traceback
+import shutil
 
 import sqlalchemy as db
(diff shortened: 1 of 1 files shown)


kmharrington (commented):
What I see here looks good but I think I'm missing part of it. How does setting `book.timing = False` in the exception lead to the books getting Rebound in a different way?

guanyilun:
> What I see here looks good but I think I'm missing part of it. How does setting `book.timing = False` in the exception lead to the books getting Rebound in a different way?

Ah sorry for the confusion! Failed books will be automatically retried in the next run, so when it reruns it will be trying with timing=False.

kmharrington (approved):
Ok, great,  one new question. `REBIND` is saying "we have those books but we want to delete and re-write them." I am guessing that is something done for debugging / testing since we're fixing lots of things right now. If so, that's cool. 

But I want to make sure nothing is automatically tagging books as rebind and it's not something we plan to use in the field. Because "we will never remake books" once they've been passed off to the librarian is a pretty big tenant of the books.

guanyilun:
> Ok, great, one new question. `REBIND` is saying "we have those books but we want to delete and re-write them." I am guessing that is something done for debugging / testing since we're fixing lots of things right now. If so, that's cool.
> 
> But I want to make sure nothing is automatically tagging books as rebind and it's not something we plan to use in the field. Because "we will never remake books" once they've been passed off to the librarian is a pretty big tenant of the books.

You are right - they are only a manual handle to remake books when there's some major non-compatible changes in bookbinder. I don't expect to use this much in production, unless it's to resolve some specific failure mode. I've seen some failure mode where g3tsmurf db indexes files while Observations are being written to disk, which leads to Imprinter registering a book that's missing one wafer, in that case I expect to manually label them as rebind and let the imprinter maintanance bot (which runs daily) to handle those cases.

Issue #353, opened 2022-12-20
make-obs-book needs to look for dropped frames 


kmharrington:
The .g3 smurf files with timing integrated have a whole lot of new information we need to track. In particular, we need to look for dropped frames in the counters, interpolate through them, and flag them somehow so that we know what's been done. 

This needs to be implemented and tested extensively on the P10R1 / P10R2 data.

kmharrington:
From discussion on the data packaging call, there are several parts to this. First: this is mostly dropped samples on the edge of frames instead of entire frames. The issues at the beginning and end of files are different than samples in the middle. 

For the beginning and end of files/observations. G3tsmurf needs to update the times of fields to reflect the timing system times when available. @guanyilun  Can I double check that the times passed to bookbinder for start/stops originate from `obs.start` and `obs.stop`? 

So, if there are missing samples at the beginning or end of the book it's because the latest `obs.start` in a multi-book observation isn't actually the latest sample time. I think these are because counter time != frame.times. I think the easiest way to fix that would be to use counter time to set `obs.start` and `obs.stop` and I can write a g3tsmurf fix to do that.

Sample drops in the middle are a whole different set of issues. We discussed @jlashner  looking at a preprocessing script that would find out the timestamps a-priori so bookbinder can easily slot in missing samples. Seems like the additional run time there isn't too bad? tbd

guanyilun:
> Can I double check that the times passed to bookbinder for start/stops originate from `obs.start` and `obs.stop`?

yes that's right

Issue #389, opened 2023-01-30
Multi-slot bookbinder sample count bugs?


mhasself:
Most multi-slot books I've looked at have problems with the sample count -- the stream_ids all have different numbers of samples.  This may simply be due to unpatched dropped frames (#127).  But in at least one case the acquisition is short enough that dropped frames are unlikely, yet still there is a sample count discrepancy.

`/mnt/so1/shared/site-pipeline/data_pkg/books/ucsd-sat1/obs/16747/obs_1674766989_sat1_1111101`
```
ufm_mv14:  N= 11370
ufm_mv18: N= 11370
ufm_mv19: N= 11369
ufm_mv22: N= 11369
ufm_mv6: N= 11369
ufm_mv9: N= 11369

algadoreo:
My guess is that the smurf-streamer based timestamps is causing some samples to fall on the wrong side of the book’s end time. Let’s see if they still persist after the timing system issue gets fixed (#387).

mhasself:
Understood.

Are the time boundaries set to be squarely _between_ adjacent samples (to minimize risks of rounding error)?

algadoreo:
No—the start/end times are pulled by the Imprinter from the database and passed on to Bookbinder. But now that you mention it, I’m not sure how those start/end times are generated.

Nonetheless, if the timing system is on (and everything works correctly), every wafer should have identical timestamps, right? Would it still be sensitive to rounding errors in that case?

algadoreo:
@guanyilun pointed out that the culprit could also be the gap checker, since it would add a sample if it thinks the gap between frames is too long (a symptom of the inaccurate smurf-streamer timestamps). Either way, we can see what happens after timing system issue is resolved.

algadoreo:
I believe this got fixed by #401—this was the earliest PR I found where the output produces the correct number of samples (N=11369) for all wafers. Although jlashner’s `get_book_times` function mentioned above returns a list with 11370 samples, the last one falls outside the book’s end time and is trimmed.

Perhaps @guanyilun can verify if that PR did anything that changed the behaviour in the right way, e.g., added precise timestamps for start/end times?

mhasself:
I think this is all fixed now...

PR #405, opened 2023-02-10
`get_book_times` helper function

jlashner:
This PR adds a new function to the imprinter that takes in an obs book (assuming a timing system), and returns a full gap-less list of timestamps that should be contained in the book.

This only takes 15 seconds to run for a 2 hour run for a two-hour timestream and I think could both:
-  simplify the gap-filling logic of the bookbinder, and allow for frame pre-allocation which may speed up current run-times
- Help the bookbinder determine whether there are missing samples at the start/end of files and frames

Relevant: #353

### DIFF
--- a/sotodlib/io/imprinter.py
+++ b/sotodlib/io/imprinter.py
@@ -688,23 +688,33 @@ class Imprinter:
 
     def get_book_times(self, book):
         """
-        Returns an array of gap-filled timestamps for a book.
+        Given a book of co-sampled data with a timing system, this function will
+        return a gap-filled list of timestamps that should be contained in the book.
+
+        Parameters
+        ----------
+        book: Book object
         """
         filedb = self.get_files_for_book(book)
-        times = {}
-        for k, files in filedb.items():
-            ts = []
-            for frame in itertools.chain(*[core.G3File(f) for f in files]):
-                if frame.type != core.G3FrameType.Scan:
-                    continue
-                ts.append(get_timestamps(frame))
-            times[k] = fill_gaps(np.hstack(ts))
+        t0 = book.start.timestamp()
+        t1 = book.stop.timestamp()
+        for i, files in enumerate(filedb.values()):
+            _files = sorted(files)
+            if i == 0:  # Get full list of timestamps for first file list
+                ts = []
+                for frame in itertools.chain(*[core.G3File(f) for f in _files]):
+                    if frame.type != core.G3FrameType.Scan:
+                        continue
+                    ts.append(get_timestamps(frame))
+                ts = fill_gaps(np.hstack(ts))
 
-        t0 = np.max([ts[0] for ts in times.values()] + [book.start.timestamp()])
-        t1 = np.min([ts[-1] for ts in times.values()] + [book.stop.timestamp()])
+                _t0, _t1 = ts[[0, -1]]
+            else:
+                _t0, _t1 = get_start_and_end(_files)
+            
+            t0 = max(t0, _t0)
+            t1 = min(t1, _t1)
 
-        # Doesn't matter which we choose because they will all be the same once trimmed
-        ts = list(times.values())[0]
         m = (t0 <= ts) & (ts <= t1)
         return ts[m]
 
@@ -715,6 +725,16 @@ class Imprinter:
 def get_timestamps(frame):
     """
     Returns an array of timestamps for data in a G3 Frame
+
+    Parameters
+    -------------
+    frame : G3Frame
+        Scan frame containing detector data
+    
+    Returns
+    -------------
+    timestamps : np.ndarray
+        Array of timestamps (seconds) for samples in the frame
     """
     c0, c2 = frame['primary'].data[[3, 5]]
     s, ns = split_ts_bits(c2)
@@ -727,6 +747,16 @@ def get_timestamps(frame):
 def fill_gaps(ts):
     """
     Fills gaps in an array of timestamps.
+
+    Parameters
+    -------------
+    ts : np.ndarray
+        List of timestamps of length `n`, potentially with gaps
+    
+    Returns
+    --------
+    new_ts : np.ndarray
+        New list of timestamps of length >= n, with gaps filled.
     """
     # Find indices where gaps occur and how long each gap is
     dts = np.diff(ts)
@@ -749,6 +779,39 @@ def fill_gaps(ts):
     new_ts[m] = interp(xs[m])
 
     return new_ts
+
+def get_start_and_end(files):
+    """
+    Gets start and end time for a list of L2 detector data files
+
+    Parameters
+    -------------
+    files : np.ndarray
+        List of L2 G3 files for a single detector set in an observation
+    
+    Returns
+    --------
+    t0 : float
+        Start time of observation
+    t1 : float
+        End time or observation
+    """
+    _files = sorted(files)
+
+    # Get start time
+    for frame in core.G3File(_files[0]):
+        if frame.type == core.G3FrameType.Scan:
+            t0 = get_timestamps(frame)[0]
+            break
+    
+    # Get end time
+    frame = None
+    for _frame in core.G3File(_files[-1]):
+        if _frame.type == core.G3FrameType.Scan:
+            frame = _frame
+    t1 = get_timestamps(frame)[-1]
+    
+    return t0, t1
 
 def create_g3tsmurf_session(config):
     """create a connection session to g3tsmurf database
--- a/sotodlib/io/imprinter.py
+++ b/sotodlib/io/imprinter.py
@@ -10,8 +10,12 @@ from sqlalchemy import or_, and_, not_
 from sqlalchemy.orm import sessionmaker
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import relationship
+from spt3g import core
+import itertools
+from scipy.interpolate import interp1d
 
 from .load_smurf import G3tSmurf, Observations as G3tObservations, SmurfStatus, get_channel_info
+from .load_smurf import split_ts_bits
 from .bookbinder import Bookbinder, TimingSystemError
 from ..site_pipeline.util import init_logger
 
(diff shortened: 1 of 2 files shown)


sotodlib/io/imprinter.py
```diff
@@ -682,9 +686,132 @@ def delete_book_files(self, book):
             self.logger.warning(f"Failed to remove {book.path}: {e}")
             self.logger.error(traceback.format_exc())
 
+    def get_book_times(self, book):
+        """
+        Given a book of co-sampled data with a timing system, this function will
+        return a gap-filled list of timestamps that should be contained in the book.
```
mhasself, line 692: So this does the "trimming" for you ... maybe mention that.

This is an opportunity to report how much time has been trimmed from the various filesets to make them all the same length... but I imagine that's been checked carefully elsewhere.

sotodlib/io/imprinter.py
```diff
@@ -682,9 +686,132 @@ def delete_book_files(self, book):
...
+        Parameters
+        ----------
+        book: Book object
+        """
+        filedb = self.get_files_for_book(book)
+        t0 = book.start.timestamp()
+        t1 = book.stop.timestamp()
+        for i, files in enumerate(filedb.values()):
+            _files = sorted(files)
+            if i == 0:  # Get full list of timestamps for first file list
```
algadoreo, line 703: A downside to using only one wafer to determine the list of timestamps is that the filled-in samples may not line up exactly with the samples that exist in other wafers. It probably doesn’t matter since we’ll flag those samples anyway, but I thought I should point that out.

I realize that checking all the wafers adds computational time, so maybe it could exist as a user-option to perform a more “comprehensive” search: the times for each wafer is assembled and trimmed, then look for unique times in the whole set. (Uniqueness only works when the timing system is on, but the correct fallback for no timing system is to make single-wafer books where it’s okay if samples don’t line up.) The vector of unique times is then filled with the missing times.

sotodlib/io/imprinter.py
```diff
@@ -682,9 +686,132 @@ def delete_book_files(self, book):
...
+    
+    Returns
+    --------
+    new_ts : np.ndarray
+        New list of timestamps of length >= n, with gaps filled.
+    """
+    # Find indices where gaps occur and how long each gap is
+    dts = np.diff(ts)
+    dt = np.mean(dts)
+    missing = np.round(dts/dt - 1).astype(int)
```
algadoreo, line 764: Would median be better here? I think a large gap (or a large number of smaller gaps) might skew the mean too much upward. On the other hand, the median would be affected by a large variance, but a timing system should keep that in check, right?

Also, probably an unlikely scenario, but if for some reason you get two samples close to each other (dts/dt < 0.5), then I think you end up with -1 missing samples, which would break the subsequent code. Maybe add a max(0, missing) or something to be safe?

sotodlib/io/imprinter.py
```diff
@@ -682,9 +686,132 @@ def delete_book_files(self, book):
...
+    """
+    c0, c2 = frame['primary'].data[[3, 5]]
+    s, ns = split_ts_bits(c2)
+    # Add 20 years in seconds (accounting for leap years) to handle
+    # offset between EPICS time referenced to 1990 relative to UNIX time.
+    c2 = s + ns*1e-9 + 5*(4*365 + 1)*24*60*60
+    timestamps = np.round(c2 - (c0 / 480000) ) + c0 / 480000
+    return timestamps
+
+def fill_gaps(ts):
```
algadoreo, line 747: Perhaps `fill_time_gaps` to make it clear it’s for time gaps?

algadoreo (commented):
Thanks for this! It’ll make my life a lot easier once this is merged.

sotodlib/io/imprinter.py
jlashner: Right. where do you think would be a good place to report this? should it make its way into the M_index file?

sotodlib/io/imprinter.py
jlashner: Not sure I understand you here. Are you talking about if the slots are not actually co-sampled from a timing system?

It kind of seems like you are talking about a case where the filled timestamps are something like:
`[0, 1, 2, 3, 4, 5]`, but one sample happens to have an extra timestamp like `[0, 1, 2, 3, 3.5, 4, 5]`.
I think the correct action here would be to not include the `3.5` sample of the second array, which I think this method will actually correctly handle.

I'm probably misinterpreting you're concern though... do you have like an example of timestamp arrays that demonstrates your thought?

algadoreo: I mean something like this: the filled-in timestamps are based on the mean/median dt, right? so if you have `[0, 1, 2, 3, 5]` in the first wafer, the algorithm would think the missing sample should be `4` when going with the median (or `4.25` when based on the mean). But it may turn out that all the other wafers actually had `[0, 1, 2, 3, 4.1, 5]`. The filled-in time value isn’t exactly what it should be. (This is a contrived example, but the idea is that your mean/median is only approximating where the true timestamps are)

But we’re going to be flagging those filled-in samples anyway, so maybe this doesn’t matter in the end.

sotodlib/io/imprinter.py
jlashner: Sure, median would probably be more correct, I will update this. I hope large gaps like your describing would trigger some sort of warning somewhere.... Maybe we should do some sort of validation that tests that all arrays are indeed co-sampled before binding. That is not the purpose of this function but it might be a natural place to put it (like you mention above) since we're already extracting one set of timestamps.

For this scenario, I think we first have to figure out how we should actually handle this sort of event. If `dts/dt < 0.5` we may want to drop the sample. Like if we have `[..., 1, 2, 2.1, 3, 4]`, I think the resulting timestamps should still be [..., 1, 2, 3, 4]`

algadoreo: Yeah those are definitely things to do down the line—no need to hold up this PR for them, I think. Nonetheless, the negative samples does mean I have to add a check in Bookbinder.

sotodlib/io/imprinter.py
jlashner: Right... I'm not exactly sure what the correct behavior would be in this case.... maybe @mhasself can comment, I'm sure this has been thought through for other projects.

As far as I'm aware, these issues where the timing sample is off by a large amount and it's not due to a dropped sample are exceedingly rare / not possible, so I'm not sure how much we want to concern ourselves with those edge cases. If we are concerned... I think the right thing to do is to first decide how we want to be handling those cases, write tests, and then afterwards modify the function so it handles them properly.

sotodlib/io/imprinter.py
```diff
@@ -682,9 +686,132 @@ def delete_book_files(self, book):
...
+                    if frame.type != core.G3FrameType.Scan:
+                        continue
+                    ts.append(get_timestamps(frame))
+                ts = fill_gaps(np.hstack(ts))
+
+                _t0, _t1 = ts[[0, -1]]
+            else:
+                _t0, _t1 = get_start_and_end(_files)
+            
+            t0 = max(t0, _t0)
```
guanyilun, line 715: I'm not sure we want to use book.start and book.stop to inform the min and max here, with the same reasoning as in the HWP PR that these timestamps are not accurate. You should only trust the timestamps from frames, so just point t0 and t1 to _t0 and _t1 in the first wafer would suffice.

guanyilun (commented):
Sorry I don't really understand this `get_book_times` function. For example, how is returning a gap-filled timestamp array more helpful for bookbinder versus just passing the start and stop time and the sampling rate?

jlashner:
We talked a bit about this on the call on Thursday, but in my opinion having a full array of all the timestamps that should be present in the book will greatly simplify the logic in the bookbinder. Currently, [the bookbinder is trying to detect and fill gaps](https://github.com/simonsobs/sotodlib/blob/c8bd0c41797dd73019fe86602340bfc0c6cd9ce9/sotodlib/io/bookbinder.py#L702-L743) after reading in each smurf frame individually, which is very difficult to get right since you have much less information. As Jason brought up, this is especially hard at the start/end of files and frames, and the current implementation may have [issues](https://github.com/simonsobs/sotodlib/issues/389).

I think it is more natural to have an initial step that first loads in the entire set of timestamps for an observation (assuming the runtime isn't too bad), and finding and filling gaps all at once using the function PR'ed here. Doing this in the imprinter, which has access to files from all UFMs, and passing the array of timestamps to the bookbinder will essentially eliminate the possibility of sampling errors like Matthew pointed out. This would be also be a natural place to add validation to check that the observations are actually co-sampled if we want to. 

Additionally, with the array of timestamps and the HK data, we have all we need to figure out the frame boundaries and the samples per frame for the entire book before we load in the smurf data. This could simplify the logic of the bookbinder, so it just needs to:
- allocate the array for a single output frame
- read in smurf data and put each sample into its closest slot
- interpolate in the slots that didn't get filled

Of course, you and Jason have thought about the bookbinder a lot more than I have, so you definitely don't have to use this tool if you think you have an easier solution to #389 that you'd rather implement.

guanyilun:
Thanks for the clarification! I definitely think this function will be helpful for bookbinder for the reasons you listed and agree that imprinter is the best place for it. My comment earlier was more of an interface question, that is, I’m trying to understand the benefits of passing the entire timestamp array (ts) versus just ts[0], ts[-1] and np.median(np.diff(ts)) with which we can approximately reconstruct the entire array whenever needed. Naively I would expect them to be equivalent because we only care about the missing samples and they are gap-filled in your timestamp arrays based sampling rate anyway, but I have not thought about it deeply, maybe the timing system is not stable enough for this to be a robust method.

jlashner:
Right, the issue is I don't think it's quite that simple to reconstruct `ts` from just `ts[0]`, `ts[-1]`, and the median sampling rate... 

If I compare the gap-filled timestamps I compute here `ts` with a reconstruction using 
```ts2 = np.arange(ts[0], ts[-1], np.median(np.diff(ts))```
you can see that the error from the median is blown up when you try to extrapolate to multi-hour long timestreams as is shown here:
image

There may be ways around this if you are more clever about how you're doing the reconstruction... for instance, computing the total number of samples in the gap-filled array and using `np.linspace` may work, but I think it is safer and easier to just pass in the full array of times, in which case you are sure that for the samples that are present, the timestamps are exactly the same as they are in the smurf files.

**Comment by guanyilun** [comment]
That’s convincing, thanks for the plot!

File: sotodlib/io/imprinter.py (hunk 1 above)
Comment (kmharrington) [reply sotodlib/io/imprinter.py:692]: I don't think I've seen the trimming amount reported anywhere yet.

File: sotodlib/io/imprinter.py [hunk 6]
```diff
@@ -682,9 +686,132 @@ def delete_book_files(self, book):
...
+            t1 = min(t1, _t1)
+
+        m = (t0 <= ts) & (ts <= t1)
+        return ts[m]
+
+
 #####################
 # Utility functions #
 #####################
+def get_timestamps(frame):
```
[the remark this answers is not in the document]
kmharrington: I approve of make this a general utility function and then only having one of them.

sotodlib/io/imprinter.py
mhasself: The fill isn't based on median, actually ... it's a linterp, which will interpolate exactly between 3 and 5.  So you'll get 4.

However -- which of these new functions are designed to work on data where: (A) the timing system is in operation, and we can probably make multi-wafer books vs. (B) the timing system is not in operation.

The (A) case will be very stable... interpolating will bring only tiny errors, and I think that's fine.  Though you could pretty easily check+patch with the other vectors of timestreams, and this would be a great place to do it.

The (B) case can have very noisy timestamps... but of course in that case you don't need to worry about consistency between multiple wafers because you'll only have one wafer.

I don't think you need to have the "same code" run in cases (A) and (B).  It may be better to have separate, specialized functions for checking / filling / preparing the timestamp vectors in the two cases.

sotodlib/io/imprinter.py
jlashner: I kind of thought this would be a good chance to enforce this, so as it is now the data will be trimmed so that all samples are within `book.start` and `book.stop`. I'm also fine with removing this though, but then we will always need to keep in mind that the `start` and `stop` are not hard limits.

guanyilun: How about replacing book.start and book.end with the correct timestamps from this function?

jlashner: I'll just leave it for now and not cut based on the book `start` and `stop`. I do think we should think about using the `get_start_and_end` function to set book.start and book.stop when the book is created though.

sotodlib/io/imprinter.py
jlashner: Ok cool, I think I'm going to leave it for now and open an issue, since I think this should be done by the book-binder, which creates the M_index file

sotodlib/io/imprinter.py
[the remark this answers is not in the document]
jlashner: Ok ya so I looked into this, and I think slight differences in the functions (i.e. Jason's function casts times to a G3VectorTime before returning) makes it difficult to use the exact same function for both purposes. I think the conversion from counters to timestamps however is a general function with a lot of hardcoded values that should probably be a shared util function. For now I separated this out into a new function `counters_to_timestamps` in the bookbinder, and use that function to construct the timestamps here.

jlashner:
Hi all, thanks for the feedback! I added a commit that I think addresses the biggest points, so I think it's ready to be merged if someone approves it. I think things like tests for handling edge-cases should be added in a separate PR.

guanyilun (approved):

algadoreo:
Wait—in `fill_time_gaps`, L773, weren’t we changing to the median dt instead of the mean?

jlashner:
Oops sorry, merged this right before I saw your comment Jason, let me get that

PR #406, opened 2023-02-13
debug hwp slot pattern subtraction

ykyohei:
Debugged encoder slot patter subtraction.
Previous code fails when the number of encoder slots in the range of interest is specific numbers.

Previous example of error 
on /mnt/so1/data/ucsd-sat1/hk/16750/1675094333.g3
```
line 516, in eval_angle
    solved['fast_time'] = solved['fast_time'] - subtract
ValueError: operands could not be broadcast together with shapes (8296086,) (8295780,) 
```


tterasaki (approved):
Thank you, Kyohei!

PR #408, opened 2023-02-14
load_smurf: reduce RAM footprint when loading a few detectors

mhasself:
load_file will take copies, instead of refs, when pulling a ~small number of dets out of a frame stream.


kmharrington (approved):
Thanks for looking at this so quickly!

PR #409, opened 2023-02-14
Readout simulation operator

keskitalo:
Simulate several readout-related pathologies:

- Electronic glitches: both independent and coincident on the bias line with separate distributions. Correlated glitches have different amplitudes
- Cosmic ray glitches with user-defined rate, amplitude and correlation length
- Jumps: independent and correlated on the bias line
- Failed bolometers: random failures to meet prescribed yield. Yield can be drawn from Gaussian distribution. Failed bolometer TOD is replaced with a randomized linear trend.
- Bolometer misidentification: Swap TOD between bolometers that share frequency, polarization type and bias line. Probability to misidentify decreases with distance based on user-defined Gaussian distribution

Add initial support for life-like hardware map files.


keskitalo:
Point source simulation with and without misidentified bolometers:
image

kmharrington:
very cool! good to know the sim is working. Does toast have a way of outputting per-detector maps?

keskitalo:
TOAST can do it but the workflow would need some changes to support it.

keskitalo:
@mhasself @smsimon Any chance you could take a look at this PR? It is fairly large but there shouldn't be anything controversial about it.

mhasself:
Seems fine... are you going to get the tests working?

When det_info is _not_ being used, does this change the default focal plane layouts?  It is important to make note when that happens.

keskitalo:
@mhasself thanks for noting the unit test failures. I pushed a fix. The default, simulated focalplane geometry has not changed.

msilvafe:
This looks like nice additions to the TOAST sims. This shouldn't hold up merging but can you produce an example params file that shows how to setup a sim using these new readout pathology operators?

keskitalo:
@msilvafe already done :) https://github.com/simonsobs/pwg-scripts/tree/master/pwg-tds/sim-readout

msilvafe:
> @msilvafe already done :) https://github.com/simonsobs/pwg-scripts/tree/master/pwg-tds/sim-readout

Wonderful, thanks @keskitalo!

mhasself (approved):

PR #412, opened 2023-02-17
Bookbinder: Fill in missing samples

algadoreo:
This improves the Bookbinder’s missing-sample checker. Rather than only checking for gaps between input frames when inputting data, it is better to just look for missing samples in the output frame that will be written to a book. This would also catch missing samples that were in the middle of frames, instead of only between frames.

Like before, two methods to check for missing samples are provided: matching with a provided reference timestamp list or using an estimate of the sampling interval to look for gaps. The missing timestamps are filled in using the `fill_time_gaps` function written by jlashner in #405; the samples are assigned the value of FLAGGED_SAMPLE_VALUE (default: -1) and flagged in the output as before.

### DIFF
--- a/sotodlib/io/bookbinder.py
+++ b/sotodlib/io/bookbinder.py
@@ -493,10 +493,12 @@ class FrameProcessor(object):
             # Use estimated sample interval to fill until end_time. Not using fill_time_gaps because
             # end_time is externally determined and does not usually line up with sample interval
             ts = np.append(ts, np.arange(data.times[-1].time+dt, end_time, dt))
+            # for consistency with smurf_timestamps
+            ts /= core.G3Units.s
 
         # Create new G3SuperTimestream with filled-in samples
         if len(data.times) < len(ts):
-            i_missing, _ = find_missing_samples(ts, np.array(data.times))
+            i_missing, _ = find_missing_samples(ts, np.array(data.times)/core.G3Units.s)
             m = np.ones(len(ts), dtype=bool)
             m[i_missing] = False
             assert np.sum(m) == len(data.times)
--- a/sotodlib/io/bookbinder.py
+++ b/sotodlib/io/bookbinder.py
@@ -585,8 +585,9 @@ class FrameProcessor(object):
         if smurf_primary is not None:
             smurf_primary = self.fill_in_missing_samples(smurf_primary, flush_time)
             f['primary'], fx['primary'] = split_supertimestream(smurf_primary, self.maxlength)
-        # Put any excess samples back in the buffer
+        # Put any excess samples (those beyond max length) back in the buffer
         self.smbundle.prepend(fx)
+        # Update (local) flush time if necessary
         if data_excess is not None:
             flush_time = f['signal'].times[-1] + 1
             self._frame_splits.append(flush_time)
@@ -1144,6 +1145,11 @@ def get_channel_data_from_name(s, channel_name):
     return s.data[idx]
 
 def counters_to_timestamps(c0, c2):
+    """
+    Convert timing counter values into timestamp values (sec)
+
+    Copied from load_smurf.py (Jan 23, 2023)
+    """
     s, ns = split_ts_bits(c2)
 
     # Add 20 years in seconds (accounting for leap years) to handle
@@ -1155,8 +1161,6 @@ def counters_to_timestamps(c0, c2):
 def get_timestamps(f, use_counters):
     """
     Calculate the timestamp field for loaded data
-
-    Copied from load_smurf.py (Jan 23, 2023)
 
     Parameters
     ----------
@@ -1205,7 +1209,8 @@ def fill_time_gaps(ts):
     missing = np.round(dts/dt - 1).astype(int)
     total_missing = int(np.sum(missing))
 
-    # Create new  array with the correct number of samples
+    # Create new array with the correct number of samples
+    # Fill with -1 assuming timestamps always > 0
     new_ts = np.full(len(ts) + total_missing, -1, dtype=ts.dtype)
 
     # Insert old timestamps into new array with offsets that account for gaps
--- a/sotodlib/io/bookbinder.py
+++ b/sotodlib/io/bookbinder.py
@@ -473,7 +473,7 @@ class FrameProcessor(object):
         # Obtain list of filled-in timestamps, whether from provided list or estimate with fill_time_gaps
         if self._smurf_timestamps is not None:
             # trim the list of timestamps to the current frame
-            m_ = np.logical_and(self._smurf_timestamps < end_time, self._smurf_timestamps > prev_frame_end)
+            m_ = np.logical_and(self._smurf_timestamps <= end_time, self._smurf_timestamps >= prev_frame_end)
             ts = self._smurf_timestamps[m_]
         else:
             # fallback when no timestamp is provided. We switch to estimate sample interval based on data
@@ -493,13 +493,11 @@ class FrameProcessor(object):
             # Use estimated sample interval to fill until end_time. Not using fill_time_gaps because
             # end_time is externally determined and does not usually line up with sample interval
             ts = np.append(ts, np.arange(data.times[-1].time+dt, end_time, dt))
-            # for consistency with smurf_timestamps
-            ts /= core.G3Units.s
 
         # Create new G3SuperTimestream with filled-in samples
         if len(data.times) < len(ts):
             dt = np.median(np.diff(ts))
-            i_missing, _ = find_missing_samples(ts, np.array(data.times)/core.G3Units.s, atol=dt/2)
+            i_missing, _ = find_missing_samples(ts, np.array(data.times), atol=dt/2)
             m = np.ones(len(ts), dtype=bool)
             m[i_missing] = False
             assert np.sum(m) == len(data.times)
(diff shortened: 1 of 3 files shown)


algadoreo:
Using mhasself’s [check_book.py](https://github.com/simonsobs/sotodlib/pull/398/files#diff-7deb946186b74042af57352bcb61afe9af331a0d4e3822297ad434357f0107d5), I’ve verified that the output book for the observation with known missing samples (`obs_1673307576_sat1_1111101`, from #387) now has the correct number of samples for all wafers (N=877847).

kmharrington (approved):
Sounds like the books now pass tests / can be loaded so going ahead and approving. 

One question I have now is the default flag value / how they're flagged. (Which can be a later PR). @jlashner  I think we can have -1's show up normally? Is there a number that can't possibly show up? Is there another flag field or something that says those values are bad?

jlashner:
I haven't looked at this much yet, but to answer Katie, -1 can definitely show up in the timestream data, and for integer types there isn't really a value that can't possibly show up. Safest bet is to use a separate "flag" array.

One question I have is that right now this seems to support running in a mode where no timestamps are provided. I think we're in a place right now where we can enforce that this should always be run with the correct timestamps provided, so I'm not sure we should be supporting the case in which they're not, since it seems to add complexity / branches that need to be tested. Is this something we want to keep in this PR?

algadoreo:
The output frames contain flag fields for missing samples for both SMuRF data (snippet below) and HK data (a little further down and needs to be improved). The “default value” is normally read in from the config file, where it is currently -1 (although its fallback value defined in the Bookbinder is also -1).

https://github.com/simonsobs/sotodlib/blob/a4f4c68bcca22c4623ff449963c2d7a6a7bf03b2/sotodlib/io/bookbinder.py#L610

Regarding the timestamp list, I always thought it was the opposite—that we would phase it out because of the extra time required to compile the list at the outset. That’s why I wrote the unit test for `fill_in_missing_samples` to go through all branches. But I’m not opposed to removing complexity if a fallback process isn’t required.

sotodlib/io/bookbinder.py
```diff
@@ -563,6 +481,91 @@ def replace_times_and_trim_frame(self, f):
...
+        if len(data.times) < len(ts):
+            dt = np.median(np.diff(ts))
+            i_missing, _ = find_missing_samples(ts, np.array(data.times), atol=dt/2)
+            m = np.ones(len(ts), dtype=bool)
+            m[i_missing] = False
+            assert np.sum(m) == len(data.times)
+            new_data = np.full((data.data.shape[0], len(ts)), self.FLAGGED_SAMPLE_VALUE, dtype=data.data.dtype)
+            new_data[:,m] = data.data
+            dataout = so3g.G3SuperTimestream(data.names, core.G3VectorTime(ts), new_data)
+            flag_gap = core.G3VectorBool(~m)[:self.maxlength]
```
jlashner, line 556: I'm confused why this `self.maxlength` limit is here but isn't applied to the data array or the flag_gap array on line 562. Is this needed? It seems like by construction flag_gap and new_data should already be the same length.

sotodlib/io/bookbinder.py
```diff
@@ -563,6 +481,91 @@ def replace_times_and_trim_frame(self, f):
...
+            ts = np.append(ts, np.arange(data.times[-1].time+dt, end_time, dt, dtype=int))
+
+        # Create new G3SuperTimestream with filled-in samples
+        if len(data.times) < len(ts):
+            dt = np.median(np.diff(ts))
+            i_missing, _ = find_missing_samples(ts, np.array(data.times), atol=dt/2)
+            m = np.ones(len(ts), dtype=bool)
+            m[i_missing] = False
+            assert np.sum(m) == len(data.times)
+            new_data = np.full((data.data.shape[0], len(ts)), self.FLAGGED_SAMPLE_VALUE, dtype=data.data.dtype)
```
jlashner, line 553: I thought we wanted to interpolate here in the case of dropped samples? Like Katie said, setting to a FLAGGED_SAMPLE_VALUE isn't really helpful since that value can be valid detector data, so it can't be used to create a gap mask.

sotodlib/io/bookbinder.py
```diff
@@ -583,21 +586,36 @@ def flush(self, flush_time=None):
...
+        fx = {'data': data_excess}  # note input data use 'data' instead of 'signal'
         if smurf_bias is not None:
-            f['tes_biases'] = smurf_bias
+            smurf_bias = self.fill_in_missing_samples(smurf_bias, flush_time)
+            f['tes_biases'], fx['tes_biases'] = split_supertimestream(smurf_bias, self.maxlength)
         if smurf_primary is not None:
-            f['primary'] = smurf_primary
-
-        # Ancillary data object to be written to disk (co-sampled HK encoder data)
+            smurf_primary = self.fill_in_missing_samples(smurf_primary, flush_time)
```
jlashner, line 598: You're calling fill_in_missing_samples three different times here, but you really only need find the missing samples once since the timestamps are the same for these three arrays.

jlashner (commented):
Ya I think we should be using the pre-computed timestamps. In general they will be more reliable than doing it on the fly, and as far as I can tell the issues with finding missing samples at frame / file boundaries still exist in this code. The timestreams will take like arround 1 min for an hour long timestream, so I don't think the run time is a serious issue, and there are plenty of other places we can optimize  if we think it's taking too long.

I had a few other questions / comments while reading the code in line, but looking good so far!

sotodlib/io/bookbinder.py
algadoreo: It’s there because the frame length may exceed maxlength after filling in missing samples, and while the data array gets passed through `split_timestream` after returning, the flag array is not. _However_, I think this is superseded by

https://github.com/simonsobs/sotodlib/blob/a4f4c68bcca22c4623ff449963c2d7a6a7bf03b2/sotodlib/io/bookbinder.py#L610

and I simply forgot to remove it (since it doesn’t complain even when len(flag_gap) < self.maxlength).

sotodlib/io/bookbinder.py
algadoreo: I’m not following why interpolated values would make it easier to create a gap mask—and isn’t that what the flag array is for?

sotodlib/io/bookbinder.py
algadoreo: If someone has time to optimize this, that would be appreciated! (This is also the reason why the same message gets printed 3 times, so it would be great if it didn’t do that)

algadoreo:
> as far as I can tell the issues with finding missing samples at frame / file boundaries still exist in this code

The timestamp list should include those, right? And that’s also the reason why the fallback block has so many branches:

https://github.com/simonsobs/sotodlib/blob/a4f4c68bcca22c4623ff449963c2d7a6a7bf03b2/sotodlib/io/bookbinder.py#L530-L544

That’s tested below, where `start_time` is already set at `3*dt` earlier than the first data sample:

https://github.com/simonsobs/sotodlib/blob/a4f4c68bcca22c4623ff449963c2d7a6a7bf03b2/tests/test_bookbinder.py#L150

Of course, none of that is needed if we have the comprehensive timestamp list.

sotodlib/io/bookbinder.py
jlashner: People will need to interpolate through the gaps to use the data anyway, so writing interpolated values to disk (with the flag mask that tells you which values were interpolated) will reduce the preprocessing people will need to do on load.

algadoreo: I’m missing something here—why do people need to interpolate through the gap to use it? For example, if I have:

```
Data:   1   2  -1   4   5
Flag:   0   0   1   0   0
```

why would having a -1 (or any other value) instead of 3 cause problems when the flag is there to say “ignore this value”?

kmharrington: If we don't put the interpolation here it's going in the loader b/c this isn't the type of operation we have folks do themselves, it's one that just gets automated somewhere. I'd rather we do it here so it only has to happen once instead of on every load. 

We do need to document the data/flags setup we end up having in the tod format doc, because it will be important to be able to trace everything through.

jlashner: If you want to do something like take an FFT of your timestream, it's very important that both there aren't random -1's in the data, and that the data is equisampled (so you can't just mask out the -1's). Gap-filling + interpolation is generally how we handle that

kmharrington: yes, if you want to do things like run a Fourier filter on the timestreams it's much easier to just have "mostly right" data points in there. Then you can run the filter without much concern able contaminating adject points. 

Otherwise you have to run Fourier filters accounting for un-equal sample spacing which is a HUGE efficiency hit.

algadoreo: I see. In that case, I suppose we can just copy what Jack already wrote in `fill_time_gaps` (the last block). I won’t have time to do this though. (And the unit tests will also need to be updated accordingly.)

kmharrington: lol Jack beat me to it (didn't see that it had been updated before my comment)

algadoreo:
Thanks everyone for all the feedback! I’ll open new issues for the outstanding items since they definitely need to be done, and merge this PR for now so we can start the next round of testing.

PR #414, opened 2023-02-27
Use HDF5 to store beam

skhrg:
Following discussion from slack, moved this over from #251 
I can add backwards compatibility with pickle files if we want. I do have a scrip for converting the old pickle files to hdf5 that I can put in pwg-scripts if needed.


Zhilei:
Thanks, Saianeesh! I don't think we need backward compatibility for pickle files, it will generate confusion.

Let's merge this quickly so that we can use the updated `main` branch for a new branch for the bcp work

Zhilei (approved):

PR #415, opened 2023-02-27
Hdf5 beam in more operators

skhrg:


Zhilei (approved):
Looks good! Thank you for the quick turnaround!

PR #416, opened 2023-02-27
Adding analyze_bright_ptsrc.py

Zhilei:
Re-organizing the `bcp_addition` branch and PR, adding the `analyze_bright_ptsrc.py` file to the `site_pipeline` folder.


kaiwen-zheng131:
I noticed that in the outputs generated, there is an issue with the detector indices. There are 860 detectors in each file as expected, but within those 860 rows, results for 4 detectors are duplicated (p214_A, p214_B, p215_A and p215_B) and the results for 4 detectors are missing (p267_B, p268_A, p268_B, p269_A.) Can you please take a look at this? Thanks.

skhrg:
Noticed the same thing. Something to note is that this is not just duplicated rows. The names are the same but not the fit values. See below:

```
In [45]: uniq, counts = np.unique(dat['dets:readout_id'], return_counts=True)

In [46]: dups = dat.subset(rows= np.isin(dat['dets:readout_id'], uniq[counts > 1]))

In [47]: dups
Out[47]: ResultSet<[amp,xi0,eta0,fwhm_xi,fwhm_eta,phi,tau,dets:readout_id,snr,rel_cal,abs_cal], 8 rows>

In [48]: np.unique(dups["xi0"]).shape
Out[48]: (8,)

In [49]: plt.scatter(dups["xi0"], dups["eta0"])
QStandardPaths: XDG_RUNTIME_DIR not set, defaulting to '/tmp/runtime-saianeesh'
Out[49]: <matplotlib.collections.PathCollection at 0x7f2c8953da20>

In [50]: plt.show()
```
image

Zhilei:
@kaiwen-zheng131 @skhrg Thanks for finding this! It turns out there was a bug in indexing the detectors, now it is fixed and I also regenerated the result file at the same location. Please check

kaiwen-zheng131:
I checked the output and the total pixel count went down from 860 to 859, with p429_B missing. Is that change intended?

Zhilei:
@kaiwen-zheng131 @skhrg Thanks for noticing this.
Currently, the fitting code only reports the detectors with a successful fit, which means if all the fittings are successful we get 860 entries; if we fail to fit one, we get 859 entries. I can see this causing problems. Alternatively, I can report all the detectors, regardless of the fitting outcome, while filling all the failed entries with a special value for you to flag.
BTW, you mentioned NaN as readout ids, can you give a specific example?

Zhilei:
I have implemented a fix that even though the number fields will be populated when a fitting fails, the det:readout_id field keeps.

kaiwen-zheng131:
@Zhilei Correct, p429_B is the last detector and it's consistently missing after the previous bug in indexing the detectors has been fixed.

Zhilei:
Thanks for the revision, especially adding this to cli and the docs! It is so much work!
One question is that you removed the wafer input in this commit https://github.com/simonsobs/sotodlib/pull/416/commits/62bcf43a919b73ae4109d53947650b190faed9d6
I thought we keep the `obs_id:wafer` as the input but not the `band`, now we have the `band` left but not the `wafer`. Did I understand it correctly? I can fix it if it is a typo.

Zhilei:
It just came to me that we need the `band` argument for the initial fitting values. That's probably why you kept it there. However, my question about the `wafer` parameter still stands.

mhasself:
> It just came to me that we need the `band` argument for the initial fitting values. That's probably why you kept it there. However, my question about the `wafer` parameter still stands.

Fair point about our intention to split the analysis by wafer.  When someone goes in, to change "ctx_idx" to obs_id, they can add in a means to select the wafer/detset to target.  I'm not sure if we have decided on a consistent way to do this, across modules.  In make-uncal-beam-map, it actually loops over all detsets found in the data, producing one result for each one.  It would be useful to be able to target one, optionally, though.

Zhilei:
I don't think we have agreed on a way to pass the obs_id and detset information. In addition, if we calibrate one wafer at a time, each obs_id only relates to one wafer as well.

I suggest we leave it as it is for now and merge. Eventually, we definitely need to retire the 'ctx_idx' argument and we can revisit this issue then.

PS: I have tested the updated script and it successfully ran on my side. Also, it only takes 1/3 of the time compared to before. Thanks for speeding it up!

I will merge unless I hear any objections by the end of EOB today. Thanks!

Zhilei:
@mhasself It seems like I need at least one approving review to merge. It will be great if you can approve it at your convenience. 
Once you approve it I will merge. Thanks!

mhasself (approved):
Thanks!

Issue #417, opened 2023-02-28
More columns when loading detmap


skhrg:
It would be useful for something I am working on (generating priors from detmap results) if [`add_detmap_info`](https://github.com/simonsobs/sotodlib/blob/master/sotodlib/io/g3tsmurf_utils.py#L325) could be made to load additional columns from the detmap file. In my case I think `fit_fr_mhz` is what I want but I think it would make sense to just let it take in a list of columns to load or something.

I assume @davidvng is the person to talk to about this?

davidvng:
@skhrg The other method of making detmaps, through databases and context, uses more columns (you can see a list [here](https://github.com/simonsobs/sotodlib/blob/master/sotodlib/site_pipeline/make_det_info_wafer.py#L99)), though it doesn't currently include fit_fr_mhz. I can add this column and help you get this set up if you'd like. If you prefer to keep using `add_detmap_info` then I can work on a change there too.

skhrg:
Whoops got sidetracked and forgot to respond to this.
Is it going to be more correct in the long term for me to use the output of `make_det_info_wafer` (I assume yes?)

Also thanks for the PR! That should be enough for me to test stuff before I worry about getting things setup properly.

davidvng:
Since the output of `make_det_info_wafer` and `make_read_det_match` will create csv files anyway, I think your method will be ok.

I also added option to set columns='all'. I just merged #418 so you should be good to go! Closing

skhrg:
Thanks for the quick turnaround on this!

PR #418, opened 2023-03-03
Added optional columns to add_detmap_info

davidvng:
Allows users to include optional columns that are part of the entire detmap csv output to `add_detmap_info`
This addresses #417.


kmharrington (approved):
approving because I like this idea/setup but one more request. Could we have an option where if `columns='all'` then it just loads everything?

PR #419, opened 2023-03-06
Adds smurf data to oper books

### LINKED ISSUE #361: Implement Operation book differences
The [TOD Format Doc](https://github.com/simonsobs/tod2maps_docs) has a couple differences between the Observation books and the Operation books. The one I know we're still missing is the "add the sodetlib metadata" into the book / index card information.

jlashner:
This PR adds smurf ancillary data to `oper` books. Currently this only copies files that have a `npy` suffix, which I believe captures everything we want to add, but we can easily change this if it doesn't. Files are currently copied to `book_output/smurf/<basename>`, but this can change. I was considering `book_folder/smurf/<stream_id>/<basename>`, but I don't think it matters since there will only be one stream-id per oper book.

A few other things this includes
- I added a test_mode kwarg to the `bind_book` function, since it is _really_ useful to be able to run that function to write an existing book to my local directory where I can check it. In test_mode, none of the changes to `book` will be committed, and it won't complain about the book already being bound.
- Removes the hwp binding option. This will definitely need to be merged in for the bookbinder to work properly, so if this PR takes too long we may want to separate it out into a separate PR.

Resolves #361

### DIFF
--- a/sotodlib/io/imprinter.py
+++ b/sotodlib/io/imprinter.py
@@ -269,7 +269,8 @@ class Imprinter:
             self.register_book(session, bid, obs_list, commit=False)
         if commit: session.commit()
 
-    def bind_book(self, book, session=None, output_root="out", message=""):
+    def bind_book(self, book, session=None, output_root="out", message="",
+                  test_mode=False):
         """Bind book using bookbinder
 
         Parameters
@@ -281,6 +282,10 @@ class Imprinter:
             output root directory
         message: string
             message to be added to the book
+        test_mode : bool
+            If in test_mode, this function will still run on already copied
+            books, and will not update any db fields in the db. This is useful
+            for testing purposes.
         """
         if session is None: session = self.get_session()
         # get book id and book object, depending on whether book id is given or not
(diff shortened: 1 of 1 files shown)


kmharrington:
My only questions/thoughts are in file formatting / placement.

1.  I don't think we need a smurf directory in the books, I think these books are so small the files can still be top level
2. `basename` has a ctime in it right? Could we strip that off for these? I can imagine many places where it will be nice to just be about to do 

```
for book in list_of_books:
      do_something_interesting( book_dir/file_pattern.npy )
```

jlashner:
ya, we can move the the files to the base directory, that's fine with me.

For the `basename` I kind of disagree with stripping the timestamp from these. In the sodetlib data files, we reference the current tune and bgmap, and if we change the basename it will be impossible to find the referenced file without a caldb. As it is now, we can do things like write a quick search function to find the referenced meta file, or check by eye that we are looking at the correct tunefile.

As an alternative, in the tod2maps  doc it says that we should add the path of the main metadata file of the operation to the `M_index` file under the key `meta_file`.   With this you could easily loop through books and do something with the metadata, as you mention above. Would it be ok with you if we keep the basename, but add the path to the M_index file?

kmharrington:
Right! I forgot I wrote that. The `meta_file` key also works for the easy for loops so let's implement that one.

jlashner:
Alright, just submitted a commit that should address feedback. Metadata is copied to the base directory, and we add the field `meta_files` to the M_index file, which looks like this:
```
book_id: oper_1673285630_sat1_1000000
end_time: 1673285657.343472
meta_files:
  bgmap: 1673285663_bg_map.npy
  bias_steps: 1673285671_bias_step_analysis.npy
n_frames: 20
n_samples: 116665
session_id: 1673285630
start_time: 1673285628.177208
```

kmharrington (approved):
looks good to me. thanks!

PR #421, opened 2023-03-09
Add UFM mapping

kmharrington:
Adding the ability to use actual UFM hardware information for the readout sims. The UFM rotations relative to each other look correct. Note: this DOES NOT work for LF detectors yet, the mappings between MF and UHF are similar enough that I can use existing MF mappings to spoof the UHF. The LF detector positions and setup will be totally different. There will be straightforward ways to build off this when we know which UFMs are installed in which wafer slots.

Example use for creating a hardware setup: 
```
lat_hw = sim_nominal()
core.hardware.sim_wafer_names(lat_hw)
sim_telescope_detectors(
    lat_hw, 
    "LAT", 
    tube_slots=["i1"],
    det_info=("/home/kmharrin/scratch/20230216_toast_scratch/det_info.h5","Cv4"),
)
```

### DIFF
--- a/sotodlib/core/hardware.py
+++ b/sotodlib/core/hardware.py
@@ -73,6 +73,28 @@ def parse_readout_id(readout_id):
     channel = smband * 8 + smchan
     return (wf, ct, channel)
 
+def sim_wafer_names( hw ):
+    """Adds SO generic UFM names to the hardware model based on the type of the wafer
+       Ex: Uv1, Mv4, Lv3, etc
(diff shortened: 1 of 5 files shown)


kmharrington:
@keskitalo  I think this is good enough to at least merge into the readout PR.

keskitalo (approved):
Looks great and certainly safe to merge with another PR. Sorry for the delay! Somehow slipped through the cracks.

Issue #420, opened 2023-03-09
bug in imprinter.get_start_and_end


guanyilun:
Here's how it looks for the ending timestamp:
```python
# Get end time
frame = None
for _frame in core.G3File(_files[-1]):
    if _frame.type == core.G3FrameType.Scan:
        frame = _frame
t1 = get_frame_times(frame)[1][-1]
```
It has an unhandled failure mode that If the last g3 file in an observation has no scan frame, frame will be None and it will error out. To reproduce this error:

```python
from sotodlib.io.imprinter import Imprinter

im = Imprinter(im_config="/mnt/so1/shared/site-pipeline/data_pkg/imprinter.yaml")

book = im.get_book("obs_1677794454_sat1_1111101")
im.bind_book(book)
```

PR #425, opened 2023-03-10
Fix get_start_and_end frame loop

jlashner:
This makes the start/end determination more robust to instances where the last file has no scan frames, as Yilun pointed out in #420.


guanyilun (approved):
looking good to me, thanks for the fix!

Issue #404, opened 2023-02-09
HKUtils: Enhance hkutils to output axis manager of cosampled hk + detector data


sanahabhimani:
It'll be useful to add a component to the hkutils that takes in the start and stop of detector data to find the hk devices that were online at that same time, and outputs an axis manager of axis managers for all detector + hk data with interpolated timestamps. This will allow for immediate analysis of detector and calibrator/housekeeping data, automating the interpolation that will happen _all the time_

PR #426, opened 2023-03-16
Make HK Amans Cosampled with Detector Timestreams

sanahabhimani:
Adding an extra function to `hkutils` that takes in an hkconfig file, a detconfig file, and an obs.id to output an HK axis manager that's cosampled with detector time streams.

Incorporates #404 

Things to note:

- could expand it to make it more general outside of detector timestamps but may need a bit more thinking
- assumes the start and stop times of an `obs.id` is all the start and stop time you need
- probably a better way to do this: right now, it takes in the detector timestamps and adds it as an axis to each individual HK axis manager. Likely, we want just one timestamps axis and need the HK axis manager's to have only interpolated data.
- it'll be good for me to add documentation to the readthedocs on using `hkutils`

### DIFF
--- a/sotodlib/io/hk_utils.py
+++ b/sotodlib/io/hk_utils.py
@@ -249,4 +249,33 @@ def get_hkaman(start, stop, config):
     data = get_grouped_hkdata(start, stop, config)
     hk_amans = make_hkaman(data)
     return hk_amans
+
+def get_detcosamp_hkaman(config, det_aman):
+    """
+    Wrapper to combine get_grouped_hkdata() and make_hkaman() to output one
+    axismanager of HK axismanagers that are cosampled to detector timestreams.
+
+    Parameters:
+        config (str): Filename of a .yaml file for loading HK fields and aliases.
+            Required for get_grouped_hkdata().
+        det_aman (AxisManager): detector data AxisManager from load_smurf().
+
+    Notes
+    -----
+
+    An example config file looks like:
+
+        data_dir: '/mnt/so1/data/chicago-latrt/hk/'
+
+        field_list:
+            'bf_4k' : 'observatory.LSA22HG.feeds.temperatures.Channel_06_T'
+            'xy_stage_x': 'observatory.XYWing.feeds.positions.x'
+            'xy_stage_y': 'observatory.XYWing.feeds.positions.y'
+    """
+    start = det_aman.timestamps[0]
+    stop = det_aman.timestamps[1]
+    data = get_grouped_hkdata(start, stop, config)
+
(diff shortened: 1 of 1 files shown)


sotodlib/io/hk_utils.py
```diff
@@ -206,4 +210,78 @@ def get_hkaman(start, stop, config):
...
+
+        data_prefix : "/mnt/so1/data/chicago-latrt"
+        hwp_prefix : "/mnt/so1/shared/site-pipeline/data_pkg/chicago-latrt/hwp/"
+        g3tsmurf_db: "/mnt/so1/shared/site-pipeline/data_pkg/chicago-latrt/g3tsmurf.db"
+    """
+    # load detector timestamps
+    SMURF = ls.G3tSmurf.from_configs(detconfig)
+    session = SMURF.Session()
+    obs = session.query(Observations).filter(Observations.obs_id == obs_id).one()
+    det_aman = ls.load_file([f.name for f in obs.files], archive=SMURF)
```
kmharrington, line 239: This be much more useful if you allow the user to load their own AxisManager first, there are way too many ways people want to customize what data they have loaded for you to do it here. I think you just need to take in `aman` and an hkconfig. You can use `aman.timestamps[0]` and `aman.timestamps[-1]` to get the time ranges.

sotodlib/io/hk_utils.py
```diff
@@ -206,4 +210,78 @@ def get_hkaman(start, stop, config):
...
+
+    grouped_data = get_grouped_hkdata(obs.start, obs.stop, hkconfig)
+
+    amans = []
+    device_names = []
+    for group in grouped_data:
```
kmharrington, line 245: This looks like you've copied a lot of code from other functions, ( I think `make_hkaman` ). Straight copies of code in different places are hard to maintain because 1 function gets updated and another doesn't. Can you come up with a way to have one function call the other? Or both functions call a helper?

kmharrington (changes requested):
In line comments!

sotodlib/io/hk_utils.py
sanahabhimani: I ended up expanding `make_hkaman()` to be able to handle the 2 cases: (1) output an aman for HK data, organizing by cosampled device where appropriate, and (2) given a detector aman, output an HK aman where the HK data is cosampled to detector timestamps.

kmharrington (approved):
I like it (just tested it on some old data). I do think a sphinx docs entry to it will be useful though, with some usage examples there and/or in pwg-tutorials. So that's on the future todo list.

PR #430, opened 2023-03-17
so-site-pipeline command line interface

mhasself:
This adds an entry point called so-site-pipeline, the purpose of which is to more easily access scripts in sotodlib.site_pipeline.

Only a couple of elements have been "registered" for now; a certain interface in the submodule is expected.

@guanyilun Since this requires formalizing entry points to site pipeline scripts, it is worth considering if it is also a good time to build in easier access from Prefect.  I forget what the exact requirements are there.

### DIFF
--- a/docs/site_pipeline.rst
+++ b/docs/site_pipeline.rst
@@ -13,6 +13,14 @@ quick-turnaround data processing at the observatory.
   here will be on operation; command line parameters and config file
   syntax.
 
+
+Command line interface
+======================
+
+
+Usage
+-----
+
 To execute a pipeline element from the command line, use the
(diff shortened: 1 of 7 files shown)


guanyilun:
prefect doesn't have a preference for command-line interface -- these commands will get composed through string formatting anyway. A functional interface is more natural to prefect since prefect interface is just decorators that wrap around them. I think the `update_g3tsmurf_database` script is a good example for this concept. If we could have more script in this form, it will make it easier for interfacing with prefect.

kmharrington (approved):
Sorry it took a minute to get to this. I just tested this by adding in the preprocess-tod script in that PR. Looks good and was pretty easy to get added!

PR #431, opened 2023-03-17
Fix 0-length buffer bug in Bookbinder

### LINKED ISSUE #422: Bookbinder 0-length buffer bug
If the buffer length is exactly maxlength, the buffer gets cleared and the Bookbinder gets stuck in a state where there are no buffered samples, while the frame split time (`event_time`) is not advanced because the condition to exit the while loop is not satisfied (`ready == False` since the length is 0):

https://github.com/simonsobs/sotodlib/blob/a4f4c68bcca22c4623ff449963c2d7a6a7bf03b2/sotodlib/io/bookbinder.py#L1063-L1065

Although a new frame is added by continuing in the loop, the frame split time is still the previous one. It then tries to return samples before the (old) split time, finds none, and throws the error.

algadoreo:
If the buffer is exactly maxlength, the Bookbinder clears the buffer without advancing the frame split time and gets stuck. This fix changes the `>= maxlength` to a strict `>`. The split time written is then changed to be the first sample of the next frame, consistent with the convention for HK-determined splits.

Fixes #422

### DIFF
--- a/sotodlib/io/bookbinder.py
+++ b/sotodlib/io/bookbinder.py
@@ -601,7 +601,7 @@ class FrameProcessor(object):
         self.smbundle.prepend(fx)
         # Update (local) flush time if necessary
         if data_excess is not None:
-            flush_time = f['signal'].times[-1] + 1
+            flush_time = data_excess.times[0]
             self._frame_splits[-1] = flush_time  # update the most recently added split time
         # Update last sample
         self._prev_smurf_sample = f['signal'].times[-1]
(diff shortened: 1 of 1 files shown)


guanyilun (approved):
This looks good to me! While your knowledge is still fresh, could you add a few more lines of comments to explain on high level what the code block from L717 to L725 is doing?

algadoreo:
Okay—let me know if these new comments are sufficient or not.

PR #432, opened 2023-03-17
Book metadata improvements

guanyilun:
Fixed issues (1,3,5) with metadata found in #388.

An example M_index.yaml file from this PR:

```yaml
book_id: obs_1677794454_sat1_1111101
book_type: obs
end_time: 1677795650.3189769
n_frames: 38
n_samples: 239315
sample_ranges:
- - 0
  - 239315
session_id: 1677794454
start_time: 1677794453.7489772
stream_ids:
- ufm_mv14
- ufm_mv18
- ufm_mv19
- ufm_mv22
- ufm_mv6
- ufm_mv9
telescope: sat1
```
An example M_book.yaml from this PR:
```yaml
book:
  book_id: obs_1677794454_sat1_1111101
  finalized_at: '2023-03-18T17:14:16.597195'
  schema_version: 0
  type: obs
bookbinder:
  codebase: /home/yguan/software/sotodlib/sotodlib/__init__.py
  context: simons1-data-pkg-dev
  version: 0.4.0+1175.gf37013d.dirty
```

Also fixed an off-by-1 bug that came up during my local testing, and propagated the corresponding changes to the unit tests.

Edit: schema_version v0 -> 0

### DIFF
--- a/sotodlib/io/bookbinder.py
+++ b/sotodlib/io/bookbinder.py
@@ -513,7 +513,7 @@ class FrameProcessor(object):
             prev_frame_end = self._prev_smurf_sample.time
         # Define a strict upper limit (exclusive) up until which to look/interpolate for missing samples
         if self.BOOK_END_TIME is not None:
-            end_time = min(flush_time.time, self.BOOK_END_TIME.time)
+            end_time = min(flush_time.time, self.BOOK_END_TIME.time + 1)
         elif self._smurf_timestamps is not None:
             end_time = min(flush_time.time, self._smurf_timestamps[-1] + 1)
         else:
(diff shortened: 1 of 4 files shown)


kmharrington (changes requested):
Awesome, that's for setting this up! Can I ask for another set of info (which I realize now isn't documented well, sorry about that). 

Each lvl2 observation will have a TuneSet that we'll want to track as well, since those are basically our "detsets" at level 3. So, per-stream id. they should be accessible in `obs.tunesets[0]`. And, assuming readout_ids are found, that should always have an entry.

kmharrington:
Oh, looking at the index file, I have another one we didn't add to the documentation. tube_slot. Each book will be 1-to-1 with tubes in the LAT/SAT and that will make matching the hardware model easier.

kmharrington (approved):
looks good, thanks for the updates

PR #438, opened 2023-03-30
Bookbinder Refactor

jlashner:
This is not complete yet, but I wanted to start a draft for this PR so that we can discuss on the call today and so people could start testing if they're interested. To run it, you can do something like:
```
config = '/mnt/so1/shared/site-pipeline/data_pkg/imprinter.yaml'
imprinter = Imprinter(config, db_args={'connect_args': {'check_same_thread': False}})
bid = 'obs_1677877436_sat1_1111101'
book = imprinter.get_book(bid)
filedb = imprinter.get_files_for_book(book)
obsdb = imprinter.get_g3tsmurf_obs_for_book(book)
bookbinder = bookbinder2.BookBinder(book, obsdb, filedb, [])
bookbinder.bind('test')
```
and it will bind the book to the directory 'test'.

A few things that have not yet been implemented...
- Writing real ACU data to books. Right now ancillary data is just an empty G3TimestreamMap with the correct timestamps.
- Determining frame splits using ACU data. Right now it is just done splitting on `max_samps_per_frame`
- Option to change readout channel names
- Interpolation of unfilled samples

However those aside this seems to be working pretty well... There's at least one rare failure mode I've found that I'm trying to track down. Of course we also need to check the data in the output books matches the data in the L2 files, and that this matches the data in the output books generated by the old bookbinder.

The big remaining item is to properly handle ACU data, which I'm hoping we can do by pretty much directly copying Jason's functions `locate_crossing_events` and `find_frame_splits` with minor tweaks.

Right now this is all saved in the `bookbinder2.py` file since I don't want to remove the current bookbinder so we can compare results.


jlashner:
Hi all, I think this is pretty much finished and ready for review. I realize this is a pretty large PR and difficult to review, so some people may want to just check out the data files. It's currently binding end-to-end data from 2 weeks ago to `/home/jlashner/scratch/2023/e2e_data/out/books`, so it would be good to check those out / run some sort of validation. So far though it seems to be performing very well. Let me know if there are any questions / suggestions!

sotodlib/io/bookbinder2.py
```diff
@@ -0,0 +1,931 @@
...
+        """
+        self.log.info("Binding ancilary data")
+
+        # Handle file writers
+        writer = None
+        cur_file_idx = None
+        out_files = []
+
+        az, el, boresight, corotation = None, None, None, None
+        if 'Corrected_Azimuth' in self.data:
```
kmharrington, line 176: what are these corrected fields?

sotodlib/io/bookbinder2.py
```diff
@@ -0,0 +1,931 @@
...
+        self.nframes = 0
+        ts = []
+        frame_idxs = []
+        frame_idx = 0
+        for frame in get_frame_iter(self.files):
+            if frame.type != core.G3FrameType.Scan:
+                continue
+
+            if self.readout_ids is None:
+                self.readout_ids = frame['data'].names
```
kmharrington, line 262: unless I'm reading this wrong, we don't want this fall back. If we can't find readout id's through g3tsmurf we want to throw errors

sotodlib/io/bookbinder2.py
```diff
@@ -0,0 +1,931 @@
+import so3g
```
kmharrington, line 1: I'm a bit hesitant to start adding in "bookbinder2." It means we'll have some reason to start having 3, 4, 5, etc in the future. I know this is a massive change to the codebase and we want to acknowledge the previous work, particularly by @algadoreo. Especially since we wouldn't have this version without having gone through making the first version and figuring out all the hard parts and interfaces. But this is all version controlled so we will still have the history.

jlashner: haha yes totally. Its convenient right now to have both this and `bookbinder.py` in the codebase for the sake of comparison, but I do plan on renaming this to `bookbinder.py` and replacing the old version before merge.

jlashner:
> 1. why do we need an option to change readout_ids?
I'm fine with making this a required parameterf

> 2. Does the ACU emulator make actual scan patterns? Is that something we can test with E2E testing?
E2E testing uses the simulator to create realistic scan patterns. Here is an example of the azimuth data for a book and the frame dividisions:
image

guanyilun:
Thanks for putting this together! I haven't gone through the codes but a high level comment is that I think compatibility changes in the imprinter interface should probably be part of this PR too to avoid breaking deployed prefect flows.

sotodlib/io/bookbinder2.py
algadoreo: Lauren’s explanation was:
>Within the `ACU_position` block, all fields labeled `Raw` are raw values from the encoders, and all fields labeled `Corrected` are encoder values that may have a correction factor applied to them from the ACU pointing model.

She also told me to use the corrected values.

mhasself: Although the ACU supports a rich TPOINT+ model for Raw -> Corrected, my expectation is that we will dial in, at most, only the main "encoder offsets", determined early on in commissioning.

If somehow Corrected becomes weird, we are recording the Raw & Corrected in the HK so we'll be able to figure out what happened and undo it in post.

I am actually quite open to the idea that we just use Raw fields everywhere... it sort of depends on how bad the Raw pointing is.  If az/el encoder offsets are large, but that's all we need, then there will be some pressure to set it and forget it.  But if the pointing is super weird, and encoder offsets only partly help with that, then we might as well not bother setting them (in which case Raw = Encoder).

The reasons I am against enabling the ACU's full pointing model: (a) some parameters will cause mixing of az and el, which we don't want to do while scanning (a "constant el" scan should be one where the elevation motor does not move); (b) we care enough about pointing that we'll need to measure and refine our physical pointing model very carefully over time, and apply it in a way that properly handles parallactic rotation, and if the ACU is doing non-trivial corrections we'll just have to undo those before applying our physical model to characterize the system.

algadoreo: Which is the more appropriate one for use in determining telescope states/turnarounds? I’m a little concerned about the “mixing of az and el”: the Bookbinder currently tags any v_el > 10^-4 deg/sec as a slew instead of a scan. (Bookbinder has access to both Raw and Corrected so it doesn’t matter too much which is used vs which is written, but we should decide how it’ll be done.)

sotodlib/io/bookbinder2.py
```diff
@@ -0,0 +1,1022 @@
...
+            t = get_frame_times(frame)[1]
+            ts.append(t)
+            frame_idxs.append(np.full(len(t), frame_idx, dtype=np.int))
+
+            self.nframes += 1
+            frame_idx += 1
+
+        self.times = np.hstack(ts)
+        self.frame_idxs = np.hstack(frame_idxs)
+        self.preprocessed = True
```
guanyilun, line 283: not used

sotodlib/io/bookbinder2.py
```diff
@@ -0,0 +1,1022 @@
...
+
+    def bind(self, outdir, times, frame_idxs, file_idxs, pbar=False, ancil=None,
+             atol=1e-4):
```
guanyilun, line 286: atol is not used: hardcoded inside this function

sotodlib/io/bookbinder2.py
```diff
@@ -0,0 +1,1022 @@
...
+                m = (oframe_idxs == oframe_idx) & (iframe_idxs == iframe_idx)
+                outsamps = out_offset_idxs[m]
+                insamps = in_offset_idxs[m]
+
+                # A fast way to copy data into a numpy array is to use direct
+                # slicing like ``arr_out[o0:o1] = arr_in[i0:i1]`` since it doesn't
+                # need to create a temporary copy of the array, and can just do a
+                # direct mem-map. Doing this speeds up the binding by a factor of
+                # 3-4x compared to other methods I've tried.
+                split_idxs = 1 + np.where(
```
guanyilun, line 385: a comment here about what split_idxs is will be helpful for understanding the subsequent code

sotodlib/io/bookbinder2.py
```diff
@@ -0,0 +1,1022 @@
...
+    def bind(self, pbar=False):
+        """
+        Binds data.
+
+        Params
+        ---------
+        outdir : str
+            Output directory to write bound files. If this does not exist, the
+            directory will be created.
+        """
```
guanyilun, line 661: docstr doesn't match function signature

sotodlib/io/bookbinder2.py
```diff
@@ -0,0 +1,1022 @@
...
+        elif pbar is False:
+            pbar = tqdm(total=self.nframes, disable=True)
+
+        pbar.set_description(f"Binding {self.stream_id}")
+
+        self.log.info(f"Binding smurf obsid {self.obs_id}")
+
+        # Mapping to input sample idx to output sample idx
+        atol = 1e-4
+        sample_map = find_ref_idxs(times, self.times)
```
guanyilun, line 322: a comment indicating times has no gap, self.times have gaps will be helpful for reading

sotodlib/io/bookbinder2.py
```diff
@@ -0,0 +1,1022 @@
...
+    Parameters
+    -------------
+    ts : np.ndarray
+        List of timestamps of length `n`, potentially with gaps
+    
+    Returns
+    --------
+    new_ts : np.ndarray
+        New list of timestamps of length >= n, with gaps filled.
+    """
```
guanyilun, line 702: missing the second return value in docstr

sotodlib/io/bookbinder2.py
```diff
@@ -0,0 +1,1022 @@
...
+    # Find indices where gaps occur and how long each gap is
+    dts = np.diff(ts)
+    dt = np.median(dts)
+    missing = np.round(dts/dt - 1).astype(int)
+    total_missing = int(np.sum(missing))
+
+    # Create new  array with the correct number of samples
```
guanyilun, line 709: extra space

sotodlib/io/bookbinder2.py
```diff
@@ -0,0 +1,1022 @@
...
+    atol : float
+        Absolute tolerance for the difference between the reference
+
+    Returns
+    -------
+    i_missing : array_like
+        indices (in refs) of the missing samples in vs
+    t_missing : array_like
+        values (in refs) of the missing samples in vs
+    """
```
guanyilun, line 803: docstr inconsistent: atol is no longer used, t_missing is not returned

guanyilun (changes requested):
I've gone through it and think it's well written! I left a few minor comments for readability. Another comment that doesn't need to be addressed in this PR is that I find `SmurfStreamProcessor.bind" to be quite complex in its logic. It may be good to factor out some of the logic into some "pure functions" that is independent of IO so we could easily develop comprehensive set of unit tests for those.

jlashner:
Hi Yilun, thanks for reading through and providing feedback!! I think I hit all of your inline comments in the most recent commit.

I do agree that the logic in `SmurfStreamProcessor.bind` is complex... however I can't really think of a way to split the logic into separate functions in an intuitive way that doesn't sacrifice speed. I think because the logic relies on so many arguments, (`iframe`, `data`, `primary`, `biases`, `filled`,  frame and offset mappings, etc.) passing all of these to an external function that's io independent seems kind of contrived to me.

Instead I propose just writing a test for the full SmurfStreamProcessor functionality, by writing a fake data file with gaps, preprocessing + binding it, reading in the bound data and comparing it against the input data. I can write that up and add it to this PR, because I do think having that sort of test is really important here.

guanyilun:
> Instead I propose just writing a test for the full SmurfStreamProcessor functionality, by writing a fake data file with gaps, preprocessing + binding it, reading in the bound data and comparing it against the input data. I can write that up and add it to this PR, because I do think having that sort of test is really important here.

Thanks! Yes this sounds like a fair solution to me.

jlashner:
Hi All,

With these last few commits I implemented some of the changes we discussed on the call last week. Namely
 - For each frame I'm pulling `azimuth_mode` from the ACU summary output. According to Matthew, ProgramTrack corresponds to Scanning, and "Preset" corresponds to slewing. I'm not sure this is clearly documented... but don't really think the bookbinder is the right place to document it. I think maybe adding a description to the ACU agent docs is probably the right place. Anyhow, if there is no az-data, I set the `azimuth_mode=None`, if there are any ProgramTrack samples, I set `azimuth_mode=ProgramTrack`, and otherwise I set to `Preset`.
 - I add `hwp_freq_mean/stdev` and `az_speed_mean/stdev` to the book metadata, where the mean/stdev are computed over the entire book. 

Let me know if there are any remaining comments! If not I believe we can start thinking about merging.
Also, tests are failing because they require `socs` to generate data. Any chance we can build that into the test docker? Or should we disable these select tests in the github workflow?

sotodlib/io/bookbinder2.py
```diff
@@ -0,0 +1,1118 @@
...
+                self.data[k] = []
+            self.data[k].append(v)
+    
+    def finalize(self):
+        if self.times:
+            self.times = np.hstack(self.times)
+            for k, v in self.data.items():
+                self.data[k] = np.hstack(v)
+        else:
+            self.times = np.array([], dtype=np.float32)
```
algadoreo, line 100: Is float32 consistent with the times defined elsewhere in `finalize` and `process_frame`—i.e., in other frames (I believe numpy defaults to float64)—or does this not matter?

sotodlib/io/bookbinder2.py
```diff
@@ -0,0 +1,1118 @@
...
+    Processor for ancillary (ACU) data
+    
+    Params
+    --------
+    files : list
+        List of HK files to process
+
+    Attributes
+    -----------
+    data : dict
```
algadoreo, line 115: Should this have `data` --> `blocks`? (same in docstring for `preprocess`)

sotodlib/io/bookbinder2.py
```diff
@@ -0,0 +1,1118 @@
...
+        for oframe_idx in np.unique(frame_idxs):
+            if file_idxs[oframe_idx] != cur_file_idx:
+                close_writer(writer)
+                cur_file_idx = file_idxs[oframe_idx]
+                fname = os.path.join(outdir, f'A_ancil_{cur_file_idx:0>3}.g3')
+                out_files.append(fname)
+                writer = core.G3Writer(fname)
+
+            m = frame_idxs == oframe_idx
+            ts = times[m]
```
algadoreo, line 202: Will we have, as in the original Bookbinder, a flag indicating which timestamps (in ancil) were interpolated? I imagine that would be passed in along with `times` and masked just like here, then propagated into `anc_data` below.

sotodlib/io/bookbinder2.py
```diff
@@ -0,0 +1,1118 @@
...
+        t0 : float
+            Start time of frame (unix time)
+        t1 : float
+            Stop time of frame (unix time)
+        """
+        bl = self.blocks.get('ACU_summary_output')
+        if bl.data:
+            az_mode = bl.data["Azimuth_mode"]
+            acu_summary_times = bl.times
+            m = (t0 < acu_summary_times) & (acu_summary_times <= t1)
```
algadoreo, line 255: What’s the reason for not including `t0`? Aren’t `t0` and `t1` both timestamps included within the frame? (same with L267 below)

sotodlib/io/bookbinder2.py
```diff
@@ -0,0 +1,1118 @@
...
+            if not np.any(m):
+                frame['azimuth_mode'] = 'None'
```
algadoreo, line 257: Is this equivalent to a stopped state?

sotodlib/io/bookbinder2.py
```diff
@@ -0,0 +1,1118 @@
...
+        self.book_id = book_id
+
+        if log is None:
+            self.log = logging.getLogger('bookbinder')
+        else:
+            self.log = log
+
+    def preprocess(self):
+        """
+        Extracts file times, nchans, and nnframes from file list
```
algadoreo, line 304: Too manny nns in `nnframes`

sotodlib/io/bookbinder2.py
```diff
@@ -0,0 +1,1118 @@
...
+
+        self.nframes = 0
+        ts = []
+        frame_idxs = []
+        frame_idx = 0
+        for frame in get_frame_iter(self.files):
+            if frame.type != core.G3FrameType.Scan:
+                continue
+
+
```
algadoreo, line 319: Is the extra empty line meant to be a comment? (Can we add one here, something like `# Populate attributes`?)

sotodlib/io/bookbinder2.py
```diff
@@ -0,0 +1,1118 @@
...
+            oframe = core.G3Frame(core.G3FrameType.Scan)
+
+            if ancil is not None:
+                t0, t1 = ts[0], ts[-1]
+                oframe['ancil'] = ancil.anc_frame_data[oframe_idx]
+                ancil.add_acu_summary_info(oframe, t0, t1)
+
+            oframe['book_id'] = self.book_id 
+            oframe['sample_range'] = core.G3VectorInt([int(i0), int(i1+1)])
+            oframe['flag_smurfgaps'] = core.G3VectorBool(filled)
```
algadoreo, line 505: Are we flagging interpolated samples (`~filled`) or input samples (`filled`)? (Or did I get these mixed up?)

sotodlib/io/bookbinder2.py
```diff
@@ -0,0 +1,1118 @@
...
+        if self.times is not None:
+            return
+
+        for stream in self.streams.values():
+            stream.preprocess()
+
+        t0 = np.max([s.times[0] for s in self.streams.values()])
+        t1 = np.min([s.times[-1] for s in self.streams.values()])
+        ts, _ = fill_time_gaps(stream.times)
+        m = (t0 < ts) & (ts < t1)
```
algadoreo, line 616: Is there a reason for excluding `t0` and `t1`? They’re samples extracted from `stream.times`, right?

sotodlib/io/bookbinder2.py
```diff
@@ -0,0 +1,1118 @@
...
+
+        self.ancil.preprocess()
+
+        # Divide up frames
+        frame_splits = find_frame_splits(self.ancil)
+        if frame_splits is None:
+            frame_idxs = np.arange(len(ts)) // self.max_samps_per_frame
+        else:
+            frame_idxs = np.digitize(ts, frame_splits)
+            frame_idxs -= frame_idxs[0]
```
algadoreo, line 627: Does this handle frames whose length exceeds `self.max_samps_per_frame`?

sotodlib/io/bookbinder2.py
```diff
@@ -0,0 +1,1118 @@
...
+            corotation = np.interp(times, block.times, block.data['Corrected_Corotation'])
+
+        anc_frame_data = []
+        for oframe_idx in np.unique(frame_idxs):
+            if file_idxs[oframe_idx] != cur_file_idx:
+                close_writer(writer)
+                cur_file_idx = file_idxs[oframe_idx]
+                fname = os.path.join(outdir, f'A_ancil_{cur_file_idx:0>3}.g3')
+                out_files.append(fname)
+                writer = core.G3Writer(fname)
```
algadoreo, line 199: Suggest adding a comment above this block, something like `# Update values if starting a new output file`

sotodlib/io/bookbinder2.py
```diff
@@ -0,0 +1,1118 @@
...
+            f'{obs.action_ctime}_{obs.action_name}'
+        )
+
+        if not os.path.exists(action_dir):
+            continue
+
+        for root, _, fs in os.walk(action_dir):
+            files.extend([os.path.join(root, f) for f in fs])
+
+    return [f for f in files if copy_to_book(f)]
```
algadoreo, line 1118: I like having newlines at the end of files but I suppose it isn’t strictly necessary. (Also applies to all the other files)

algadoreo (commented):
Sorry I took so long to review this! I really like the new structure of looping over output frames rather than input frames, and the logic is easy to follow (instead of the hacked mess of the original).

My one concern is with the precision of using seconds (floats) vs. G3Time units (ints). This is crucial for any `<` and `>` evaluations with timestamps and I was having trouble with getting samples in the right frames in the past (because `t == t-1` is `True` when the number is ~O(1e17)), so we should ensure this isn’t causing problems. Other than that, I’ve put minor comments/questions at the relevant lines below.

Thanks for your hard work putting this together—it looks really good!

sotodlib/io/bookbinder2.py
jlashner: Are you asking if we will have a flag-mask for which `bookbinder.times` values are based on smurf data and which are interpolated? I don't think we really need this since the drop rate is very small, especially if we use timestamps from multiple slots to fill in gaps. @kmharrington do you have an opinion on whether we need to include a flag for this?

sotodlib/io/bookbinder2.py
jlashner: I think it's good to keep a distinction between having no `az` data, and being stopped.

sotodlib/io/bookbinder2.py
jlashner: ah yep, good catch

sotodlib/io/bookbinder2.py
jlashner: Yep, this can be inclusive

sotodlib/io/bookbinder2.py
kmharrington: tbh I'm usually in favor of just "save all the masks" just in case. But IDK what level of complication this adds here. 

Is there ever a point where we interpolate the timestamps here and don't also interpolate the signal data? I.e. if the timesample mask would always be 100% overlapped by the signal mask then I'm not worried about it.

sotodlib/io/bookbinder2.py
jlashner: No... I guess we should do that here, though makes the logic a bit more complicated.

sotodlib/io/bookbinder2.py
algadoreo: Yeah—I’m referring to the samples that are missing in the SMuRF file(s) used to make the reference timestamp list. I believe it’s equivalent to the intersect of the `flag_smurfgaps` from the D-files for those wafers (used to make the list), so we wouldn’t be losing info by not including it (I don’t think). It’s more for convenience than anything—in case someone were to only load the A-files.

sotodlib/io/bookbinder2.py
algadoreo: Ah whoops I misread the if condition. What’s marker for a stopped state? when both `azimuth_velocity_mean` and `elevation_velocity_mean` are within atol?

jlashner: There are more detailed az-states such as "Stop" / more in the HK books. I decided not to include them in this summary since based on our conversation it seemed like we really only cared about "Scanning" vs "Not-Scanning" (or slewing). Unless there's a reason to go into more depth here, I don't think we really need to, since that involves making decisions about what the 'summary' should be when there's a combination of different modes within a frame.

As it is now, the azimuth_mode summary is:
- 'None' --> No az-mode data in the frame
- 'Program Track' --> There exists at least one sample where it's scanning
- 'Preset' --> Not scanning

jlashner:
Hi Jason, thanks for going through the code! I think you caught a bunch of things that may have caused issues in the future, so I really appreciate it. I believe I addressed all the comments you made, and I changed the book-timestamp determination to take into account all streams instead of just one.

I am not personally concerned with the G3Time precision. The way the refactor works the bookbinder doesn't really attempt any direct comparisons of timestamps, with the exception of `find_ref_idxs`, which I don't believe requires sub-microsecond precision, but if there is a particular line of code you're concerned about, definitely let me know!

I am currently trying to work out the tests so they don't fail, but think this is ready for a final consideration so if you or @guanyilun could re-review that would be great.

algadoreo (approved):
What’s the plan for merging `bookbinder2` (and the test)? Otherwise, everything looks good to me so I’ll approve it. The index-based approach changed the time precision issue from a code-breaking bug to a minor binning bug (if it still exists at all), so I’m actually quite happy with it. We might want to test the `np.digitize` binning at some point to be sure, but that’s easy.

Thanks Jack for taking this to the next level, and to everyone else too for your valuable comments!

guanyilun (approved):
Go for it!

PR #439, opened 2023-04-03
Make Position Match

skhrg:
Has a ton of improvements both in terms of performance and interface and has now been tested on data from the LATRt.

Main thing that needs to be settled before this is merged is the interface with analyze-bright-ptsrc. @Zhilei do you have the format of your results written down somewhere (or can we discuss and write it down on the next BCP-call).


Zhilei:
Thanks! I have an updated result file saved on our server, I will Slack you with more details.

Zhilei (approved):

docs/site_pipeline.rst
```diff
@@ -276,6 +276,108 @@ Command line arguments
...
+    - "polang_res_2.h5"
+  detmap: "detmap_results.csv"
+  bias_map: "bias_map_results.npz"
+  
+  g3tsmurf:
+    obs_id: "obs_ufm_uv8_1677778438"
+    paths:
+        archive_path: "/mnt/so1/data/chicago-latrt/timestreams/"
+        meta_path: "/mnt/so1/data/chicago-latrt/smurf/"
+        db_path: "/mnt/so1/shared/site-pipeline/data_pkg/chicago-latrt/g3tsmurf.db"
```
kmharrington, line 316: let's not add private paths to public docs.

skhrg:
Ok, I have made another round of improvements most critically loading inputs from context. I would like to merge this and focus more in `finalize-focal-plane`. @Zhilei let me know it you have any objections.

sotodlib/site_pipeline/make_position_match.py
```diff
@@ -0,0 +1,883 @@
...
+            polang.append(det.angle_actual_deg)
+            det_ids.append(det.detector_id)
+            template_bg.append(det.bias_line)
+            is_north.append(det.is_north)
+        det_ids = np.array(det_ids)
+        template_bg = np.array(template_bg)
+        template_msk = np.isin(template_bg, valid_bg)
+        template_n = np.array(is_north) & template_msk
+        template_s = ~np.array(is_north) & template_msk
+        template_bg[template_bg % 2 == 1] *= -1
```
mhasself, line 530: This `*= -1` occurs in several places.  I cannot figure out what the point of that is.

sotodlib/site_pipeline/make_position_match.py
```diff
@@ -0,0 +1,883 @@
...
+        mapping = row_ind[np.argsort(col_ind)]
+
+    outliers = np.array([])
+    if out_thresh > 0:
+        outliers = np.where(P[mapping, range(P.shape[1])] < out_thresh)[0]
+
+    return mapping, outliers, P, TY
+
+
+def main():
```
mhasself, line 423: This is a monster of a function.  It needs more narrative comments, breaking things up and explaining what subblocks are doing.

Please take some of the little stand-alone blocks (such as the bit that assigns each det to a bias_group) and make then separate functions.  It may seem silly to have a function that is only called once, but the side benefits are:
- any local variables do not stick around forever in the main scope -- it's easier to tell what things are important to carry forward
- it encourages use of cohesive data structures that are easy to pass in and out of functions (e.g. dicts or AxisMans instead of a bunch of named arrays of the same size)
- it is easier to appreciate the flow of the main function, as a whole.  (A well chosen function name will document what the function is doing, so you don't even have to comment about it.)

In such small, stand-alone functions ... don't worry about docstrings; mark them private by naming them with a leading `_`.

mhasself (changes requested):
Thanks for this -- this looks like a nice solution.

I took a good look through the whole thing, but I admit I am not following a lot of the smurf / detmap stuff in detail.  I have left a bunch of requests for documentation and explaining some of the code a bit more.

I know you want to press forward, and we can do that if you prefer (and make my comments an "issue" to be resolved).  But it might be easier to make changes now while the whole thing is fresh.

skhrg:
These changes should be pretty easy to do. A lot of them are things that I had intending on fixing a some point but kinda forget anyways.

On the units. Nothing in this code really cares about the units. The units of the outputs are the units of the inputs. So I think nominally thats radians for pointing and mm for xy positions. Polarization I'm more unsure of, the angles from detmap (equivalent of xy in mm) is in degrees. Not really sure what the measured polarization angles are going to look like.

sotodlib/site_pipeline/make_position_match.py
```diff
@@ -0,0 +1,883 @@
...
+        outliers: Indices of points that are outliers.
+                  Note that this is in the basis of mapping and focal_plane, not template.
+
+        P: The likelihood array without priors applied.
+
+        TY: The transformed points.
+    """
+    if not bias_lines:
+        focal_plane = focal_plane[:, 1:]
+        template = template[:, 1:]
```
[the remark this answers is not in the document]
skhrg: This excludes bias line information when computing the transformation. The docstring might be a bit confusing.

sotodlib/site_pipeline/make_position_match.py
skhrg: Ah, that is a bit of a dumb trick. All it really does it punish the fitting algorithm for mixing up adjacent bias lines. Causes a moderate increase in my agreement with detmap. I could instead strictly enforce that the bias lines from `bgmap` but the current approach can handle some amount of error in `bgmap`. I actually think that the most "correct" thing to do is apply some non-affine transformation to the bias line information so it can guide the transformation in a way that cannot be reversed by the transformation itself. But that is something I want to think more deeply about before implementing.

sotodlib/site_pipeline/make_position_match.py
```diff
@@ -0,0 +1,883 @@
...
+        priors: The 2d array of priors in the basis of the focal plane and template that are to be matched.
+    """
+    likelihoods *= normalization
+    priors = 1 + likelihoods
+
+    missing = np.setdiff1d(final_template_det_ids, template_det_ids)
+    template_det_ids = np.concatenate(missing)
+    priors = np.concatenate((priors, np.ones((len(missing), len(fp_readout_ids)))))
+    asort = np.argsort(template_det_ids)
+    template_map = np.argsort(np.argsort(final_template_det_ids))
```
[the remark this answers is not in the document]
skhrg: Only equivalent off the top of my head is using `searchsorted` which does seem to be slightly faster:

```
In [25]: %timeit np.sort(b)[np.argsort(np.argsort(a))]
5.01 µs ± 29.9 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [26]: %timeit b[np.argsort(b).take(b.searchsorted(a, sorter=np.argsort(b)))]
4.64 µs ± 326 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

BUt if there is something more clever I am missing let me know

sotodlib/site_pipeline/make_position_match.py
```diff
@@ -0,0 +1,883 @@
...
+    Do an approximate transformation of the pointing back to
+    the focal plane using the mapping from the loaded detmap.
+
+    Arguments:
+
+        aman: AxisManager containing both pointing and datmap results.
+
+        pointing_name: Name of sub-AxisManager containing pointing info
+    """
+    xi_0 = np.nanmedian(aman[pointing_name].xi0)
```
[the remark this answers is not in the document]
skhrg: Yeah that was something I did when testing and then totally forgot to make it better. Does something smarter now. Also cleaned it up and added an explanation in the docstring. Let me know if what its doing is still unclear.

sotodlib/site_pipeline/make_position_match.py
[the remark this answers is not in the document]
mhasself: Right, nevermind ... I made a mistake.  I'll take another look at this and make a better suggestion that would have helped me be less confused...

skhrg:
Think that I addressed all the things we talked about. It also runs on the results of passing the simmed planet data through `analyze-bright-ptsrc` now.

sotodlib/site_pipeline/make_position_match.py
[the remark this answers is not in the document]
mhasself: The docstring says focal_plane is (4, n) or (3, n).  But these array ops are as if the focal_plane had shape (n, 4) or (n, 3).

docs/site_pipeline.rst
```diff
@@ -572,6 +572,115 @@ API
 .. autoclass:: sotodlib.site_pipeline.monitor.Monitor
     :members:
 
+make-position-match
```
mhasself, line 575: Move this up, before make-source-flags (i.e. alphabetical in the other elements).  Right now it's rendering inside "QDS Monitor".

sotodlib/site_pipeline/make_position_match.py
```diff
@@ -0,0 +1,1014 @@
...
+        polang.append(det.angle_actual_deg)
+        det_ids.append(det.detector_id)
+        template_bg.append(det.bias_line)
+        is_north.append(det.is_north)
+    det_ids = np.array(det_ids)
+    template_bg = np.array(template_bg)
+    template_msk = np.isin(template_bg, valid_bg)
+    template_n = np.array(is_north) & template_msk
+    template_s = ~np.array(is_north) & template_msk
+    template_bg[template_bg % 2 == 1] *= -1
```
mhasself, line 439: You need to write a comment in the code about why you're doing this. :)

One self-documenting way is to define a function, _scramble_bgs(bgs), and explain in there what the point of this is.  Then everywhere you need to do this, it's like template_bg = _scramble_bgs(template_bg) and an editor can much more easily find/check all the places this is being done.

mhasself (approved):
Approving but please resolve the minor doc things I mention below, before merging.

I'm making a list of other stuff to add, soon, but this is a good place to merge and come back later I think.

Thanks again.

PR #440, opened 2023-04-05
Preprocess updates

kmharrington:
Updates to `preprocess_tod` so that it can be used in an automated analysis setup. Big things are adding an entry point / argparser that looks for observations in Context that don't have matching entries in the preprocessing archive.

### DIFF
--- a/sotodlib/site_pipeline/preprocess_tod.py
+++ b/sotodlib/site_pipeline/preprocess_tod.py
@@ -199,8 +199,9 @@ def load_preprocess_tod(obs_id, configs="preprocess_configs.yaml", context=None 
     return aman
 
 
-def get_parser():
-    parser = argparse.ArgumentParser()
+def get_parser(parser=None):
+    if parser is None:
+        parser = argparse.ArgumentParser()
     parser.add_argument('configs', help="Preprocessing Configuration File")
     parser.add_argument(
         '--query', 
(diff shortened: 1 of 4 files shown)


sotodlib/site_pipeline/preprocess_tod.py
```diff
@@ -170,3 +197,40 @@ def load_preprocess_tod(obs_id, configs="preprocess_configs.yaml", context=None
...
+
+
+def get_parser(parser=None):
+    if parser is None:
+        parser = argparse.ArgumentParser()
+    parser.add_argument('configs', help="Preprocessing Configuration File")
+    parser.add_argument(
+        '--query', 
+        help="Query to pass to the observation list", 
+        default=2,
```
mhasself, line 209: Heh, worried that query(1) won't quite get everything?  Is None not equivalent to 1?  That would be a bug in my books...

mhasself (approved):
Thanks for the updates.  Approving, but with a couple soft requests...

Please consider adding two additional args:
--force, or --overwrite, to trigger reanalysis even if it targets exist in the output db already.
--target <obs>, to select a particular obs to work on

Those are useful in testing.  But also useful if we go the route of having Prefect run elements in sequence on each obs rather than relying on lots of cronjob mentality.

Also, I think @guanyilun wants to have the logger settable through main().  Can happen later, just mentioning it here since you're modifying the entry point already.

sotodlib/site_pipeline/preprocess_tod.py
kmharrington: yeah fair, I put in the query because I was doing something with a database that needed to exclude entries based on a calibration column. But between this and your other suggestions I'll expand the options / flags a but more.

kmharrington:
Ok, I think I've got all the soft requests as well. Going to go ahead and merge once the tests pass.

Issue #304, opened 2022-09-14
G3tSmurf Observations based on .g3 tags within other actions.


kmharrington:
Working through data from the LATR made me realize we need to look at expanding how calibration observations are made. 

The biggest example is "take_noise" which is being run inside of things like uxm_setup and uxm_relock. We could find these observations based on the presence of their files, but that will require additional logic, etc.

Based on the [flow_to_librarian plan](https://github.com/simonsobs/tod2maps_docs), we need to be able to connect the tags inside the .g3 files with observations. This will add an additional step/check in the observation making.

Issue #259, opened 2022-04-29
Only create TuneSet when it's found in streamed data files


kmharrington:
Right now we create a lot of extra TuneSets (one gets made for every `setup_notches` which means we get `N_bands-1` extra every time we retune an array). Re-work these so we only  make TuneSets off tune files we see in streamed `.g3` files.

kmharrington:
While doing this, Tuneset and Tune names need to include stream_id

Issue #218, opened 2022-02-03
G3tsmurf: Fix Tags Population


kmharrington:
For some reason, the Tags table is only getting the last tag input into the table instead of all of them. This needs to be fixed.

PR #448, opened 2023-04-21
G3tsmurf Updates 2023-04

kmharrington:
Going through a bunch of g3tsmurf updates at once. Now ready for review!

Biggest part of this PR is re-ordering the database updating and reworking some of the logic based on existing issues. To test this I made a databases covering all of SAT1's P10R2 cooldown using both the current master branch and this branch. All the differences between the two databases are because of fixing existing issues. Making note of what I checked here because this really should go into a test structure somehow:

- Size Old: 126Mb, New: 112Mb
- Channels: Old: 42572. New: 42288.
- Channel Assignments: Old: 198, New: 198
- Tunes: Old: 197, New: 197
- TuneSets: Old: 197. New: 25 ( Expected because of #259 )
- Observations: Old 2010. New 2222 ( Expected, with caveats because of #304 )
- Tags: Old 2006. New. 6728 (Expected because of #218 )
- Stray Files: Old: 206, New: 0 (Expected because of #304 )
- There are 6 observations in the new database that are categorized as observations based on tag when they were previously categorized as operations based on action name. This is likely an issue with tagging in sodetlib but the re-categorization is correct based on what is in the files. 
- The first 7 observations in the new database are not the old database because of how I ran the updates
- All other new observations are recovered noise timestreams in other actions.
- There are 4 old observations that did not have a tuneset but now do have one in the new dataset. Every new observation has a tuneset if the old observations have them.
- All new observations have the same tuneset as their respective old observations 
- All new TuneSets have the same set of readout_ids as the old TuneSets

I've also gone back in time and looked at some LATRt databases that did not have the final tagging implemented and have verified these updates do not break backward compatibility.

### DIFF
--- a/sotodlib/io/load_smurf.py
+++ b/sotodlib/io/load_smurf.py
@@ -583,7 +583,7 @@ class G3tSmurf:
         session : SQLAlchemy Session
             The active session
         """
-        name = tune_path.split("/")[-1]
+        name = stream_id+"_"+tune_path.split("/")[-1].strip(".npy")
         tune = (
             session.query(Tunes)
             .filter(Tunes.name == name, Tunes.stream_id == stream_id)
--- a/sotodlib/io/load_smurf.py
+++ b/sotodlib/io/load_smurf.py
@@ -267,16 +267,13 @@ class G3tSmurf:
             session.commit()
         
         status = SmurfStatus.from_file(db_file.name)
-        if len(status.tags)>0:
-            if status.tags[0]=="obs" or status.tags[0]=="oper":
-                ## this is where I tell it to make an observation
-                print(f"file {db_file.name} is an observation")
 
         if status.tune is None or status.tune == "":
             my_tune = None
         else:
-            tune_name = status.stream_id+"_"+status.tune.strip(".npy")
-            my_tune = session.query(Tunes).filter(Tunes.name==tune_name).one_or_none()
+            my_tune = session.query(Tunes).filter(
+                Tunes.name==Tunes.get_name_from_status(status)
+            ).one_or_none()
         if my_tune is not None:
             ## tune has been found in streamed data. 
             self.add_tuneset_to_file(my_tune, db_file, session)
@@ -377,7 +374,14 @@ class G3tSmurf:
         db_file.n_channels = total_channels
         db_file.n_frames = frame_idx
         db_file.timing = timing
+        session.commit()
 
+        if len(status.tags)>0:
+            if status.tags[0]=="obs" or status.tags[0]=="oper":
+                ## this is where I tell it to make an observation
+                logger.debug(f"file {db_file.name} is an observation")
+                self.add_new_observation_from_status(status, session)
+    
     def index_archive(
         self,
         stop_at_error=False,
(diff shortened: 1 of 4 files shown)


sotodlib/io/load_smurf.py
```diff
@@ -265,6 +265,18 @@ def add_file(self, path, session, overwrite=False):
...
+
+        if status.tune is None or status.tune == "":
+            my_tune = None
+        else:
+            my_tune = session.query(Tunes).filter(
+                Tunes.name==Tunes.get_name_from_status(status)
+            ).one_or_none()
+        if my_tune is not None:
+            ## tune has been found in streamed data. 
+            self.add_tuneset_to_file(my_tune, db_file, session)
```
jlashner, line 280: Does this create the requirement that the Tune must exist before the g3 file is indexed? I feel like this is kind of problematic, especially since it's very possible for a G3 file to be synced before the corresponding tune is synced (since there's a separate suprsync agent for the g3 timestreams and the smurf files). This means that even if `index_metadata`and `index_archive` are called in the correct order there can definitely still be cases where a g3-file is indexed before its corresponding tune...

In the case where `my_tune is None` can we create an empty tune that is populated when the metadata is indexed?

kmharrington: It does create that requirement but that's why I changed the order of indexing in the update script. The `index_metadata` runs through just channel assignment and tune files and adds all of them to the database. So if all the data transfer is up-to-date then this will always work. 

If the data transfer stops at some point we'll have the monitor telling us observations don't have readout ids. There is a way to re-run the updates to fix those. Although I've just realized I should add one more check to the update script to catch them automatically.

kmharrington: ok, new commit will let us re-check for tunes if data transfer between `smurf` and `timestreams` gets delayed.

jlashner: Nice, I still don't love the soft requirement that the data transfer is up-to-date since that adds some strange race conditions, but if exceptions are automatically handled by the update script that's good enough for me!

sotodlib/io/load_smurf.py
```diff
@@ -662,35 +748,53 @@ def add_new_observation(
...
+                    f" at {action_ctime}. Not Making Observation"
+                )
+                return
+
+            session_id = int((x.name[:-3].split("/")[-1]).split("_")[0])
+        
+        if calibration:
+            obs_id=f"oper_{stream_id}_{session_id}"
+        else:
+            obs_id=f"obs_{stream_id}_{session_id}"
```
jlashner, line 776: Noticing the `Observations` docstrings aren't up to date and don't have the `obs/oper` prefix.

sotodlib/io/load_smurf.py
kmharrington: yeah but there have always been these race conditions they were just in another place. Misses can be dealt with automatically and this way we have a lot lower overhead on the number of tunesets we need to track / make readout/detector id matching for.

sotodlib/io/load_smurf.py
```diff
@@ -1134,14 +1234,36 @@ def index_metadata(
             max_ctime=max_ctime,
             stop_at_error=stop_at_error,
         )
+        
+        session.close()
+
+    def index_action_observations(
+        self, min_ctime=16000 * 1e5, max_ctime=None, stop_at_error=False
+    ):
```
jlashner, line 1242: Not sure if I understand this one, why create a separate function that wraps `index_observations`? Is this only so that you don't need to pass in a session object?

jlashner (approved):
Thanks Katie, its looking good to me (with a couple of small questions/comments in line)!

I'm still slightly concerned about race conditions between the syncing and indexing, but I trust that you understand how everything fits together better than I do... Do you think there's any way for us to write books based on incomplete observation information? That's the main thing I'm worried about.

guanyilun (approved):
It looks good to me!

PR #455, opened 2023-04-28
context: bug-fix recent det_info changes

mhasself:
Oops, should put that det_info into aman.det_info, not into root level.

### DIFF
--- a/sotodlib/core/context.py
+++ b/sotodlib/core/context.py
@@ -305,7 +305,7 @@ class Context(odict):
                 # merge.  Duplicate keys should be avoided, because
                 # checking the values are the same is annoying.
                 _det_info = aman['det_info']
-                aman.move('det_info', None)
+                del aman['det_info']
                 _det_info.restrict_axes([meta.dets])
                 for k in meta.det_info._fields:
                     if k in _det_info._fields:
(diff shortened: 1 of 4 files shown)


kmharrington:
This gets us to `detset` and `detset_book` being in `aman.det_info`. That also seems confusing, those two fields had better be identical and if they aren't we have massive traceability problems.

mhasself:
> This gets us to detset and detset_book being in aman.det_info. That also seems confusing, those two fields had better be identical and if they aren't we have massive traceability problems.

Ya, it's not my favorite to have them there.  I will drop _book, and replace it with some internal consistency check in the loader. 

> `detset_book` and `stream_id` are also not explicitly along the dets axis.

I don't understand what you mean.

kmharrington:
When I print the AxisManager it said (for example) `stream_id[64]` instead of `stream_id[dets]` that means it's not assigned to the `dets` axis.

mhasself:
> When I print the AxisManager it said (for example) `stream_id[64]` instead of `stream_id[dets]` that means it's not assigned to the `dets` axis.

I did not think that could possibly have happened!  Working on it...

kmharrington (approved):
Approving this because the two things I wanted to change I just went in and changed. @mhasself definitely double check my changes but I think we've mostly talked about them.

mhasself:
Thanks!

PR #457, opened 2023-05-02
Detrend Speedup

kmharrington:
The hour+ SAT books are giving us a good opportunity to profile the `tod_ops` code. There are several we'll need to work on but this is the start. 

Detrend on 1 UFM x 1.2 hours of data was taking ~110 seconds on simons1. The longest parts were subtracting the slopes and recasting at then end (in case data types had changed). Moving to the `-=`, using the `for` loop if possible, and only recasting if the data type had changed got it down to ~6 seconds. Added some comments so we hopefully remember why it's done this way.

Also add another argument to `get_glitch_flags` which lets us speed that up a little bit but that one will take some more careful work to speed up in the future (it's currently at ~240 seconds). 

Edit: Added some quick changes to `calc_psd` to let us just reduce the samples we call for it. I'm not planning any other changes on this branch.

### DIFF
--- a/sotodlib/tod_ops/detrend.py
+++ b/sotodlib/tod_ops/detrend.py
@@ -38,6 +38,7 @@ def detrend_data(tod, method='linear', axis_name='samps',
     n_samps = signal.shape[axis_idx]
     
     # Ensure last axis is the one to detrend.
+    # note that any axis re-ordering is not the slow part of this function
     if axis_idx != signal.ndim - 1:
         # Note this reordering is its own inverse; e.g. [0, 1, -1, 3, 4, 2]
         axis_reorder = list(range(signal.ndim))
@@ -52,14 +53,24 @@ def detrend_data(tod, method='linear', axis_name='samps',
         x = np.linspace(0, 1, n_samps)
         count = max(1, min(count, signal.shape[-1] // 2))
         slopes = signal[...,-count:].mean(axis=-1)-signal[...,:count].mean(axis=-1)
-        signal = signal - slopes[...,None] * x
+        if len(signal.shape)>2:
+            signal -= slopes[...,None] * x
+        ## the 2d loop is significantly faster if possible
+        else:
+            for i in range(signal.shape[0]):
+                signal[i,:] -= slopes[i]*x
         signal -= np.mean(signal, axis=-1)[...,None]
     else:
         raise ValueError("method flag must be linear, mean, or median")
 
     if axis_idx != signal.ndim - 1:
         signal = signal.transpose(tuple(axis_reorder))
-    return signal.astype(dtype_in)
+
+    if signal.dtype != dtype_in:
+        # .astype takes awhile on long arrays, only cast if necessary
+        return signal.astype(dtype_in)
+    else:
+        return signal
 
 def detrend_tod(tod, method='linear', signal_name='signal', axis_name='samps', count=10, out_name=None):
     """simple wrapper: to be more verbose"""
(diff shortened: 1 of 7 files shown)


mhasself (changes requested):
Looks fast but the docstring needs to be updated -- this now does modify the signal in place.  Does that break any in-tree usage?  I'm ok with breaking the default behavior, but it needs to be propagated.

Related augmentation proposal: add a "wrap=None" argument, which by default will simply detrend the signal in place ... but if passed as a string (wrap="signal_detrended") will cause the input data to be copied and stored in the axisman under that key.

sotodlib/tod_ops/detrend.py
```diff
@@ -52,14 +53,24 @@ def detrend_data(tod, method='linear', axis_name='samps',
...
 
     if axis_idx != signal.ndim - 1:
         signal = signal.transpose(tuple(axis_reorder))
-    return signal.astype(dtype_in)
+
+    if signal.dtype != dtype_in:
+        # .astype takes awhile on long arrays, only cast if necessary
+        return signal.astype(dtype_in)
+    else:
+        return signal
```
mhasself, line 73: It might be worth guarding against the first case entirely.  If you construct x and slopes to have the same type as signal, that should be enough.  Then you can assert signal.dtype == dtype_in as a fail-safe that will hopefully never ring.

kmharrington:
Ok @mhasself  I think I've addressed your comments. In the process I've realized that we've gotten to the point where `detrend_data` and `detrend_tod` had effectively become redundant, so I've renamed `detrend_data` to `detrend_tod`. (most people outside of sotodlib have been using `detrend_tod`, so I expect that's the least disruptive. 

I've updated the tests as well.

sotodlib/tod_ops/detrend.py
```diff
@@ -1,10 +1,19 @@
...
+    signal_name='signal', 
+    in_place=True,
+    wrap_name=None,
+    count=10
+):
     
-    """Returns detrended data. Decide for yourself if it goes
-        into the axis manager. Generally intended for use before filter.
+    """Returns detrended data. Detrends data in place by default but pass
+        in_place=False if you would like a copied array (such as if you're just
```
mhasself, line 14: Un-indent second (and all subsequent) lines of docstring.

In argument descriptions, please capitalize start of sentences after periods.

The docstring style here isn't quite numpy nor google... but sphinx seems ok with it, and it matches the style of other tod_ops docs.  So I won't fight that.

sotodlib/tod_ops/detrend.py
```diff
@@ -49,28 +66,26 @@ def detrend_data(tod, method='linear', axis_name='samps',
     elif method == 'median':
         signal = signal - np.median(signal, axis=-1)[...,None]
     elif method == 'linear':
-        x = np.linspace(0, 1, n_samps)
+        x = np.linspace(0, 1, n_samps, dtype=dtype_in)
         count = max(1, min(count, signal.shape[-1] // 2))
-        slopes = signal[...,-count:].mean(axis=-1)-signal[...,:count].mean(axis=-1)
-        signal = signal - slopes[...,None] * x
+        slopes = signal[...,-count:].mean(axis=-1,dtype=dtype_in)-signal[...,:count].mean(axis=-1,dtype=dtype_in)
```
mhasself, line 71: Please flake8 this line... the suppression of spaces has not, alas, led to it all being readable on a single line.

Other lines below have arithmetic spacing violations too.

mhasself (approved):
Happy to see those detrend functions merged!  I'd appreciate some attention to formatting though ... see comments on a couple lines, but there is trouble of this sort throughout tod_ops.  If we'd rather do a bulk reformat in separate PR, I will tolerate a punt on this.

PR #460, opened 2023-05-04
Metadata Loading Changes

kmharrington:
While working with a bunch of preprocessing metadata I was reminded that I dislike the current behavior of the metadata loader. Currently, if the loader encounters a ManifestDb with _missing_ information from some detectors, it just quietly drops those detectors from the returned metadata object. 

This is guaranteed to confuse folks at some point when we're going through pipeline-like processes where we're depending on something having been run before-hand. Especially if we load an observation, get a "partial" observation back, and then assume we're "done" with that observation.

This update should at least mostly fix this issue. As we're going through and loading the different ManifestDbs, if the newly added fields _are not_ options for data-selection but reduce the overall number of detectors, the code will throw an error unless `ignore_missing_dets` is set to True (where it will then give a warning instead). I would prefer having the default behavior be `ignore_missing_dets=False`, because I really do thing the `True` option will hide more problems then it helps. 

The code works, this is just a draft because the context tests were absolutely essential for getting this working correctly so that means I definitely need to add some for the new behavior.

### DIFF
--- a/sotodlib/core/context.py
+++ b/sotodlib/core/context.py
@@ -177,6 +177,7 @@ class Context(odict):
                 detsets=None,
                 meta=None,
                 ignore_missing=None,
+                ignore_missing_dets=False,
                 free_tags=None,
                 no_signal=None,
                 loader_type=None,
@@ -211,6 +212,8 @@ class Context(odict):
             obs_colon_tags fields for detector restrictions.
           ignore_missing (bool): If True, don't fail when a metadata
             item can't be loaded, just try to proceed without it.
+          ignore_missing_dets (bool): If True, don't fail when metadata entries
+            are missing information for a subset of detectors.
           no_signal (bool): If True, the .signal will be set to None.
             This is a way to get the axes and pointing info without
             the (large) TOD blob.  Not all loaders may support this.
@@ -279,7 +282,8 @@ class Context(odict):
         """
         meta = self.get_meta(obs_id=obs_id, dets=dets, samples=samples,
                              filename=filename, detsets=detsets, meta=meta,
-                             free_tags=free_tags, ignore_missing=ignore_missing)
+                             free_tags=free_tags, ignore_missing=ignore_missing,
+                             ignore_missing_dets=ignore_missing_dets)
 
         # Use the obs_id, dets, and samples from meta.
         obs_id = meta['obs_info']['obs_id']
@@ -335,6 +339,7 @@ class Context(odict):
                  free_tags=None,
                  check=False,
                  ignore_missing=False,
+                 ignore_missing_dets=True,
                  det_info_scan=False):
         """Load supporting metadata for an observation and return it in an
         AxisManager.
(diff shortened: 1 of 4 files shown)


kmharrington:
Ok, new tests are up so this should be ready for review.

mhasself (commented):
Sorry to be slow getting to this.  This looks well-constructed; thanks for enhancing the tests.

Unfortunately this sent me down an analytic rabbit hole, and now I want us to consider a refactoring.

A minor point -- we have wrapped ACT data that relies on simply tossing detectors that have no metadata.  I'd like to preserve that behavior as an option configurable in the context.yaml somehow, even if the default is to be more strict.

More importantly: I think these two parameters, `ignore_missing` and `ignore_missing_dets`, don't really let us access all the behaviors we want.  As a test case, imagine an obs X with two detsets [detset1, detset2], and a metadata product Y that only has a ManifestDb entry for detset1, but not detset2.  That means:

These succeed:
```
  # Result: only load detset1, including metadata Y.
  ctx.get_obs(X, ignore_missing_dets=True)
  ctx.get_obs(X, detsets=["detset1"])
  
  # Result: only load detset2, excluding metadata Y.
  ctx.get_obs(X, detsets=["detset2"], ignore_missing=True)
```
  
These fail:
```
  # Fail because no metadata Y for detset2.
  ctx.get_obs(X)
  ctx.get_obs(X, ignore_missing=True)
  ctx.get_obs(X, detsets=["detset2"])
  ctx.get_obs(X, detsets=["detset2"], ignore_missing_dets=True)
```
  
Importantly there is _no way_ to achieve the following:
```
  # Result: load detset1 and detset2 (excluding metadata Y)
  ???
```

I.e. the incompleteness of Y forces us to either load detset1 with Y or load detset2 without Y.  There is no way to push through and load _something_ for detset1+detset2!

So I think we should revisit what `ignore_missing` and `ignore_missing_dets` are doing, and get a complete, useful picture together.  This is what makes sense right to me today:

1. We should probably treat "totally absent" as an extreme form of "incomplete"; i.e. if an obs has no metadata entries, that should be treated the same way as if it had partial coverage (only some detsets, only some dets).

2. If incomplete metadata are encountered, I can imagine cases where I'd want to do each of the following:
  - Fail.
  - Do not fail.  Cancel the metadata load of Y (for all dets).
  - Do not fail.  Toss the dets for which the metadata Y is missing and carry the others forward (even if there are none left).

So what if those two arguments were replaced with this one:
   - on_missing="fail" (or False)
   - on_missing="skip"
   - on_missing="trim"

This approach would make it impossible to request the equivalent of `get_obs(X, ignore_missing=True, ignore_missing_dets=False)`, a.k.a. "don't fail if the metadata are entirely absent, but do fail if the metadata are present and incomplete."  But I think that case is a bit pathological

kmharrington:
I see what you mean in your comments, and I think we're getting to a point where we'll want these metadata loading options to be configurable on a per-ManifestDb basis. Any objection to turning these into a key we can add to each metadata entry to set behavior? 

I think this is a large enough change on top of this setup that it's going to be a week or two to get to it.

mhasself:
> I see what you mean in your comments, and I think we're getting to a point where we'll want these metadata loading options to be configurable on a per-ManifestDb basis. Any objection to turning these into a key we can add to each metadata entry to set behavior?

No objection!

PR #463, opened 2023-05-09
add GlitchAggregate preprocessing module

guanyilun:
WIP


guanyilun:
This branch has been worked on for quite some time. @jsyleung has done a lot of work polishing and cleaning up this branch. This is more or less stable now for people who routinely work with this so we think it's a good idea to have people to look at it and possibly merge to master. I think it shouldn't break any existing preprocess modules except adding a few more. @jsyleung can answer more detailed questions about what goes into this PR.

mhasself:
I'd like to take a look at this.

New dependency scikit-learn does not support py3.10, but this library does.  So that's a concern.

msilvafe:
> I'd like to take a look at this.
> 
> New dependency scikit-learn does not support py3.10, but this library does. So that's a concern.

FYI @mhasself I am waiting for your review before I review.

jsyleung:
@mhasself py3.10 is scheduled reach end of life in ~4 months (Oct 31). Will we be bumping versions? If not, perhaps the solution is to fix `"sklearn=1.7.0"`, which supports Python >=3.10. (For the record, sklearn 1.8 and 1.9 both require Python >=3.11)

There is merit to fixing a version, as models aren’t necessarily forward-compatible. We’ll probably want to have a version with the longest support life though.

docs/tod_ops.rst
```diff
@@ -232,3 +232,36 @@ Module containing functions for FFTs and related operations.
 
 .. automodule:: sotodlib.tod_ops.fft_ops
    :members:
+
+tod_ops.glitch_ml
```
msilvafe, line 236: Can you cite the ACT paper which this is based off of as a reference in this documentation?

sotodlib/tod_ops/glitch_ml.py
```diff
@@ -0,0 +1,648 @@
...
+
+    Returns
+    -------
+    stats_arr : numpy.ndarray
+        Array of shape ``(n_stats,)`` with the computed summary statistics.
+    """
+    if stats is None:
+        stats = get_default_stats()
+
+    signal = detrend_tod(snippet, method='median')
```
msilvafe, line 260: Perhaps you want to expose the method argument to compute_summary_stats so you could compare to using a linear or mean detrending?

sotodlib/tod_ops/glitch_ml.py
```diff
@@ -0,0 +1,648 @@
...
+
+    roll_corr = -np.mean(snippet.boresight.roll)  # roll correction
+    xi, eta = snippet.focal_plane.xi, snippet.focal_plane.eta
+    x_wnans = np.rad2deg(xi * np.cos(roll_corr) - eta * np.sin(roll_corr))
+    y_wnans = np.rad2deg(eta * np.cos(roll_corr) + xi * np.sin(roll_corr))
```
msilvafe, line 265: I think these are SAT specific rotations, should double check this for LAT and make code flexible to platform.

sotodlib/tod_ops/glitch_ml.py
```diff
@@ -0,0 +1,648 @@
...
+
+    x_t = x_wnans[~np.isnan(x_wnans)]
+    y_t = y_wnans[~np.isnan(y_wnans)]
+
+    data = {'signal': signal, 'x_pos': x_t, 'y_pos': y_t}
```
msilvafe, line 270: How do you manage x_t and y_t not matching the axes in signal bc of the masking?

sotodlib/tod_ops/glitch_ml.py
```diff
@@ -0,0 +1,648 @@
...
+    return forest
+
+
+#: Column names for the prediction arrays returned by :func:`classify_data_forest`
+#: and :func:`classify_snippets`.  Column 0 is the predicted class label (0–3);
+#: columns 1–4 are the class probabilities.
+PRED_COLUMNS = ['Glitch Prediction', 'Probability of being a Point Source',
+                'Probability of being a Point Source + Other',
+                'Probability of being a Cosmic Ray',
+                'Probability of being an Electronic Glitch']
```
msilvafe, line 331: Are we pretty confident that this is the set of classifications you want for all platforms/for SO. My understanding was this was based on ACT and it's reasonable to think we may want different classifications for SO. Can some of this be configurable in a config file/configuration class instead of hardcoded in? (ok to save for later but open an issue for this if you intend to make this update later).

sotodlib/tod_ops/glitch_ml.py
```diff
@@ -0,0 +1,648 @@
...
+    """
+    if stats is None:
+        stats = get_default_stats()
+
+    if X_train.shape[1] != len(stats):
+        raise ValueError(
+            f"Number of columns in X_train ({X_train.shape[1]}) does not "
+            f"match number of stat names ({len(stats)})")
+
+    forest = RandomForestClassifier(criterion='entropy', n_estimators=n_trees, random_state=1, n_jobs=2, max_depth=max_depth)
```
msilvafe, line 317: Why is `n_jobs=2` hardcoded here? Does this require a specific MPI setup or is it OpenMP? I guess just want to double check on conflicts dependent on compute site. Perhaps @mhasself would have more thoughts on this particular point?

sotodlib/tod_ops/glitch_ml.py
```diff
@@ -0,0 +1,648 @@
...
+    Smooths the per-sample maximum/mean across detectors with a 3-wide
+    boxcar, selects prominent samples, and counts peaks with
+    :func:`scipy.signal.find_peaks`.
+    """
+
+    name = "Number of Peaks"
+    requires = ("signal",)
+
+    def calc(self, *, signal):
+        kernel = np.ones(3) / 3
```
msilvafe, line 637: We have this sort of thing baked in in a few places in sotodlib not just in this module but when you do things like this based on samples you create inflexibility to sample rate changes. Perhaps this should be a time based boxcar?

msilvafe (changes requested):
Please resolve merge conflicts. Change the status from WIP since I believe you are ready to merge (according to slack exchange). @mhasself do you have any opinions about adding in the scikit-learn dependency? Also `glitch_ml` module has a lot of helper functions which are a single line (i.e. `ranged_from_n_flagged`, `ranges2slices`, `build_snippet_layout`, etc.) I'm not sure that's needed and it could all just go into `get_snippets` or the other heavier functions (i.e. `compute_summary_stats`,  `train_forrest` etc.). Also it seems like there's an overall lack of logging/print statements to help track code progress/debug. I suggest expanding that (but if this is working for you that's optional). I've also provided some inline comments to address. Thanks and sorry for the delay! I'll make sure that the next iteration cycle on this review is short!

PR #464, opened 2023-05-10
Update to newer python version in readthedocs

BrianJKoopman:
This should fix the docs builds that broke recently.


kmharrington (approved):
seems like it did something?

PR #467, opened 2023-05-16
Sat mapmake

tterasaki:
Adding map-making functions in coords.sat_mapmake.

Validation is done in 
`/home/tterasaki/workspace/sat_mapmake_debug/230509_compare_mapmake.ipynb` at simons1.

The attached figure is result of the mapmaking. Each column corresponds to T,Q, and U.
First row is from Matthew's demonstration code. Second is from coords.sat_mapmake.make_map.
Third is difference between 1st and 2nd. They agrees perfectly.
download


sotodlib/coords/sat_mapmake.py
```diff
@@ -0,0 +1,102 @@
...
+        If None, no cuts will be applied.
+    det_weights_dsT : array-like or None, optional
+        The detector weights to use in the map-making for the dsT timestream.
+    det_weights_demod : array-like or None, optional
+        The detector weights to use in the map-making for the demodulated Q and U timestreams.
+        If both of `det_weights_dsT` and `det_weights_demod` are None, uniform detector weights will be used.
+        If only one of two are provided, the other weight is provided by `det_weights_dsT` = 2 * `det_weights_demod`.
+
+    Returns
+    -------
```
mhasself, line 43: I think it would be good to return a dict of maps here.  See coords.planets.make_map, for example, and imitate that structure.  Returning a dict makes it easier to enrich the output (by adding in the projection operator, for example) without breaking the interface.

sotodlib/coords/sat_mapmake.py
```diff
@@ -0,0 +1,102 @@
...
+    if wcs_kernel is None:
+        wcs_kernel = coords.get_wcs_kernel('car', 0, 0, res)    
+    if dsT is None:
+        dsT = tod['dsT']
+    if demodQ is None:
+        demodQ = tod['demodQ']
+    if demodU is None:
+        demodU = tod['demodU']
+    
+    PQU = coords.P.for_tod(tod=tod, wcs_kernel=wcs_kernel, cuts=cuts, comps='QU')
```
mhasself, line 60: Can you make PQU an optional argument to this function (but call it `P`).  The reason is that we might want to map in a different coordinate system (such as source-centered or planet-scan).  If P is provided as an argument, you don't need to check wcs_kernel.

mhasself (changes requested):
Thanks!  See a few minor comments, most of which we discussed already.

sotodlib/coords/sat_mapmake.py
```diff
@@ -0,0 +1,102 @@
...
+    
+    # Q/U maps and weights
+    mQ_weighted = PQU.to_map(tod=tod, signal=demodQ, det_weights=det_weights_demod)
+    mU_weighted = PQU.to_map(tod=tod, signal=demodU, det_weights=det_weights_demod)
+    mQU_weighted = PQU.zeros()
+    
+    #### CAUTION: Here the definition of mQU_weighted uses a wrong way of definition, as toast simulation defines that in the wrong way.
+    mQU_weighted[0][:] = mQ_weighted[0] - mU_weighted[1] # (= Q_{flipped detector coord}*cos(2 theta_pa) - U_{flipped detector coord}*sin(2 theta_pa) )
+    mQU_weighted[1][:] = mQ_weighted[1] + mU_weighted[0] # (= Q_{flipped detector coord}*sin(2 theta_pa) + U_{flipped detector coord}*cos(2 theta_pa) )
+    #### In field, you should use instead ####
```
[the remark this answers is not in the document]
tterasaki: I have revisited the definitions and I think this function is correct.
The calculation in code is consistent with  the pointing matrix with detector gamma -> -gamma. 
Each component in the code is as follows:
- mQ_weighted[0] = Q_{demod}*cos(2gamma)
- mQ_weighted[1] = Q_{demod}*sin(2gamma)
- mU_weighted[0] = U_{demod}*cos(2gamma)
- mU_weighted[1] = U_{demod}*sin(2gamma)

So calculations for the wrong definitions:
- mQU_weighted[0][:] = mQ_weighted[0] - mU_weighted[1]
- mQU_weighted[1][:] = mQ_weighted[1] + mU_weighted[0]

, and calculations for the correct definitions:
- mQU_weighted[0][:] = mQ_weighted[0] + mU_weighted[1]
- mQU_weighted[1][:] = - mQ_weighted[1] + mU_weighted[0]

, are consistent with the equations in [this link](https://docs.google.com/presentation/d/1Qym2md0zIVxSRKcgVpWswsgBimxQYY_QoEPDXHGoxcQ/edit#slide=id.g24bb2e11f96_0_5)

mhasself (approved):
Ok, I am convinced.  Thanks for this!

Issue #468, opened 2023-05-16
G3HWP load_data and load_file do  not work interchangeably


kmharrington:
While helping set up some data access for TSAT I'm noticing that the two different modes of operation for G3HWP don't work interchangeably. I can load a single file from HWP data but if I try to load via `load_data` I get an error that fields don't exist. Example below: 

```
import datetime as dt
import sotodlib.hwp.g3thwp as g3thwp

fname = "/mnt/so1/data/ucsd-sat1/hk/16667/1666732499.g3"
cfg_file = "/mnt/so1/users/kmharrin/hwp_test/update_hwp_config.yaml"

hwp = g3thwp.G3tHWP(cfg_file)
## works
#data = hwp.load_file(fname)
## does not work
data = hwp.load_data( 
    start = dt.datetime(2022,10,25,21,14), 
    end = dt.datetime(2022,10,25,21,16),
)
solved = hwp.analyze(data)
print( solved["fast_time"].shape, solved["angle"].shape )
```

PR #469, opened 2023-05-17
G3tHWP load_data fix

yukisakurai:
I have fixed a bug in the load_data function in g3thwp that was due to the datetime format.
(Connected to issue #468)


kmharrington (approved):
Thanks!

mhasself:
Are you going to check the timezone of such datetime objects?  See discussion [here](https://github.com/simonsobs/sorunlib/pull/61).

If you don't associate a timezone with a datetime, it will quietly assume the computer's local timezone.

PR #471, opened 2023-05-18
Adding timezone check to G3tHWP

yukisakurai:
I added time zone check for datetime input arguments in g3thwp, setting timezone to utc with logger.warning if there is no timezone info.
https://github.com/simonsobs/sotodlib/pull/469#issuecomment-1551701763


mhasself (approved):
Thanks!

PR #473, opened 2023-05-23
Sss module

msilvafe:
This adds the ability to fit out Legendre polynomials in az-signal space for ground/sss template subtraction. I added the basic fitter and signal subtraction functions that interface with the axis managers, much like the hwpss module. There's some basic docs and a simple unit test in here as well. However, I think there are a few things worth discussing and/or adding before we move into final PR review:

1. Do we want the option to bin in az before fitting?
2. Do we need to include some fit statistics? 
3. Is this the right place in the repository for this to live?
4. We should add these processes into the preprocess module as well
5. Do we have a target time this should run in? I haven't done any realistic timing tests but on the toast data I ran it had 860  detectors for 14 minutes and took 6.5 minutes to perform the fitting.

I have tested that this works on both a toast sim on simons one found here: `/home/msilvafe/mnt/flp_dev/sat_sims/20221213/out_sss` and a dummy axis manager I created that just has az-legendre polynomials and white noise (function is in the unit test script to generate this). Some examples of those below:
| TOAST Sim Test  | Dummy AxisManager Test |
| ------------- | ------------- |
| TOAST_test_SSS | Dummy_test_SSS |
| Seems like TOASTs default template to use for the Az signal is not really well modeled by these legendre polynomials but given this is what ABS used that's what I started with. I'm also fairly certain the TOAST default isn't very physical given previous conversations with Reijo.  | The fit recovers the input as expected when the input signal is just noise + legendre polynomials in az.  |

Both of these are generated in a notebook on simons1: `'/mnt/so1/users/msilvafe/notebooks/flp_dev/20230518_TOD_Processing.ipynb'`


tterasaki:
Thank you for grateful work! I think binning will be helpful.
When we want to assess SSS itself, such as making ground template by coadding SSS of all observation, the full sampled time-ordered SSS data will be too much if not binned.

msilvafe:
> Thank you for grateful work! I think binning will be helpful. When we want to assess SSS itself, such as making ground template by coadding SSS of all observation, the full sampled time-ordered SSS data will be too much if not binned.

Thanks for taking a look @tterasaki. Feel free to commit to a binning function and any statistics you want. We could probably implement a linreg function as we did in the hwpss module as well using [`scipy.special.legendre`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.legendre.html) since they're orthogonal in this -1 to 1 space if we end up needing it to run faster.

Now that you point out that we want to be able to bin over _all of the observations_ in order to make a ground template, I'm realizing that my conversion from az pointing -> x won't work. It takes x to +/-1 for the az range in the given TOD _not_ the full 360 deg az range. I suspect that's not what we want to do either. We want x to be az normalized to have +/-180 = +/-1 always and have the bins always be defined over that full +/-180 range. That should be a relatively simple fix code change, but without testing, I'm not sure the fits won't be worse when fitting a subset of the az-range instead of the full +/-180.

kmharrington:
An incomplete list of thoughts:

1. Yes binning with masking (masking flags) would be a good option. This option should also run significantly faster than a fit to the whole time series. Could you and @tterasaki look at making a more generic version of [get_binned_signal ](https://github.com/simonsobs/sotodlib/blob/de684da4c1cd10a576ac9ff21051f52706eeb3d6/sotodlib/hwp/hwp.py#L178) that we could put in TOD ops and call in all places we need to bin data? This is the type of operation where c++ acceleration is likely to become helpful, so if we start by having all functions call the same binning function then it will be easier to optimize later.

2. Fit statistics will be helpful in the future if they're saved as a data product so we can see what sort of distributions we get.

3. Just move this file into tod_ops and pin the `.rst` into the `tod_ops` section. 

4. Yes please.

5. Ya, very slow. Binning will likely be helpful in speeding things up if the slow speed is just the fits. You can try linearizing the fits like the HWP ones but I'm a bit worried that's not the issue b/c the docs for the function you're calling says it's doing basically the same thing. I could also now be asking for a generalized linear fitting function that can be called on many different times of linear fits.

More thoughts on binning / data products:

* I'm thinking we could be using this in two different ways, one for instrument characterization of azimuth synched pickup and one for just cleaning low frequency modes from the data. What you have now is great for the second thing but we'd need more dev to use this for the first thing. 

* Binning with the median instead of the mean is more robust to junk in the data but takes longer to calculate.

* For SSS instrument characterization, we're going to have to deal with more than just az ranges. We'll also be combining as a function of elevation and boresight rotation. It is safe to say the elevation and boresight will be constant over the observation, but mostly pointing out that we'll need to be recombining this data into different groups later. For that, I think saving the binned data (not the fits) on a standard set of bins (configurable in the preprocess module?) will be the most useful. If the binned data / errors are saved then we can combine them from different observations and then fit. 

* You are right that if we make bins from -180 -> +180 then the fits will change. In particular, for single observations there will be modes that we have no constraining power for. Low order modes will all look similar to each other. This is another reason for  saving the binned data then doing fits across many observations, at least for when we want to build up the actual SSS from things like ground pickup. 

* I would also leave what you have as an option, because it is what we could use for just cleaning "any low frequency mode" from the data.

tterasaki:
I have added the binning (and removed the tod fitting).
Now binning function is under the tod_ops, and get_hwpss uses the general binning function as well.
After the binning, two method can be specified: 'interpolation' and 'fitting'.
In 'interpolation', use the binned signal as SSS template directly.
The good point of 'interpolation' is that this method does not depend on fitting.
So I think we will use 'interpolation' usually.
The `fit` method is the same as before.

One note is that default of binning [-np.pi, np.pi].
In this case, binning is done over [-np.pi, np.pi], which means there is some bins which do not contains samples.
And the full range of [-np.pi, np.pi] is converted to [-1, 1] of `x_Lugendre`.
So fitting is applied for data in the partial region which contains non-zero samples.
If you want to do fitting using full range from minimum az to maximum az, you can use `bin_range = None`.

You can see my demonstration in `/mnt/so1/users/tterasaki/workspace/sss_dev/230531_debug02_1_bin_and_fit.ipynb`

tterasaki:
@msilvafe Can we merge this PR into master? I mean, although there are something to be updated around preprocessing, the core of this PR looks working with the site data. This function is useful for current site data with SSS. So I want to merge it if it is possible and work on the preprocessing stuff in another branch.

msilvafe:
> @msilvafe Can we merge this PR into master? I mean, although there are something to be updated around preprocessing, the core of this PR looks working with the site data. This function is useful for current site data with SSS. So I want to merge it if it is possible and work on the preprocessing stuff in another branch.

I have one last set of local changes to this to pull the fitting part out of the main sss function into its own function. I think that's the last thing I wanted to do before converting to a real PR from draft. Can you wait for that?

msilvafe:
@mhasself and @tterasaki I think this is ready for review so we can merge and start using it on the commissioning data.

I moved the fitting function out of the main `get_sss` function to `fit_sss` so it can be called independently. My thinking on this follows my comment above about using `get_sss` to output a binned version of the sss over many observations which can be averaged into a new `sss_stats`-like axis manager and then passed into `fit_sss` for studying ground template stability over time, el ranges, etc. 

One detail which we can decide to update later is if we want this merged to use the `binned_signal_sigma` as weights in the legendre fitting. The function we are using now `numpy.polynomial.legendre.legfit` takes in x of size m and y of size m, k but does not allow weights of size m, k. I tried it fitting a single detector (i.e. x size m, y size m, and weights size m) and it did seem to improve the fits (by eye) but if we want to run that way we will have to loop over detectors which sacrifices some of the speedup we got using this function. However if the fit part is not run too often (only on these binned averages of many observations types of datasets) maybe this slow down is fine.

tterasaki:
Looks nice. Thank you Max!

tterasaki (approved):

sotodlib/tod_ops/binning.py
```diff
@@ -0,0 +1,67 @@
...
+    ----------
+    aman : TOD
+        The Axismanager object to be binned.
+    bin_by : array-like
+        The array by which signal is binned. It should has the same samps as aman.
+    signal : str, optional
+        The name of the signal to be binned. Defaults to aman.signal if not specified.
+    bin_range : list or None
+        A list specifying the bin range ([min, max])
+        Default is None, which means bin range is set to [min(bin_by), max(bin_by)] automaticaly by np.histogram.
```
mhasself, line 20: These two lines, in sphinx, will render as a single paragraph.  So end the first line with a period.  (And I would prefer that your wrap it all into a single paragraph rather than using the newline to imply a paragraph break.)

sotodlib/tod_ops/binning.py
```diff
@@ -0,0 +1,67 @@
...
+    bins : int, optional
+        The number of bins to use. Default is 100.
+    flags : RangesMatrix, optinal
```
mhasself, line 23: optional

sotodlib/tod_ops/binning.py
```diff
@@ -0,0 +1,67 @@
...
+        The Axismanager object to be binned.
+    bin_by : array-like
+        The array by which signal is binned. It should has the same samps as aman.
+    signal : str, optional
+        The name of the signal to be binned. Defaults to aman.signal if not specified.
+    bin_range : list or None
+        A list specifying the bin range ([min, max])
+        Default is None, which means bin range is set to [min(bin_by), max(bin_by)] automaticaly by np.histogram.
+    bins : int, optional
+        The number of bins to use. Default is 100.
```
mhasself, line 22: In other binning functions, such as numpy.histogram, "bins" can be a number of bins (like this) or an array with the bin edges.

It would be nice to imitate that interface -- if bins is passed in as an array, it is used as the bins.

sotodlib/tod_ops/binning.py
```diff
@@ -0,0 +1,67 @@
...
+                                          weights=signal[i][m[i]])[0][mcnts] / bin_counts[mcnts]
+
+        binned_signal_squared_mean[i][mcnts] = np.histogram(bin_by[m[i]], bins=bins, range=bin_range,
+                                                       weights=signal[i][m[i]]**2)[0][mcnts] / bin_counts[mcnts]
+
+    # get sigma of each bin
+    binned_signal_sigma[:, mcnts] = np.sqrt(np.abs(binned_signal_squared_mean[:,mcnts] - binned_signal[:,mcnts]**2)
+                                 ) / np.sqrt(bin_counts[mcnts])
+
+    return {'bin_centers': bin_centers, 'binned_signal': binned_signal, 'binned_signal_sigma': binned_signal_sigma}
```
mhasself, line 67: Nice thing about returning a dict is you can put all the useful stuff in there.  Can you put the bin edges in there too?  That is a useful thing one can pass to other binning functions (or back into this one, maybe, someday).

sotodlib/tod_ops/binning.py
```diff
@@ -0,0 +1,67 @@
...
+    # bin signal
+    binned_signal = np.full([aman.dets.count, bins], np.nan)
+    binned_signal_squared_mean = np.full([aman.dets.count, bins], np.nan)
+    binned_signal_sigma = np.full([aman.dets.count, bins], np.nan)
+
+    # get mask from flags
+    if flags is None:
+        m = np.ones([aman.dets.count, aman.samps.count], dtype=bool)
+    else:
+        m = ~flags.mask()
```
mhasself, line 53: It is less RAM intensive, and thus possibly faster, to do this bool mask creation within the detector loop.

sotodlib/tod_ops/binning.py
```diff
@@ -0,0 +1,67 @@
...
+
+    # binning tod
+    for i in range(aman.dets.count):
+        binned_signal[i][mcnts] = np.histogram(bin_by[m[i]], bins=bins, range=bin_range,
+                                          weights=signal[i][m[i]])[0][mcnts] / bin_counts[mcnts]
```
mhasself, line 58: `bin_counts[mcnts]` is not the right divisor, if `m[i]` is non-trivial.

sotodlib/tod_ops/binning.py
```diff
@@ -0,0 +1,67 @@
...
+
+    Parameters
+    ----------
+    aman : TOD
+        The Axismanager object to be binned.
+    bin_by : array-like
+        The array by which signal is binned. It should has the same samps as aman.
+    signal : str, optional
+        The name of the signal to be binned. Defaults to aman.signal if not specified.
+    bin_range : list or None
```
mhasself, line 18: Why not call this `range`, to match numpy binning functions?

sotodlib/tod_ops/sss.py
```diff
@@ -0,0 +1,216 @@
...
+        az = aman.boresight.az
+    binning_dict = bin_signal(aman, bin_by=az, signal=signal,
+                               bin_range=bin_range, bins=bins, flags=flags)
+    return binning_dict
+
+def fit_sss(az, sss_stats, nmodes, fit_range=None):
+    """
+    Function for fitting legendre polynomials to signal binned in azimuth.
+
+    Paramaters
```
mhasself, line 51: Parameters

sotodlib/tod_ops/sss.py
```diff
@@ -0,0 +1,216 @@
+import numpy as np
```
mhasself, line 1: Add comment at the top about what this module is for.  Specifically State Something about what SSS stands for, cause that's not obvious to everyone.

I've seen SSS most often on my boarding pass, and it means I'm getting extra security checks.

sotodlib/tod_ops/sss.py
```diff
@@ -0,0 +1,216 @@
...
+        signal = aman.signal
+    if az is None:
+        az = aman.boresight.az
+    binning_dict = bin_signal(aman, bin_by=az, signal=signal,
+                               bin_range=bin_range, bins=bins, flags=flags)
+    return binning_dict
+
+def fit_sss(az, sss_stats, nmodes, fit_range=None):
+    """
+    Function for fitting legendre polynomials to signal binned in azimuth.
```
mhasself, line 49: I cannot help noticing that you carefully capitalized "Legendre" in your variable names, below, but not in the docstring.  That seems backwards!

sotodlib/tod_ops/sss.py
```diff
@@ -0,0 +1,216 @@
...
+
+    Paramaters
+    ----------
+    az: array-like
+        azimuth array from main axis manager
+    sss_stats: AxisManager
+        Axis manager containing binned signal and azimuth used for fitting.
+        Created by ``get_sss`` function.
+    nmodes: integer, required
+        Highest order legendre polynomial to include in the fit.
```
mhasself, line 59: Is this the number of modes, or the highest order?  Because those differ by 1.  Please clarify.  max_mode would be a better name for the second thing.

If it is the max_mode, remind the user somehow that max_mode+1 modes will be fit.

Personally I find "nmodes", representing the _number_ of modes, rather than the order of the highest mode, to be a much easier thing to deal with.

"required" is not neccessary.

sotodlib/tod_ops/sss.py
```diff
@@ -0,0 +1,216 @@
...
+    Returns
+    -------
+    sss_stats: AxisManager
+        Returns updated sss_stats with added fit information
+    model_sig_tod: array-like
+        Model fit for each detector size ndets x n_samps
+    """
+    bin_width = sss_stats.binned_az[1] - sss_stats.binned_az[0]
+    m = ~np.isnan(sss_stats.binned_signal[0]) # masks bins without counts
+    if np.count_nonzero(m) < nmodes:
```
mhasself, line 74: This is check is inaccurate if nmodes is the _order of the highest mode_.  You should fail on count(m) < nmodes + 1 (unless you change change the meaning of nmodes as discussed above).

sotodlib/tod_ops/sss.py
```diff
@@ -0,0 +1,216 @@
...
+        aman.wrap(sss_model_name, model_sig_tod, [(0, 'dets'), (1, 'samps')])
+    
+    return sss_stats, model_sig_tod
+
+def subtract_sss(aman, signal=None, sss_template=None,
+                 subtract_name='sss_remove'):
+    """
+    Subtract the scan synchronous signal (SSS) template from the
+    signal in the given axis manager.
+
```
mhasself, line 189: Clarify that this does not modify the signal array directly, but rather produces and returns a new array that results from subtracting the template from the provided signal.

I think it would be more useful to have a function (or an option in this function) that modifies the signal _in place_.  Remember that signal arrays are often big, and it is expensive to create and store copies of them.

docs/tod_ops.rst
```diff
@@ -172,3 +172,15 @@ an unit step, so it is acting as a matched filter.
 
 .. automodule:: sotodlib.tod_ops.jumps
    :members:
+
+
+
+tod_ops.sss
+=============
+
+Function for fitting legendre polynomials in az-vs-signal space. This has been used in ABS and other experiments to get a scan synchronous signal (SSS) template largely due to polarized ground signal to remove from the data.
```
mhasself, line 181: Please wrap this line to ~78 characters, like the rest of the docs.

sotodlib/hwp/hwp.py
```diff
@@ -203,48 +204,21 @@ def get_binned_signal(aman, signal=None, hwp_angle=None,
         * estimated sigma of binned signal
     """
     if signal is None:
-        signal = aman.signal
+        signal = aman.signal
```
mhasself, line 207: Cleanup stray whitespace

sotodlib/preprocess/processes.py
```diff
@@ -245,6 +245,23 @@ def process(self, aman, proc_aman):
         hwp.demod_tod(aman, **self.process_cfgs)
 
 
+class EstimateSSS(_Preprocess):
+    """Fits the amplitude of legendre polynomials in sig vs Az space.
+
+    .. autofunction:: sotodlib.tod_ops.sss.get_sss
```
mhasself, line 251: Define acronym SSS.

docs/tod_ops.rst
```diff
@@ -172,3 +172,15 @@ an unit step, so it is acting as a matched filter.
 
 .. automodule:: sotodlib.tod_ops.jumps
    :members:
+
+
```
mhasself, line 176: Can you add binning.py functions as autodocs somewhere in here, too?

mhasself (changes requested):
Lotta comments here ... especially in "binning.py", which I agree can be a useful general tool for different binning applications.

I know it's a big ask, but can we consider the wisdom of the acronym "SSS"?  I find that non-obvious ... to an outsider, it's opaque -- but even to an ~insider like me I struggle to remember whether the first S is Scan or Spin.  AzSS or ScanSS would be my alternative proposals.  I'm not going to insist on this change, just bringing it up.

sotodlib/tod_ops/sss.py
tterasaki: I think this style of function, get-model and subtract-model, is defined from preprocessing. It is used  also used in hwp synchronous signal derivation/subtraction. Maybe there is a way to satisfy both.

sotodlib/tod_ops/sss.py
tterasaki: Yeah, let's change it to azss (Azimuth synchronous signal).

BTW, there are 253 definitions for 'SSS', lol (https://www.abbreviations.com/SSS)

sotodlib/tod_ops/binning.py
tterasaki: You are right. I am working on this now.

tterasaki: Done.

sotodlib/tod_ops/binning.py
tterasaki: Done. Mask will be used only in the case of flags is not None.

sotodlib/tod_ops/sss.py
tterasaki: Modified.

sotodlib/tod_ops/sss.py
tterasaki: Replaced all SSS I found with AzSS.

sotodlib/tod_ops/sss.py
tterasaki: Done.

sotodlib/tod_ops/sss.py
tterasaki: I have added `subtract_in_place` optinon in get_azss. So user can just subtract sss with `subtract_in_place`=True and `merge_model`=False without increasing the size of data.

sotodlib/preprocess/processes.py
tterasaki: Added a description.

msilvafe:
> @msilvafe @mhasself OK, I have reflected all comments from Matthew. Can you check it?

I looked over this, and it all looks good! Thanks for @tterasaki. I ran the sphinx-build on the docs to see how they looked and added some minor things related to typos, missing returned parameters, formatting, etc.

sotodlib/preprocess/processes.py
```diff
@@ -245,22 +245,25 @@ def process(self, aman, proc_aman):
         hwp.demod_tod(aman, **self.process_cfgs)
 
 
-class EstimateSSS(_Preprocess):
-    """Fits the amplitude of legendre polynomials in sig vs Az space.
+class EstimateAZSS(_Preprocess):
```
mhasself, line 248: Did you mean "AZSS"?  I would have thought "AzSS".

sotodlib/tod_ops/azss.py
```diff
@@ -0,0 +1,225 @@
+"""Module for estimating Aziumth Synchronous Signal (azss)"""
```
mhasself, line 1: Aziumth -> Azimuth.

mhasself (approved):
Looks great!  I think class name "EstimateAzSS" would be more consistent with our other applications of PascalCase; I don't really like "EstimateAZSS".  But merging is in your hands now.

PR #474, opened 2023-05-23
Resample aman

msilvafe:
Added in submodule to `tod_ops` for resampling axis manager onto a new set of timestamps. This does not necessarily cover all general cases since there are lots of different data types in an axis manager, so I assume there will be some discussion of how this could be expanded or restructured. My main use case, however has been downsampling after hwp demodulation. I've tested this the most and wanted it to be PR'd so we could start discussing what we'd want to add before merging.

So far I tested on a toast sim: obs_id: `1692085774_ST1_f090_1000000` from this context: `'/home/tterasaki/workspace/sat_mapmake_debug/run3/context-hwpss/context.yaml'` in @tterasaki's workspace which is a sim link to some of @mhasself 's planet sims. Since I don't have any filtering in this module I ran this TOD through the preprocess module to get out demodulated Q/U then ran this new resample module to downsample to 2x the LPF cutoff from the demodulation (8 Hz). You can see the results here:

<p align="center">

</p>

The plot was generated in this notebook on simons1: `/home/msilvafe/notebooks/flp_dev/20230523_TOD_Processing_Demo.ipynb`

I'm sure @kmharrington will have additional thoughts for things to test/expand but my initial ones are:

1. Adding + testing this in preprocess
2. Writing a unit test script (that catches the exceptions for unsupported data types)
3. Testing this with decimation on an axis that is not samps (i.e a binned samples axis)
4. Testing an up-sampling option
5. Closer attention to how csr_arrays and other non-array style data types are handled. They're currently dropped, or a ValueError is raised depending on if it's a known datatype we're intentionally dropping (csr_arrays) or something I haven't tested before.

### DIFF
--- a/docs/tod_ops.rst
+++ b/docs/tod_ops.rst
@@ -172,3 +172,15 @@ an unit step, so it is acting as a matched filter.
 
 .. automodule:: sotodlib.tod_ops.jumps
    :members:
+
+tod_ops.resample
+================
+
+Support for resampling an axis manager to a new set of timestamps. Useful for downsampling
+or upsampling data. This does not include any filtering required to reduce aliasing in
+this step. Data must first be filtered with appropriate nyquist filter using the
+``sotodlib.tod_ops.fourier_filter`` module.
+
+.. automodule:: sotodlib.tod_ops.resample
+   :members:
+
--- a/sotodlib/tod_ops/resample.py
+++ b/sotodlib/tod_ops/resample.py
@@ -1,0 +1,126 @@
+import numpy as np
+from sotodlib import core
+from sotodlib.core import flagman as fm
+from so3g.proj.ranges import Ranges, RangesMatrix
+from scipy import interpolate
+
+try:
+    from scipy.sparse import csr_array
+except ImportError:
+    from scipy.sparse import csr_matrix as csr_array
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+def spl_int(t1, t0, y):
+    """
+    Wrapper for scipy.interpolate spline interpolation to take the same
+    arguments as np.interp
+    Args:
+    -----
+    t1 (ndarray): Timestamps to interpolate onto.
+    t0 (ndarray): Timestamps of the data to interpolate.
+    y (ndarray): Data to interpolate
+
+    Returns:
+    --------
+    Interpolated values.
+    """
+    tck = interpolate.splrep(t0, y)
+    return interpolate.splev(t1, tck)
+
+
+def interp_aman(aman, t0, t1, axis='samps', interp_type='linear'):
+    """
+    Function for interpolating an axis manager to some new set of timesamples.
+    This is useful for upsampling, downsampling, or resampling between two
+    datasets taken at similar rates. It handles interpolating numpy arrays,
+    and RangesMatrices but currently does not interpolate csr matrices
+    (it just drops these).
+
+    Args:
+    -----
+    aman (AxisManager): Axis manager to interpolate
+    t0 (ndarray): Timestamps of axis manager.
+    t2 (ndarray): Timestamps to interpolate onto.
+    axis (str): Name of axis to interpolate along (defaults to samps).
+                I have not tested setting thing to something else!!!
+    interp_type (str): Either 'linear' or 'spline' type interpolation.
+
+    Returns:
+    --------
+    dest (AxisManager): New axis manager interpolated to t1 timestamps.
+    """
+    new_axes = []
+    for k, v in aman._axes.items():
+        if k == axis:
+            new_axes.append(core.OffsetAxis(axis, len(t1)))
+        else:
+            new_axes.append(v)
+    dest = core.AxisManager(*new_axes)
+
+    for k, assign in aman._assignments.items():
+        if axis in assign:
+            if isinstance(aman[k], core.AxisManager):
+                dest.wrap(k, interp_aman(aman[k], t0, t1, axis=axis,
+                          interp_type=interp_type))
+            elif isinstance(aman[k], RangesMatrix):
+                shape = list(aman[k].shape)
+                dest.wrap_new(k, shape=shape, dtype=type(aman[k]))
+                dest[k] = RangesMatrix([Ranges.from_mask(
+                                       np.round(np.interp(t1, t0,
+                                        aman[k][det].mask()), 0).astype(bool))\
+                                       for det in range(aman[k].shape[0])])
+            elif isinstance(aman[k], csr_array):
+                logger.warning('csr matrix is not supported in downsampling.' +
+                               f' {k} is a csr_matrix and is being dropped ' +
+                               'from the returned axis manager')
+                continue
+            elif isinstance(aman[k], np.ndarray):
+                shape = list(aman[k].shape)
+                for i, a in enumerate(assign):
+                    if a is not None:
+                        shape[i] = a
+                dest.wrap_new(k, shape=shape, dtype=aman[k].dtype)
+                if len(shape) == 1:
+                    if interp_type == 'linear':
+                        dest[k] = np.interp(t1, t0, aman[k])
+                    if interp_type == 'spline':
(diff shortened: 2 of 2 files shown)


kmharrington:
Do you mean this as a draft or something ready for a full PR? 

I'm generally good with how this is set up but haven't tried any detailed tests yet, I'll note that I'm looking at this as "a first option" for resampling/downsampling because we're definitely going to want expanded options (of course, 1 option is infinity times better than zero options so major thanks for starting this). In particular for signal-type axes, FFT based decimation should be much faster but is only possible for limited ranges of samp rate 1 --> samp rate 2. I think the `decimate_aman` function is pretty generally named, so we might want to set that up assuming there will be additional available methods in the future.

I'd like to have more options for what to do with RangesMatricies. Being able to pass a flag for "any", "all", or the "avg" (what it looks like you're currently doing) would be awesome.  

Other note is I'd like all the `interp_aman` kwargs exposed for `decimate_aman` as well.

msilvafe:
> Do you mean this as a draft or something ready for a full PR?
> 
> I'm generally good with how this is set up but haven't tried any detailed tests yet, I'll note that I'm looking at this as "a first option" for resampling/downsampling because we're definitely going to want expanded options (of course, 1 option is infinity times better than zero options so major thanks for starting this). In particular for signal-type axes, FFT based decimation should be much faster but is only possible for limited ranges of samp rate 1 --> samp rate 2. I think the `decimate_aman` function is pretty generally named, so we might want to set that up assuming there will be additional available methods in the future.
> 
> I'd like to have more options for what to do with RangesMatricies. Being able to pass a flag for "any", "all", or the "avg" (what it looks like you're currently doing) would be awesome.
> 
> Other note is I'd like all the `interp_aman` kwargs exposed for `decimate_aman` as well.

To be honest I'm not 100% sure what the dividing line between draft and full PR is but I sort of thought that since we needed to add in at the least the preprocess process and unit test scripts this wasn't ready to merge so it wasn't ready for a full PR. It sounds like you're ok saving items 3, 4, & 5 from my initial PR comment for later so I think once we address:
1. Adding a process to preprocess module for decimation
2. Adding a unit test script
3. Pass the `interp_aman` args to `decimate_aman`
4. Modify the ranges matrix interpolation to include an "any", "all", or "avg" logic/argument for passing flags to the resampled axis manager.

then we can move this from draft -> full PR review. I'm not sure I will be able to do this in the short term. From my recent experience writing the unit test script takes a bit of thought and time, and I think getting the logic to work right for the modified `RangesMatrix` interpolation will also take some work and testing. So if any others wanna hop on this and add some commits to address these in the meantime I'm totally happy for that.

kmharrington:
@msilvafe  I went around and thought about the cut resampling for awhile. I've come to the conclusion that the "all" / "avg" versions of combining cuts probably aren't that useful or correct. So I've gone ahead and written a nicer version of Ranges resampling that won't have to go through the masking/averaging and have updated your function to use it.

sotodlib/core/flagman.py
```diff
@@ -278,6 +279,74 @@ def has_all_cut(flag):
...
+            if e == rng.shape[0]:
+                ne = new_rng.shape[0]
+            else:
+                ne = np.where( t1 >= t0[e] )[0]
+                if len(ne) > 0:
+                    ne = ne[0]
+                else:
+                    ne = new_rng.shape[0]
+                    
+            if ns == 0 and ne == 0:
```
msilvafe, line 330: @kmharrington, what is this catching? Should this maybe raise a warning?

kmharrington: it's catching flags that happen outside the range of t1.

tterasaki:
Sorry I did not respond to this for a long time. I have taken a look at this, and realized a problem. I mean the current implementation does not have a anti-alias filter.  And also, using interpolation for down-sampling is not good, that can distort the original signal.

I think those problems due to that the function has application for either up and down sampling. I definitely want split the resampling function into up-sampling and down-sampling. I will write some in another branch.

msilvafe:
> Sorry I did not respond to this for a long time. I have taken a look at this, and realized a problem. I mean the current implementation does not have a anti-alias filter. And also, using interpolation for down-sampling is not good, that can distort the original signal.
> 
> I think those problems due to that the function has application for either up and down sampling. I definitely want split the resampling function into up-sampling and down-sampling. I will write some in another branch.

Hey @tterasaki my intention for this to be a separate function from downsample or upsample. So we still want a downsample function which does the prefiltering (low pass) before resampling (that's why I called it resample not down or upsample). What is the issue you have with interpolations? I first tried to resample just using an integer number of samples (like the smurf downsampler where it just takes every 20th point or Nth point where N is the chosen decimation factor) but saw some weird issues with the timestamp sampling and found that interpolation onto an even grid of points worked better that's why I chose the interpolation onto an arbitrary set of input timestamps.

amaurea:
Matthew pointed me at this, since I was about to implement it myself.

I second @tterasaki's concern about downsampling. It's fine `interp_aman` if this is just meant to be one step in the full resampling process, though in that case we should probably make the full function very prominent and hide the interp one away as an implementation detail. Otherwise it's very easy to get something that seems like it's running correctly but actually ends up throwing away most of our statistical power. Personally I would just do the whole operation in one function.

I already have an implementation of fft_downsampling, where one passes in an array and a target length, and it does fourier space truncation to match that. This is equivalent to an antialiasing filter followed by fourier-interpolation, except that it's faster. How about something like your `interp_aman`, but which does `add_slope_back(fft_downsample(remove_slope(arr)))` for all floating-point-valued axis manager members with a sample dimension? The point of the slope stuff is to make the arrays approximately periodic so fourier methods work safely. Fourier downsampling doesn't work for variable sample spacing, though, so it's less general than the current function (but we'll surely want equi-spaced samples in practice anyway).

Or, if the signal will be the only noisy array in practice, maybe it should just be given special treatment.

ykyohei:
I heard about this PR and tested to use it, and I didn't find problems so far.
I think this is very useful for down sampling the demodulated and low pass (anti-alias) filtered aman.
The proposed fft_downsampling sounds great, but we can implement it in a separate PR?
I would like to have this in master branch.

msilvafe:
> I heard about this PR and tested to use it, and I didn't find problems so far.
> 
> I think this is very useful for down sampling the demodulated and low pass (anti-alias) filtered aman.
> 
> The proposed fft_downsampling sounds great, but we can implement it in a separate PR?
> 
> I would like to have this in master branch.

I'm ok with that. If you've tested it and are ok with it as is perhaps convert from draft to full and request review.

iparask:
@msilvafe how is your code different from this downsampling: https://github.com/simonsobs/sotodlib/blob/master/sotodlib/mapmaking/utilities.py#L481?

There are a bunch of downsampling functions in that file.

If all functions produce the same result for downsampling, I think moving the more general ones to a `sotodlib.utils` and use those in both cases.

msilvafe:
> @msilvafe how is your code different from this downsampling: https://github.com/simonsobs/sotodlib/blob/master/sotodlib/mapmaking/utilities.py#L481?
> 
> There are a bunch of downsampling functions in that file.
> 
> If all functions produce the same result for downsampling, I think moving the more general ones to a `sotodlib.utils` and use those in both cases will reduce code duplication.

This is a general resample operation that can be used for downsampling or upsampling or resampling on to a different set of samples and handles all of the datatypes and sub axis managers within our main axis manager. The function you link to in the mapmaking module (which I've never seen before because I typically don't venture into that part of sotodlib) is making a new axis manager which has only a few essential axes aligned along samples (signal, boresight, and some subset of the flags) downsampled using different methods for the flags + signal downsampling.

I agree we could unify things here but I think this is not the only place where the mapmaking module has code that overlaps with code in the rest of the repo so I was hoping that we could avoid this holding us up. @ykyohei asked to try to get something that works now for downsampling so I suggested he tested this out and make whatever required changes to get it to work for him and I think this was his attempt to push this forward. But maybe you're suggesting we don't merge this until we unify those things. If so, I can work on that with Kyohei but I think we'll need to get buy in from Sigurd on whatever common fn's we write.

PR #475, opened 2023-05-23
Restore backward compatibility on Tuneset naming

kmharrington:
I missed that these queries would not be backwards compatible for older databases.


jseibert575 (approved):
Looks good, tested on the previously failing observation and it seems to work! Thanks, Katie!

PR #476, opened 2023-05-24
Gapfill wrapper

msilvafe:
I'm adding a wrapper for the `tod_ops.gapfill` module that @bkelle and I have been using in our testing of the LAT + SAT preprocessing. I also added in a unit test script, and process to preprocess module. I tested the wrapper function on its own and running through preprocess. My own testing was following the same test in the unit test script which generates an axis manager with all of the detectors showing the same polynomial signal but with different glitches in each detector. Currently the polynomial gap filler is working very well for this test case but the pca model filling is failing so for now I have just commented out that test in the unit test script.
| Testing Function w/ Polynomial Filler  | Testing Function w/ PCA Filler  | Testing running through preprocess module |
| ------------- | ------------- | ------------- |
| gapfill_demo_poly | gapfill_demo_pca  | gapfill_demo_poly_preprocess |
| Glitch filling works well using polynomial filler. | PCA filling failing, @mhasself suggests poly filling first before running the PCA model fill but this returns a `TypeError` when run with the simple AxisManager that can be constructed as is done in the unit test script.  | Polynomial filler used in the preprocess module works well. |

All of the plots here and testing can be found in a notebook on simons1: `/home/msilvafe/notebooks/flp_dev/20230524_Test_Gapfill.ipynb` and the preprocess configs can also be found there.


msilvafe:
@mhasself figured out that to avoid the poor gap filling from the pca module we first needed to 1) gapfill with the poly gap filler 2) run the pca 3) gap fill with the pca model. So I changed to that. Now both methods are working well:

| Testing Function w/ Polynomial Filler  | Testing Function w/ PCA Filler  |
| ------------- | ------------- |
| gapfill_demo_poly_post_fix | gapfill_demo_pca_post_fix |
| Glitch filling works well using polynomial filler. | Glitch filling works well using PCA filler. |

I was also getting a TypeError when running the pca gap filling because with all of the signals the same the pca would return complex eigenvalues, which couldn't be added to regular float64. So I updated the pca module to only return real eigenvalues per Matthew's suggestion. So everything should be working and ready for review post commit [f82eb6b](https://github.com/simonsobs/sotodlib/pull/476/commits/f82eb6b268732c795716196f6ba632a9af50cb8f).

sotodlib/tod_ops/gapfill.py
```diff
@@ -1,6 +1,9 @@
 import numpy as np
 import so3g
+from sotodlib.tod_ops import pca
```
mhasself, line 3: Use relative references for internal imports, `from . import pca`.

sotodlib/tod_ops/gapfill.py
```diff
@@ -380,3 +383,69 @@ def get_contaminated_ranges(good_flags, bad_flags):
...
+        Returns ndarray with gaps filled from input signal.
+    """
+    if signal is None:
+        sig = np.copy(aman.signal)
+    else:
+        sig = np.copy(signal)
+    
+    if glitch_flags is None:
+        glitch_flags = aman.flags.glitches
+    if use_pca is True:
```
mhasself, line 427: Remove " is True"

sotodlib/tod_ops/gapfill.py
```diff
@@ -380,3 +383,69 @@ def get_contaminated_ranges(good_flags, bad_flags):
...
+        # fill with poly fill before PCA
+        gaps = get_gap_fill(aman, nbuf=nbuf, flags=glitch_flags,
+                            signal=np.float32(sig))
+        sig = gaps.swap(aman, signal=sig)
+        # PCA fill
+        mod = pca.get_pca_model(tod=aman, n_modes=modes,
+                                signal=sig)
+        gfill = get_gap_model(tod=aman, model=mod, flags=glitch_flags)
+        sig = gfill.swap(aman, signal=sig)
+    if use_pca is False:
```
mhasself, line 442: Not "else:"?

sotodlib/tod_ops/gapfill.py
```diff
@@ -380,3 +383,69 @@ def get_contaminated_ranges(good_flags, bad_flags):
                 rs.add_interval(int(i0), int(i1))
     return contam
 
+
+def fill_glitches(aman, nbuf=10, use_pca=False, modes=3, signal=None,
+                  glitch_flags=None, merge_filled=True,
+                  filled_name='gap_filled'):
+    """
+    Top level function for filling glitches assuming glitch flags exist.
+    Wraps the other functions in the ``tod_ops.gapfill`` module.
```
mhasself, line 392: "assuming glitch flags exist" -> "using pre-computed glitch flags provided by the caller."
"Top level function for filling glitches"  -> "Fill glitches in time-ordered data"

Mention that it uses polynomial filling, by default, but can use PCA optionally.

This function fills glitches, using either a polynomial or PCA-based approach.

sotodlib/tod_ops/gapfill.py
```diff
@@ -380,3 +383,69 @@ def get_contaminated_ranges(good_flags, bad_flags):
...
+    signal : ndarray or None
+        Array of data to fill glitches in. If None then uses ``aman.signal``.
+        Default is None.
+    glitch_flags : RangesMatrix or None
+        RangesMatrix containing flags to use for gap filling. If None then
+        uses ``aman.flags.glitches``.
+    merge_filled : bool
+        Merge gap filled data into input axis manager.
+    filled_name : str
+        Axis name to merge gap filled data if merge_filled is True.
```
mhasself, line 413: In other places we have used "wrap" to support the thing these two arguments are doing.  E.g.:

wrap=False -> no, don't merge it just return it
wrap=True -> yes, merge the new array and call it the default thing ("gap_filled" apparently)
wrap="my_other-name" -> yes, merge the new array and cal it "my_other_name"

Can the preprocess system cope with that setup?  Can you use generic "wrap" instead of specific "merge_filled, filled_name"?

sotodlib/tod_ops/gapfill.py
```diff
@@ -380,3 +383,69 @@ def get_contaminated_ranges(good_flags, bad_flags):
...
+        sig = gaps.swap(aman, signal=sig)
+        # PCA fill
+        mod = pca.get_pca_model(tod=aman, n_modes=modes,
+                                signal=sig)
+        gfill = get_gap_model(tod=aman, model=mod, flags=glitch_flags)
+        sig = gfill.swap(aman, signal=sig)
+    if use_pca is False:
+        gaps = get_gap_fill(aman, nbuf=nbuf, flags=glitch_flags,
+                            signal=np.float32(sig))
+        sig = gaps.swap(aman, signal=sig)
```
mhasself, line 445: This bit, for polynomial filling, is the same code that is run in the PCA block.  So you could reduce code duplication and eliminate the `else` by doing:
```
    [process args]
    [polyfill]
    if use_pca:
       [pcafill]
    [wrap and return]
```

sotodlib/tod_ops/gapfill.py
```diff
@@ -380,3 +383,69 @@ def get_contaminated_ranges(good_flags, bad_flags):
...
+        glitch_flags = aman.flags.glitches
+    if use_pca is True:
+        if modes > aman.dets.count:
+            logger.warning(f'modes = {modes} > number of detectors = ' +
+                           f'{aman.dets.count}, setting modes = number of ' +
+                           'detectors')
+            modes = aman.dets.count
+        # fill with poly fill before PCA
+        gaps = get_gap_fill(aman, nbuf=nbuf, flags=glitch_flags,
+                            signal=np.float32(sig))
```
mhasself, line 435: It is annoying this needs to be converted to float32 ... will create an issue to resolve that.

mhasself (changes requested):
Thanks for this -- will be very useful.

This PR seems to have mixed in resample.py as well.  I thought that was intended to be in a separate PR, so I've not reviewed it for now.

tests/test_tod_ops.py
```diff
@@ -121,6 +153,28 @@ def test_basic(self):
...
+                                                      wrap=mg)
+        self.assertTrue(np.max(np.abs(glitch_filled-aman.inputsignal)) < 1e-3)
+
+        # test pca fill
+        up, mg = True, False
+        glitch_filled = tod_ops.gapfill.fill_glitches(aman, use_pca=up,
+                                                      wrap=mg)
+        print(np.max(np.abs(glitch_filled-aman.inputsignal)))
+
+        # test in_place
```
mhasself, line 173: This isn't really "in_place".  "in_place" would mean that tod.signal is modified directly, without creating a whole new vector for glitch_filled.  Also, there's no harm in return glitch_filled if it was also wrapped.

sotodlib/preprocess/processes.py
```diff
@@ -244,6 +244,35 @@ class Demodulate(_Preprocess):
     def process(self, aman, proc_aman):
         hwp.demod_tod(aman, **self.process_cfgs)
 
+
+class Glitchfill(_Preprocess):
```
mhasself, line 248: Sorry, nitpicking now -- looks like the naming convention for preprocess modules is "ThingAction", so I would think this would be GlitchFill.

mhasself (approved):
A couple minor things (but approving pending those fixes).  Thanks for this, and for including tests.

tests/test_tod_ops.py
msilvafe: Do you want me to make an in_place option or just change the comment to "test wrap new field"? Would you prefer I always return glitch_filled so instead of checking if the returned value is None here instead check if the new field exists in the axis manager?

mhasself: Just change the comment.

Yes, I think it's better to always return glitch_filled, and instead check that the correct new field was created in the axisman.

PR #477, opened 2023-05-25
Another metadata example

mhasself:
Basic example of including time-ordered data vectors as metadata.


kmharrington (approved):
yay more documentation!

PR #478, opened 2023-05-26
t_glitch argument wasn't used

msilvafe:
This is a simple 1-line change I found when reviewing how our glitch detection works for my thesis. In `flags.get_glitch_flags`, the `t_glitch` argument wasn't used, and 0.002 was hardcoded into the gaussian filter.


kmharrington (approved):
oops! Thanks for catching this

PR #480, opened 2023-05-26
Trending slopes

skhrg:
As discussed on slack this uses slope instead of max-min to find trending detectors.
The value for `max_trend` is from a quick look at some data from last year. As I look at the data from the most recent LATR cooldown I will refine the default.

### DIFF
--- a/sotodlib/flags.py
+++ b/sotodlib/flags.py
@@ -13,8 +13,10 @@ from . import core
 from .tod_ops import filters
 from .tod_ops import fourier_filter
 
-def get_turnaround_flags(tod, qlim=1, az=None, merge=True,
-                         overwrite=False,name='turnarounds'):
+
+def get_turnaround_flags(
+    tod, qlim=1, az=None, merge=True, overwrite=False, name="turnarounds"
+):
     """Flag the scan turnaround times.
 
     Args:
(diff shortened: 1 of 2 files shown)


sotodlib/flags.py
```diff
@@ -82,115 +85,136 @@ def get_glitch_flags(
...
+            s = s[:, ::n]
+        _slopes = ((t * s).mean(axis=1) - t.mean() * s.mean(axis=1)) / (
+            (t**2).mean() - (t.mean()) ** 2
+        )
+        cut = np.hstack(
+            (cut, np.tile((np.abs(_slopes) > max_trend)[..., np.newaxis], samps))
+        )
+        if full_output:
+            slopes = np.hstack((slopes, _slopes[..., np.newaxis]))
+            samp_edges.append(samp_edges[-1] + len(t))
```
kmharrington, line 197: `len(t)` needs to be `samps` here in case the array has been cut down

sotodlib/flags.py
```diff
@@ -82,115 +85,136 @@ def get_glitch_flags(
...
+
+    slopes = np.zeros((len(signal), 0))
+    cut = np.zeros((len(signal), 0), dtype=bool)
+    samp_edges = [0]
+    for t, s in zip(
+        np.array_split(timestamps, n_pieces), np.array_split(signal, n_pieces, 1)
+    ):
+        samps = len(t)
+        # Cheap downsampling
+        if len(t) < max_samples:
```
kmharrington, line 185: this should be if samps > max_samples. We don't need to down sample if the array is shorter

skhrg: Good catch

kmharrington (changes requested):
The two inline comments look like it wasn't fully tested for cases where the arrays are longer or shorter than max samples. Those definitely need to be fixed. 

While we're at it, could I ask you to write test cases for this function? We have zero coverage here and I had to stare at this enough that it might be a good place to start it.

skhrg:
Is there a specific place I should put the test? I don't see a `test_flags.py`

kmharrington:
yeah go ahead and make it!

skhrg:
@kmharrington  let me know if you want the test to cover a wider scope of cases.

kmharrington (approved):
Looks pretty good to me. One more test case would be running with n_pieces > 1. Make sure that all gets the right things as well.

PR #481, opened 2023-05-31
update_obsdb.py

AERAdler:
Script that updates a database of books, takes in command line arguments base_dir from which to search, recency (in days), book_type (obs/oper) and a config file.


sotodlib/site_pipeline/update_obsdb.py
```diff
@@ -0,0 +1,160 @@
...
+    if "obsdb" in configs:
+        bookcartobsdb.to_file(configs["obsdb"])
+    else:
+        bookcartobsdb.to_file("obsdb_from{}_to{}".format(tback, tnow))
+
+
+def get_parser(parser=None):
+    if parser is None:
+        parser = argparse.ArgumentParser()
+    parser.add_argument('--base_dir', type=str,
```
kmharrington, line 142: At a minimum we need base-dir to be a required argument, because we can't add things without it. But I think I would actually prefer for base-dir to be something specified in the config file because it shouldn't change once we've set up databases.

sotodlib/site_pipeline/update_obsdb.py
```diff
@@ -0,0 +1,160 @@
...
+    parser.add_argument("--config", help="ObsDb configuration file",
+        default="config.yaml", type=str)
+    parser.add_argument('--recency', default=2, type=float,
+        help="Days to subtract from now to set as minimum ctime")
+    parser.add_argument("--verbosity", default=2, type=int,
+        help="Increase output verbosity. 0:Error, 1:Warning, 2:Info(default), 3:Debug")
+    parser.add_argument("--booktype", default="both", type=str,
+        help="Select book type to look for: obs, oper, both(default)")
+    return parser
+def main():
```
kmharrington, line 153: empty line between end of function and start of new one.

sotodlib/site_pipeline/update_obsdb.py
```diff
@@ -0,0 +1,160 @@
...
+    if booktype=="both":
+        accept_type = ["obs", "oper"]
+    else:
+        accept_type = [booktype]
+
+    configs = yaml.safe_load(open(config, "r"))
+    if "obsdb" in configs:
+        if os.path.isfile(configs["obsdb"]):
+            bookcartobsdb = ObsDb.from_file(configs["obsdb"])
+    if "cols" in configs:
```
kmharrington, line 68: could we rename "cols" to "obsdb_cols"? This way it's a bit more obvious in a config file where we're defining both obsdb and obsfiledb parameters

sotodlib/site_pipeline/update_obsdb.py
```diff
@@ -0,0 +1,160 @@
...
+            The extra files give us a bit of trouble. We'll assume that index,
+            book, and bookbinder_log are always here, and edit our config file
+            to cover the rest as extra files.
+            """
+            book_file_list = os.listdir(bookpath)
+            book_file_list.remove("M_index.yaml")
+            book_file_list.remove("M_book.yaml")
+            book_file_list.remove("Z_bookbinder_log.txt")
+            book_file_list = [file_name for file_name in book_file_list if not file_name.endswith(".g3")]
+            if len(book_file_list)>0:
```
kmharrington, line 99: I think I understand why this is here. It's because the `oper` books have extra files. @mhasself  I'd rather update `check_book` to be a bit nicer about the presence of extra files in the books. Can we do that?

sotodlib/site_pipeline/update_obsdb.py
```diff
@@ -0,0 +1,160 @@
...
+           
+        else:
+            bookcart.remove(bookpath)
+    if "obsdb" in configs:
+        bookcartobsdb.to_file(configs["obsdb"])
+    else:
+        bookcartobsdb.to_file("obsdb_from{}_to{}".format(tback, tnow))
+
+
+def get_parser(parser=None):
```
kmharrington, line 139: I realize, your parser is written to have only optional arguments, but configs should really be non-optional. (update-g3tsmurf should have examples of both).

sotodlib/site_pipeline/update_obsdb.py
```diff
@@ -0,0 +1,160 @@
...
+    if parser is None:
+        parser = argparse.ArgumentParser()
+    parser.add_argument('--base_dir', type=str, 
+        help="base directory from which to look for books")
+    parser.add_argument("--config", help="ObsDb configuration file",
+        default="config.yaml", type=str)
+    parser.add_argument('--recency', default=2, type=float,
```
kmharrington, line 146: I think we'll want to expand the "time range of searching" in the future but for this PR I'd just like to start with this one but make it optional. (have this be optional with default=None and have None be the default for the keyword argument above. If recency is None we then search the whole folder system.

sotodlib/site_pipeline/update_obsdb.py
```diff
@@ -0,0 +1,160 @@
+"""update_obsdb
+The config file could be of the form:
+.. code-block:: yaml
```
kmharrington, line 3: I ended up needing to expand this a bit to work with the obsdb + obsfiledb together. Could you expand this example to this:

```
obsdb: obsdb.sqlite

obsfiledb: obsfiledb.sqlite
root_path: '/'

extra_files: ['M_index.yaml', 'M_book.yaml']
extra_extra_files: ['Z_bookbinder_log.txt']

cols:
  start_time: float
  stop_time: float
  n_samples: int
  telescope: str
  tube_slot: str
  type: str
  subtype: str
```

sotodlib/site_pipeline/update_obsdb.py
```diff
@@ -0,0 +1,160 @@
...
+                 recency=2., 
+                 verbosity=2, 
+                 booktype="both"):
+    """
+        Create or update an obsdb for observation or operations data.
+    Argument
+    ----------
+    base_dir : str
+        The base directory in which to look for books
+    Parameters
```
kmharrington, line 47: you don't need the parameters heading if you have the argument heading

sotodlib/site_pipeline/update_obsdb.py
```diff
@@ -0,0 +1,160 @@
...
+    ----------
+    config : str
+        Path to config file
+    recency : float
+        How far back in time to look for databases, in days. (default: 2.)
+    booktype : str
```
kmharrington, line 53: missing verbosity

kmharrington (changes requested):
Ok this is looking really great and my comments (all in-line) should be some fast changes. I do expect we'll expand this to even more functionality in the future but I'd prefer to get this working version in and then build from there.

I went over and tested this on the end-to-end system (which I realize you don't have access to but we can get you access to if you want it). I was able to build a `config.yaml` file that nicely worked to create databases over there and then tested them with a context/loader. All good!

I like the configurable columns, I tested adding some of them in.

sotodlib/site_pipeline/update_obsdb.py
AERAdler: You're right, I'll put it in the config file

sotodlib/site_pipeline/update_obsdb.py
AERAdler: The problem actually arose when one `obs` book had an extra file called "frame_splits.txt". But you're right that check_book is unable to ignore extra files like this.

AERAdler:
It should be done now, i fixed the "won't run twice" bug, rewrote the check book call to reflect the changes to that script and included Matthew's request for timestamp and duration columns.

sotodlib/site_pipeline/update_obsdb.py
```diff
@@ -0,0 +1,170 @@
...
+                bookcartobsdb.add_obs_columns(col_list)
+
+            start = index.get("start_time")
+            end = index.get("end_time") 
+            if None not in [start, end]:
+                bookcartobsdb.add_obs_columns(["timestamp float", "duration float"])
+                very_clean["timestamp"] = start
+                very_clean["duration"] = end - start
+            
+            if tags != ([] or [""]):
```
mhasself, line 137: That's probably not what you want ... I think you want `if tags != [] and tags != [""]`.  Or perhaps `if (len([t for t in tags if t != '']) > 0`

AERAdler: Good catch I'll use the `and` formulation.

kmharrington (approved):
I realize I had not come back and approved this. Let's do it!

PR #483, opened 2023-06-01
update_smurf_caldbs

jseibert575:
I'm adding the update_smurf_caldbs script here, with bias group maps and IVs added, like we discussed on today's data-flow call. The bias steps I'm holding out just while I figure out whether or not I should be adding all the files I am (should be fast). I built all of these using the SAT data from P10R2, and an example context file can be found at `'/home/jseibert/pipeline-dev/update-smurf-caldbs/context.yaml'`. Similarly, a config file can be found at `'/homejseibert/pipeline-dev/update-smurf-caldbs/config.yaml'`. Right now, there is a bug with the IV data when actually loading into axis managers. I'm working on that now, but I didn't want to delay this any farther, since I've pushed this back so far. I'll commit a change if I can figure out the problem, but it should be working for bias group maps...

Marking this as draft for now, but will convert when done.


kmharrington:
implemented elsewhere

PR #484, opened 2023-06-02
check-book: settings to tolerate arbitrary extra files in book

mhasself:
For oper books, especially, there will be ancillary files and sometimes analysis output that needs to be accepted.  You can still explicitly ban files that match some pattern, such as frame_splits.txt...


kmharrington (approved):
Thanks!

PR #485, opened 2023-06-02
Updated focal plane coords to newer 2d interp

skhrg:
Turns out it had been a long time since I updated scipy and `interp2d` is old news. This is just an update to the newer `bisplrep` and `bisplev` functions.


skhrg:
Piggybacking on this is a change that closes #487

mhasself (approved):

PR #486, opened 2023-06-05
Wg dev

atakeuchi634:
ver1 of wire grid analysis code
- need to change the format of output from cvs to hdf
- need to determine how to identify detectors
This is a temporal review. Other reviewers will be added or new pull request will be created later.


sotodlib/wiregrid/wiregrid_analysis.py
```diff
@@ -0,0 +1,308 @@
+import numpy as np
+import matplotlib.pyplot as plt 
+import pandas as pd
+import csv 
+
+from iminuit import Minuit, cost
```
tterasaki, line 6: I do not know sotodlib use the iminuit. I'll check it.

PR #491, opened 2023-06-09
Finalize focal plane

skhrg:
First pass at the core functionality of finalize focal plane.

Two main things that I need to take care of before it is ready for a full review:
1. There is no gamma right now. Before I add support for it I need to add support for the physical optics stuff to also give me a nominal gamma. Also want to verify that it is ok to use the measured polangs as gamma (or if they need some coordinate transform).
2. Need to figure out how to index multi observation products. This is also true of `make_position_match`. 

Happy to take any feedback on my current approach in the meanwhile.


sotodlib/site_pipeline/finalize_focal_plane.py
```diff
@@ -0,0 +1,342 @@
...
+    if np.isclose(scale, np.pi / 180.0).any() or np.isclose(scale, 180.0 / np.pi).any():
+        logger.warning(
+            (
+                "Scale factor looks like a deg/rad conversion."
+                " Someone may have used the wrong units somewhere."
+            )
+        )
+
+    if "plot" in config and config["plot"]:
+        _mk_plot(nominal, measured, affine, shift)
```
mhasself, line 331: Add ability to save plot to disk!

sotodlib/site_pipeline/finalize_focal_plane.py
```diff
@@ -0,0 +1,342 @@
...
+        nominal = nominal[:2]
+        measured = measured[:2]
+    affine, shift = get_affine(np.vstack(nominal), np.vstack(measured))
+    scale, shear, rot = decompose_affine(affine)
+    rot = decompose_rotation(rot)
+
+    if np.isclose(scale, np.pi / 180.0).any() or np.isclose(scale, 180.0 / np.pi).any():
+        logger.warning(
+            (
+                "Scale factor looks like a deg/rad conversion."
```
mhasself, line 325: In messages like this is helpful to print out the value you triggered on.

In this program though, you should probably print out the scale / shear / rot regardless.

docs/site_pipeline.rst
```diff
@@ -177,6 +177,104 @@ including amplitude and FWHM.
...
+    - "path_to_h5"
+  # These should be one per provided result
+  # If a given has context setup to load this information set the value to null
+  # Order is the ones from obs_id first, then query, then multi_obs
+  detmaps:
+      - null
+      - "path_to_detmap"
+  
+  # Configuration options
+  use_matched: False # Set to True to use the mapping from make_position_match
```
mhasself, line 218: Please explain what use_matched does in more detail (perhaps not in the comments but rather in the main text of the help).  I.e. it takes some important parameter (the det_id) from X instead of from Y.

mhasself (changes requested):
Partial review; there's a lot in here ... I focused on finalize_focal_plane.py.  Mostly documentation concerns.

I think perhaps coords/focal_plane.py should actually be called coords/optics.py.

Main comment on the approach is:  if I read things correctly, you load a bunch of measured positions, and then do per-detector medians to get positions.  I am concerned that systematic shifts in the whole focal plane, from one obs to the next, will mess up this averaging approach.  (The median sort of doesn't care about that, _unless_ different detectors are missing from one observation to the next.)

I think a safer approach is to use some template to measure a basic shift (and _optionally_ a rot + shear) for each contributing observation, and include that shift before performing your per-detector reduction.

skhrg:
A lot of the inline comments have to do with dimensionality so I am writing this to address this in general.
Most of this code treats `gamma` as  third dimension that forms an orthogonal basis with `xi` and `eta`. I recognize that `gamma` is actually the rotation of the detector rather than a sort of `z` axis but I don't see a reason why I can't treat it as such. If that isn't true let me know.

The consequence of this is if we have information to determine `gamma` the affine matrix (`A`) is 3x3 so we end up with 3 rotation angles, 3 scale factors, 3 shear params, and a shift in 3 dimensions.

The shift in `gamma` is the one that is most useful and meaningful in the most obvious way.
The `scale` in gamma should nominally be 1, but it can capture issues like angles being passed in degrees instead of radians.
The rotation about the `gamma` axis is the rotation of the `xi-eta` plane which is the one we care about most. The other two rotations (of the `xi-gamma` and `eta-gamma` planes) aren't particularly meaningful physically but if they aren't ~0 then it could be an indicator that something went wrong.

As you noted, even though the inputs should be limited to 3 dimensions some of these functions are written to work in n dimensions. I did this for three reasons:
1. I think the math is forced to be cleaner in the general case (look at the older version of the decomposition function where I was explicitly in 2 dimensions)
2. Even though I expect in the field to have all the inputs, right now I don't have `gamma` for everything I use to test so I at least need things that work in both 2d and 3d
3. I don't know what we would use them for, but in case we want these sorts of transformations for other use cases it is nice for them to be as general as possible.

mhasself:
> A lot of the inline comments have to do with dimensionality so I am writing this to address this in general. Most of this code treats `gamma` as third dimension that forms an orthogonal basis with `xi` and `eta`. I recognize that `gamma` is actually the rotation of the detector rather than a sort of `z` axis but I don't see a reason why I can't treat it as such. If that isn't true let me know.

I think gamma shouldn't ever be mixed in like this.  For one thing, its values live on a compact range, 0-360 deg, and the linalg doesn't know that.  So if it tries to push something at 359 deg -> 361 deg, that exerts undue pressue on the 1 deg it's comparing against.  For another thing, there's no physical meaning behind "rotation" from xi or eta into gamma -- you would never want to allow that freedom, because it doesn't tell you anything physical.  Third, we'd actually expect correlations in the transformation acting on (xi, eta) and the corresponding transformation on gamma; as (xi, eta) rotates into (xi', eta') by angle h, we'd expect gamma to change to gamma' = gamma + h (or -h or something).

> 
> The consequence of this is if we have information to determine `gamma` the affine matrix (`A`) is 3x3 so we end up with 3 rotation angles, 3 scale factors, 3 shear params, and a shift in 3 dimensions.
> 
> The shift in `gamma` is the one that is most useful and meaningful in the most obvious way. The `scale` in gamma should nominally be 1, but it can capture issues like angles being passed in degrees instead of radians. The rotation about the `gamma` axis is the rotation of the `xi-eta` plane which is the one we care about most. The other two rotations (of the `xi-gamma` and `eta-gamma` planes) aren't particularly meaningful physically but if they aren't ~0 then it could be an indicator that something went wrong.

All the shift / scale stuff can be studied on gamma directly, without mixing it in with xi, eta.  Furthermore, you need to account for the periodicity issue or that will confuse your transformation.  I really thing xi-gamma and eta-gamma don't tell you anything.  If they are non-zero, it is probably telling you that you've permitted weird degrees of freedom and your data aren't quite uncorrelated enough to average them out to zero.  This could happen when dealing with a small number of measurements, for example.

> As you noted, even though the inputs should be limited to 3 dimensions some of these functions are written to work in n dimensions. I did this for three reasons:
> 
>     1. I think the math is forced to be cleaner in the general case (look at the older version of the decomposition function where I was explicitly in 2 dimensions)
> 
>     2. Even though I expect in the field to have all the inputs, right now I don't have `gamma` for everything I use to test so I at least need things that work in both 2d and 3d
> 
>     3. I don't know what we would use them for, but in case we want these sorts of transformations for other use cases it is nice for them to be as general as possible.

On 1 & 3 -- ya, go for it.  Maybe put these into the library (focal_plane.py / optics.py) then.  They are generically pretty useful.  On (2) ... see above.  I'm going to go back and read the gamma handling bit more carefully.  Early on all we'll be able to do is transform our _expected_ gamma angles, from the template, into xi-eta compatible coordinates relative to the measured detector positions.  That is an important thing that needs to be supported and should be possible here (if it isn't already).

skhrg:
> I think gamma shouldn't ever be mixed in like this. For one thing, its values live on a compact range, 0-360 deg, and the linalg doesn't know that. So if it tries to push something at 359 deg -> 361 deg, that exerts undue pressue on the 1 deg it's comparing against. For another thing, there's no physical meaning behind "rotation" from xi or eta into gamma -- you would never want to allow that freedom, because it doesn't tell you anything physical. Third, we'd actually expect correlations in the transformation acting on (xi, eta) and the corresponding transformation on gamma; as (xi, eta) rotates into (xi', eta') by angle h, we'd expect gamma to change to gamma' = gamma + h (or -h or something).

Yeah this is a good point. It is pretty straight forward for me to compute a seperate gamma transform (I think just shift and scale).

With that now instead of 3 shifts, 3 scales, 3 rotations, and 3 shears I will instead have 3 shifts, 3 scales, 1 rotation, and 1 shear. So the 4x3 table I was storing things in before now is jagged which I am not a fan of. Do you think it might make more sense to instead save these params as attributes in the h5 file? I went with the table before because I figured that a `ResultsSet` would be easier to load since we already have the machinery for it but I can just include a function that will load the attrs for you as a convenience thing.

skhrg:
> I think a safer approach is to use some template to measure a basic shift (and _optionally_ a rot + shear) for each contributing observation, and include that shift before performing your per-detector reduction.

If I do this won't my final shift always be 0? Since I am essentially just centering every measurement on the nominal pointing in that case. I could add in a shift that puts all of the measurements at some fiducial center but I think that is equivalent to averaging right?

I could compute the transformation for each measurement individually and then combine those but I do not know how one would average transformation matrices.

mhasself (approved):
Hooray!

skhrg:
Hooray!

Issue #490, opened 2023-06-09
Branch-unsafe code in sim_sso


mhasself:
When subtracting longitude angles representing on sky positions, it is essential to put the difference into the range (-pi, pi) before doing anything with it.

https://github.com/simonsobs/sotodlib/blob/4922b42323d02be6a38103d0284695c056b1c22a/sotodlib/toast/ops/sim_sso.py#L440

The symptom was that Jupiter signal didn't show up in some planet sims in some cases.  It took a long time to trace this.

The x= daz*cos(el) and y = del results are used to look up the beam shape.  In SO, focal planes are quite large, and the telescopes can point near zenith.  The (daz * cos(el), del) approximation has significant curvature as el approaches 90 degrees.  So I think this code should be rewritten to use (xi, eta) coords instead of (daz * cos el, el).  The route via quaternions (it appears qarray has everything you'd need for this) will make the az branch irrelevant.

mhasself:
Main problem was addressed in #492 (in sso and a couple other places).

PR #492, opened 2023-06-09
Pick the correct branch for source-centric coordinates

keskitalo:
Address #490 .


mhasself (changes requested):
Thanks for the quick attention, but you need, in general:
```
x = (x + np.pi) % (2 * np.pi) - np.pi
```
Furthermore you can't do the branch fixing _after_ scaling by cos(el), you need to do it before.

mhasself (approved):
Thanks!

PR #494, opened 2023-06-14
hk book types

guanyilun:
Example M_index file from a hk book:
```
book_id: hk_16866_ocs3
start_time: 1686600000.0
stop_time: 1686699999.0
telescope: ocs
tube_slot: '3'
type: hk
```

### DIFF
--- a/sotodlib/io/imprinter.py
+++ b/sotodlib/io/imprinter.py
@@ -355,8 +355,7 @@ class Imprinter:
                         # dummy start and stop times
                         'start_time': float(first5)*1e5,
                         'stop_time': (float(first5)+1)*1e5,
-                        'telescope': book.tel_tube[:3].lower(),
-                        'tube_slot': book.tel_tube.lower().replace("sat","satst")[3:],
+                        'telescope': book.tel_tube.lower(),
                         'type': 'hk',
                     }
                 def bind(self, pbar=False):
--- a/sotodlib/io/imprinter.py
+++ b/sotodlib/io/imprinter.py
@@ -353,8 +353,8 @@ class Imprinter:
                     return {
                         'book_id': book.bid,
                         # dummy start and stop times
-                        'start_time': float(first5+"00000"),
-                        'stop_time': float(first5+"99999"),
+                        'start_time': float(first5)*1e5,
+                        'stop_time': (float(first5)+1)*1e5,
                         'telescope': book.tel_tube[:3].lower(),
                         'tube_slot': book.tel_tube.lower().replace("sat","satst")[3:],
                         'type': 'hk',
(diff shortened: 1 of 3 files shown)


mhasself:
Nitpicking the schema:
- The telescope should be the platform name -- lat, satp1/2/3.  Or "site" for the site-wide instance.
- The tube_slot should not be populated in this case.
- Semi-closed intervals are So Much Better than closed intervals.  The stop time, in this case, should be 1686700000.0.

kmharrington:
Could we run these books on the E2E system to verify? I don't see any in the `hk` folder there.

sotodlib/io/imprinter.py
```diff
@@ -277,6 +276,96 @@ def register_books(self, bids, obs_lists, commit=True, session=None):
...
         if commit: session.commit()
 
+    def register_hk_books(self, commit=True, session=None):
+        """Register housekeeping books to database
+
+        """
+        session = session or self.get_session()
+
+        # fixme: tel_tube and daq_node may not be the same
+        for daq_node in self.config['sources']:
```
kmharrington, line 286: It is each daq-node that has an aggregator running on it. Should be one for each telescope and one for the whole site. (following the folder layout here: https://simonsobs.atlassian.net/wiki/spaces/PRO/pages/88244229/Site+Data+Storage+Outline#Level-2-Smurf-Data-%2F-HK-Data)

sotodlib/io/imprinter.py
```diff
@@ -277,6 +276,96 @@ def register_books(self, bids, obs_lists, commit=True, session=None):
...
+            class _HKBinder:  # dummy class to mimic baseline bookbinder
+                def __init__(self, indir, outdir):
+                    self.indir = indir
+                    self.outdir = outdir
+                def get_metadata(self):
+                    return {
+                        'book_id': book.bid,
+                        # dummy start and stop times
+                        'start_time': float(first5+"00000"),
+                        'stop_time': float(first5+"99999"),
```
kmharrington, line 357: I agree with Matthew on the [start, stop) intervals. Those are much nicer to work with.

guanyilun:
> 



> Could we run these books on the E2E system to verify? I don't see any in the `hk` folder there.

Here's one example: `/home/yguan/data_pkg/books/hk`

I think I've resolved the comments above. Here's the current M_index.yaml file:
```yaml
book_id: hk_16878_ocs3
start_time: 1687800000
stop_time: 1687900000
telescope: ocs3
type: hk
```
Note that the telescope name (ocs3) is specific to e2e system; for running at the site, it will be base on a different config file that will have the right names that Matthew suggested above.

kmharrington (approved):
I think this looks good to me.

PR #496, opened 2023-06-20
Add missing requirement

keskitalo:


keskitalo:
I'm seeing another dependency failure. `so3g/proj/coords.py` requires `qpoint`. We could make it an `sotodlib` requirement but it seems to me that it should have installed when I did `pip install so3g`.

PR #500, opened 2023-06-28
Make hwp solutions

17-sugiyama:
`make_hwp_solutions.py` returns the hwp angle AxisManager for each obs_id.
`g3thwp.py` is also updated for this script.
There are still remaining works around flags, name convention, pep8 and docstring.


sotodlib/hwp/g3thwp.py
```diff
@@ -621,7 +629,140 @@ def write_solution(self, solved, output=None):
...
+        aman.wrap_new('timestamps', shape=('samps', ))
+        aman.wrap_new('hwp_angle_ver1', shape=('samps', ))
+        aman.wrap_new('hwp_angle_ver2', shape=('samps', ))
+        aman.wrap_new('stable', shape=('samps', ))
+        aman.wrap_new('locked', shape=('samps', ))
+        aman.wrap_new('hwp_rate', shape=('samps', ))
+        aman.wrap_new('eval', shape=('samps', ))
+
+        aman.timestamps = tod.timestamps
+        aman.stable = np.zeros(len(tod.timestamps))
```
kmharrington, line 645: this line `aman.stable = np.zeros(len(tod.timestamps))` is identical to `aman.wrap_new('stable', shape=('samps', ))`. The first is more correct, so I would remove all the zeros you have down here. `wrap_new` has a `cls` argument that defaults to `np.zeros`.

We don't like doing `aman.field = new_thing` because that could break the axis tracking. If you have to, use `aman.field[:] = new_thing` to make sure the sizes are the same. But in these cases you do not need to reset these fields to zeros.

sotodlib/hwp/g3thwp.py
```diff
@@ -621,7 +629,140 @@ def write_solution(self, solved, output=None):
...
+            logger.warning('Input data is empty.')
+            return 
+        
+        start = int(tod.timestamps[0])-self._margin
+        end = int(tod.timestamps[-1])+self._margin
+        try:
+            data = self.load_data(start, end)
+        except Exception as e:
+            logger.error(f"Exception '{e}' thrown while loading HWP data. The specified encoder field is missing.")
+            self._write_empty_solution_h5(tod, output, h5_address)
```
kmharrington, line 709: I worry this will build us a whole lot of data files filled with just zeros. Maybe we should have another way of indicating the HWP has no solutions? Curious if @mhasself  has ideas.

sotodlib/hwp/g3thwp.py
```diff
@@ -621,7 +629,140 @@ def write_solution(self, solved, output=None):
...
+        aman.wrap_new('timestamps', shape=('samps', ))
+        aman.wrap_new('hwp_angle_ver1', shape=('samps', ))
+        aman.wrap_new('hwp_angle_ver2', shape=('samps', ))
+        aman.wrap_new('stable', shape=('samps', ))
+        aman.wrap_new('locked', shape=('samps', ))
+        aman.wrap_new('hwp_rate', shape=('samps', ))
+        aman.wrap_new('eval', shape=('samps', ))
+        
+        aman.timestamps = tod.timestamps
+        aman.stable = scipy.interpolate.interp1d(solved['slow_time'], solved['stable'], kind='linear', bounds_error=False)(tod.timestamps)
```
kmharrington, line 750: Change these to all be `aman.field[:] = new_thing` to make sure the array shapes aren't changing.

sotodlib/site_pipeline/make_hwp_solutions.py
```diff
@@ -0,0 +1,103 @@
...
+    man_db = core.metadata.ManifestDb(scheme=scheme)
+    
+    # Get file + dataset from policy.
+    # policy = util.ArchivePolicy.from_params(config['archive']['policy'])
+    # dest_file, dest_dataset = policy.get_dest(obs_id)
+    # Use 'output_dir' argument for now    
+    man_db_filename = os.path.join(output_dir, 'hwp_angle.sqlite')
+    output_filename = os.path.join(output_dir, 'hwp_angle.h5')
+    
+    # load observation data
```
kmharrington, line 79: This is a good initial sketch for how to calculate and save the angles, but right now it will go and recalculate all of them every time we run it. For this script we'll want more options than that.  Could you copy more of the setup that is in preprocess-tod? https://github.com/simonsobs/sotodlib/blob/master/sotodlib/site_pipeline/preprocess_tod.py#L239 

The logic in the function goes:
1. if obs-id is passed as an argument, just run the calculation for that observation
2. otherwise, query context with min/max ctimes (if given) or any other query parameters to build a list of observations to calculate on
3. for each observation, if there is already an entry in the metadata AND overwrite=False, skip. Otherwise, calculate+save.

glitter17: Thank you for the really helpful example! I added some args (overwrite, obs-id, min-/max-time, query) to specify the calculation range.

sotodlib/hwp/g3thwp.py
```diff
@@ -621,7 +629,134 @@ def write_solution(self, solved, output=None):
...
+        aman.wrap_new('hwp_angle_ver1', shape=('samps', ))
+        aman.wrap_new('hwp_angle_ver2', shape=('samps', ))
+        aman.wrap_new('stable', shape=('samps', ))
+        aman.wrap_new('locked', shape=('samps', ))
+        aman.wrap_new('hwp_rate', shape=('samps', ))
+        aman.wrap_new('eval', shape=('samps', ))
+        
+        aman.timestamps[:] = tod.timestamps
+        aman.stable[:] = scipy.interpolate.interp1d(solved['slow_time'], solved['stable'], kind='linear', bounds_error=False)(tod.timestamps)
+        aman.locked[:] = scipy.interpolate.interp1d(solved['slow_time'], solved['locked'], kind='linear', bounds_error=False)(tod.timestamps)
```
mhasself, line 745: Remember that "stable" and "locked" should be booleans.  It looks like interp1d will convert the input bools to 0.0 and 1.0 ... then interpolate those and return floats.

I think we want to convert the interpolated result to a boolean, either with  "... == 1.0" or "... > .999".

sotodlib/hwp/g3thwp.py
```diff
@@ -621,7 +629,134 @@ def write_solution(self, solved, output=None):
...
+        aman.wrap_new('timestamps', shape=('samps', ))
+        aman.wrap_new('hwp_angle_ver1', shape=('samps', ))
+        aman.wrap_new('hwp_angle_ver2', shape=('samps', ))
+        aman.wrap_new('stable', shape=('samps', ))
+        aman.wrap_new('locked', shape=('samps', ))
+        aman.wrap_new('hwp_rate', shape=('samps', ))
+        aman.wrap_new('eval', shape=('samps', ))
+        
+        aman.timestamps[:] = tod.timestamps
+        aman.stable[:] = scipy.interpolate.interp1d(solved['slow_time'], solved['stable'], kind='linear', bounds_error=False)(tod.timestamps)
```
mhasself, line 744: Don't we want bounds_error=True?  Or is this just a way to make the call faster?

sotodlib/hwp/g3thwp.py
```diff
@@ -621,7 +629,134 @@ def write_solution(self, solved, output=None):
...
+        aman.locked[:] = scipy.interpolate.interp1d(solved['slow_time'], solved['locked'], kind='linear', bounds_error=False)(tod.timestamps)
+        aman.hwp_rate[:] = scipy.interpolate.interp1d(solved['slow_time'], solved['hwp_rate'], kind='linear', bounds_error=False)(tod.timestamps)
+        if 'fast_time_raw' in solved.keys():
+            aman.hwp_angle_ver1[:] = scipy.interpolate.interp1d(solved['fast_time_raw'], solved['angle'], kind='linear',bounds_error=False)(tod.timestamps)
+            aman.hwp_angle_ver2[:] = scipy.interpolate.interp1d(solved['fast_time'], solved['angle'], kind='linear',bounds_error=False)(tod.timestamps)
+            aman.eval[:] = np.ones(len(tod.timestamps))
+        else:
+            logger.info('Template subtraction failed')
+            aman.hwp_angle_ver1[:] = scipy.interpolate.interp1d(solved['fast_time'], solved['angle'], kind='linear', bounds_error=False)(tod.timestamps)
+            aman.hwp_angle_ver2[:] = scipy.interpolate.interp1d(solved['fast_time'], solved['angle'], kind='linear', bounds_error=False)(tod.timestamps)
```
mhasself, line 754: You have to do something to handle the discontinuities in the angle due to the periodic domain (0 - 360).

numpy.unwrap can be helpful here.

There are probably other places in sotodlib where there has been done.

mhasself (changes requested):
Thanks for this!  A few notes about data types and processing.

sotodlib/hwp/g3thwp.py
```diff
@@ -621,7 +629,134 @@ def write_solution(self, solved, output=None):
             start_time += frame_length
 
         return
+    
+    def _write_empty_solution_h5(self, tod, output=None, h5_address=None):
+        logger.info('Writing empty solutions')
+        aman = sotodlib.core.AxisManager(tod.dets, tod.samps)
+        aman.wrap_new('timestamps', shape=('samps', ))
```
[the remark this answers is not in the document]
yukisakurai: fixed

sotodlib/hwp/g3thwp.py
```diff
@@ -621,7 +629,134 @@ def write_solution(self, solved, output=None):
...
+        -----------
+        Output file format \n
+
+        - timestamp:
+            SMuRF synched timestamp
+        - hwp_angle_ver1: float
+            SMuRF synched HWP angle (calculated from raw encoder signals) in radian
+        - hwp_angle_ver2: float
+            SMuRF synched HWP angle after the template subtraction (the systematics from the non-uniform encoder slot pattern is subtracted ) in radian. \n
+            if 'eval' is zero, template subtraction is not completed and its value is same as 'hwp_angle_ver1'.
```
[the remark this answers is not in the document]
yukisakurai: fixed

sotodlib/hwp/g3thwp.py
```diff
@@ -621,7 +629,134 @@ def write_solution(self, solved, output=None):
...
+        output: str or None
+          output path + file name, overwrite config file
+
+        Notes
+        -----------
+        Output file format \n
+
+        - timestamp:
+            SMuRF synched timestamp
+        - hwp_angle_ver1: float
```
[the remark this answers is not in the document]
yukisakurai: set all angles as float32

sotodlib/hwp/g3thwp.py
```diff
@@ -621,7 +629,134 @@ def write_solution(self, solved, output=None):
...
+            SMuRF synched HWP angle (calculated from raw encoder signals) in radian
+        - hwp_angle_ver2: float
+            SMuRF synched HWP angle after the template subtraction (the systematics from the non-uniform encoder slot pattern is subtracted ) in radian. \n
+            if 'eval' is zero, template subtraction is not completed and its value is same as 'hwp_angle_ver1'.
+        - slow_time: timestamp
+            time list of slow block (stable, locked. hwp_rate)
```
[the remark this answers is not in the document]
yukisakurai: removed

sotodlib/hwp/g3thwp.py
```diff
@@ -621,7 +629,134 @@ def write_solution(self, solved, output=None):
...
+        # calculate template subtracted angle
+        try:
+            self.eval_angle(solved)
+        except Exception as e:
+            logger.error(f"Exception '{e}' thrown while the template subtraction")
+            pass
+        
+        # write solution
+        aman = sotodlib.core.AxisManager(tod.dets, tod.samps)
+        aman.wrap_new('timestamps', shape=('samps', ))
```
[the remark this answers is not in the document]
yukisakurai: added dedicated function

sotodlib/hwp/g3thwp.py
17-sugiyama: Fixed by adding a function to interpolate the bool array in [line 663](https://github.com/simonsobs/sotodlib/blob/make-hwp-solutions/sotodlib/hwp/g3thwp.py#L663).

sotodlib/hwp/g3thwp.py
17-sugiyama: "bounds_error=False" is set not to run the extrapolations when the HWP data don't exist at the beginning or the end of the input time range. 
If "bounds_error=True", the process will stop in such cases.
If "bounds_error=False", where is out of input time range will be filled with None, instead of the extrapolation.
I added the error message when the None-filling process is performed.

sotodlib/hwp/g3thwp.py
17-sugiyama: Fixed. We can select the output type (continuous/discontinuous) of the angle solution with mod2pi argument in "analyze" function.

17-sugiyama:
Debugged make_hwp_solutions.py -- observation id is now declared properly.

sotodlib/hwp/g3thwp.py
mhasself: I'm not sure I understand... the call to analyze() is still in "discontinuous" mode, which means you're interpolating a curve with discontinuities.

This is what I'm worried about -- the black points are your solution points (fast_time, angle) and the blue points are tod.timestamps.  The interpolation is not valid at the points where angle wraps from 2pi to 0 (or 0 to 2pi).

image

17-sugiyama: Yes, discontinuous output of the angle can arise the undesired blue points.
But in my environment, the angle solution is now output in the continuous mode with the call:
`solved = self.analyze(data, mod2pi=False)`,
on L745 of g3thwp.py.
I'm sorry but I would like you to try to run the script again.
image

sotodlib/hwp/g3thwp.py
```diff
@@ -630,22 +638,38 @@ def write_solution(self, solved, output=None):
...
         return
 
+    def _bool_interpolation(self, timestamp1, data, timestamp2):
+        interp = scipy.interpolate.interp1d(timestamp1, data, kind='linear', bounds_error=False)(timestamp2)
+        result = []
+        for i in interp:
+            if i > .999:
+                result.append(True)
+            else:
+                result.append(False)
```
mhasself, line 670: A loop like this is slow ... easier to use numpy:
```
  # Create a bool array
  result = (interp > 0.999)
```

PR #501, opened 2023-06-28
Fix use of csr_matrix.hstack.vstack

mhasself:
This was causing tests to fail on py39 and py310, I guess because scipy on the CI systems just ticked up into version 1.11.


kmharrington (approved):
well that's annoying. Thanks for tracking this down!

PR #503, opened 2023-07-13
site_pipeline: some refactoring to better fit prefect

guanyilun:


mhasself:
This is ok but requires modifications to cli.py and to the site_pipeline docs (section "Wrapping a pipeline script").

In your current version (I know this is a draft), several files are missing newlines at the end.  Also the `__main__` block for update_g3tsmurf doesn't actually run the main code.

Is it needed/helpful for the main function to be called <element_name> rather than main?  That just means we need more configuration in cli.py.

guanyilun:
Thanks for the quick feedback! Using a descriptive name rather than main was the most natural thing to begin with since prefect does some auto-inference based on the function name and signature. I've now worked around it by manually replacing the `__name__` attribute of the function when I wrap it in prefect so prefect is happy. We can stay with the `main` naming convention here now!

PR #505, opened 2023-07-17
Simulated focalplane spacing

tskisner:
This PR is the matching one for the [upstream PR in toast](https://github.com/hpc4cmb/toast/pull/674).  This ensures that focalplane layout widths are always given in terms of pixel centers rather than field of view.

### DIFF
--- a/sotodlib/scripts/hardware_plot.py
+++ b/sotodlib/scripts/hardware_plot.py
@@ -45,6 +45,11 @@ def main():
     )
 
     parser.add_argument(
+        "--show_centers", required=False, default=False, action="store_true",
+        help="Add labels with pixel center coordinates."
+    )
+
+    parser.add_argument(
         "--xieta", required=False, default=False, action="store_true",
         help="Plot in Xi / Eta coordinates."
     )
@@ -93,6 +98,7 @@ def main():
         xieta=args.xieta,
         lat_corotate=args.lat_corotate,
         lat_elevation=args.lat_elevation_deg * u.degree,
+        show_centers=args.show_centers
     )
 
     return
--- a/sotodlib/sim_hardware.py
+++ b/sotodlib/sim_hardware.py
@@ -366,15 +366,15 @@ def sim_nominal():
     ltube_cryonames=["c1", "i5", "i6", "i1", "i3", "i4", "o6"]
 
     lat_ufm_slot = [
-        2,
-        1,
         0,
+        1,
+        2,
     ]
 
     lat_ufm_thetarot = [
-        120.0,
-        0.0,
         240.0,
+        0.0,
+        120.0,
     ]
 
     for tindx in range(7):
@@ -405,6 +405,8 @@ def sim_nominal():
         tb["receiver_name"] = ""
         tube_slots[nm] = tb
 
+    # These are taken from:
+    # https://simonsobs.atlassian.net/wiki/spaces/PRO/pages/101974017/Focal+Plane+Coordinates#UFM-Layout.1
     hex_to_ufm_slot = [
         0,
         2,
@@ -444,8 +446,6 @@ def sim_nominal():
         tb["wafer_slot_angle"] = [
             hex_to_ufm_thetarot[tw] for tw in range(7)
         ] # Degrees
-        # These are taken from:
-        # https://simonsobs.atlassian.net/wiki/spaces/PRO/pages/101974017/Focal+Plane+Coordinates#UFM-Layout.1
         # The "slot" here is the relative slot (ws0 - ws6) within the tube.
         tb["wafer_ufm_slot"] = list()
         # The "loc" here is the compass direction name (e.g. NO, NE, SW, etc.)
--- a/sotodlib/sim_hardware.py
+++ b/sotodlib/sim_hardware.py
@@ -394,6 +394,7 @@ def sim_nominal():
             for w, props in cnf["wafer_slots"].items():
                 if props["type"] == ttyp:
                     if off == woff[ttyp]:
+                        props["tube_index"] = tw
                         tb["wafer_slots"].append(w)
                         tb["wafer_ufm_slot"].append(lat_ufm_slot[tw])
                         woff[ttyp] += 1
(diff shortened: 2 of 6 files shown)


tskisner:
This work attempts to fix some issues with the synthetic focalplane layout, and also move some common helper functions upstream to reduce future maintenance burden.  Below is a detailed examination of the code in this PR (and the corresponding toast PR).  Note that there are differences when compared to current toast and current sotodlib master, so it is important to verify that the current code is correct.

**NOTE:  All plots in boresight X/Y coordinates looking from sky into the instrument**.  Recall that the positive "X" axis points "down" in elevation.

# LAT

First looking at the LAT layout.  Tube spacing is 359.6mm with a platescale of 0.00495 deg/mm.  In a hexagonal packing scheme, we would expect the LF tube to be located at Xi = (2 * tube_spacing) / (sqrt(3) / 2), which corresponds to +3.08 degrees.  The spacing in Eta between the other tubes should be 1.78 degrees.  These are supported by the overall plot here.
lat pdf


## LF 30GHz

Looking now at the 30GHz detectors in the LF tube, the wafer spacing is 128.4mm * 0.00495 deg/mm = 0.63558 deg.  From the pixel center coordinates on the zoom-in we see that the bottom two wafers are at Xi = +/- 0.3181 deg, which agrees within plot resolution.  For each wafer, the pixel spacing should be 18.0mm * 0.00495 deg/mm = 0.0891 deg.  This spacing can be verified on the plots by computing the distance between the center coordinates in degrees.  The FWHM for these detectors is 7.4' = 0.123 deg.  So it makes sense visually that the circles (which are the FWHM) extend about 2/3 the distance to the neighboring pixels.
lat_f030 pdf
lat_f030_zoom1
lat_f030_zoom2


## MF 90GHz

Looking at the 90GHz detectors, the wafer spacing is again 128.4mm * 0.00495 deg/mm = 0.63558 deg.  The zoom in of this graph is more cluttered, but I checked that this spacing is also true.  For each wafer, the pixel spacing should be 5.3mm * 0.00495 deg/mm = 0.0262 deg.  Also note that the distance across the rhombus gap is equal to the intra-rhombus pixel spacing.  The FWHM for these detectors is 2.2' = 0.037 deg.  So it makes sense visually that the circles (which are the FWHM) overlap with neighboring pixels.
lat_f090 pdf
lat_f090_zoom1
lat_f090_zoom2


## UHF 230GHz

Moving on to the 230GHz detectors, same wafer spacing of 128.4mm * 0.00495 deg/mm = 0.63558 deg.  Again, hard to read coordinates on the zoomed in view, but I checked this.  The pixel spacing is the same as the MF at 5.3mm * 0.00495 deg/mm = 0.0262 deg.  The FWHM is 1.0' = 0.0167 deg, which matches visually with the plot.
lat_f230 pdf
lat_f230_zoom1


# SAT

For the SAT check I looked at the plots of SAT1.  The wafer spacing is 128.4mm * 0.09668 deg/mm = 12.414 deg.  Along the Xi axis, the center pixels of the hex wafers are indeed very close to this, and the extra ~0.1 degree is due to the ARC projection used in the plot and the large size of the projected focalplane.

## LF 30GHz

The pixel spacing should be 18.0mm * 0.09668 deg/mm = 1.740 deg.  This can be seen directly on the zoomed in plot.  The FWHM is 91.0' = 1.517 degrees, which makes sense visually.
sat_f030 pdf
sat_f030_zoom1


## MF 90GHz

The pixel spacing should be 5.3mm * 0.09668 deg/mm = 0.512 deg, which agrees with the plot.  The FWHM is 30.0' = 0.5 degrees, which also makes sense given the plot.
sat_f090 pdf
sat_f090_zoom1
sat_f090_zoom2


## UHF 230GHz

The pixel spacing should be 5.3mm * 0.09668 deg/mm = 0.512 deg, which agrees with the plot.  The FWHM is 11.0' = 0.183 degrees, which also makes sense given the plot.
sat_f230 pdf
sat_f230_zoom1

tskisner:
Note that the unit tests here (and the code in this branch) will not work before [toast PR 674](https://github.com/hpc4cmb/toast/pull/674) is merged.  Now that this work is ready, I'll get that upstream PR merged.

smsimon (approved):
These layouts all look correct to me

smsimon:
One additional question--where is the capability to change individual wafer clockings now?

tskisner:
About wafer rotation angles within the tube- will push a change for that soon which stores that angle in the hardware dictionary (default to zero).  Will wait until the upstream toast wheel is in place so that tests will run.

mhasself:
I have reached out on other channels with a question about per-optics tube rotation due to reimaging optics.

mhasself (changes requested):
As discussed out of band, the on-sky footprint of each optics tube should be "flipped", or rotated by 180 deg.

tskisner:
The latest changes now implement the wafer rotations outlined in the [Focal Plane Coordinates document](https://simonsobs.atlassian.net/wiki/spaces/PRO/pages/101974017/Focal+Plane+Coordinates) on confluence.  The plots below are now in Xi / Eta sky projection coordinates for easier visual comparison.

## SAT

I used the SAT diagram and theta angles and mapped wafers locations between conventions.  I also used the fact that the `ws2` / `NE` wafer slot has a rotation of "zero", and also the fact that the pin hole in the "A" rhombus is pointing to the "Northwest" at zero theta rotation.  After getting that to match, I applied the optical flip to each tube.  Here are 3 figure showing the progression from (1) all wafers at zero rotation, (2) wafer theta angles to match the figure on the confluence, and (3) the overall flip.
sat_f090 pdf
sat_f090 pdf
sat_f090 pdf

## LAT

Similarly I modeled this after the UFM diagram / table on confluence and the rhombus pin location.  The wafers are rotated so that their pin holes point towards the tube centers.  Here are two figures (1) before the flip and (2) after the flip.
lat_f090 pdf
lat_f090 pdf

So just to be clear, the final plot in both cases are with the current code- I was just showing the progression of application of angles to make it easier to see the intermediate steps.

tskisner:
Ok, this PR was approved in other channels (slack), and the up-to-date plots are now on confluence.  Merging.

PR #506, opened 2023-07-20
Add stream_id to det_info

jseibert575:
This adds the current stream_id to det_info when loading from load_smurf tools. Also merges in det_info when loading via context.


kmharrington (approved):

kmharrington:
this has been overwritten at this point

PR #507, opened 2023-07-21
Refactoring workflows into submodules such that it can be called from within Python

ickc:
This pull request simply refactor `workflows/` into `sotodlib.workflows` so that they can be called from within Python, not only from command lines.

Scripts within `workflows/` are preserved (and calling `sotodlib.workflows`) for backward compatibility reason.

For those scripts that are already exported as scripts in `setup.py` are now setup as `console_scripts`. For those that has not previously exported, they are left untouched. I.e. `make_hardware_maps` is not included there as `make_hardware_maps.py` was never there. And the followings are equivalent:

```sh
# preserved for backward compatibility
get_wafer_offset.py -h
# console_scripts entry point
get_wafer_offset -h
# calling as module
python -m sotodlib.workflows.get_wafer_offset -h
```


tskisner:
In general, I am supportive of this change.  I had actually been thinking of further refactoring some of the high-level tools that are used in common between the workflows, [modeled after the structure used in other packages](https://github.com/hpc4cmb/litebirdtask/tree/imo_native/litebirdtask/workflow).  The changes in this PR would be compatible with that.  A few notes:
1.  There is a longstanding policy decision handed down from DM that toast must be an optional dependency of sotodlib.  Because of that, we should make sure that setup.py console scripts entries are installable even if toast is not present.  I do not think that these scripts get imported when running setup.py, but we should check and guard the toast imports if needed.
2.  We could have this top level `sotodlib.workflow` package for the `main()` functions, and then the further refactor of common code could go in `sotodlib.toast.workflow`.  I like it!
3.  We should wait to confirm that this does not break any of the scripting done by the sim group.

ickc:
For (3), I think this PR is completely backward compatible. Once installed, the following can be used to check the entry points are setup correctly:

```bash
for name in toast_so_sim toast_so_map toast_so_convert get_wafer_offset; do
    $name.py -h > /dev/null
    [[ "$?" == 0 ]] || echo "Failed to run $name.py -h"
    $name -h > /dev/null
    [[ "$?" == 0 ]] || echo "Failed to run $name -h"
    python -m sotodlib.workflows.$name -h > /dev/null
    [[ "$?" == 0 ]] || echo "Failed to run python -m sotodlib.workflows.$name -h"
done

for name in make_hardware_maps mapmaker_test; do
    python -m sotodlib.workflows.$name -h > /dev/null
    [[ "$?" == 0 ]] || echo "Failed to run python -m sotodlib.workflows.$name -h"
done
```

mhasself:
I'm in favor of this sort of thing but I would prefer the workflows be placed as a submodule of toast, i.e. in sotodlib.toast.workflows instead of sotodlib.workflows.

Thanks!

setup.py
```diff
@@ -24,9 +24,15 @@
         "so_hardware_info = sotodlib.scripts.hardware_info:main",
         "so-metadata = sotodlib.core.metadata.cli:main",
         "so-site-pipeline = sotodlib.site_pipeline.cli:main",
+        "toast_so_sim = sotodlib.workflows.toast_so_sim:cli",
+        "toast_so_map = sotodlib.workflows.toast_so_map:cli",
+        "toast_so_convert = sotodlib.workflows.toast_so_convert:cli",
+        "get_wafer_offset = sotodlib.workflows.get_wafer_offset:main",
```
tskisner, line 30: Should these be changed to `sotodlib.toast.workflows` as well?

tskisner (changes requested):
Looks good after changing one remaining set of paths in setup.py

tskisner (approved):
Looks good, thanks!

PR #509, opened 2023-07-21
Fix logic when parsing det_info tuple.

### LINKED ISSUE #508: Sim focal plane footprints are weird now
@keskitalo I believe #409 did change the default focal plane.  (cc @nad-dch , @kmharrington .)

The control of detailed orientation of subwafers is cool but this isn't how the tube is assembled.

SAT1 focal plane from commit 4d660fc (shows proper interlocking of wafers):

image

After https://github.com/simonsobs/sotodlib/pull/409 (commit cf6221f) (shows fun detailed orientations, but incorrect interlocking):

image

The LAT is also affected:
image

tskisner:
This fixes an immediate bug in SOFocalplane.  However, the unit test for this is pending as part of #505.

Closes #508.

### DIFF
--- a/sotodlib/toast/sim_focalplane.py
+++ b/sotodlib/toast/sim_focalplane.py
@@ -629,7 +629,7 @@ def sim_telescope_detectors(hw, tele, tube_slots=None, det_info=None, no_darks=F
             np.zeros(7) * u.degree,
         )
         centers = np.zeros((7, 4), dtype=np.float64)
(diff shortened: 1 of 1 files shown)


mhasself (approved):
Thanks for the quick fix.

The docstring description of the det_info argument is pretty vague.  I assume it's explained somewhere else, such that all these None checks on the first element make some kind of sense.

PR #510, opened 2023-07-25
Level 2 file Deletion

kmharrington:
Automated tools for deleting level 2 files that have already made it into books. 

Currently running on the e2e system. 

It's been running for ~48 hours now and I've been finding edge cases that are now fixed. I believe we've now had every time of book made correctly w/ correctly deleted level 2 data (modulo some errors in how the the file emulators are running).

### DIFF
--- a/sotodlib/site_pipeline/make_book.py
+++ b/sotodlib/site_pipeline/make_book.py
@@ -1,5 +1,6 @@
 from typing import Optional
 import traceback
+import argparse
 
 from ..io.imprinter import Imprinter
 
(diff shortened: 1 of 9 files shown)


sotodlib/site_pipeline/make_book.py
```diff
@@ -1,15 +1,10 @@
 from typing import Optional
```
guanyilun, line 1: this file is missing argparse, it seems

guanyilun (changes requested):
caught some obvious typos, otherwise I'm ok with this!

kmharrington:
Ok @guanyilun  I think I've got those bugs fixed. Thanks for the catch!

guanyilun (approved):

PR #511, opened 2023-07-25
Imprinter db schema update

guanyilun:
- imprinter db schema change to make book status comparison easier: use sequential integers to represent different stages of bookbinding, starting from 0 for unbound, to 2 for bound, and then to 3 for when it is uploaded to librarian, and then to 4 when level 2 files are removed, that's when the record can be safely removed from the database, if necessary.
- also includes some utility functions to answer questions like till what time are all book bound, etc.

### DIFF
--- a/sotodlib/io/imprinter.py
+++ b/sotodlib/io/imprinter.py
@@ -1153,15 +1153,6 @@ class Imprinter:
                 return book.start
         return book.start  # last book
 
-    def all_bound_until_hk(self):
-        session = self.get_session()
-        books = session.query(Books).filter(Books.type == "hk").order_by(Books.start).all()
-        for book in books:
-            if book.status < BOUND:
-                return book.start
-        return book.start  # last book
-        
-
     def delete_book_files(self, book):
         """Delete all files associated with a book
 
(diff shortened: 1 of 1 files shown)


sotodlib/io/imprinter.py
```diff
@@ -1099,6 +1140,28 @@ def get_g3tsmurf_obs_for_book(self, book):
...
+        session = self.get_session()
+        # sort by start time and find the start time by which
+        # all books are bound
+        books = session.query(Books).order_by(Books.start).all()
+        for book in books:
+            if book.status < BOUND:
+                return book.start
+        return book.start  # last book
+
+    def all_bound_until_hk(self):
```
kmharrington, line 1156: is there anything with HK here that you should also be looking at for smurf and stray books?

guanyilun: hmm that's right, there's nothing special about hk. It's the same function for smurf or stray. I can remove this redundant function.

kmharrington (approved):
Other than the in-line comment, these look good. I'm expecting I won't pull these changes for a day or two on the E2E system setup I'm running on.

PR #517, opened 2023-07-31
Hk_utils Update

sanahabhimani:
This update expands `hk_utils` to be flexible beyond the requirement of a config file. This is useful because it can be used as a complementary tool to output housekeeping axis managers from `g3thk` queries at the site for immediate low-level HK (+ detector) analysis. 

It has been tested for all use cases:
1. loading HK axismanagers given a config file (same as the one used for `load_range()`)
2. loading HK axisamanagers given field names only (no aliases provided)
3. loading HK axismanagers given field names + corresponding aliases
4. all 3 cases above repeated for when HK data needs to be cosampled with detector data

### DIFF
--- a/sotodlib/io/hk_utils.py
+++ b/sotodlib/io/hk_utils.py
@@ -402,7 +402,7 @@ def get_hkaman(start, stop, config=None, alias=None, fields=None, data_dir=None)
     """
     if config is not None:
         data = sort_hkdata_fromconfig(start=start, stop=stop, config=config)
-        hkamans = make_hkaman(grouped_data=data, alias_exists=True, 
+        hkamans = make_hkaman(grouped_data=data, alias_exists=True,
                               det_cosampled=False)
         return hkamans
 
@@ -418,7 +418,7 @@ def get_hkaman(start, stop, config=None, alias=None, fields=None, data_dir=None)
         else:
             data = sort_hkdata(start=start, stop=stop, fields=fields,
                                data_dir=data_dir, alias=alias)
-            hkamans = make_hkaman(grouped_data=data, alias_exists=True, 
+            hkamans = make_hkaman(grouped_data=data, alias_exists=True,
                                   det_cosampled=False)
             return hkamans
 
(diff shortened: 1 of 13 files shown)


sotodlib/io/hk_utils.py
```diff
@@ -49,58 +174,67 @@ def get_grouped_hkdata(start, stop, config):
...
-    # feed/device, ready to input to an axismanager
-    grouped_data = []
-    for group in grouped_feeds:
-        device_data = {}
-        for field in group:
-            field_dict = {field: fields_data[field]}
-            device_data.update(field_dict)
+    """
+    if alias is None:
+        hkdata = load_range(start, stop, fields, data_dir=data_dir)
```
kmharrington, line 216: when calling functions that have a lot of configurable options like `load_range`, I would always do `kwarg=kwarg`. That way if someone changes the order of the kwargs at some point these functions won't need to be changed.

sotodlib/io/hk_utils.py
```diff
@@ -49,58 +174,67 @@ def get_grouped_hkdata(start, stop, config):
...
+        _check_hkdata(hkdata)
 
-        grouped_data.append(device_data)
+        grouped_data = _group_data(hkdata, fields_only=True)
 
-    return grouped_data
+        return grouped_data
+
+    elif alias is not None:
+        hkdata = load_range(start, stop, fields, alias, data_dir)
```
kmharrington, line 224: kwarg=kwarg

sotodlib/io/hk_utils.py
```diff
@@ -273,8 +436,17 @@ def get_detcosamp_hkaman(config, det_aman):
...
-    return hkamans_detcosamp
-# TODO: insert a progress bar for get_grouped_hkdata
+    if config is not None:
+        data = sort_hkdata_fromconfig(start, stop, config)
+        amans = make_hkaman(data, det_aman=det_aman, alias_exists=True, det_cosampled=True)
+        return amans
+    elif fields is not None:
+        if alias is None:
+            data = sort_hkdata(start, stop, fields, data_dir)
+            amans = make_hkaman(data, det_cosampled=True, det_aman=det_aman)
```
kmharrington, line 447: similarly to other comment. Functions like `make_hkaman`, I wouldn't depend on a default kwarg parameter, if `alias_exists` needs to be false then explicitly set it to false here.

kmharrington (approved):
Going to approve this because my in-line comments are more for code longevity than functionality right now. But I do want those sorts of changes made (never trust a future coder when you don't have to, even when the future coder is you 😁 ). I probably didn't mark every place you're depending on kwarg order or default arguments, worth giving the whole file a look.

I haven't run this to test it through all the different methods of calling things. A future to-do is probably figuring out how to run some test cases.

sanahabhimani:
I will merge this shortly; I'm just going through all the cases where it's useful to have `kwarg=kwarg`

PR #519, opened 2023-08-03
g3tsmurf indexing deployment bugfix

guanyilun:
`update_g3tsmurf_db` needs an argument reordering. I found this fixed the bug in prefect auto-deployment but I still need to understand why. It may have to do with how I modify function signature by hand. Before I figure out the reason, we should make sure that `logger` is always the last argument in each site-pipeline module script.


kmharrington (approved):
I don't mind changing it to get things working. But I do think functions that call functions with lots of configurable kwargs like this should be written to be robust to parameter re-ordering.

PR #521, opened 2023-08-05
fix bug in make_map function in coords/demod.py

chervias:
The make_map function is missing a center_on=None parameter in the inputs. If center_on is not defined it fails at some point.


mhasself (approved):
Thanks!

PR #522, opened 2023-08-08
Preprocess debug

tterasaki:
Towards the first light, I have minor-updated some functions which are mainly related to check whether the preprocessing is going well or not.

Main modifications are as follows.

1. PSD calculation update
2. plot PSD before/after preprocessing
3. plot HWPSS fittiing result

# 1. PSD calculation update
Current PSD calculation can be executed only once and can only manage aman.signal, which will be annoying when we want to compare before/after preprocessing.
I have modified the way that `signal` defaults in `tod_ops.fft_ops.calc_psd`  ([link](https://github.com/simonsobs/sotodlib/blob/9f72a6ec721e752f284f9b7417dca476d5c20bcc/sotodlib/tod_ops/fft_ops.py#L222-L225C30p))
From
```python
if signal is None:
        signal = aman.signal
```
To
```python
if signal is None:
    signal = aman.signal
elif isinstance(signal, str):
    signal = aman[signal]
```
And modifed `preprocess.PSDCalc` so that the psd result should be wrapped into aman and proc_aman as `psd_{self.process_cfgs['signal']}` ((link)[https://github.com/simonsobs/sotodlib/blob/9f72a6ec721e752f284f9b7417dca476d5c20bcc/sotodlib/preprocess/processes.py#L128-L131])

# 2. plot PSD before/after preprocessing
To compare the PSD easily, added `sotodlib.hwp.hwp_utils.plot_preprocess_PSDs` ((link)[https://github.com/simonsobs/sotodlib/blob/9f72a6ec721e752f284f9b7417dca476d5c20bcc/sotodlib/hwp/hwp_utils.py#L45])
The plot below is output of `sotodlib.hwp.hwp_utils.plot_preprocess_PSDs`.
image

# 3. plot HWPSS fitting
To check the HWPSS fitting, added `sotodlib.hwp.hwp_utils.plot_hwpss_fit_status` ((link)[https://github.com/simonsobs/sotodlib/blob/9f72a6ec721e752f284f9b7417dca476d5c20bcc/sotodlib/hwp/hwp_utils.py#L6])
The plot below is output of `sotodlib.hwp.hwp_utils.plot_hwpss_fit_status`.

image

### DIFF
--- a/sotodlib/coords/demod.py
+++ b/sotodlib/coords/demod.py
@@ -7,7 +7,7 @@ from .. import core, tod_ops, coords
 
 def make_map(tod,
              P=None, wcs_kernel=None,
-             res=0.1 * coords.DEG,
+             res=0.1 * coords.DEG, center_on=None,
              dsT=None, demodQ=None, demodU=None,
              cuts=None,
              det_weights=None, det_weights_demod=None,
--- a/sotodlib/flag_utils.py
+++ b/sotodlib/flag_utils.py
@@ -5,7 +5,7 @@ import time
 from scipy import stats
 
 def plot_glitch_stats(tod, glitches=None, N_bins=30, save_path='./',
-                      save_name='glitch_flag_stats.png', save_plot=True):
+                      save_name='glitch_flag_stats.png', save_plot=False):
     """
     Function for plotting the glitch flags/cut statistics using the built in stats functions
     in the RangesMatrices class.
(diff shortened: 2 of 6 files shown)


davidvng:
Hi @tterasaki 
I've been assigned to this PR and have tested your code and it looks good! After talking with @msilvafe I have some comments and would like to contribute by adding to this PR but would like your thoughts first.

- The structure you have set up here requires the "psd" process to be added multiple times to the configs to create and populate `aman` with `psd,` `psd_dsT`, `psd_demodQ`, and `psd_demodU`. Since the names are hard-coded into `plot_preprocess_PSDs` it is difficult unless the config is correctly set up. I think we'd like to be able to work directly with the configs without writing to them. Perhaps the function could insert the correctly named psd whenever `process` appears in the configs.
- I'd like to create a high-level function in `site_pipeline.preprocess_tod` similar to `run_preprocess` that would be called something like `debug_preprocess` This function would essentially call your function whenever a `process` is found in the pipe, or create different graphs whenever a `calc_and_save` is found (e.g. glitch plots). Let me know if you think this structure makes sense and if you have any thoughts.

tterasaki:
Hi David! Yeah, all make sense to me. I gonna make the `debug_preprocess`.

PR #523, opened 2023-08-15
Change format to CLI interface works

kmharrington:
Needed to reorganize how some functions were named / called to get this script working through the site pipeline command line interface. I think this is the preferred format?


guanyilun:
If you want to run it through prefect, you should also add type annotations to the main function, so that prefect UI assigns the right input widgets for them.

kmharrington:
Thanks @guanyilun. I've added that so this should work through prefect

mhasself (approved):

PR #531, opened 2023-08-18
Put so3g into docs build requirements.

mhasself:
This should fix the docs build.

The error is from so3g being mocked, so astropy.units can't handle it.


keskitalo:
Awesome! Thanks.

PR #533, opened 2023-08-22
Split HK sources from Detector data sources

kmharrington:
This update, matched with [site-pipeline-config changes here](https://github.com/simonsobs/site-pipeline-configs/pull/9) make the HK bookbinding separate from the detector book binding. This is necessary for the LAT data, which has the HK and detector data going to separate places and for the site hk data, which has no detectors. I've tested this on some of the site level 2 data and looks like it all works.

### DIFF
--- a/sotodlib/io/imprinter.py
+++ b/sotodlib/io/imprinter.py
@@ -604,9 +604,9 @@ class Imprinter:
             raise BookBoundError(f"Book {bid} is already bound")
         assert book.type in VALID_OBSTYPES
 
-        # find appropriate binder for the book type
-        binder = self._get_binder_for_book(book, output_root=output_root)
         try:
+            # find appropriate binder for the book type
+            binder = self._get_binder_for_book(book, output_root=output_root)
             binder.bind(pbar=pbar)
 
             # write M_book file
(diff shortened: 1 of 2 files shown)


sotodlib/io/imprinter.py
```diff
@@ -148,6 +148,35 @@ class Imprinter:
...
+              g3tsmurf: g3tsmurf config file
+
+        The different tel-tubes will be bound into separate books containing 
+        only the stream_ids listed under slots. While the hk_sources take a 
+        g3tsmurf config key, that is just for symmetry, those files only need
+        to include a data_prefix and g3thk_db entry if only hk books are being 
+        created from that configuration. The g3tsmurf configurations are assumed
+        to be one per daq node, meaning this example could have the same config
+        file repeated several times.
+
```
mhasself, line 179: I assume the config file isn't documented elsewhere.  Questions to add answers for (?):

- These notes are helpful.  Maybe two paragraphs, first for "sources", then for hk_sources (pointing out the differences).
- What determines the number of slots in the slotmask (i.e. 111 vs 1111111)
- Are you supposed to put a None as a slot entry, to have a gap?
- The "tel-tube" and "daq-node" keys end up getting used for ... something.  In the code, what is that called?  Just "teltube"?  And that in turn becomes part of the book name.  (What I'm getting at is that "daq-node" sounds like the name of a computer when really it's just the label we're going to apply to all the data from some particular source.  Some clarification would be helpful.)

mhasself (approved):
I know you might need to press forward on this but I have some suggestions for the docs!

guanyilun (approved):
Looking good to me!

Issue #535, opened 2023-08-23
make-book continually retries on failed books


kmharrington:
There's one book from the SAT cooldown, `obs_1691876375_satp1_0111000` that takes about 2 hours for the bookbinder to fail on (related to this issue: https://github.com/simonsobs/sotodlib/issues/527). 

It looks like make-book is trying to re-run it every time, so the logic in make-book doesn't actually work out to "try twice then stop" like the comments suggest.

kmharrington:
fixed with #533

PR #538, opened 2023-08-25
make_book: not retry failed books and add option to specify book id

guanyilun:
- make_book pipeline will not automatically retry failed books now, related to #535 
- add an option to specify a particular id, this will make it easier to rebind a failed book later.

### DIFF
--- a/sotodlib/site_pipeline/make_book.py
+++ b/sotodlib/site_pipeline/make_book.py
@@ -2,10 +2,10 @@ from typing import Optional
 import traceback
 import argparse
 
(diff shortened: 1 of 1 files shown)


kmharrington (changes requested):
I came in and made a change on the last PR I merged to do this as a first pass. Going to be worth rebasing off of master. I also had to pull out the source kwarg because it was going to conflict with the changes needed for the HK books and the LATR OT detector books.

kmharrington:
already implemented elsewhere

PR #539, opened 2023-08-27
Add new fields to update_obsdb linked to scan amplitude and telescope flavors

AERAdler:
See commit eafca00 in branch obs-det-info-docs


tskisner:
I have a question about the obsdb column names related to the scan bounds.  [In a separate PR, I am adding code](https://github.com/simonsobs/sotodlib/pull/444/files#diff-ad7de80c6fae2b6914ea32408c806cb3990b9253c8bbd79bfb831a5aa91c723bR314-R333) from @mhasself that seems to indicate the intended names are az|el|roll "nom" and "span".  

Is there document somewhere with the full set of proposed columns in the obsdb?  Or maybe even just a google sheet with the column names, types, and description?  I know they are in flux, but just having a snapshot somewhere would be useful.

mhasself:
> Is there document somewhere with the full set of proposed columns in the obsdb? Or maybe even just a google sheet with the column names, types, and description? I know they are in flux, but just having a snapshot somewhere would be useful.

Ahh, thanks for asking ... see [this PR](https://github.com/simonsobs/sotodlib/pull/493) for the most up-to-date specification.

kmharrington (changes requested):
Ok, a couple things we'll need updated to get the running at the site. 

(1) Prefect needs the input arguments to be typed. You can look at `update_librarian` or `update_g3tsmurf_db` for example of this. 
(2) Prefect also needs the last argument in the function to be `logger=None`. You should then use this logger in place of all the print statements in the function. 

The last thing is for the LAT and also a question for @mhasself. I'm assuming we want to have one `ObsDb` for all the `LAT` optics tubes? If so, this script needs to be able to take a list of base directories instead of just a singular directory.

AERAdler:
Pinging the two of you, commit ee52045 should be prefect friendly now.

sotodlib/site_pipeline/update_obsdb.py
```diff
@@ -68,10 +74,28 @@ def main(config:str,
         Output verbosity. 0:Error, 1:Warning, 2:Info(default), 3:Debug
     overwrite : bool
         if False, do not re-check existing entries
+    logger : logging
+        When to output the print statements
+
+
     """
+
```
mhasself, line 82: The function arg logger now conflicts with the global "logger" variable.  We will need to investigate later whether all of this is really necessary ... but for now please fix it like this:
```
if logger is None:
    logger = globals()['logger']
else:
    globals()['logger'] = logger
```

mhasself (changes requested):
Looks good -- one thing to fix though.  Have you tested this on site books?

sotodlib/site_pipeline/update_obsdb.py
AERAdler: I'll fix and run a test on the site books. I hope I survive contact with data...

AERAdler: We're good now.

sotodlib/site_pipeline/update_obsdb.py
```diff
@@ -143,14 +175,48 @@ def main(config:str,
...
+            stream_file = os.path.join(bookpath,"*{}*.g3".format(stream_ids[0]))
+            stream = load_book.load_book_file(stream_file, no_signal=True)
+
+            for coor in ["az", "el", "boresight"]:
+                try:
+                    coor_enc = stream.ancil[coor+"_enc"]
+                    bookcartobsdb.add_obs_columns([f"{coor}_center float", 
+                                                   f"{coor}_throw float"])
+                    very_clean[f"{coor}_center"]=.5*(coor_enc.max()+coor_enc.min())
+                    very_clean[f"{coor}_throw"]=.5*(coor_enc.max()-coor_enc.min())
```
mhasself, line 215: Please add spaces around your operators.

sotodlib/site_pipeline/update_obsdb.py
```diff
@@ -29,12 +30,16 @@
...
+from sotodlib.io import load_book
 import os
 import glob
 import yaml
 import numpy as np
 import time
 import argparse
+import logging
+from sotodlib.site_pipeline import util
+logger = util.init_logger(__name__, 'update-obsdb: ')
```
mhasself, line 42: Move this code line below all imports (with at least one blank line before it).

sotodlib/site_pipeline/update_obsdb.py
```diff
@@ -102,20 +130,24 @@ def main(config:str,
...
-            _, book_id = os.path.split(dirpath)
-            if book_id in existing and not overwrite:
+    #Check if there are one or multiple base_dir specified
+    if isinstance(base_dir,str):
+        base_dir = [base_dir]
+    for bd in base_dir:
+        #Find folders that are book-like and recent
+        for dirpath,_, _ in os.walk(bd):
+            last_mod = max(os.path.getmtime(root) for root,_,_ in os.walk(dirpath))
+            if last_mod<tback:#Ignore older directories
```
mhasself, line 140: spaces after commas and around operators (`last_mod < tback`)

mhasself (approved):
Thanks.  Some minor formatting issues (please try to follow those principles in all sotodlib code) -- please fix those and then feel free to merge.

kmharrington (approved):

PR #544, opened 2023-09-04
Fix bug introduced in #415

tskisner:
Looks like this was just a typo where `h5py` was used instead of `h5py.File`.


skhrg:
Whoops yeah that was a bad copy and paste. I thought we fixed all of them but I guess one slipped through.

skhrg (approved):

PR #547, opened 2023-09-08
Bilinear mapmaking support

amaurea:
This pull request adds support for bilinear mapmaking to sotodlib. This functionality depends on the bilinear branch of so3g, but I've added some temporary scaffoling in this pull request to still make standard nearest neighbor mapmaking work while using the old version of so3g. This scaffolding is quite ugly, so I suggest we remove it as soon as possible. The scaffolding involves working around the shape of the threads/rangs/ivals (I wish this thing had a more consistent name in the code!) object having 4D in new so3g and 3D in the old one; as well as the Projectionist accepting a new interpol argument in the new version but not in the old version.

### DIFF
--- a/sotodlib/coords/pmat.py
+++ b/sotodlib/coords/pmat.py
@@ -85,6 +85,8 @@ class P:
     - det_weights (optional): weights (one per detector) to apply to
       time-ordered data when binning a map (and also when binning a
       weights matrix).  [dets]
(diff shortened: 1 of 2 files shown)


mhasself (changes requested):
(review text pending, standby)

mhasself (approved):
I tweaked a docstring to fix the docs build (you have to have blank lines around a nested list).  Thanks!

PR #548, opened 2023-09-08
Site Updates for Data Packaging

kmharrington:
* The librarian upload should definitely be throwing errors when if fails instead of just passing
* I realized the timecode finalization should be based on servers instead of on stream_ids because that's actually how the files are made. This is also now possible based on the "one per daq-node" setup that is now enforced with imprinter
* Added a function that will now tell us why the timecode books aren't being made.

### DIFF
--- a/sotodlib/site_pipeline/update_book_plan.py
+++ b/sotodlib/site_pipeline/update_book_plan.py
@@ -190,7 +190,7 @@ def get_parser(parser=None):
         '--update-delay-timecodes', type=float, 
         help= "Days to subtract from now to set as minimum ctime "
               "for timecode books",
-        default=1
+        default=7
     )
 
     return parser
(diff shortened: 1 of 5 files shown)


sotodlib/site_pipeline/update_book_plan.py
```diff
@@ -144,6 +178,21 @@ def get_parser(parser=None):
...
+    )
+    parser.add_argument(
+        '--max-ctime-timecodes', type=float, 
+        help="Maximum creation time for timecode books"
+    )
+    parser.add_argument(
+        '--update-delay-timecodes', type=float, 
+        help= "Days to subtract from now to set as minimum ctime "
+              "for timecode books",
+        default=1
```
guanyilun, line 193: should be 7?

guanyilun (approved):

sotodlib/site_pipeline/update_book_plan.py
kmharrington: yup, thanks for the catch

Issue #525, opened 2023-08-17
Shorthand for common wcs_kernels


mhasself:
wcs_kernels are used in sotodlib.coords to represent pixelization schemes (a way of dividing up the sky into a rectangular grid of pixels) without specifying a full map footprint.  It is basically a WCS representing a single pixel map; with the idea that a mapmaking code can continue the pixelization and create a footprint that contains the data (but is pixel-aligned with any other map that did the same continuation).

Although the full space of pixelizations supported by FITS / WCS is quite complex, we mostly deal with one of these:

- For source-centered maps, a TAN projection centered at "ra, dec" = (0, 0), with square pixels (res_x = res_y).
- For sky maps, CAR projection (in the natural orientation), with pixel edges crossing ra, dec = (0, 0), and pixels that are square on the equator.

It would be handy to be able to construct these common kernels  of constructing these kernels by passing in a simple string, e.g.
```
  wcsk = coords.get_wcs_kernel('car-0.5arcmin')
  wcsk = coords.get_wcs_kernel('tan-0.25deg')
```
Moreover it would be nice to be able to pass those strings into functions that want a wcs_kernel ... for one-liners like:
```
  p = coords.P.for_tod(tod, wcs_kernel='car-0.5arcmin')
```

tanaybhandarkar:
@mhasself are these the only resolutions we would care about? Or would we potentially want something more general, like the option to pass something like `car-1.0arcmin` too?

mhasself:
I think we'd want at least a few of these.  You could start with [1, 1.5, 2, 3, 4, 5, 6, 8, 10, 12, 15, 20, 25, 30, 40, 45] arcsec / arcmin / deg, with deg probably only going up to 10.

Fine with me to just parse the string and extract the resolution; though there's something nice about making it hard for people to have resolutions of, like, 1.23452 arcmins or whatever.

mhasself:
Closed via #552 , thanks.

PR #552, opened 2023-09-21
Change get_wcs_kernel to be more stringent

tanaybhandarkar:
I've tried to make changes to `get_wcs_kernel` as pointed out in [issue 525.](https://github.com/simonsobs/sotodlib/issues/525)  I'm not sure this is what was envisioned, but I've tested it on an various projections and resolutions and it yields expected results. The function will throw an error in cases where the beam is too large, or doesn't come from a pre-approved set of sizes. 

If this is acceptable, I can propagate the change into `get_footprint`, which is the only function that I can find that will need to be changed.

### DIFF
--- a/sotodlib/coords/helpers.py
+++ b/sotodlib/coords/helpers.py
@@ -221,16 +221,15 @@ def get_wcs_kernel(proj, ra=None, dec=None, res=None):
 
     Args:
       proj (str): Either the name of a FITS projection to use (e.g. 'car',
-                  'cea', 'tan'), in which case "res" must also be specified,
-                  or a string containing both the projection and the
-                  resolution, in this format:
+        'cea', 'tan'), in which case "res" must also be specified, or a
+        string containing both the projection and the resolution, in this
+        format:
 
                       proj-res
 
-                  where proj is a projection type (e.g. 'car', 'tan'
-                  'gnom') and res is the resolution, in appropriate
-                  units (deg, arcmin or arcsec).Examples of acceptable
-                  args:
+        where proj is a projection type (e.g. 'car', 'tan, 'gnom') and res
+        is the resolution, in appropriate units (deg, arcmin or arcsec).
+        Examples of acceptable args:
 
                       'car-0.5deg'
                       'tan-3arcmin'
--- a/sotodlib/coords/helpers.py
+++ b/sotodlib/coords/helpers.py
@@ -210,7 +210,7 @@ def get_horiz(tod, wrap=False, dets=None, timestamps=None, focal_plane=None,
         tod.wrap(wrap, output, [(0, 'dets'), (1, 'samps')])
     return output
 
-def test_wcs_kernel(proj, ra=None, dec=None, res=None):
+def get_wcs_kernel(proj, ra=None, dec=None, res=None):
     """Construct a WCS.  This fixes the projection type (e.g. CAR, TAN),
     centered with respect to a reference point (at ra,dec = (0,0)), and 
     resolution of a pixelization, without specifying a particular grid 
--- a/sotodlib/coords/helpers.py
+++ b/sotodlib/coords/helpers.py
@@ -222,7 +222,9 @@ def get_wcs_kernel(proj, ra=None, dec=None, res=None):
     Args:
       proj (str): Name and resolution of projection to be used. It
                   must adhere to the following format:
+
                       proj-res
+                  
                   where proj is a projection type (e.g. 'car', 'tan'
                   'gnom') and res is the resolution, in appropriate
                   units (deg, arcmin or arcsec).
--- a/sotodlib/coords/helpers.py
+++ b/sotodlib/coords/helpers.py
@@ -280,7 +280,7 @@ def get_footprint(tod, wcs_kernel, dets=None, timestamps=None, boresight=None,
     big enough to contain all data from tod.  Returns (shape, wcs).
 
     """
-    if type(wcs_kernel) == str:
+    if isinstance(wcs_kernel, str):
         wcs_kernel = get_wcs_kernel(wcs_kernel)
 
     dets = _valid_arg(dets, tod.dets.vals, src=tod)
--- a/sotodlib/coords/helpers.py
+++ b/sotodlib/coords/helpers.py
@@ -224,14 +224,16 @@ def get_wcs_kernel(proj, ra=None, dec=None, res=None):
                   must adhere to the following format:
 
                       proj-res
-                  
+
                   where proj is a projection type (e.g. 'car', 'tan'
                   'gnom') and res is the resolution, in appropriate
-                  units (deg, arcmin or arcsec).
-                  Examples of acceptable args:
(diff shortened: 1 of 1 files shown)


mhasself:
Good start  -- please leave the old interface in place, though (that will make the tests pass again).  The idea was to still be flexible about wcs kernels but to make it easier to declare simple ones.  Perhaps rewrite the function signature as
```
get_wcs_kernel(proj, ra=None, dec=None, res=None)
```
and then detect that user is using a shorthand because only proj has been set.

Given that you've set it up to parse out the values in a flexible way, I think you don't have to be so strict about the allowed values.

String parsing is actually way more fun if you learn python regular expressions.  Try this:
```
import re
m = re.match('(?P<proj>car|tan)-(?P<val>[0-9]*.?[0-9]*)(?P<unit>arcsec|arcmin|deg)', 'car-0.5arcmin')
print(m.groupdict())
```

There are various places in the coords code that require a wcs_kernel as an argument, and I was imagining altering those functions so that if a string is passed in it will automatically convert it with `wcs_kernel = get_wcs_kernel(wcs_kernel)`.

tanaybhandarkar:
Thanks for the comments. I thought about using regular expressions, but didn't see the `re` library in the imports- before I keep going, are there any rules about libraries in this repo? I don't see anything about it in the developer guidelines, so can I import something if I need it?

mhasself:
I am definitely sensitive to adding new requirements.  If it's pure python and on PyPI, it's not much of a concern.  Weird big compiled packages are to be avoided unless essential.

re is in the standard library tho

sotodlib/coords/helpers.py
```diff
@@ -228,11 +243,33 @@ def get_wcs_kernel(proj, ra, dec, res):
     Returns a WCS object that captures the requested pixelization.
 
     """
-    assert np.isscalar(res)  # This ain't enlib.
-    _, wcs = enmap.geometry(np.array((dec, ra)), shape=(1,1), proj=proj,
-                            res=(res, -res))
+    if len(proj) > 3:
+        m = re.match('(?P<proj>car|tan)-(?P<res>[0-9]*.?[0-9]*)(?P<unit>arcsec|arcmin|deg)', proj)
```
mhasself, line 247: If the expression doesn't match, then m will be None.  Check for that, and raise a ValueError saying the string didn't parse.  Otherwise the next line will error and the user will just be confused.

sotodlib/coords/helpers.py
```diff
@@ -209,16 +210,30 @@ def get_horiz(tod, wrap=False, dets=None, timestamps=None, focal_plane=None,
...
+    resolution of a pixelization, without specifying a particular grid
+    of pixels.
 
     This interface is subject to change.
 
     Args:
-      proj (str): Name of the projection to use, as processed by
-        pixell.  E.g. 'car', 'cea', 'tan'.
+      proj (str): Name and resolution of projection to be used. It
+                  must adhere to the following format:
```
mhasself, line 224: The old thing is still allowed, so should still be documented.  "Either the name of a FITS projection to use (e.g. 'car', 'cea', 'tan'), in which case "res" must also be specified, or a string containing both the projection and the resolution, in this format:"...

sotodlib/coords/helpers.py
```diff
@@ -228,11 +243,33 @@ def get_wcs_kernel(proj, ra, dec, res):
...
+
+        # Convert to radians
+        res = res * np.pi/180.
+        if unit == 'arcmin':
+            res /= 60.
+        if unit == 'arcsec':
+            res /= 3600.
+
+        if unit == 'deg' and res > 10:
+            raise ValueError("Beam is too big")
```
mhasself, line 260: I think remove this... no need to police.

sotodlib/coords/helpers.py
```diff
@@ -228,11 +243,33 @@ def get_wcs_kernel(proj, ra, dec, res):
...
+
+        _, wcs = enmap.geometry(np.array((0, 0)), shape=(1, 1), proj=proj,
+                                res=(res, -res))
+
+    else:
+        assert np.isscalar(res)  # This ain't enlib.
+        _, wcs = enmap.geometry(np.array((dec, ra)), shape=(1, 1), proj=proj,
+                                res=(res, -res))
+
     return wcs
```
mhasself, line 270: Before returning, tweak reference pixel offset to 0.5.  But only for CAR projection.  This makes it Fejer1 compatible (good for SHTs).
```
if proj == 'car':
    wcs.wcs.crpix = [1.0, 0.5]
```

mhasself (changes requested):
Thanks.  A few more notes :)

sotodlib/coords/helpers.py
```diff
@@ -220,8 +220,10 @@ def get_wcs_kernel(proj, ra=None, dec=None, res=None):
     This interface is subject to change.
 
     Args:
-      proj (str): Name and resolution of projection to be used. It
-                  must adhere to the following format:
+      proj (str): Either the name of a FITS projection to use (e.g. 'car',
+                  'cea', 'tan'), in which case "res" must also be specified,
+                  or a string containing both the projection and the
+                  resolution, in this format:
```
mhasself, line 226: This formatting is technically acceptable but it is the convention in this file to indent the description only by 2 spaces.  i.e.
```
    Args:
      proj (str): Either the name of a FITS projection to use (e.g. 'car',
        'cea', 'tan'), in which case "res" must also be specified,
```

mhasself (changes requested):
One last comment then please convert to "ready for review".

mhasself (changes requested):
You promised to update get_footprint ... please do so :)

Almost there!

tanaybhandarkar:
> You promised to update get_footprint ... please do so :)
> 
> Almost there!

I sure did! I'm assuming that the only two options we want for `wcs_kernel` in `get_footprint` is a string like we use above, or an actual wcs object. Let me know if there should be more flexibility.

sotodlib/coords/helpers.py
```diff
@@ -228,17 +244,45 @@ def get_wcs_kernel(proj, ra, dec, res):
...
+
 def get_footprint(tod, wcs_kernel, dets=None, timestamps=None, boresight=None,
                   focal_plane=None, sight=None, rot=None):
     """Find a geometry (in the sense of enmap) based on wcs_kernel that is
     big enough to contain all data from tod.  Returns (shape, wcs).
 
     """
+    if type(wcs_kernel) == str:
+        wcs_kernel = get_wcs_kernel(wcs_kernel)
+
```
mhasself, line 285: Python people try not to be too strict about classes.  We don't need this to be a str, we just need it to be sufficiently str-like.  I.e. any subclass of a str is acceptable.  So ... replace the check with `if isinstance(wcs_kernel, str)`.

This is not likely to ruin anyone's day in this case, but generally it's a good practice when all you're trying to do is distinguish between broad classes.

tanaybhandarkar: Makes sense. Done

mhasself (approved):

Issue #536, opened 2023-08-24
Speed up glitch flags


kmharrington:
This line takes a very long time to run: https://github.com/simonsobs/sotodlib/blob/2d6e0467bf2457b0f3a71a84217069c36802c42f/sotodlib/flags.py#L98 

We need to add some logic to look at how long the arrays are and only use a fraction of the array to get these statistics when they are very long.

mhasself:
@tanaybhandarkar Here's one.

tanaybhandarkar:
I have been playing around with the downsampling option that @msilvafe suggested. Right now, `get_glitch_flags` takes 30-50s to run on a AxisManager of shape (2580, 264000). About 10s is spent on the Fourier filtering, 20-40s on the `stats.iqr` step. To study the effect of downsampling, I just ran the `stats.iqr` step on the calculated `fvec` and looked at the distributions.  I also had to set `merge=False`, so I'm not sure how much time that step takes. 

![image](https://github.com/simonsobs/sotodlib/assets/28115016/5f59fb2d-d349-4640-8b65-b15232c4e66d)

Here, "ds" refers to the downsampling factor (i.e., ds= 50 means I used every 50th point: `fvec[::50]`). ds=1 means no downsampling, and `t` is elapsed time. Based on the distribution, ds=2 seems to remain fairly close to the original, non-downsampled distribution, yet only takes half as much time.  

I think setting ds=2 would help a lot here, and I could add a check to downsample when arrays are larger than some number.

mhasself:
I want to advocate for a higher ds factor.  Like, 20.  Actually I think you should set the downsampling (or decimation, really) so that the vector you pass in to stats.iqr is small-but-not-too-small.  In this example 264000  / 20 is 13200.  So you could pick ds so that your sampling vector is ~20000 points.

I think a few percent bias or noise in iqr_range is acceptable.  Really what we're trying to compute here is something proportional to the standard deviation (but with outlier rejection).  It shouldn't take many data points to estimate that number.  Think of the stdev of a white noise timestream -- the error goes down as 1/sqrt(Npts)).  You should be able to get a 1% estimate (which is better than we need) using only 10000 points.

mhasself:
Addressed in #555

PR #555, opened 2023-10-05
Faster glitch flagging

tanaybhandarkar:
Based on the discussions in issue #536, I have put in a catch to supply a decimation factor for the Fourier filtered signal if there are over 50,000 samples in the AxisManager. The decimated array will now be around 20,000 samples, which will speed up the `stats.iqr` step.

### DIFF
--- a/sotodlib/flags.py
+++ b/sotodlib/flags.py
@@ -95,7 +95,10 @@ def get_glitch_flags(
     )
     # get the threshods based on n_sig x nlev = n_sig x iqu x 0.741
     fvec = np.abs(fvec)
-    iqr_range = 0.741 * stats.iqr(fvec, axis=1)
+        if fvec.shape[1] > 50000:
+        ds = int(fvec.shape[1]/20000)
+    else: ds = 1
+    iqr_range = 0.741 * stats.iqr(fvec[:,::ds], axis=1)
     # get flags
     msk = fvec > iqr_range[:, None] * n_sig
     flag = RangesMatrix([Ranges.from_bitmask(m) for m in msk])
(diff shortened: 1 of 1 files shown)


msilvafe:
That's great, roughly how much speedup do you get when you still estimate the iqr reasonably?

sotodlib/flags.py
```diff
@@ -95,7 +95,10 @@ def get_glitch_flags(
     )
     # get the threshods based on n_sig x nlev = n_sig x iqu x 0.741
     fvec = np.abs(fvec)
-    iqr_range = 0.741 * stats.iqr(fvec, axis=1)
+    if fvec.shape[1] > 50000:
+        ds = int(fvec.shape[1]/20000)
+    else: ds = 1
```
mhasself, line 100: two lines please

mhasself (changes requested):
This is fine except for I don't like "if a: b" and "else: c" on one line!

tanaybhandarkar:
> That's great, roughly how much speedup do you get when you still estimate the iqr reasonably? 


Without this adjustment, on a 264000 sample AxisManager, it takes about 60-90s for `get_glitch_flags` to fully run. When adding this downsampling option, it cuts that down to 20s. 

Caveat: There's a lot of variation on the timing, even when running the function without changing anything. I've chalked this up to a simons1 computing quirk. I never saw a 5 min runtime, like you mentioned to me earlier- if you know which context file you were working with, I can test with that one, too.

mhasself (approved):
Looks good to me but let Max or Katie confirm before merging.

msilvafe:
Looks good to me I'm happy for you to merge.

PR #556, opened 2023-10-05
Add logger import.

msilvafe:
When the `sotodlib.site_pipeline.preprocess_tod` method `_build_pipe_from_configs` is called and the config dictionary includes a process name which doesn't have a corresponding class defined in `sotodlib.preprocess.processes` a log warning should be raised but currently that logging is not imported properly so that log statement breaks the code. I now added that import and tested that it works.

### DIFF
--- a/sotodlib/site_pipeline/preprocess_tod.py
+++ b/sotodlib/site_pipeline/preprocess_tod.py
@@ -6,7 +6,8 @@ import argparse
 from sotodlib import core
 import sotodlib.site_pipeline.util as sp_util
 from sotodlib.preprocess import _Preprocess, PIPELINE, processes
-
+import logging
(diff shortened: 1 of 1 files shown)


mhasself:
Unfortunately the logger will sometimes be passed into the main function (e.g. if this is being called through prefect).

That means the module global logger can be set when the module loads (like you did here, though maybe look at whether you should use utils like main() does), BUT the main function must be able to override that global variable, before doing stuff.

Here's a demo of how you can do it:
```
logger = 'hi'

def update_logger(logger):
    globals()['logger'] = logger

print(logger)
update_logger('test')
print(logger)
```

It's not great, but it's what we're working with.

mhasself (approved):
Looks like it will work, thanks.

PR #563, opened 2023-10-13
allow passing loggers into some databasing classes for working with prefect

guanyilun:
The issue with logging in prefect is that, although prefect loggers are passed in as an argument to the flow, they didn't get passed all the way into each class that does the actual work such as G3tsmurf or Imprinter classes. This PR is a step towards fixing it.

### DIFF
--- a/sotodlib/io/bookbinder.py
+++ b/sotodlib/io/bookbinder.py
@@ -18,7 +18,7 @@ if not log.hasHandlers():
     init_logger('bookbinder')
 
 
-def setup_logger(logfile=None):
+def setup_logger(logfile=None, log=None):
     """
     This setups up a logger for bookbinder. If a logfile is passed, it will
     write to that file as well as stdout. It is useful to create a one-off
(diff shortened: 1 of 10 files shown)


mhasself:
I am very uncomfortable with having to propagate an external logger to *every* place where we might want to capture output from a script.  This is practically disqualifying, in my mind.  (I realized the other day that it was a very bad sign that we *needed* to pass in a logger to our entry points...)

Does the feature described [here](https://github.com/PrefectHQ/prefect/pull/7580) help us at all?  I really think we need a system that captures stdout/stderr without having to propagate prefecty objects inside our pipeline code.  If we have to write our own wrappers to capture and pipe out the output, that is more appealing to me than having to tunnel an external logger deep into our libraries.

guanyilun:
ok that's a fair point. I can experiment with that

kmharrington:
what's the status of this? has it been supplanted?

guanyilun:
this is no longer being worked on. We used a different logging solution, as pointed out by Matthew.

mhasself:
> this is no longer being worked on. We used a different logging solution, as pointed out by Matthew.

I don't recall learning that!

Does that mean we don't have to pass "logger" in anymore?  Because that would make the site_pipeline scripts a lot easier to write.

guanyilun:
yes, I think we can plan to phase it out

kmharrington:
right now the setup has things still throwing errors if logger isn't the last kwarg. what is it that's requiring that?

guanyilun:
logger no longer required

PR #564, opened 2023-10-13
Add function to derive the az el of planets

tterasaki:
image

To visualize when boresight hits the planet, I want a function to get the trajectory of planet.

### DIFF
--- a/sotodlib/coords/planets.py
+++ b/sotodlib/coords/planets.py
@@ -236,15 +236,61 @@ def filter_for_sources(tod=None, signal=None, source_flags=None,
         tod.wrap(wrap, signal, [(0, 'dets'), (1, 'samps')])
     return signal
 
+def _get_astrometric(source_name, timestamp, site='_default'):
+    """
+    Derive skyfield's Astrometric object of a celestial source at a 
+    specific timestamp and observing site, which is used to derive
+    radec/azel in get_source_pos/get_source_azel.
+    
+    Note that it will download a 16M ephemeris file on first use.
+    
+    Args:
+      source_name: Planet name; in capitalized format, e.g. "Jupiter",
+        or fixed source specification.
+      timestamp: unix timestamp.
+      site (str or so3g.proj.EarthlySite): if this is a string, the
+        site will be looked up in so3g.proj.SITES dict.
 
+    Returns:
+      astrometric: skyfield's astrometric object
+    """
+    # Get the ephemeris -- this will trigger a 16M download on first use.
+    de_url = RESOURCE_URLS['de421.bsp']
+    de_filename = au_data.download_file(de_url, cache=True)
+
+    planets = jpllib.SpiceKernel(de_filename)
+    for k in [
+            source_name,
+            source_name + ' barycenter',
+    ]:
+        try:
+            target = planets[k]
+            break
+        except (ValueError, KeyError):
+            pass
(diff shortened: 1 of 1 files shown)


sotodlib/coords/planets.py
```diff
@@ -297,6 +305,58 @@ def get_source_pos(source_name, timestamp, site='_default'):
...
+    """
+    Get the apparent azimuth and elevation of a celestial source at a specific timestamp and observing site.
+
+    This function calculates the apparent azimuth and elevation of a celestial source, accounting for the geographical position
+    on Earth but assuming no atmospheric refraction. Note that it will download a 16M ephemeris file on first use.
+
+    Args:
+        source_name: Planet name; in capitalized format, e.g. "Jupiter"
+        timestamp (float): The Unix timestamp representing the time for 
+        which to calculate azimuth and elevation.
```
mhasself, line 319: indent "which..."

sotodlib/coords/planets.py
```diff
@@ -297,6 +305,58 @@ def get_source_pos(source_name, timestamp, site='_default'):
...
+        site (str or so3g.proj.EarthlySite): if this is a string, the
+        site will be looked up in so3g.proj.SITES dict.
```
mhasself, line 321: indent "site will ..."

sotodlib/coords/planets.py
```diff
@@ -297,6 +305,58 @@ def get_source_pos(source_name, timestamp, site='_default'):
...
+        site = so3g.proj.SITES[site]
+
+    observatory = site.skyfield_site(planets)
+
+    timescale = skyfield_api.load.timescale()
+    sf_timestamp = timescale.from_datetime(
+        datetime.datetime.fromtimestamp(timestamp, tz=skyfield_api.utc))
+    amet0 = observatory.at(sf_timestamp).observe(target)
+    el, az, distance = amet0.apparent().altaz()
+    return az.to(units.rad).value, el.to(units.rad).value, distance.to(units.au).value
```
mhasself, line 357: Up to this point, this function is basically identical to get_source_pos.  Please either:
- move the duplicated code into a helper function, for both get_source_pos and get_source_azel to use, OR
- simply add an argument to get_source_pos that returns az,el instead of ra,dec (and you can have get_source_azel be an alias that calls into get_source_pos)

mhasself (changes requested):
Thanks -- looks useful.  However, please eliminate the code duplication.  The code in get_source_pos is carefully written, and when you copy-paste that elsewhere you are setting us up for divergence, and different bugs and their fixes ending up in two places instead of one.

Also ... in the future, please do not include bulk reformatting along with content changes. 
Bulk reformatting should be avoided unless the situation is really bad.  When reformatting is deemed necessary, it should be in a commit by itself, and probably in a PR by itself.  That will help people navigate PRs and inspect the commit history in an efficient way.

tterasaki:
Thank you for your thoughtful comments. Yeah you are rights.
Now I made a helper function `_get_astrometric` which is used by both get_source_pos and get_source_azel.

sotodlib/coords/planets.py
mhasself: Still need spaces here!

mhasself (approved):
Thanks, just fix the docstring where noted and then it's ok to merge.

PR #566, opened 2023-10-14
use joinedload instead of lazyload to speed up db queries in hk

guanyilun:
improve performance of hk agent loading by eager loading all children fields.


kmharrington (approved):
Thanks!

PR #567, opened 2023-10-15
Save make_position_match animations

skhrg:
Just a minor change to allow users to save the animations produced while matching. May be useful now that we about to start running on real data.


mhasself (approved):
Thanks -- approving but please:
- update description of "vis" in the docstring
- mention there, and/or in the config example, whether the output directory needs to exist already, or not.

PR #569, opened 2023-10-17
ObsDb error handling for site books

kmharrington:
This PR adds a few things 

* fixes the miss-ordering between `stream_ids` and `detsets`. which were causing the creation of a lot of weird and wrong detsets.
* adds an option in check book to ignore the requirement of this ordering being the same
* adds an option in check book to remove fixed tones from detector lists
* adds options to just skip adding failed books to an obsfiledb / obsdb. 

Using all of these, I have now obsfiledb and obsdb making running on the site uploaded books.

### DIFF
--- a/sotodlib/site_pipeline/update_obsdb.py
+++ b/sotodlib/site_pipeline/update_obsdb.py
@@ -91,7 +91,9 @@ def main(config:str,
         for col, typ in config_dict["obsdb_cols"].items():
             col_list.append(col+" "+typ)
         bookcartobsdb.add_obs_columns(col_list)
-
+    if "skip_bad_books" not in config_dict:
+        config_dict["skip_bad_books"] = False
+        
     #How far back we should look
     tnow = time.time()
     if recency is not None:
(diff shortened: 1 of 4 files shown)


sotodlib/site_pipeline/update_obsdb.py
```diff
@@ -116,8 +118,16 @@ def main(config:str,
...
+
+            try:
+                #obsfiledb creation
+                checkbook(bookpath, config, add=True, overwrite=True)
+            except Exception as e:
+                if config_dict["skip_bad_books"]:
+                    print(f"failed to add {bookpath}")
+                    continue
+                else:
+                    raise e
```
jlashner, line 130: Is there a way we can mark failed books so we don't continuously retry them?

jlashner (approved):
Looks good to me, if this gets us an obsdb I'm happy with it for now. I think we will probably want to mark bad books somewhere so we don't continuously retry them forever, but we may also just be happy with running with using a min_ctime in the update_obsdb script.

sotodlib/site_pipeline/update_obsdb.py
kmharrington: here, no. But I'm actively working on adding running of check_book to the end of the book binding process so hopefully they fail then when we're actually tracking fails

PR #572, opened 2023-10-18
Changes to sim_catalog.py

tanaybhandarkar:
`get_beam_map` would crash because it was incorrectly reading in h5 files. `observe_catalog` crashed due to unit errors. Both of these issues have been fixed here.

### DIFF
--- a/sotodlib/toast/ops/sim_catalog.py
+++ b/sotodlib/toast/ops/sim_catalog.py
@@ -247,7 +247,7 @@ class SimCatalog(Operator):
             # We have already read the single beam file.
             beam_dic = self.beam_props["ALL"]
         else:
-            with h5py(self.beam_file, 'r') as f_t:
+            with h5py.File(self.beam_file, 'r') as f_t:
                 beam_dic = {}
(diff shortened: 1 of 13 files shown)


sotodlib/toast/ops/sim_catalog.py
```diff
@@ -438,7 +438,7 @@ def _observe_catalog(
                 x = phi_diff * np.cos(np.pi / 2 - det_theta[hit])
                 y = det_theta[hit] - source_theta
                 # Rotate into the beam frame
-                psi = det_psi[hit] - det_psi_pol.to_value(u.rad)
+                psi = det_psi[hit] - det_psi_pol.to_value(u.rad**2)
```
keskitalo, line 441: This is incorrect.  Please remove `* u.deg` on line 317 instead

keskitalo (changes requested):
Can we drop sotodlib/ml_mapmaking/mapmaker.py from this PR?

tanaybhandarkar:
> Can we drop sotodlib/ml_mapmaking/mapmaker.py from this PR?

Yeah I will do that, that was mistakenly staged for this.

PR #573, opened 2023-10-18
Fixes to units and syntax in sim_catalog

tanaybhandarkar:
Ok, perhaps this time I haven't slaughtered a branch. 

`get_beam_map` would crash because it was incorrectly reading in h5 files. `observe_catalog` crashed due to unit errors. Both of these issues have been fixed here.


keskitalo (approved):
Thanks a lot for fixing these issues!

Issue #479, opened 2023-05-26
analyze-bright-ptsrc todo


mhasself:
In lieu of a full gating review of #416, here are some to-do items for analyze-bright-ptsrc.  Please comment here as they get fixed.

1. The script should consume a config file, where things like default fitting parameters (including their dependence on telescope and  tube_type) can live.  A few things provided on command line should instead have defaults set in the config file.
2. Documentation (there is a stub now) should included an example config file and a description of the output structures.
3. The --ctx_idx command line argument should be changed to mandatory argument obs_id.  (While obs_id should normally be provided as a string obs_id, the ability to accept an integer and convert that to an index in the obsdb is pretty useful for testing... so ok to keep that part.)
4. The program still requires telescope and band from the command line ... the telescope should be inferred from the obsdb.  The band is not known; only tube_type (e.g. SAT_MF) will be known, and available in obsdb.
5. sso_name should also be inferred, at least by default, from the obsdb.
6. Like other elements, this one should be set up to analyze and write out analysis per-detset (a.k.a. per-wafer).  The filename and dataset name need to be somewhat configurable or adaptable to that.
7. Remove hard-coded paths to files on simons1.  Where needed, files should be specificed in config file.
8. The functions tod_sim_* are very similar and should be condensend to a single function.  When calling curve_fit, lambdas or a similar simple function wrapper should be used instead.
9. The abscal stuff is beyond the scope of this script, and should be made optional.  Especially given that one should not assume to know the band center at this time.

mhasself:
:musical_note: _chime_ :musical_note:  Please turn to page #736.

PR #574, opened 2023-10-20
Analyze bright pt src updates

nad-dch:
Updates 1,3,4,5,7,8,9 from previous issue #479. For the 2nd bullet point of the required updates, an example file is included at https://github.com/simonsobs/pwg-scripts/blob/master/pwg-bcp/sotodlib_staging/data/initial_parameters_fit_pt_src.yaml.  Related to item number 6: does it make sense to simply add a parsing argument for the `detset` argument of `core.context.get_obs()` or do we assume we are completely agnostic as to which wafers/tubes are included in the observation until we load it?

### DIFF
--- a/sotodlib/site_pipeline/analyze_bright_ptsrc.py
+++ b/sotodlib/site_pipeline/analyze_bright_ptsrc.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+import matplotlib.pyplot as plt
 import so3g
 from so3g.proj import coords, quat
 import sotodlib.coords.planets as planets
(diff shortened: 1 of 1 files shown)


sotodlib/site_pipeline/analyze_bright_ptsrc.py
```diff
@@ -290,214 +226,301 @@ def fit_params(
...
+    ax.plot(np.degrees(xi_hw), np.degrees(eta_hw), '.')
+
+    csl = so3g.proj.CelestialSightLine.az_el(
+        tod.timestamps, tod.boresight.az, tod.boresight.el, weather="typical")
+    q_bore = csl.Q
+    planet = planets.SlowSource.for_named_source(sso, tod.timestamps[0] * 1.)
+    ra0, dec0 = planet.pos(tod.timestamps[0])
+    planet_q = so3g.proj.quat.rotation_lonlat(ra0, dec0)
+    q_total = ~q_bore * planet_q
+    xi_total, eta_total, _ = quat.decompose_xieta(q_total)
```
kmharrington, line 400: it looks like nothing in this math accounts for boresight rotation of the telescope. (I'm noticing it while looking at these plots for a scan of Jupiter that has non-zero boresight).

mhasself (approved):
Apologies that this sat for such a long time.

I've captured some next steps in #574; but I propose we get this merged and keep working on it from here.

PR #579, opened 2023-10-24
g3thwp debug with SCR1 data

yukisakurai:
I have checked the HWP site highbay cooldown data (SCR1). 
There are relatively large number of packet drops compared with lab. data. 
This required additional exception handling and debugging of packet drop filling in G3tHWP.
Specifically, the following functions have been added and modified:
_process_counter_index_reset
_fill_dropped_packets
I have verified that the angle output is stable even with packet drops at least SCR1 (timestamp until 1692000000). 
Once HWP rotates, I plan to check the latest SATP1 data as well.
(FYI: IRIG timing shift is fixed from Oct. 11th)

### DIFF
--- a/sotodlib/hwp/g3thwp.py
+++ b/sotodlib/hwp/g3thwp.py
@@ -811,7 +811,7 @@ class G3tHWP():
         
         # check packet drop
         self._encoder_packet_sort()
-        self._find_dropped_packets()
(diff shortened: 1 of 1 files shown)


kmharrington (approved):
Seems good to me, hopefully get a time to test this on site soon!

PR #582, opened 2023-10-25
Sub filter

atakeuchi634:
Two kinds of subscan poly-filters are added in sotodlib.tod_ops.sub_polyf.py .

One is subscan_polyfilter_onlystat(), which is a poly-filter applied only into non-turnaround part.
The other is subscan_polyfilter(), which is a polyfilter applied into entire TOD.

In order to find subscan ranges, three methods are prepared.
- get_turnaround_flags()
  - based on sotodlib.flags.get_turnaround_flags
  - function that returns boolean array
- get_turnaround_flags_by_dazdt()
  - function that returns boolean array
  - turnaround part is determined looking at daz/dt
- get_leftright_array()
  - function that returns list of [start, end] of left/right  scans 

I found that turnaround mask is defined looking at azimuth itself in current sotodlib.
I guess this can be a problem in raster scans.
I added "get_turnaround_flags_by_dazdt()" in sotodlib.flags.
(But this should be modified)

example : 
"/so/scratch/atakeuchi/231025_dev_subpolyfilter.ipynb"


tterasaki:
This is just a suggestion, but can we use a step-function-like matched filter and `scipy.signal.find_peaks` as a more robust turnaround detection?

I have written an rough example in `/so/tterasaki/MF1/2310/231025_atsuto_gitreview.ipynb`.
image


```Python
# Matched filter + find_peaks
az = aman.boresight.az
daz = np.diff(az)

# derive approx daz
approx_daz = np.mean(daz[daz>np.percentile(daz, 95)])

# derive approx aprox daz/dt and rough number of samples in one_scan period 
dt = np.mean(np.diff(aman.timestamps))
approx_dazdt = approx_daz / dt
approx_samps_onescan = int(np.ptp(az) / approx_dazdt / dt)

# Make a step-function-like matched filter
kernel_size = 400
kernel = np.zeros(kernel_size)
# normalize with approx_daz so that peaks height be ~1
kernel[:kernel_size//2] = 1/approx_daz
kernel[kernel_size//2:] = - 1/approx_daz
kernel /= kernel_size

# convolve signal with the kernel
matched = np.convolve(np.diff(aman.boresight.az), kernel, mode='same')
matched[:kernel_size] = 0
matched[-kernel_size:] = 0 # We have to treat the initial/last samples properly later

# find peaks in matched daz, height=0.1, distance=approx_samps_onescan//3 would be robust enough
from scipy.signal import find_peaks
peaks, properties = find_peaks(matched, height=0.1, distance=approx_samps_onescan//3)
```

sotodlib/tod_ops/sub_polyf.py
```diff
@@ -0,0 +1,190 @@
+import numpy as np 
+import matplotlib.pyplot as plt 
+
+from sotodlib import flags
+
+
+
+def get_turnaround_flags_by_dazdt(
+    tod, qlim=5.,az=None
```
kmharrington, line 9: I'm not a fan of making copies of these functions that do the same math with different outputs. Especially not with flags / ranges objects. `ranges.mask()` will get you the boolean array for the flag and that code is written in C++ so it should be very fast. I don't see a reason to have these copies.

atakeuchi634: Thanks for comments and a helpful information. 
I have not known that we have function `ranges.mask()`.
Now I deleted copies,  and that function is used in the current version.

atakeuchi634:
Thank you for your comments, Tomoki and Katie.

I modified scripts based on your comments.
Main updates are : 
- turn-around ranges are searched for using matched filter.
- copies of turnaround flagging functions in sub_poly.py are deleted.
- left/right flagging function (get_leftright_flags_by_dazdt) is added.

You can find a jupyter with which I tested the scripts in 
/so/scratch/atakeuchi/231026_dev_subpolyfilter.ipynb

Any comments and suggestions are welcome.
Thanks!

sotodlib/tod_ops/sub_polyf.py
```diff
@@ -0,0 +1,122 @@
...
+    
+    # negative range
+    negative_indices = np.where(signs <= 0)[0]
+    ranges = np.split(negative_indices, np.where(np.diff(negative_indices) != 1)[0]+1)
+    index_ranges_n = [(rng[0], rng[-1]) for rng in ranges]
+    
+
+    index_ranges_p.extend(index_ranges_n)
+    
+    return index_ranges_p
```
tterasaki, line 30: Now, you have flag of right and left scans. So this function should be removed.

sotodlib/tod_ops/sub_polyf.py
```diff
@@ -0,0 +1,122 @@
...
+    starts = np.insert(ones[np.where(diff != 1)[0]], 0, ones[0])
+    ends = np.append(ones[np.where(diff != 1)[0] - 1], ones[-1])
+    indices = list(zip(starts, ends))
+    indices = [(start, end) for start, end in indices if (end - start + 1 >= _min)]
+
+    return indices
+
+
+def subscan_polyfilter_tamasked(aman, degree, signal='signal', default_flag=False, wrap_flag=True) : 
+
```
tterasaki, line 49: We often use a flags=None as default. and 
```Python
if flag is None:
    flag = aman.flags.turnarounds
```
With this configuration, users can use there own flags to mask.

tterasaki: So I prefer to merge `subscan_polyfilter_tamasked` and `subscan_polyfilter`.
And masking for glitch should be available for this function.

sotodlib/flags.py
```diff
@@ -14,6 +15,222 @@
...
+    # derive approx daz
+    approx_daz = np.mean(daz[daz>np.percentile(daz, 95)])
+
+    # derive approx aprox daz/dt and rough number of samples in one_scan period 
+    dt = np.mean(np.diff(aman.timestamps))
+    approx_dazdt = approx_daz / dt
+    approx_samps_onescan = int(np.ptp(az) / approx_dazdt / dt)
+
+    # Make a step-function like matched filter
+    kernel_size = 400
```
tterasaki, line 149: Make kernel_size an arguments.

sotodlib/flags.py
```diff
@@ -14,6 +15,222 @@
...
+    # search for edges of square waves looking at the peaks
+    for p in peaks : 
+        start = prev
+        end = p
+        length = p-prev
+        middle = start + length//2
+        prev = p
+
+        left[start+nbuffer:middle-nbuffer] = True
+        right[middle+nbuffer:end-nbuffer] = True
```
tterasaki, line 192: Let's find negative peaks instead.

sotodlib/flags.py
```diff
@@ -14,6 +15,222 @@
...
+    
+    # treatment of the begining
+    if first_is_positive : 
+        left[:peaks[0]+nbuffer] = False
+        right[:peaks[0]+nbuffer] = False
+    else : 
+        right[int(peaks[0]-mean_length//2):peaks[0]-nbuffer] = True
+        left[:peaks[0]+nbuffer] = False
+        
+    if w_turnaround :
```
tterasaki, line 210: Add discription for `w_turnaround`

sotodlib/flags.py
```diff
@@ -14,6 +15,222 @@
...
+            raise ValueError("Flag name {} already exists in aman.flags".format(name))
+        elif name in aman.flags:
+            aman.flags[name] = ta_flag
+        else:
+            aman.flags.wrap(name, ta_flag)
+    return ta_flag
+    
+
+def get_leftright_flags_by_dazdt(
+    aman, t_buffer=1.8, w_turnaround=True, az=None, merge=True, overwrite=False, name=["left", "right"]
```
tterasaki, line 118: Can we integrate `get_leftright_flags` and `get_turnaround_flags` since some of lines are duplicated.

sotodlib/tod_ops/sub_polyf.py
```diff
@@ -0,0 +1,122 @@
...
+    
+
+    index_ranges_p.extend(index_ranges_n)
+    
+    return index_ranges_p
+
+
+
+def get_range_list(ta_flag,_min=0):
+
```
tterasaki, line 35: Change the name of function. Start from '_' (underscore).

sotodlib/tod_ops/sub_polyf.py
```diff
@@ -0,0 +1,122 @@
...
+    
+    starts = np.insert(ones[np.where(diff != 1)[0]], 0, ones[0])
+    ends = np.append(ones[np.where(diff != 1)[0] - 1], ones[-1])
+    indices = list(zip(starts, ends))
+    indices = [(start, end) for start, end in indices if (end - start + 1 >= _min)]
+
+    return indices
+
+
+def subscan_polyfilter_tamasked(aman, degree, signal='signal', default_flag=False, wrap_flag=True) :
```
tterasaki, line 48: Can we remove the `_tamasked`. Instead give flag of mask to be masked as an argument?

sotodlib/tod_ops/sub_polyf.py
```diff
@@ -0,0 +1,122 @@
...
+        signal: Which tod is used
+
+    Returns:
+        -
+    """   
+    
+    time = aman.timestamps
+    time = time - time[0]
+    
+    ranges_list = flags.get_leftright_array(aman)
```
kmharrington, line 102: this is no longer in flags

atakeuchi634:
Thanks both.
 
In the latest scripts, non-necessary functions, arguments, variables are removed or integrated.
In the new ones, we can 
-  select the method to define subscan (az or daz/dt, use entire TOD or non-turnarounds parts only), 
-  mask some specified parts in TODs in filtering (e.g. point source timing, glitches, etc...)

msilvafe:
> Thanks both.
> 
>  
> 
> In the latest scripts, non-necessary functions, arguments, variables are removed or integrated.
> 
> In the new ones, we can 
> 
> -  select the method to define subscan (az or daz/dt, use entire TOD or non-turnarounds parts only), 
> 
> -  mask some specified parts in TODs in filtering (e.g. point source timing, glitches, etc...)

Thanks for the work on this! If you're ready for review can you convert this from draft to full PR?

msilvafe:
> @kmharrington FYI I marked this as ready for review because I think all of the substantive docs + tod processing updates have been made now. The docs build test is failing, I can build the docs on simons1 without an issue but I can reproduce the failure when I try to build in a fresh venv. My last few commits are implementing some changes to the github workflow code for the docs build test that @BrianJKoopman suggested a couple weeks ago but that didn't seem to work so I'll have to work with him (or someone else familiar with the github workflow infrastructure) to fix the build before we merge.

I got the docs build test to pass by removing the `-W` in the docs build portion of the workflow which failed the test if a warning is raised. But I still don't understand why that particular warning is raised for this module: 

```WARNING: Failed to get a function signature for sotodlib.tod_ops.sub_polyf: <module 'sotodlib.tod_ops.sub_polyf' from '/home/msilvafe/repos/sotodlib/sotodlib/tod_ops/sub_polyf.py'> is not a callable object```

I'll leave it up to you if you ok merging with this mystery.

kmharrington (approved):
This is a massive set of updates that are super useful and I'm glad to see being added into `tod_ops`. So well done all! 

I'm approving this on the condition that the one issue I've found gets fixed before the merge. That's that the in_place option doesn't actually work as written. Assume you get that fixed feel free to merge!

tterasaki (approved):

kmharrington (approved):

PR #583, opened 2023-10-25
Reorganize toast workflow tools

tskisner:
This extracts large groups of functionality from the monolithic toast workflows and enables their use from interactive sessions.


tskisner:
This is now ready for review @keskitalo .  Since the current monolithic workflows have no unit tests, I tested the refactored high-level workflow (`toast_so_sim`) on a small dataset.  Although not part of this PR, I am separately making some notebooks in `pwg-tutorials` that use these new workflow tools.  Attached here is an export of a test notebook showing how these same workflow tools can be used interactively.  This saves us a lot of duplicated boilerplate.

[workflows.pdf](https://github.com/simonsobs/sotodlib/files/13198900/workflows.pdf)

tskisner:
Some offline discussion is ongoing about this PR, so converting back to a draft.

sotodlib/toast/workflows/proc_quality.py
```diff
@@ -0,0 +1,242 @@
...
+        )
+        job_ops.filtered_statistics.apply(data)
+        log.info_rank(
+            "Calculated filtered statistics in", comm=data.comm.comm_world, timer=timer
+        )
+        job_ops.mem_count.prefix = "After filtered statistics"
+        job_ops.mem_count.apply(data)
+
+
+def setup_hn_map(operators):
```
keskitalo, line 106: The `hn` and `cadence` maps are more about data characterization than quality.

sotodlib/toast/workflows/sim_optical_pickup.py
```diff
@@ -0,0 +1,106 @@
...
+    timer.start()
+
+    # Configured operators for this job
+    job_ops = job.operators
+
+    if job_ops.sim_hwpss.enabled and job_ops.sim_ground.hwp_angle is not None:
+        job_ops.sim_hwpss.hwp_angle = job_ops.sim_ground.hwp_angle
+        job_ops.sim_hwpss.detector_pointing = job_ops.det_pointing_azel
+        job_ops.sim_hwpss.stokes_weights = job_ops.weights_azel
+        log.info_rank("Running HWPSS simulation...", comm=data.comm.comm_world)
```
keskitalo, line 100: I like it when the workflow reports both when if begins an operation and when it ends. That makes it more apparent to a general user what is going on. However, since SO is acting as a model for these workflow tools, we should pick one convention and stick to it.

sotodlib/toast/workflows/so_convert.py
keskitalo, line 1: Are we happy with the workflows and their building blocks residing in the same directory?

sotodlib/toast/workflows/so_map.py
```diff
@@ -0,0 +1,253 @@
...
+
+    wrk.setup_load_data_hdf5(operators)
+    wrk.setup_load_data_books(operators)
+    wrk.setup_load_data_context(operators)
+
+    wrk.setup_pointing(operators)
+    wrk.setup_simple_noise_models(operators)
+    wrk.setup_flag_noise_outliers(operators)
+
+    wrk.setup_filter_hwpss(operators)
```
keskitalo, line 170: This is obviously infinitely better than the old, super long workflow script. I still wonder, if we could set up all the canonical data reduction operators with a single `wrk.setup_data_reduction()`? I'm concerned that when we develop a new data reduction operator, we will have to edit a large number of workflows that all do data reduction.

keskitalo (approved):
This is a very valuable improvement to our workflow building.

I made a bunch of suggestions in the review but none of them are critical.

There are lots of unused timers in workflow tools. I wonder if we could define a helper method function or a decorator that handled timing and memory profiling without so many repeated lines in the workflow methods.

sotodlib/toast/workflows/data.py
```diff
@@ -0,0 +1,217 @@
...
+        job (namespace):  The configured operators and templates for this job.
+        otherargs (namespace):  Other commandline arguments.
+        runargs (namespace):  Job related runtime parameters.
+        data (Data):  The data container.
+
+    Returns:
+        None
+
+    """
+    raise NotImplementedError("LoadContext operator not yet merged.")
```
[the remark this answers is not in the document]
tskisner: Currently the setup is a no-op, but sure, can add that.

sotodlib/toast/workflows/proc_mapmaker_filterbin.py
```diff
@@ -0,0 +1,90 @@
...
+    timer.start()
+
+    # Configured operators for this job
+    job_ops = job.operators
+
+    job_ops.filterbin.binning = job_ops.binner_final
+    job_ops.filterbin.det_data = job_ops.sim_noise.det_data
+    job_ops.filterbin.output_dir = otherargs.out_dir
+
+    if job_ops.filterbin.enabled:
```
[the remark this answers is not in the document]
tskisner: The reason I did this was so that the operator was configured with information from other operators, even if the actual run was disabled.  In principle one could then re-enable and run it, although agreed that would likely not be done very often.

sotodlib/toast/workflows/proc_quality.py
tskisner: Fair point, I can put those in a separate file.

keskitalo: Or just rename this since noise estimation is also characterization

sotodlib/toast/workflows/sim_optical_pickup.py
tskisner: Yes, my intention was to have a log message at the start of each thing and at the end with the timing info.  I'll go back through all the functions and make sure that pattern is followed.

sotodlib/toast/workflows/so_convert.py
tskisner: No strong preference- I tried to make these follow their own naming prefix, but they could be placed anywhere.  Maybe `sotodlib.toast.scripts`?

sotodlib/toast/workflows/so_map.py
tskisner: In the model where we have a few giant scripts doing everything, there are a couple places to edit.  In the model where we have many smaller scripts doing specific things, my thought was that these would have a very specific set of simulation or reduction steps and we would not necessarily add new operators to everything (otherwise we would end up with the giant script model).  Taking this to the logical extreme, one could write a single workflow that just took a bunch of config files and executed things in a particular order and then stop writing scripts and just edit toml/json files.  I was aiming for some middle ground, where we can describe (different) types of high-level data flow in each script.

sotodlib/toast/workflows/so_convert.py
keskitalo: I think the workflows are in an intuitive place but perhaps the helper methods could be in `toast/workflow_tools` or `toast/workflows/tools`?

sotodlib/toast/workflows/so_map.py
keskitalo: Fair enough. So in this model we anticipate a relatively small number of workflows that could require synchronous editing when the canonical set of data reduction operators is augmented.

tskisner: I think we will have both.  Some people prefer a giant script that does everything with pieces enabled / disabled by config files.  Others prefer a script with only the steps for a specific use case (e.g. SAT transfer function with loading from a context and calling wrapped preprocessing, or simulated observing with 3 different mapmakers running on the same data, etc).  So in the do-everything-script case there are just a couple places to add new things.  In the other case there might be one or two specific scripts that need updating, but not all of them.

Perhaps later in the project, when there is a more stable reduction sequence, we could make another helper function that wraps even more things.  Or if we find many workflows have the same sequence of things, we could extract that out.  I am just advocating not doing that until it becomes clear what those blocks are.

sotodlib/toast/workflows/data.py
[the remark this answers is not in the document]
tskisner: Done

sotodlib/toast/workflows/proc_mapmaker.py
```diff
@@ -0,0 +1,171 @@
...
+    """
+    templates.append(
+        toast.templates.Offset(
+            name="baselines",
+            step_time=2.0 * u.second,
+        )
+    )
+    templates.append(
+        toast.templates.Periodic(
+            name="sss",
```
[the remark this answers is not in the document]
tskisner: Changed to "azss"

sotodlib/toast/workflows/proc_mapmaker_filterbin.py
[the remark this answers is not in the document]
tskisner: I ended up implementing exactly what you suggest- the important point that I missed is the case where multiple workflow functions are coupled together and expect their traits to be set consistently.  So it is safer to not set traits based on other operators unless we need to.

sotodlib/toast/workflows/proc_quality.py
tskisner: Done

sotodlib/toast/workflows/so_convert.py
tskisner: I moved the entrypoints (not all of which use the workflow tools) to `sotodlib/toast/scripts/`.

sotodlib/toast/workflows/sim_optical_pickup.py
tskisner: I implemented a decorator with logs the start / end time of each workflow function and the memory usage if it is enabled.  In some cases, when a workflow function is running several operators, I left the sub-timing information since it is useful.

keskitalo:
Looks great!

PR #584, opened 2023-10-25
fix bug in sim_source

chervias:
Fixing a bug that allows to use scan_type=fixed


tskisner:
It looks like there are some git history things going on here.  The drone_scans PR should probably be first rebased and merged before any more changes.  That was waiting on me to test some WCS coordinate swapping in toast, but I would vote for just doing the rebase and merge now instead of waiting.  @gabrielecoppi does that sound ok?  @chervias was this PR just intended to fix that one source file?

chervias:
Hi Ted, yes I'm trying to add a couple of lines to `sim_source.py` in the `drone_scans` branch. I have a fork in my username, which I use for SAT mapmaking stuff. I created a branch in that fork for drone work, I added and committed the single file `sim_source.py`, but now in the pull request is trying to push everything that has changed since the last update to the drone_scans branch.

PR #587, opened 2023-10-26
SMuRF detset info

kmharrington:
Long awaited first set of smurf related caldbs. I'm also starting this script with the assumption that I'll get expanded later. 

Right now this copies over all the interesting stream_id, wafer/tube-slot info, and channel tuning info and sets things up for it to load the metadata as desired. 

Test it out for yourself at: `'/so/scratch/kmharrin/202310_smurf_detsets/context.yaml'`


mhasself (approved):
Hooray!

PR #588, opened 2023-10-28
`selectinload` instead of `joinedload` for performance improvement.

dixilo:
Using `selectinload` seems much faster than `joinedload` for querying small amount of entries out of the big database.


mhasself:
@guanyilun Please comment on whether you think this is an acceptable change or not.

guanyilun:
I think this is acceptable.

kmharrington (approved):
This solves a current issue with the data packaging. thanks for working this out, sorry the review took forever.

PR #589, opened 2023-10-30
Transforming mapmaking.py into a module

chervias:
The code in mapmaking.py has been split into several sub-modules and mapmaking is a proper module now. Some dependencies might be broken, I tried to get all of them.


mhasself (approved):
Thanks!

PR #590, opened 2023-10-30
Det-match code and docs

jlashner:
Adds module for the det-matching algorithm, and corresponding documentation. This code relies on functions from the finalize-focal-plane branch, so that would have be merged first.


sotodlib/coords/det_match.py
```diff
@@ -0,0 +1,670 @@
...
+                )
+
+                if int(d['smurf_channel']) != -1:
+                    r.smurf_channel = int(d['smurf_channel'])
+
+                if fp_pars is not None:
+                    x, y = float(d['det_x']), float(d['det_y'])
+                    xp = x * np.cos(theta) - y * np.sin(theta) + dx
+                    yp = x * np.sin(theta) + y * np.cos(theta) + dy
+                    xi, eta, gamma = fp_func(None, x=xp, y=yp, pol=0)
```
skhrg, line 267: The LAT and SAT focal plane functions don't have the same signature as the LAT one also needs the path to the data file from Zemax. Maybe do a `functools.partial` to take care of that?

sotodlib/coords/det_match.py
```diff
@@ -0,0 +1,670 @@
...
+                    is_north=is_north
+                )
+
+                if int(d['smurf_channel']) != -1:
+                    r.smurf_channel = int(d['smurf_channel'])
+
+                if fp_pars is not None:
+                    x, y = float(d['det_x']), float(d['det_y'])
+                    xp = x * np.cos(theta) - y * np.sin(theta) + dx
+                    yp = x * np.sin(theta) + y * np.cos(theta) + dy
```
skhrg, line 266: There is a function in the `optics` module for this, might as well use it to keep duplicated code to a minimum

sotodlib/coords/det_match.py
```diff
@@ -0,0 +1,670 @@
...
+        m = bgs_mismatch & (~bgs_unassigned)
+        mat[m] += self.match_pars.assigned_bg_mismatch_pen
+
+        # If pointing, add cost if assigned too far
+        dd = np.sqrt(
+              (ra1['xi'][:, None] - ra2['xi'][None, :])**2 \
+            + (ra1['eta'][:, None] - ra2['eta'][None, :])**2)
+        m = ~np.isnan(dd)
+        c = np.exp((np.abs(dd / self.match_pars.dist_width)) ** 2)
+        mat[m] += np.exp((np.abs(dd[m] / self.match_pars.dist_width)) ** 2)
```
skhrg, line 460: c is unused right?

skhrg (commented):
Looks good to me.

Not sure if `coords` is the most intuitive place for it to live.  It could reasonably live in the root of sotodlib (ie: `sotodlib.det_match`) but maybe that is a Matthew decision.

My one generic complaint is that I have no intuition for the cost values so I would probably have a hard time tuning penalties. One of the reasons I like the likelihood approach it they have a well defined finite scale of 1, but costs can be infinite. I think it still makes sense to use the cost based approach anyways, but I think this warrants good bookkeeping on what params were used in different match scenarios so that people have good data points to start from.

sotodlib/coords/det_match.py
```diff
@@ -0,0 +1,670 @@
...
+import numpy as np
+import matplotlib.pyplot as plt
+from scipy.optimize import linear_sum_assignment
+
+from sotodlib.coords import optics
+from sotodlib.core import metadata
+from sotodlib.io.metadata import write_dataset
+
+warnings.filterwarnings(action='ignore',
+                        message='overflow encountered in exp*', category=RuntimeWarning)
```
mhasself, line 16: Can you use a more targeted context manager instead of just turning these off for everything?

sotodlib/coords/det_match.py
```diff
@@ -0,0 +1,670 @@
...
+            The resonator to take design properties from
+        in_place (bool):
+            If True, the src_res will be modified. Otherwise, a new Resonator
+            will be created and returned.
+    """
+    if in_place:
+        r = src_res
+    else:
+        r = deepcopy(src_res)
+    r.det_x = src_res.det_x
```
mhasself, line 102: I think you want dst_res on RHS of all the lines from here down, right?  Either that or the construction of r is wrong...

sotodlib/coords/det_match.py
```diff
@@ -0,0 +1,670 @@
...
+    position and polarization from the other.
+
+    Args:
+        src_res (Resonator):
+            The resonator to take smurf properties from
+        dst_res (Resonator):
+            The resonator to take design properties from
+        in_place (bool):
+            If True, the src_res will be modified. Otherwise, a new Resonator
+            will be created and returned.
```
mhasself, line 96: The naming here is src and dst, reading as source and destination, but apparently "in_place" modifies src, not dst...

So I think maybe these should be called smurf_res or meas_res, and design_res...

mhasself (changes requested):
Partial review ... this is very interesting so probably more is coming ...

jlashner:
Thanks for the review saianeesh, I'll add this to the docs, but to comment on the intuition behind param-tuning, I hardly ever touch anything other than `freq_offset`, `freq_width`, and `dist_width`, which is essentially just directly setting at what freq/dist differences I want the cost function to start blowing up. Other params generally don't need to be touched, and if they do its kind of a sign that the existing data is not good enough to match with.

mhasself (approved):

docs/det_match.rst
```diff
@@ -0,0 +1,92 @@
...
+
+Usage
+--------
+Below is an example of how to match a resonator-set based on a smurf tune-file
+and sodetlib bgmap to a resonator-set based on a handmade solution file,
+and save the output.
+
+.. code-block:: python
+
+    from sotodlib.io.coords import det_match as dm
```
tterasaki, line 32: Was cyberstoking,
from sotodlib.io.coords import det_match as dm
>>
from sotodlib.coords import det_match as dm

sotodlib/coords/det_match.py
jlashner: Sorry I don't really see what function you're referring to... I see that `SAT_focal_plane` takes "mapping_data" that might be used for this, but not exactly sure how to use it or what its doing.

I think I'd rather just keep this here for now since its a simple transformation, but we can generalize later if we want.

skhrg: Was talking about [`ufm_to_fp`](https://github.com/simonsobs/sotodlib/blob/f82b2823e874bd6f0532375c5e4f4ec5eea2bc0b/sotodlib/coords/optics.py#L277). But its not a big deal. I have (unmerged) [a function](https://github.com/simonsobs/sotodlib/blob/c75cacd5ae4f487ac0c88c1fa3b4996636b35c30/sotodlib/coords/optics.py#L559) that provides a single interface for both the LAT and SAT that should eventually replace stuff like this, but that can wait until I finish that branch.

docs/det_match.rst
jlashner: Thanks tomoki!

jlashner:
Sorry taking so long to get back to this, tried to address some of @skhrg's comments, but wasn't super successful. It's kind of difficult to make this properly support the LAT without an example of how to use the focal_plane functions. I think since it is going to be a while since we have LAT pointing, it might be worth just merging this and fixing the LAT focal-plane function later when we need it.

Do we think this is ok to merge?

skhrg:
To use it with the LAT you just need something like 
```
ufm_to_fp_pars = get_ufm_to_fp_pars(telescope, slot, config_path)
x_fp, y_fp, pol_fp = ufm_to_fp(aman, x=x, y=y, pol=pol, **ufm_to_fp_pars)
xi, eta, gamma = LAT_focal_plane(
            aman, zemax_path, x=x_fp, y=y_fp, pol=pol_fp, rot=rot, tube=tube
        )
```

But yeah we won't have LAT pointing for a while and by then I should get a more convenient interface merged in so I think its fine to punt for now.

docs/index.rst
```diff
@@ -24,6 +24,7 @@ Welcome to SOTODLib's documentation!
    site_pipeline.rst
    reference.rst
    dev.rst
+   det_match.rst
```
mhasself, line 27: Move this to be above site_pipeline please

docs/det_match.rst
```diff
@@ -0,0 +1,92 @@
+.. py:module:: sotodlib.coords.det_match
```
mhasself, line 1: Please add some autofunction and/or autoclass at the bottom of this file.  See tod_ops.rst for lots of examples

This not only dumps in the docstrings, directly, it also provides links to the source code (which is often what one ultimately needs, to understand the significance of some parameter / attribute).

mhasself (changes requested):
Two things about the docs to make this perfect ...

mhasself (approved):
Approving but do fix the docs build

PR #591, opened 2023-10-30
Jump improvements

skhrg:
Improvements to the algorithm to get more jumps on the first pass and performance improvements.

### DIFF
--- a/sotodlib/tod_ops/jumps.py
+++ b/sotodlib/tod_ops/jumps.py
@@ -85,10 +85,8 @@ def _jumpfinder(x, min_chunk, min_size, win_size, nsigma, max_depth=-1, depth=0)
     if np.all(msk):
         return jumps.reshape(orig_shape)
 
-    # TODO: Use msk to save time when only a few chans may have jumps
-
     # Take cumulative sum, this is equivalent to convolving with a step
-    x_step = np.cumsum(x, axis=-1)
+    x_step = np.cumsum(x[msk], axis=-1)
 
     # Smooth and take the second derivative
     sg_x_step = np.abs(sig.savgol_filter(x_step, win_size, 2, deriv=2, axis=-1))
@@ -138,7 +136,7 @@ def _jumpfinder(x, min_chunk, min_size, win_size, nsigma, max_depth=-1, depth=0)
     jump_rows = jump_rows[size_cut]
     jump_cols = jump_cols[size_cut]
 
-    jumps[jump_rows, jump_cols] = True
+    jumps[np.flatnonzero(msk)[jump_rows], jump_cols] = True
 
     # If no jumps found return
     if not np.any(jumps):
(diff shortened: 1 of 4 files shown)


msilvafe:
When ready to convert from draft to ready for review tag me. I would like to test this + review it.

skhrg:
So lesson from all this is that: iterating over things that are vectorized >>> recursing over things that are not (although the latter uses less memory). 

Test script can be found at `/so/home/saianeesh/code/scratch/jump_test.py`

sotodlib/tod_ops/jumps.py
```diff
@@ -61,58 +62,111 @@ def _jumpfinder(x, min_chunk, min_size, win_size, max_depth=-1, depth=0, **kwarg
...
+    x = np.atleast_2d(x)
 
-    # If std is basically 0 no need to check for jumps
-    if np.isclose(x.std(), 0.0) or (std_est(x) == 0):
-        return np.array([], dtype=int)
+    jumps = np.zeros(x.shape, dtype=bool)
+    if x.shape[-1] < min_chunk:
+        return jumps.reshape(orig_shape)
+
+    size_msk = (np.max(x, axis=-1) - np.min(x, axis=-1)) < min_size
```
msilvafe, line 77: np.ptp(x, axis=-1)

kmharrington:
I grabbed a random "CMB" observation and looked at running this jump detector on it like 
```
j, fixed = jumps.find_jumps(
    aman, min_sigma=10, max_iters=2, 
    fix=jumps.jumpfix_median_sub, inplace=False
)
```

It's detecting a whole lot of jumps that aren't jumps even with `min_sigma=10`

Script / details here: https://site.simonsobs.org/~kmharrin/scratch/202311_tod_proc_reviews/test_jump_fixes.py

Another thing I notice is that this is running off a point in sotodlib before we had changed the book loading units to phase. So if anything in here is dependent on units and was tuned off level 2 data you should rebase off current master branch.

kmharrington:
On further review, it looks like what's not working is the `in_place=False` argument.

kmharrington:
yeah I believe that's what happened

skhrg:
Hmm, the inplace thing doesn't seem to be the issue. Does that data happen to be a source scan? I am catching a lot of things that are much larger than the noise that look like this 
image

I can get rid of these by making the window larger or using `min_size` rather than `min_sigma`. The window thing can be risky because if you make it too big something like this can look like a jump
image

Maybe something like `min_size = .5*np.ptp(signal, axis=-1)` would be a good way of dealing with this case? If we recompute the ptp on each iter then it could work fine even on long TODs with several jumps (may need more iters or use of the sliding window jumpfinder).

skhrg:
Update: Looks like at least some of the errors come from floating point errors when numbers get too big. Doing a median subtraction ahead of time fixes this

skhrg:
Ack no, something went funny in that merge, will try to fix but may just make a new branch

msilvafe:
Ok @kmharrington, I have been testing this in person with @skhrg and I'm ready for this to merge from my end.

kmharrington (approved):
Excited to get this into the pipeline!

skhrg:
Ok going to clean up the commit history and then merge

PR #595, opened 2023-11-01
Linearize times in bookbinder if Low Precision

jlashner:
This PR changes the bookbinder such that if data has Low-precision (software) timestamps, it will linearize it using the frame-counter.

This fixes issues that we're having binding LAT data, which is running without a timing system, which breaks the bookbinder, since it detects and tries to fill gaps where there are none.

This has been tested on `obs_1698478281_lati6_100`, which is currently failing.

### DIFF
--- a/sotodlib/io/bookbinder.py
+++ b/sotodlib/io/bookbinder.py
@@ -320,6 +320,8 @@ class SmurfStreamProcessor:
 
         self.nframes = 0
         ts = []
+        smurf_frame_counters = [] # smurf frame-counter
+        fc_idx = None
         frame_idxs = []
         frame_idx = 0
         for frame in get_frame_iter(self.files):
@@ -330,6 +332,10 @@ class SmurfStreamProcessor:
             if self.nchans is None:
                 self.nchans = len(self.readout_ids)
                 self.primary_names = frame['primary'].names
+                for i, n in enumerate(self.primary_names):
+                    if n == "FrameCounter":
+                        fc_idx = i
+                        break
                 self.bias_names = frame['tes_biases'].names
                 self.timing_paradigm = frame['timing_paradigm']
                 self.session_id = frame['session_id']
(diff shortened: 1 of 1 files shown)


sotodlib/io/bookbinder.py
```diff
@@ -330,6 +332,10 @@ def preprocess(self):
             if self.nchans is None:
                 self.nchans = len(self.readout_ids)
                 self.primary_names = frame['primary'].names
+                for i, n in enumerate(self.primary_names):
+                    if n == "FrameCounter":
+                        fc_idx = i
+                        break
```
mhasself, line 338: Does `fc_idx = list(self.primary_names).index('FrameCounter')` not work?

mhasself (approved):
Makes sense.  Will that log message show up in the Z_bookbinder.log?

sotodlib/io/bookbinder.py
```diff
@@ -340,14 +346,25 @@ def preprocess(self):
...
+        self.smurf_frame_counters = np.hstack(smurf_frame_counters)
         self.frame_idxs = np.hstack(frame_idxs)
 
+        # If low-precision, we need to linearize timestamps in order for
+        # bookbinder to work properly
+        if self.timing_paradigm == 'Low Precision':
+            self.log.info(
+                "Timestamps are Low Precision, linearizing from frame-counter"
+            )
+            fsamp, offset = np.polyfit(self.smurf_frame_counters, self.times, 1)
```
mhasself, line 365: Maybe not `fsamp` ... it's sample_period, or step, or dt or something.

sotodlib/io/bookbinder.py
jlashner: yep that works

sotodlib/io/bookbinder.py
jlashner: done

jlashner:
Ya, the log statement will end up in stdout along with the logfile

Issue #562, opened 2023-10-12
depth1 mapmaking - a "preprocess" module that bins a TOD into a map set


mhasself:
As suggested by @tskisner during DRP telecon.

tskisner:
The reason I asked about this issue (it was more of a question than a suggestion) was in the context of a toast `SitePipeline` operator that I am designing.  In this work I had been planning to use a preprocess "Pipeline" and its config on each toast process to operate on its local set of detectors.  This assumes that the pre-processing is a set of serial operations.  Since your mapmaker things are MPI parallel, this would need some additional logic to see if a preprocess operator was using MPI or not and if so, run that collectively with the toast process group working on a given observation.  It sounds like I should at least add a stub for that possibility (that some pre-process ops use MPI).

mhasself:
The mapmaker does not naturally split single observations across different MPI tasks (because, in the Max-L model anyway, single-obs detectors are correlated and need to be accessible in the single task).   Rather, MPI is used in multi-observation mapmaking (especially iterative mapmaking).

So in this problem, where we're incorporating a single obs (or a single obs-wafer) into a map, MPI would not be implicated (at least not in the current plan).  I certainly had no plans to run preprocess pipelines in MPI with an observation split across tasks.

Indeed a sensible boundary for problems that preprocess pipelines should solve is "stuff you could do with a single observation, in a single process".

So I wouldn't prioritize making this MPI aware.

PR #597, opened 2023-11-02
SAT mapmaking

chervias:
This PR includes 3 main things regarding SAT mapmaking
 - Create new sub-module `demod_mapmaker.py` in `mapmaking`. These classes are based on the original `MLmapmaker` classes but work with a 3-timestreams demodulated TOD. Also a new model of white noise in `noise_model.py`
 - New sub-module `tod_process.py` that groups the tods for mapping. The functions here create a dictionary that is fed to the depth-1 mapmaker for making maps, but also there are modes to map _per observation_ or by _time interval_. This function is meant to be used in the SAT mapmaker as well as the LAT depth-1 mapmaker, decoupling the TOD grouping from the mapmakers.
 - The filter+bin mapmaker for the SAT as a script to be run within the `so-site-pipeline`.


chervias:
Now I have made a few modifications to run with real data from SATp1 in site compute11

chervias:
This PR also addresses issue #562 with the `mapmaking/tod_grouping.py` functions.

PR #599, opened 2023-11-03
update-obsdb: sanitize tags list (no empty strings)

mhasself:
Also removed a (now) unnecessary fork in how `update_obs` is called.


kmharrington (approved):

AERAdler (approved):

kmharrington:
I found the bookbinder one as well and added that in.

PR #601, opened 2023-11-07
Adding the maximum likelihood mapmaker to Site-Pipeline

tanaybhandarkar:
Integrating the ML Mapmaker into site-pipeline. User can either provide a `config.yaml` file, or use flags through the CLI. There are also default values that are assumed for the non-critical arguments. 

Some things to work on before we can merge:

1. Docs- I can add a section to `site-pipeline.rst` if that's the correct place to do so. I'll add an example `config.yaml` file there too.
2. I had to add a `default=True` to `--nocal` [here](https://github.com/simonsobs/sotodlib/blob/a987bbf86ec61ca6e6284be3ea7ad1092d079b9c/sotodlib/site_pipeline/ml_mapmaking.py#L36). If this is not set to `True`, when [we loop through ](https://github.com/simonsobs/sotodlib/blob/a987bbf86ec61ca6e6284be3ea7ad1092d079b9c/sotodlib/site_pipeline/ml_mapmaking.py#L86-L89)the `args` dictionary, `nocal` is set to `False` unless explicitly called out. This will crash the script later on when no `relcal` is provided.
3. I have not tested every flag, partially because I'm not sure how best to test them. If problems arise with a flag, maybe the procedure should be to raise an issue.


sotodlib/site_pipeline/ml_mapmaking.py
```diff
@@ -0,0 +1,312 @@
+from argparse import ArgumentParser
+import numpy as np, sys, time, warnings, os, so3g
+from sotodlib.core import Context, AxisManager, IndexAxis
+from sotodlib import mapmaking
+from sotodlib.io import metadata   # PerDetectorHdf5 work-around
+from sotodlib import tod_ops
+from sotodlib.tod_ops import filters
+from pixell import enmap, utils, fft, bunch, wcsutils, mpi
+#import moby2.analysis.socompat
```
mhasself, line 9: You can just delete this line.  It was in here so that context system would be able to load ACT data properly, but that has a different work-around now (should the need arise).

sotodlib/site_pipeline/ml_mapmaking.py
```diff
@@ -0,0 +1,312 @@
...
+
+
+def main(config_file=None, **args):
+    cfg = _get_config(config_file)
+    
+    # Certain fields are required. Check if they are all supplied here
+    required_fields = ['query','freq','area','odir','prefix']
+    for req in required_fields:
+        if req not in cfg.keys():
+            raise KeyError("Please supply a {}".format(req))
```
mhasself, line 58: It looks to me like:
- the config_file must exist (so can't use cli only?)
- the "required_fields" must all be specified in the config file (I can't specify --freq on command line instead?)

I think those should both be loosened.  At the end of the process of reading a config file, and processing command line args ... check that all the things you need are defined.

That error message is a bit vague -- supply how?
-

sotodlib/site_pipeline/ml_mapmaking.py
```diff
@@ -0,0 +1,312 @@
...
+
+    defaults = {"comps":"T",
+                "ntod": None,    
+                "tods": None,
+                "nset": None,
+                "nmat": "corr",
+                "max_dets": None,
+                "verbose": 0,
+                "quiet": 0,
+                "center-at": None,
```
mhasself, line 68: center_at

mhasself (commented):
Couple comments inline.

It would probably be a good idea to:
- move the "defaults" to root level in the submodule, so it can be accessed directly
- make the argument processing, where default + config_file + args are processed, a separate function.  That will simplify testing and declutter main() a bit.

sotodlib/site_pipeline/make_ml_map.py
```diff
@@ -0,0 +1,322 @@
...
+    for key, item in args.items():
+        if item is not None:
+            cfg[key] = item
+    # Certain fields are required. Check if they are all supplied here
+    required_fields = ['query','freq','area','odir','prefix','context']
+    for req in required_fields:
+        if req not in cfg.keys():
+            raise KeyError("{} is a required argument. Please supply it in a config file or via the command line".format(req))
+
+    args = cfg
```
tanaybhandarkar, line 101: To clarify the intended workflow here:
1. Load in the default values into `cfg`
2. Check for a config file. If one is found, load it in.
3. If the config file had key/values not found in the default dict, add them
4. If the config file had values that do not match the default dict's values, update `cfg` with those values
5. Update `cfg` with values supplied in the command line
6. Throw an error if one of the crucial required values is not supplied
7. Continue to mapmaking

sotodlib/site_pipeline/make_ml_map.py
```diff
@@ -0,0 +1,322 @@
...
+        # Check which fields have been supplied in config.yaml
+        for field, file_val in cfg_from_file.items():
+            # If an optional arg is supplied, in the config file
+            # add it here
+            if field not in cfg.keys():
+                cfg[field] = cfg_from_file[field]
+            # If the value in the config file is different from the 
+            # default, update it
+            elif cfg[field] != file_val:
+                cfg[field] = file_val
```
mhasself, line 87: I think lines 79-87 are equivalent to:
```
cfg.update(cfg_from_file)
```
Do you agree?

The only modification I might make, is to skip any settings from cfg_from_file, if they are None.  Because then a user setting "something: null" in their config file would be equivalent to saying "just use the default".

The one-liner for that is:
```
cfg.update({k: v for k, v in cfg_from_file.items() if v is not None})
```

sotodlib/site_pipeline/make_ml_map.py
```diff
@@ -0,0 +1,322 @@
...
+    else:
+        print("No config file provided, assuming default values") 
+
+    # Merge flags from config file and defaults with any passed through CLI
+    for key, item in args.items():
+        if item is not None:
+            cfg[key] = item
+    # Certain fields are required. Check if they are all supplied here
+    required_fields = ['query','freq','area','odir','prefix','context']
```
mhasself, line 96: Reasonable default query is "1" (all tods).  I think odir and prefix could have reasonable defaults too ("output", "").

sotodlib/site_pipeline/make_ml_map.py
```diff
@@ -0,0 +1,322 @@
...
+
+
+def _get_config(config_file):
+    return yaml.safe_load(open(config_file,'r'))
+
+
+
+def main(config_file=None, defaults=defaults, **args):
+    
+    cfg = defaults
```
mhasself, line 73: Note that this will make cfg an alias for defaults.  If instead you want to take a copy, do
```
cfg = dict(defaults)
```

This can matter, if someone calls `main()` in a loop for some reason.  They might expect to be starting from the stated `defaults` each time, not the `defaults` updated with the previous config.

sotodlib/site_pipeline/make_ml_map.py
tanaybhandarkar: I think that makes sense. Question about area: right now I run your `get_geom` script to generate a an `area` from a `context.yaml`. Where do we anticipate this step will happen when we move to `site-pipeline`? In principle, that could be added to the map-making module, but maybe it needs to be separate for some reason?

sotodlib/site_pipeline/make_ml_map.py
tanaybhandarkar: I'm guessing I can do something similar when merging values from the command line too (lines 91-94)

```   
# Merge flags from config file and defaults with any passed through CLI
    for key, item in args.items():
        if item is not None:
            cfg[key] = item
```
  Can then become:
  ```
  cfg.update({k: v for k, v in args.items() if v is not None})
  ```

docs/site_pipeline.rst
```diff
@@ -493,6 +493,107 @@ Command line arguments
...
+        comps: 'T' # TQU
+        inject: null 
+        nocal: True # No relcal or abscal
+        downsample: 1 # Downsample TOD by this factor
+        tiled: 0 # Tiling boolean (0 or 1)
+        nmat-dir: './nmats/' # Dir to save or load nmat
+        nmat: 'corr' # 'corr' or 'uncorr' 
+        maxiter: 500 # Max number of iterative steps
+        nmat-mode: 'build' # 'cache', 'build', 'load' or 'save'
+        center-at: null
```
mhasself, line 589: I think these need to use underscores, not hyphens ...

sotodlib/site_pipeline/make_ml_map.py
mhasself: The library level code can definitely live in mapmaking module, or perhaps in coords.helpers.

There some argument that make-ml-map should be able to generate this file itself.  It could be configurable like  the noise matrix args; with the option of computing it (on the fly as part of the preprocessing), or loading a saved one from disk.  We probably want a stand-alone way to generate this file though, which does the minimal thing needed.  That could be a short site-pipeline script (make-footprint-file?), or just a special sub-mode of make-ml-map.

mhasself (approved):
Thanks for adding the docs.  One more minor fix but then merge away.

Issue #600, opened 2023-11-06
obs*db improvements


mhasself:
1. obsfiledb: would benefit from calls to help get the file root for an obs_id (or equivalently, ways to construct arbitrary subpaths such as {prefix}/{obs_id}/smurf/ ...)
2. It is not possible to use tags in so-metadata cli obsdb queries

PR #603, opened 2023-11-10
obsfiledb fix, minor metadata improvements

mhasself:
Major thing here is a change to obsfiledb to loosen the requirement on detector names.  Previously a detector name could belong to only one detset ... but now it can belong to more than one.  (Within a single _observation_, such overlapping detsets must not be used!)

An existing obsfiledb can be upgraded to the new format, like this:
```
  so-metadata obsfiledb  /path/to/obsfiledb.sqlite  fix-db --overwrite
```

Couple of other fixes:
- addreses half of #600 -- so-metadata obsdb can handle queries that use tags.
- makes context.get_obs a bit more permissive, if an observation is not registered in obsdb yet.  fail later!


kmharrington (approved):
Tested this on some site obsfiledbs and looks like it's working. thanks!

PR #605, opened 2023-11-14
Fixes for time, duration, and detector/tube labels

AERAdler:
Fixed errors in the codes for start/stop labels so now we can get duration properly. Wrote some hardcoded labels for detectors, but needs polishing.

### DIFF
--- a/sotodlib/site_pipeline/update_obsdb.py
+++ b/sotodlib/site_pipeline/update_obsdb.py
@@ -69,10 +69,10 @@ def telescope_lookup(telescope:str):
                 "tube_flavor": "mf", "detector_flavor": "tes"}
     elif telescope == "satp2":
         return {"telescope": "satp2","telescope_flavor": "sat",
-                "tube_flavor": "mf", "detector_flavor": "tes"}
+                "tube_flavor": "uhf", "detector_flavor": "tes"}
     elif telescope == "satp3":
         return {"telescope": "satp3", "telescope_flavor": "sat",
-                "tube_flavor": "uhf", "detector_flavor": "tes"}
+                "tube_flavor": "mf", "detector_flavor": "tes"}
     elif telescope == "lati1":
         return {"telescope": "lati1", "telescope_flavor": "lat",
                 "tube_flavor": "mf", "detector_flavor": "tes"}
--- a/sotodlib/site_pipeline/update_obsdb.py
+++ b/sotodlib/site_pipeline/update_obsdb.py
@@ -53,6 +53,37 @@ def check_meta_type(bookpath):
     else:
         return meta["type"]
 
+
+def telescope_lookup(telescope:str):
+    """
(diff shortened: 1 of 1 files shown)


sotodlib/site_pipeline/update_obsdb.py
```diff
@@ -180,29 +211,32 @@ def main(config:str,
...
+            try:
+                telescope = index["telescope"]
+                flavors = telescope_lookup(telescope)
+                for flav in flavors:
+                    bookcartobsdb.add_obs_columns([flav+" str"])
+                    very_clean[flav] = flavors[flav]
+
+            except KeyError:
+                logger.warning("No telescope key in index file")
+                pass
```
mhasself, line 223: So we're not going to make this a fatal error?  I guess that's ok, but maybe in the failed case, at least fill in very_clean['telescope_flavor'] = 'unknown', so that we can sort of see what happened in the output obsdb.

In an except block, you don't need "pass" if there are other statements.

sotodlib/site_pipeline/update_obsdb.py
```diff
@@ -180,29 +211,32 @@ def main(config:str,
...
+            try:
+                start = index["start_time"]
+                end = index["stop_time"] 
                 bookcartobsdb.add_obs_columns(["timestamp float", "duration float"])
                 very_clean["timestamp"] = start
                 very_clean["duration"] = end - start
-
+            except KeyError:
+                logger.error("Incomplete timing information for obs_id {obs_id}")
+                pass
```
mhasself, line 239: Also not fatal... fine.  Remove the "pass".

sotodlib/site_pipeline/update_obsdb.py
```diff
@@ -218,17 +252,18 @@ def main(config:str,
                     logger.error(f"No {coor} pointing in some streams for obs_id {obs_id}")
                     pass
             try:
-                if index.get("telescope_flavor")=="SAT":
+                if very_clean.get("telescope_flavor")=="sat":
```
mhasself, line 255: With the change above, you can just do `if very_clean['telescope_flavor'] == 'sat':`.  Note again that == is an operator and gets spaces around it.

mhasself (changes requested):
Thanks -- only one substantive request.  Mostly formatting!

mhasself (approved):
This is technically good to go, but we're thrashing around a bit on the formatting.  Please take one more pass then merge.

You should probably review [this section](https://peps.python.org/pep-0008/#other-recommendations) on operators and argument formatting.

The two violations that I see now are spaces around kwargs (e.g. should be `get_parser(parser=None)`) and missing spaces after colons in type annotations (this is new to me, I haven't used type annotations), i.e. it should be `check_meta_type(bookpath: str)`.  This is all covered in PEP8 now.

PR #608, opened 2023-11-15
Add detector calibration info to update_caldb script

jlashner:
This update adds a new function to the `update_smurf_caldbs` script that will write detector calibration info pulled from sodetlib IVs and bias steps to a Manifestdb + hdf5 archive. I have run this and tested in my home directory on a subset of the full observation list. I think it should take ~ an hour or so to run completely, so we may want to run directly as the pipeline user before adding to prefect, so we don't have a massive hour-long operation.

### DIFF
--- a/sotodlib/site_pipeline/update_smurf_caldbs.py
+++ b/sotodlib/site_pipeline/update_smurf_caldbs.py
@@ -40,6 +40,7 @@ import sotodlib.site_pipeline.util as sp_util
 
 # stolen  from pysmurf, max bias volt / num_bits
 DEFAULT_RTM_BIT_TO_VOLT = 10 / 2**19
+DEFAULT_pA_per_phi0 = 9e6
 
 default_logger = sp_util.init_logger("smurf_caldbs")
 
@@ -245,12 +246,14 @@ def get_cal_resset(ctx: core.Context, obs_id, logger=None):
     iv_obsids = get_cal_obsids(ctx, obs_id, 'iv')
 
     rtm_bit_to_volt = None
+    pA_per_phi0 = None
     ivas = {dset: None for dset in iv_obsids}
     for dset, oid in iv_obsids.items():
         if oid is not None:
             ivas[dset] = _load_smurf_npy(ctx, oid, 'iv')
             if rtm_bit_to_volt is None:
                 rtm_bit_to_volt = ivas[dset]['meta']['rtm_bit_to_volt']
+                pA_per_phi0 = ivas[dset]['meta']['pA_per_phi0']
         else:
             logger.debug("missing IV data for %s", dset)
 
@@ -261,11 +264,14 @@ def get_cal_resset(ctx: core.Context, obs_id, logger=None):
             bsas[dset] = _load_smurf_npy(ctx, oid, 'bias_step_analysis')
             if rtm_bit_to_volt is None:
                 rtm_bit_to_volt = bsas[dset]['meta']['rtm_bit_to_volt']
+                pA_per_phi0 = ivas[dset]['meta']['pA_per_phi0']
         else:
             logger.debug("missing bias step data for %s", dset)
 
     if rtm_bit_to_volt is None:
         rtm_bit_to_volt = DEFAULT_RTM_BIT_TO_VOLT
+    if pA_per_phi0 is None:
+        pA_per_phi0 = DEFAULT_pA_per_phi0
 
     # Add IV info
     for i, cal in enumerate(cals):
@@ -330,7 +336,7 @@ def get_cal_resset(ctx: core.Context, obs_id, logger=None):
         cal.r_frac = bsa['Rfrac'][ridx]
         cal.p_bias = bsa['Pj'][ridx]
         cal.s_i = bsa['Si'][ridx]
-        cal.phase_to_pW = 9e6 / (2*np.pi) / cal.s_i * cal.polarity
+        cal.phase_to_pW = pA_per_phi0 / (2*np.pi) / cal.s_i * cal.polarity
 
     rset = core.metadata.ResultSet.from_friend(np.array(
         [astuple(c) for c in cals], dtype=cal_dtype
--- a/sotodlib/site_pipeline/update_smurf_caldbs.py
+++ b/sotodlib/site_pipeline/update_smurf_caldbs.py
@@ -238,8 +238,6 @@ def get_cal_resset(ctx: core.Context, obs_id):
             if rtm_bit_to_volt is None:
                 rtm_bit_to_volt = ivas[dset]['meta']['rtm_bit_to_volt']
 
-
-
     bias_step_obsids = get_cal_obsids(ctx, obs_id, 'bias_steps')
     bsas = {dset: None for dset in bias_step_obsids}
     for dset, oid in bias_step_obsids.items():
@@ -249,8 +247,6 @@ def get_cal_resset(ctx: core.Context, obs_id):
                 rtm_bit_to_volt = bsas[dset]['meta']['rtm_bit_to_volt']
 
     rtm_bit_to_volt = DEFAULT_RTM_BIT_TO_VOLT
-
-
 
     # Add IV info
     for i, cal in enumerate(cals):
--- a/sotodlib/site_pipeline/update_smurf_caldbs.py
+++ b/sotodlib/site_pipeline/update_smurf_caldbs.py
@@ -1,25 +1,27 @@
-""" Script to import tuning and readout id channel mapping into a manifest
-database for book loading. At present this just works in the configuration where
-it has access to both level 2 and level 3 indexing. This is technically possible
-with just level 3 data / indexing but requires some still non-existant tools. 
+""" 
+Script to import tuning and readout id channel mapping, and detector calibration
+information into manifest dbs for book loading.  At present this just works in
+the configuration where it has access to both level 2 and level 3 indexing. This
+is technically possible with just level 3 data / indexing but requires some
+still non-existant tools. 
 
-Configuration file required:
+Configuration file required::
 
-config = {
-    'context': 'context.yaml',
-    'archive': {
-        'detset': {
-            'index': 'manifest.sqlite',
-            'h5file': 'manifest.h5'
+    config = {
+        'context': 'context.yaml',
+        'archive': {
+            'detset': {
+                'index': 'detset.sqlite',
+                'h5file': 'detset.h5'
+            },
+            'det_cal': {
+                'index': 'det_cal.sqlite',
+                'h5file': 'det_cal.h5
+            },
         },
-        'det_cal': {
-            'index': 'det_cal.sqlite',
-            'h5file': 'det_cal.h5
-        },
-    },
-    'g3tsmurf': g3tsmurf_hwp_config.yaml',
-    'imprinter': imprinter.yaml,
-}
+        'g3tsmurf': g3tsmurf_hwp_config.yaml',
+        'imprinter': imprinter.yaml,
+    }
 """
 
 import traceback
@@ -53,10 +55,7 @@ def main(config: Union[str, dict],
         overwrite:Optional[bool]=False,
         logger=None):
     smurf_detset_info(config, overwrite, logger)
-
-
-
-    run_update_det_caldb(config, logger=logger)
+    run_update_det_caldb(config, logger=logger, overwrite=overwrite)
 
 def smurf_detset_info(config: Union[str, dict], 
         overwrite:Optional[bool]=False,
@@ -202,6 +201,42 @@ cal_dtype = [
 
 @dataclass
 class CalInfo:
+    """
+    Class that contains detector calibration information that will go into the
+    caldb.
+
+    Attributes
+    ------------
+    readout_id: str
+        Readout id of detector
+    r_tes: float
+        Detector resistance [ohms], determined through bias steps while the
+        detector is biased
+    r_frac: float
+        Fractional resistance of TES, given by r_tes / r_n
+    p_bias: float
+        Bias power on the TES [J] computed using bias steps at the bias point
+    s_i: float
+        Current responsivity of the TES [1/V] computed using bias steps at the
+        bias point
+    phase_to_pW: float
+        Phase to power conversion factor [pW/rad] computed using s_i,
+        pA_per_phi0, and detector polarity
+    bg: int
+        Bias group of the detector. Taken from IV curve data, which contains
+        bgmap data taken immediately prior to IV. This will be -1 if the
+        detector is unassigned
+    polarity: int
+        Polarity of the detector response for a positive change in bias current
+        while the detector is superconducting. This is needed to correct for
+        detectors that have reversed response. 
+    r_n: float
+        Normal resistance of the TES [Ohms] calculated from IV curve data
+    p_sat: float
+        "saturation power" of the TES [J] calculated from IV curve data.
+        This is defined  as the electrical bias power at which the TES
+        resistance is 90% of the normal resistance.
+    """
     # Fields must be ordered like cal_dtype!
     readout_id: str = ''
 
@@ -220,7 +255,11 @@ class CalInfo:
 
 
 def get_cal_resset(ctx: core.Context, obs_id, logger=None):
-    """Returns calibration ResultSet for a given ObsId"""
+    """
+    Returns calibration ResultSet for a given ObsId. This pulls IV and bias step
+    data for each detset in the observation, and uses that to compute CalInfo
+    for each detector in the observation
+    """
     if logger is None:
         logger=default_logger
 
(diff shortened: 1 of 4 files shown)


sotodlib/site_pipeline/update_smurf_caldbs.py
```diff
@@ -109,15 +123,252 @@ def smurf_detset_info(config:str | dict,
...
+
+def get_cal_obsids(ctx, obs_id, cal_type):
+    """
+    Returns set of obs-ids corresponding to the most recent calibration
+    operations for a given obsid:
+
+    Returns
+    ----------
+        obs_ids: dict
+            Dict of obs_ids for each detset in specified operation
```
msilvafe, line 145: Did you test the sphinx build for this? I think you want a period at the end of line 140 and one less indent on lines 144 & 145. Should also have a block for input args.

sotodlib/site_pipeline/update_smurf_caldbs.py
```diff
@@ -109,15 +123,252 @@ def smurf_detset_info(config:str | dict,
...
+            break
+    else:
+        raise FileNotFoundError("Could not find npy file")
+    res = np.load(fpath, allow_pickle=True).item()
+    return res
+
+
+def get_cal_resset(ctx: core.Context, obs_id):
+    """Returns calibration ResultSet for a given ObsId"""
+    am = ctx.get_obs(obs_id, samples=(0, 1), ignore_missing=True)
```
msilvafe, line 227: I think you could also use `no_signal = True` instead of `samples = (0,1)`.

jlashner: This is required so we get a single sample that we can use to check if the biases have changed haha. I should add a comment

mhasself: I think it will save time to also add no_signal=True though -- won't have to unpack the compressed detector data, only the "primary" stuff.

mhasself (changes requested):
Great to see this, thanks.

sotodlib/site_pipeline/update_smurf_caldbs.py
```diff
@@ -109,15 +123,252 @@ def smurf_detset_info(config:str | dict,
...
+            if rtm_bit_to_volt is None:
+                rtm_bit_to_volt = ivas[dset]['meta']['rtm_bit_to_volt']
+
+    bias_step_obsids = get_cal_obsids(ctx, obs_id, 'bias_steps')
+    bsas = {dset: None for dset in bias_step_obsids}
+    for dset, oid in bias_step_obsids.items():
+        if oid is not None:
+            bsas[dset] = _load_smurf_npy(oid, 'bias_step_analysis')
+            if rtm_bit_to_volt is None:
+                rtm_bit_to_volt = bsas[dset]['meta']['rtm_bit_to_volt']
```
msilvafe, line 247: what's the point of getting `rtm_bit_to_volt` from the bsa and iva if you just overwrite it in line 249?

sotodlib/site_pipeline/update_smurf_caldbs.py
```diff
@@ -109,15 +123,252 @@ def smurf_detset_info(config:str | dict,
...
+@dataclass
+class CalInfo:
+    # Fields must be ordered like cal_dtype!
+    readout_id: str = ''
+
+    # From bias steps
+    r_tes: float = np.nan   # Ohm
+    r_frac: float = np.nan
+    p_bias: float = np.nan  # J
+    s_i: float = np.nan     # 1/V
```
msilvafe, line 192: Why do you get `s_i` only from the IV? We know the current IV + BS analysis differ and I think we believe that the BS calibration is more valid at least for rfrac ~< 0.6. Can we include both? Maybe `s_i_iv` and `s_i_bs`

sotodlib/site_pipeline/update_smurf_caldbs.py
```diff
@@ -109,15 +123,252 @@ def smurf_detset_info(config:str | dict,
...
+        bg = cal.bg
+        bsa = bsas[detset]
+
+        if bsa is None or bg == -1:
+            continue
+
+        bl_label = f'{stream_id}_b{bg:0>2}'
+        # If observation bias differs from bias-steps by more than 0.1 V,
+        # don't include bias step calibration info
+        if np.abs(obs_biases[bl_label] - bsa['Vbias'][bg]) > 0.1:
```
msilvafe, line 288: Can you log that this condition is met for debugging?

sotodlib/site_pipeline/update_smurf_caldbs.py
```diff
@@ -109,15 +123,252 @@ def smurf_detset_info(config:str | dict,
...
+            (bsa['bands'] == band) & (bsa['channels'] == chan)
+        )[0]
+        if not ridx: # Channel doesn't exist in bias step analysis
+            continue
+
+        ridx = ridx[0]
+        cal.r_tes = bsa['R0'][ridx]
+        cal.r_frac = bsa['Rfrac'][ridx]
+        cal.p_bias = bsa['Pj'][ridx]
+        cal.s_i = bsa['Si'][ridx]
```
msilvafe, line 301: Can we add a field that is `cal.dac_to_pW = (9e6/2/np.pi) / (bsa['Si'][ridx]*bsa['polarity'][ridx])` I feel like for folks not knowledgeable of the detectors it would be good to just multiply by one obviously named parameter to get into pW units.

sotodlib/site_pipeline/update_smurf_caldbs.py
```diff
@@ -109,15 +123,252 @@ def smurf_detset_info(config:str | dict,
...
+    operations for a given obsid:
+
+    Returns
+    ----------
+        obs_ids: dict
+            Dict of obs_ids for each detset in specified operation
+    """
+    obs = ctx.obsdb.query(f"obs_id == '{obs_id}'")[0]
+    detsets = ctx.obsfiledb.get_detsets(obs_id)
+    min_ct = obs['start_time'] - 3600*24*7
```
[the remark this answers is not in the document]
jlashner: I thought 1-week timescale for cal info was a bit overly generous, I'm not sure if we want to accidentally pull from IV info thats over a week old. Happy to set this as a configurable parameter though. If no cal files exist, this function will just return the dict where each detset is assigned to `None`, so don't think we need a try-catch statement, but I can add a log debug statement in the `get_cal_resset` function that tells you if your missing cal files.

sotodlib/site_pipeline/update_smurf_caldbs.py
jlashner: oops was missing an if statement

sotodlib/site_pipeline/update_smurf_caldbs.py
```diff
@@ -109,15 +123,252 @@ def smurf_detset_info(config:str | dict,
...
+cal_dtype = [
+    ('dets:readout_id', '<U40'),
+    ('dets:cal.r_tes', float),
+    ('dets:cal.r_frac', float),
+    ('dets:cal.p_bias', float),
+    ('dets:cal.s_i', float),
+    ('dets:cal.bg', int),
+    ('dets:cal.bg_polarity', int),
+    ('dets:cal.r_n', float),
+    ('dets:cal.p_sat', float),
```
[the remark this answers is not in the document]
jlashner: can do. Does readout_id still need to look like `"dets:readout_id"`?

sotodlib/site_pipeline/update_smurf_caldbs.py
jlashner: Sure, this should be `phase_to_pW` though, right?

sotodlib/site_pipeline/update_smurf_caldbs.py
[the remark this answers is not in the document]
mhasself: yes, that one still needs "dets:" prefix.

mhasself (approved):
Thanks!

sotodlib/site_pipeline/update_smurf_caldbs.py
msilvafe: Ya that sounds like a reasonable name.

sotodlib/site_pipeline/update_smurf_caldbs.py
```diff
@@ -109,15 +124,289 @@ def smurf_detset_info(config:str | dict,
...
+            'obs:obs_id': obs_id,
+            'dataset': obs_id,
+        }, filename=h5_path,)
+
+
+def run_update_det_caldb(config, log=None):
+    ctx = core.Context(config['context'])
+    update_det_caldb(
+        ctx, 
+        config['archive']['det_cal']['index'],
```
kmharrington, line 403: could you check for these keys / provide error messaged about what keys should exist and what they should be?

msilvafe:
Besides my inline comments this looks good to me.

I think it could use a little more docs that explain what fields get populated here, and where they come from (i.e. the most recently available smurf files in the last week) and maybe point to the appropriate iv analysis and bs analysis code in sodetlib. But I was looking in the docs and there's not an obvious place for that to go, it seems like most of the docs are nice tutorials from Matthew on how to setup and use the dB's but not necessarily explaining the contents of particular types of dBs so I don't think we need to make that documentation a prereq for merging this in.

sotodlib/site_pipeline/update_smurf_caldbs.py
```diff
@@ -109,15 +124,289 @@ def smurf_detset_info(config:str | dict,
...
+    for obs_id in tqdm(remaining_obsids, disable=(not show_pb)):
+        try:
+            rset = get_cal_resset(ctx, obs_id)
+        except Exception as e:
+            logger.error("Failed on %s: %s", obs_id, e)
+            if format_exc:
+                logger.error(traceback.format_exc())
+            continue
+
+        logger.info("Writing metadata for %s", obs_id)
```
kmharrington, line 391: can we make this function and `run_update_det_caldb` accept an overwrite option? those usually come in handy and the detset code accepts that.

sotodlib/site_pipeline/update_smurf_caldbs.py
```diff
@@ -109,15 +124,289 @@ def smurf_detset_info(config:str | dict,
...
+
+
+def get_obs_with_detsets(ctx, detset_idx):
+    """Gets all observations with detset data"""
+    db = core.metadata.ManifestDb(detset_idx)
+    detsets = db.get_entries(['dataset'])['dataset']
+
+    obs_ids = set()
+    for dset in detsets:
+        cur = ctx.obsfiledb.conn.execute(
```
kmharrington, line 354: Could we move this query into `obsfiledb`? This seems like a more general function so it'd be nice to have it available there. It's also the exact inverse of a function in there so it would match nicely.

sotodlib/site_pipeline/update_smurf_caldbs.py
```diff
@@ -109,15 +124,289 @@ def smurf_detset_info(config:str | dict,
...
+    phase_to_pW: float = np.nan  # pW/rad
+
+    # From IV
+    bg: int = -1
+    polarity: int = 1
+    r_n: float = np.nan  
+    p_sat: float = np.nan   # J
+
+
+def _load_smurf_npy(ctx, obs_id, substr):
```
kmharrington, line 214: Could we move this function into `sotodlib.io.load_book`? This also seems like a function that other folks might want to use / something that's more general than site_pipeline

sotodlib/site_pipeline/update_smurf_caldbs.py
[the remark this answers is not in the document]
kmharrington: I think a week is too generous. I wouldn't go more than a day back.

kmharrington (changes requested):
Thanks for setting this up! A couple in-line comments, mostly looking to take some of the more general functions out of site pipeline and into other parts of sotodlib. 

Otherwise, could you add a docstring somewhere saying what information is coming from what calibrations? (ie. Psats and Rn from IV and phase_to_pW from Bias steps). I realize I put the detset on at the top of this file, could be there or in the functions you're adding.

mhasself:
> I think it could use a little more docs that explain what fields get populated here, and where they come from

I second this -- the site-pipeline tools are documented in the site_pipeline.rst part of the docs.  It looks like there isn't anything there yet for this script.

If you want to keep the field detail docs / comments _with the code_ (e.g. documenting your CalInfo data class), I am sympathetic,  but make it compatible with sphinx autodoc features, at least at the level where a link to the source code can be provided, and reference those classes / functions from site_pipeline.rst.

msilvafe:
> > I think it could use a little more docs that explain what fields get populated here, and where they come from
> 
> I second this -- the site-pipeline tools are documented in the site_pipeline.rst part of the docs. It looks like there isn't anything there yet for this script.
> 
> If you want to keep the field detail docs / comments _with the code_ (e.g. documenting your CalInfo data class), I am sympathetic, but make it compatible with sphinx autodoc features, at least at the level where a link to the source code can be provided, and reference those classes / functions from site_pipeline.rst.

I should have stuck with my gut on the docs request :laughing:

jlashner:
Hi all, thanks for the great feedback! Just finished implementing and adding docs. I have a caldb built for existing operations at `det_cal.h5/sqlite`. Do you think we can copy this to the site-pipeline user and repath the manifest db, so we don't have to re-run in its entirely? Also I think this is ready to merge as soon as @kmharrington approves

msilvafe:
Can you also add in links to the bias step analysis and iv analysis code in sodetlib in the docs (site-pipeline.rst preamble bit)?

sotodlib/site_pipeline/update_smurf_caldbs.py
```diff
@@ -109,15 +131,309 @@ def smurf_detset_info(config:str | dict,
...
+            rset = get_cal_resset(ctx, obs_id)
+        except Exception as e:
+            logger.error("Failed on %s: %s", obs_id, e)
+            if format_exc:
+                logger.error(traceback.format_exc())
+            continue
+
+        logger.info("Writing metadata for %s", obs_id)
+        write_dataset(rset, h5_path, obs_id, overwrite=overwrite)
+        db.add_entry({
```
kmharrington, line 418: this line will throw an error if overwrite=true and the entry already exists.

sotodlib/site_pipeline/update_smurf_caldbs.py
```diff
@@ -109,15 +131,309 @@ def smurf_detset_info(config:str | dict,
...
+    if not os.path.exists(idx_path):
+        scheme = core.metadata.ManifestScheme()
+        scheme.add_exact_match('obs:obs_id')
+        scheme.add_data_field('dataset')
+        db = core.metadata.ManifestDb(scheme=scheme)
+        db.to_file(idx_path)
+    db = core.metadata.ManifestDb(idx_path)
+    
+    # detset_db = metadata.Manifest(detset_idx)
+    existing_obsids = set(db.get_entries(['dataset'])['dataset'])
```
kmharrington, line 397: don't bother to check for existing entries if overwrite=True?

kmharrington (changes requested):
Sorry one more thing. I don't think that `overwrite` argument is actually doing anything at the moment, which is good because it would be throwing errors if it was 😂

sotodlib/site_pipeline/update_smurf_caldbs.py
jlashner: isn't it the opposite, and it would throw an error if overwrite=False?

sotodlib/site_pipeline/update_smurf_caldbs.py
jlashner: If overwrite=True, do we want to re-run the calibration analysis for all obs_ids? I thought that we just wanted it to not care if the h5 entry already existed...

kmharrington: I think we're thinking of overwrite doing different things. I really do mean re-run the calibration analysis 

There have been times when I've done this for other manifests, that's why I like to have the option around. But I realize that has usually be for analyzes where it's usually only running on sections of the database so there are times when I have to go say "rerun on all of last weeks data" but I wouldn't want to just delete the files and start again. 

I don't think just overwriting the .h5 entry is something that should ever happen in the current setup.

jlashner: ah I see.... i think most recent commit fixes it so that overwrite means the same thing as the detset function

kmharrington (approved):
Thanks!

sotodlib/site_pipeline/update_smurf_caldbs.py
```diff
@@ -109,15 +131,320 @@ def smurf_detset_info(config:str | dict,
...
         )
         # add new entries to database
         if ts.name not in added_detsets:
             db_data = {'dets:detset': ts.name,
                     'dataset': ts.name}
-            db.add_entry(db_data, config['archive']['h5file'])
+            db.add_entry(db_data, config['archive']['detset']['h5file'])
+
+
+def get_cal_obsids(ctx, obs_id, cal_type):
```
kmharrington, line 145: this is also general enough that we could move it into `load_book`?

PR #609, opened 2023-11-17
adjusted instance name of g3thwp

ykyohei:
Changed default instance name of g3thwp for site data.
Modified g3thwp to handle both site-data and lab-data instance name conventions.


yukisakurai (approved):
Good catch. Please merge.
We should note that the default here is satp1.

kmharrington (approved):
I think it might even be better to require instance to be passed instead of going to a default. But I like this more than the older version at least.

ykyohei:
Thank you for your quick review.
Let me merge this for now. I'll make additional improvement later.

Issue #79, opened 2020-07-07
Standardized interface for operations on an AxisManager


mhasself:
This is a proposal and request-for-comment on the internal interface for low-level TOD processing functions, such as filters, flagging, coordinates & projections, etc.  Potentially anything that deals with TOD as stored in an AxisManager.

We are starting to have lots of functions that take "tod", an AxisManager, as their main argument.  A function that needs to look at several different vectors (e.g. the detector signals and the timestamps) can grab what it needs from tod (tod.signal and tod.timestamps).  But such functions will have limited usefulness if one cannot override those sensible defaults (e.g. we may want to apply a filter to tod.signal_detrended instead of tod.signal).  Conversely, the code will be annoying to use if each submodule or function accommodates overrides in a different way.

I propose a partial solution, below.

### Parameters to functions

Functions that operate on data taken from an AxisManager will have 3 classes of parameters:

1. AxisManagers -- probably only one or two of these; bundles of arrays from which default data for processing will be drawn.
2. Processing parameters -- these control some aspect of the processing but do not explicitly refer to data arrays that are inputs for the processing.
3. Data overrides -- these are optional, allowing the user to provide an array to stand in for some purpose.  The parameter names, here, should be the same as the default vector name that would otherwise be used.  For example, if tod.signal would be used by default, then the override argument should be signal=... .  Passing None should cause the default data to be used; passing False should cause the specified data to be treated as absent.

In the function signature, parameters should be ordered consistent with the order above -- AxisManagers first (perhaps just a tod), then processing parameters (which need not have defaults), and then data overrides, which presumably default to None.

Rationale for passing override vectors rather than override vector names: It might seem convenient to allow the user to specify an override data vector _by name_, e.g. do_something(tod, signal='signal_clean') would cause something to be done to tod.signal_clean instead of tod.signal.  This is not really less typing than do_something(tod, signal=tod.signal_clean), and yet it requires additional logic in the function to handle both string and array arguments.  I think it will lead to clearer code on both sides if we only specify overrides as data arrays, and do not bother with vector names.  (Pipeline-style operator functions, where a callable is instantiated with some parameters, but then actually executed on data later, would break this... but those are not what we're talking about.)

### Stashing results

[...]

PR #610, opened 2023-11-20
Detmatch bugfix and from_ctx constructor

jlashner:
Bugfix from last commit of the det-match PR.

Also adds a utility function that can be used to load a resonator set from a context that contains det_cal metadata.

### DIFF
--- a/sotodlib/coords/det_match.py
+++ b/sotodlib/coords/det_match.py
@@ -43,6 +43,18 @@ def map_band_chans(b1, c1, b2, c2, chans_per_band=512):
     return mapping
 
 
+def get_north_is_highband(bands, bgs):
+    """
+    Checks if north is highband based on bgmapping. This will tell you if
+    the majority of dets on the north side of the ufm (bgs 0-5) belong to
+    highband (bands 4-7).
+    """
+    highband = (bands > 3).astype(int) * 2 - 1   # 1 if highband, -1 if not
+    north = (bgs < 6).astype(int) * 2 - 1        # 1 if north, -1 if not
+    north[bgs == -1] = 0
+    return np.mean(highband * north) > 0
+
+
 @dataclass
 class Resonator:
     """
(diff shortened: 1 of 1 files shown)


sotodlib/coords/det_match.py
```diff
@@ -167,6 +179,54 @@ def as_array(self):
...
+        Args
+        ----------
+        ctx: Context
+            Context object
+        obs_id: str
+            Obs id to base ResSet off
+        stream_id: str
+            Stream id for ResSet to load
+        cal_name: str
+            Attribute of the axismanager that holds detector calibration info
```
mhasself, line 196: [As per this](https://github.com/simonsobs/sotodlib/issues/79), a more "sotodlib" way to do this is to have an argument that is the same name as the default field within am, and default it to None.  I.e. `det_cal=None`.  The None will trigger the default to be looked up in `am`; if it isn't None then `det_cal` will be used in place of am[det_cal].  An optional extension to this behavior is to also detect when det_cal is a string and use det_cal = am[det_cal] in that case.

I won't insist on this change, here... depends on how much you see it getting used, how flexible you want to be.

sotodlib/coords/det_match.py
```diff
@@ -167,6 +179,54 @@ def as_array(self):
             data.append(tuple(getattr(r, name) for name, _ in dtype))
         return np.array(data, dtype=dtype)
 
+    @classmethod
+    def from_ctx(cls, ctx, obs_id, stream_id, cal_name='det_cal', am=None):
```
mhasself, line 183: "aman" is a more standard arg name than `am` here.

mhasself (approved):
I have no serious issue with this but see comments about arg conventions.

jlashner:
Made changes based on @mhasself's suggestions, and replaced `from_ctx` to `from_aman`, which is more generally applicable.

PR #612, opened 2023-11-25
G3thwp: debug packet drop filling to make it compatible with reference slot filling.

ykyohei:
The original packet drop filling method fails when the reference slots are dropped.
`/so/level2-daq/satp1/hk/16917/1691705690.g3` is an exmple.

Changed the filling method to use the counters one before or one after rotation.

Advantage
- This method works even when the reference slots are dropped.

Disadvantage
- This method assumes that packet drop is not too frequent.
- This method assumes the drift of rotation frequency is not fast compared to the rotation period.

Before fix
before (1)
After fix
after (1)

### DIFF
--- a/sotodlib/hwp/g3thwp.py
+++ b/sotodlib/hwp/g3thwp.py
@@ -678,7 +678,7 @@ class G3tHWP():
         Notes
         -----
 
-        Output file formatt
+        Output file format
 
         - timestamp:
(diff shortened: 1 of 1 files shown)


yukisakurai (approved):
OK I crosschecked that this modification is valid for both satp1&3 data.
Looks good to me.

PR #615, opened 2023-11-27
Select input noise model to be demodulated

tskisner:
Small fix to select the noise model passed to the demodulation operator (either estimated, loaded from files, or the nominal elevation weighted one).  The demodulation operator then creates the demodulated noise model.


keskitalo (approved):
Looks good.

PR #619, opened 2023-12-01
make-det-wafer-info modernization

mhasself:
The code to generate `det_info.wafer` datasets is now entirely internal and parses the layout CSVs directly.  This can be run as a site-pipeline script.

The formatting of the output `wafer_info` has changed in a few ways.
- `angle` is now in degrees, not radians.
- 1-letter code fields (`pol`, `rhombus`) use 'X' to represent N/A, rather than 'None'.
- `det_x` and `det_y` renamed to `x`, `y`.
- `mux_band` is an int, not a str-that-could-be-safely-cast-to-an-int.
- new field "crossover", giving the 1-letter detector pad class, TBLR.  Open to a better name here; I don't totally know what this is I just know it could be important in the future.  Maybe "pad_type"?

### DIFF
--- a/docs/site_pipeline.rst
+++ b/docs/site_pipeline.rst
@@ -135,7 +135,7 @@ resulting ManifestDbs should work for both level 2 and level 3 SMuRf data.
 
 make_det_info_wafer
 ```````````````````
-This script uses the DetMap software package to build detector IDs for a set of
+This script uses based array construction inputs to build detector IDs for a set of
 UFMs and save them in a ManifestDb / HDF5 file. The formatting of the ResultSet 
 saved in HDF5 file will map all this information into ``det_info.wafer`` when used 
 with a correctly formatted context file and a readout to detector id mapping.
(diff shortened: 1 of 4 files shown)


mhasself:
Sample contents, for `Mv15_f090_Ar11c10A`
```
array                mv15
bond_pad             22
mux_band             1
mux_channel          55
mux_subband          d5
mux_position         0
design_freq_mhz      4261.15
bias_line            1
pol                  A
bandpass             f090
det_row              11
det_col              10
rhombus              A
type                 OPTC
x                    -25.379977999999998
y                    53.139281000000004
angle                232.5
crossover            T
coax                 N
```

Also please know that I really carefully checked that this all agreed with the previous output, for all 6 UFMs in SATP1 for which it was possible to do so.

kmharrington (approved):
Thank you for going through all of this and setting it up. (Sorry took me so long to look at it)

PR #620, opened 2023-12-01
Sort bookcart

kmharrington:
Decided to go back and update boresight / other columns from the beginning of the SATp1 cooldown and _really_ wanted it do go in time order. 

And then started passing a logger around to make it quiet up a bit.


AERAdler (approved):
Looks good, thanks for quieting check_book. We should probably take stock of how many versions of check_book are in sotodlib at some future point

PR #621, opened 2023-12-01
Mumux crosstalk

keskitalo:
Implement a new, nonlinear readout crosstalk operator for TOAST simulation workflows. Uses DetMap to get representative resonator frequencies for each channel and derive the crosstalk strength between them.

The operator requires `DetMap` and `pandas` to run.

Analysis and method were developed by Toby Satterthwaite with Shawn Henderson, John Groh, Max Silva-Feaver and Zeesh Ahmed.

See https://github.com/simonsobs/pwg-scripts/tree/master/pwg-tds/mumux_crosstalk for further details.

Here is a simulated Jupiter SAT map as a test case of the operator:

jupiter_sat
rator:

The grid size is 1 degree and the data are scaled so the reference map peaks at unity.


mhasself (approved):
Very interesting.  Approving but a few comments:

1) It looks like key detector parameters (the resonator frequency must be the most important one) are looked up in DetMap based on array name and position in the wafer.  It would be nice to have a way to pass in a simple external file with just the required info. This isn't important for now but in the long term I suspect our best "solution" results won't end up back in detmap. 

2) There are a lot of unused imports / pre-amble items (XAXIS ...) in the ops code.

3) The plot is cool and but the real damage of crosstalk is going to be in the polarization leakage -- it would nice to see the Q and U parts of the map.

keskitalo:
Thanks for the review! Here are some responses.

> 1. It looks like key detector parameters (the resonator frequency must be the most important one) are looked up in DetMap based on array name and position in the wafer.  It would be nice to have a way to pass in a simple external file with just the required info. This isn't important for now but in the long term I suspect our best "solution" results won't end up back in detmap.

I think ideally these detector parameters should be available through the hardware map so what ever we use in the mean time is temporary. Mainly, I just wanted a solution that didn't require me to commit data files or build a separate library of files that are necessary to run the operator.

> 2. There are a lot of unused imports / pre-amble items (XAXIS ...) in the ops code.

Thanks. I removed the unnecessary imports and sorted the remaining ones with `isort`.

> 3. The plot is cool and but the real damage of crosstalk is going to be in the polarization leakage -- it would nice to see the Q and U parts of the map.

That is the next step. Once the the PR is merged we plan to run a SAT+LAT simulation campaign and study the effect on CMB itself.

tpsatt (approved):
Looks good for implementing the mumux crosstalk. Thanks!

tskisner:
All looks fine as a first implementation.  Obviously both the wafer and array name should be in the focalplane detector table, obtained either from [nominal values in the hardware model](https://github.com/simonsobs/sotodlib/blob/a91ea161060a5a40ebeb2ab5004e79edb1bb2229/sotodlib/sim_hardware.py#L410) or when converting axismanager metadata.

PR #622, opened 2023-12-03
Book binding errors

kmharrington:
Trying to understand why the LAT books aren't coming out correctly (all zeros in the signal). I found and fixed some timing indexing but that is not the root cause.

### DIFF
--- a/sotodlib/io/bookbinder.py
+++ b/sotodlib/io/bookbinder.py
@@ -507,6 +507,7 @@ class SmurfStreamProcessor:
                     f"No samples properly mapped in oframe frame {oframe_idx}!"
                     "Cannot properly interpolate for this frame."
                 )
+                raise ValueError(f"Cannot finish binding {self.obs_id}")
             elif np.any(~filled):
                 self.log.debug(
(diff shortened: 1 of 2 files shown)


sotodlib/io/imprinter.py
```diff
@@ -1222,20 +1222,37 @@ def update_bookdb_from_g3tsmurf(
...
                                 continue
-                            # add all of the possible overlaps
-                            output.append(
-                                ObsSet(
-                                    obs_list,
-                                    mode="obs",
-                                    slots=self.tubes[tube]["slots"],
-                                    tel_tube=tube,
+
+                            if np.all([o.timing for o in obs_list]):
```
mhasself, line 1231: Shouldn't one split the obs_list into two lists?:
- the ones with timing can be bound, _together_, into a single book
- the ones without must be individually bound

Someday we might have one slot being weird or misconfigured; need not throw out everything in that case.

kmharrington: makes sense. From a configuration standpoint I think it's more of an all or nothing thing. But easy enough to fix.

guanyilun (approved):
looks good to me too!

PR #624, opened 2023-12-06
Det cuts

msilvafe:
Adds in flags function and preprocess support for making detector level cuts based on the IV + bias step results.

### DIFF
--- a/sotodlib/tod_ops/flags.py
+++ b/sotodlib/tod_ops/flags.py
@@ -32,7 +32,7 @@ def get_det_bias_flags(aman, detcal=None, rfrac_range=(0.1, 0.7),
         Tuple (lower_bound, upper_bound) for rfrac det selection.
     psat_range : Tuple
         Tuple (lower_bound, upper_bound) for P_SAT from IV analysis.
-        P_SAT in the IV analysis is the bias power at 90% Rn in Watts.
+        P_SAT in the IV analysis is the bias power at 90% Rn in pW.
     merge : bool
         If true, merges the generated flag into aman.
     overwrite : bool


msilvafe:
Ok @kmharrington and @mhasself I have added the changes to change the name from cuts -> flags and wrap a RangesMatrix instead of a boolean mask. I think this is ready for review. Once merged I can run on all observations + create a next context to load in these cuts.

kmharrington (approved):
My only question is if we want 0.01 or 0.1 to be the minimum acceptable on-transition value. I kind of assume a detector actually at 1% Rn is latched or going to latch very soon.

msilvafe:
> My only question is if we want 0.01 or 0.1 to be the minimum acceptable on-transition value. I kind of assume a detector actually at 1% Rn is latched or going to latch very soon.

I switched it to 0.1 and will merge after build tests.

sotodlib/tod_ops/flags.py
```diff
@@ -15,6 +15,66 @@
...
+    Parameters
+    ----------
+    aman : AxisManager
+        Input axis manager.
+    detcal : AxisManager
+        AxisManager containing detector calibration information
+        from bias steps and IVs. If None defaults to aman.det_cal.
+    rfrac_range : Tuple
+        Tuple (lower_bound, upper_bound) for rfrac det selection.
+    merge : bool
```
mhasself, line 33: psat_range is not documented.  Especially important to say the units.

sotodlib/tod_ops/flags.py
```diff
@@ -15,6 +15,66 @@
...
+        If true, merges the generated flag into aman.
+    overwrite : bool
+        If true, write over flag. If false, don't.
+    name : str
+        Name of flag to add to aman.flags if merge is True.
+
+    Returns
+    -------
+    mask : Boolean Array
+        Boolean mask that selects detector cuts.
```
mhasself, line 43: Clarify whether True means cut or uncut.

mhasself (changes requested):
See two doc requests.

msilvafe:
> See two doc requests.

Ok, I've updated those doc requests. Assuming your happy can you approve those changes? Thanks.

mhasself (approved):

PR #627, opened 2023-12-07
Downsampling mode into timing logic

kmharrington:
The last PR fixed things so that books with `obs.timing = False` got shoved in to single wafer books. 

This PR rounds out the logic to get timing = False for the two ways we could have non-cosampled UFMs. The first, which was already implemented, is low-precision timestamps in the files. The second is when the downsample mode is set to internal. Both need to be true for UFMs to be co-sampled.

I've checked that the SATp1 files get `status.downsample_external = True` and the current LAT and SATp3 files have `status.downsample_external = False`

### DIFF
--- a/sotodlib/io/g3tsmurf_db.py
+++ b/sotodlib/io/g3tsmurf_db.py
@@ -125,7 +125,7 @@ class Observations(Base):
         across stream_ids.
     timing : bool
         If true, the files of the entry observation were made with times
-        referenced to the external timing system and high precision timestamps.
+        aligned to the external timing system and high precision timestamps.
     duration : float
         The total observation time in seconds
     n_samples : integer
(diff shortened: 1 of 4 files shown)


sotodlib/io/g3tsmurf_db.py
```diff
@@ -125,7 +125,7 @@ class Observations(Base):
         across stream_ids.
     timing : bool
         If true, the files of the entry observation were made with times
-        referenced to the timing system.
+        referenced to the external timing system and high precision timestamps.
```
jlashner, line 128: nitpicky, but "with times downsampled using the external timing system, and high precision timestamps"

With downsample_mode='internal', timing is still referenced to external timing system, but downsampled data is not aligned

jlashner (approved):
Looks good to me! Small nitpicky comment inline, feel free to take it or leave it

Issue #618, opened 2023-11-30
Change book loader to remove context error message?


kmharrington:
Even since we started loading with the smurf/detset ManifestDbs we're now getting this error on the AxisManager load: 
```
ERROR: sotodlib.core.context: Key "smurf" is present in det_info returned by observation loader as well as in metadata databases; The two versions are not comparable. dropping the loader version.
```

This really isn't an error, it's that we wanted that smurf key when we had no other information. I think we can just not load those things from the books now?

mhasself:
> This really isn't an error, it's that we wanted that smurf key when we had no other information. I think we can just not load those things from the books now?

Sounds good... just so I'm clear, we can remove this (and everything else that make it possible)?

Perhaps we leave the code there, but disable it when using `load_obs_book` but enable it when using `load_book_file`?

kmharrington:
I like the idea of disabling when loading from context and enabling when loading from `load_book_file`

mhasself:
Resolved in #629

PR #629, opened 2023-12-08
Book loader fixes

### LINKED ISSUE #628: `samples` not working in `get_obs` function?
Hi, I'm trying to run the following code:

```
def load_l3_aman(stream_ids=None):
    # Restrict to good time-range
    t0, t1 = 1699972473, 1699972538
    am0 = ctx.get_obs(obs_id, no_signal=True)
    i0 = int(np.argmin(np.abs(am0.timestamps - t0)))
    i1 = int(np.argmin(np.abs(am0.timestamps - t1)))

    if stream_ids is None:
        stream_ids = np.unique(am0.det_info.stream_id)
    stream_ids = np.atleast_1d(stream_ids)

    dmask = np.logical_or.reduce(
        [am0.det_info.stream_id == sid for sid in stream_ids]
    )
    dets = am0.dets.vals[dmask]

    am = ctx.get_obs(obs_id, dets=dets, samples=(i0, i1))
    print(f"Num samples: {am.samps.count}")
    return am
```
However when I do, I get that num samples is 0. When I manually do the restriction like this:
```
    am = ctx.get_obs(obs_id, dets=dets)
    am.restrict('samps', (i0, i1))
```
I get the correct number of samples, 13000. This is for obs_id: obs_1699972342_satp1_1111111

Am I misunderstanding how to use "samples" or is this a bug?

mhasself:
Resolves #628 and #618 

Bugs fixed:
- get_obs won't load anything other than the iir_params from the G3 frames; no more innocuous ERROR message
- samples=(a, b) should work, even if a!=0.

### DIFF
--- a/sotodlib/io/load_book.py
+++ b/sotodlib/io/load_book.py
@@ -136,6 +136,7 @@ def load_obs_book(db, obs_id, dets=None, prefix=None, samples=None,
         timestamps = _res['timestamps']
 
     return _concat_filesets(results, ancil, timestamps,
+                            sample0=samples[0], obs_id=obs_id,
                             signal_buffer=signal_buffer,
                             get_frame_det_info=False)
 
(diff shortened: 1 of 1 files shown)


kmharrington (approved):
Thanks!

PR #632, opened 2023-12-14
Better toast defaults

keskitalo:
Disable all unnecessary TOAST operators by default.

Extend support for `--full_pointing` command line option.

### DIFF
--- a/sotodlib/toast/workflows/proc_demodulation.py
+++ b/sotodlib/toast/workflows/proc_demodulation.py
@@ -45,7 +45,7 @@ def setup_demodulate(operators):
         toast.ops.FitNoiseModel(
             name="demod_noise_estim_fit",
             out_model="demod_noise_fit",
-            enabled=False,
+            enabled=True,
         )
     )
 
@@ -73,7 +73,10 @@ def demodulate(job, otherargs, runargs, data):
         # The pre-demodulation noise model to use
         if job_ops.noise_estim.enabled and job_ops.noise_estim_fit.enabled:
             # We have a noise estimate
-            log.info_rank("Demodulation using estimated noise model", comm=data.comm.comm_world)
+            log.info_rank(
+                "Demodulation using estimated noise model",
+                comm=data.comm.comm_world,
+            )
             noise_model = job_ops.noise_estim_fit.out_model
         else:
             have_noise = True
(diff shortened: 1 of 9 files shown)


sotodlib/toast/workflows/proc_demodulation.py
```diff
@@ -38,12 +38,14 @@ def setup_demodulate(operators):
...
             nsum=1,
             naverage=64,
+            enabled=False,
         )
     )
     operators.append(
         toast.ops.FitNoiseModel(
             name="demod_noise_estim_fit",
             out_model="demod_noise_fit",
+            enabled=False,
```
tskisner, line 48: For the noise estimation and fitting "sets of operators", I had been disabling the "estimation" operator by default and leaving the "fitting" and "flagging" operators enabled.  This allows the user to just enable the estimation operator and then the fitting / flagging will just happen.  The raw noise estimates are usually too jagged to use directly as a noise model / filter.  [Here is the workflow logic for cutting noise outliers](https://github.com/simonsobs/sotodlib/blob/304ff59c87203faba7a427a2a532c43a07fc1b08/sotodlib/toast/workflows/proc_flagging.py#L107-L117) and [here is the logic](https://github.com/simonsobs/sotodlib/blob/304ff59c87203faba7a427a2a532c43a07fc1b08/sotodlib/toast/workflows/proc_noise_est.py#L65-L69) for running noise estimation for mapmaking.

sotodlib/toast/workflows/proc_flagging.py
```diff
@@ -72,13 +72,15 @@ def setup_flag_noise_outliers(operators):
         toast.ops.FitNoiseModel(
             name="noise_cut_fit",
             out_model="noise_cut_fit",
+            enabled=False,
```
tskisner, line 75: Same comment as above, let's leave the fitting / flagging enabled and just toggle the estimation operator.

sotodlib/toast/workflows/proc_flagging.py
```diff
@@ -72,13 +72,15 @@ def setup_flag_noise_outliers(operators):
...
         )
     )
     operators.append(
         toast.ops.FlagNoiseFit(
             name="noise_cut_flag",
             sigma_NET=5.0,
             sigma_fknee=5.0,
+            enabled=False,
```
tskisner, line 83: Same with this...

sotodlib/toast/workflows/proc_noise_est.py
```diff
@@ -37,6 +37,7 @@ def setup_noise_estimation(operators):
         toast.ops.FitNoiseModel(
             name="noise_estim_fit",
             out_model="noise_estim_fit",
+            enabled=False,
```
tskisner, line 40: Another one to leave enabled.

tskisner (approved):
Thanks for making sure that the full_pointing option is propagated correctly!  If you could just re-enable the noise model fit / flag operators by default then I am fine with merging.

PR #633, opened 2023-12-14
Update trending_flags to work with preprocess.

msilvafe:
The version of turnaround_flags added in the sub_polyf PR only included a process operation for preprocess. This PR adds in a calc_and_save operation and updates the flags returned to be RangesMatrices size n_dets x n_samps.

### DIFF
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -327,6 +327,26 @@ class FlagTurnarounds(_Preprocess):
     """
     name = 'flag_turnarounds'
     
+    def calc_and_save(self, aman, proc_aman):
+        if self.calc_cfgs['method'] == 'scanspeed':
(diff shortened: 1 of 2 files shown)


kmharrington (changes requested):
One change, can you add something in the docstrings somewhere to say that "method" is a required key for the turnaround configuration file?

msilvafe:
> One change, can you add something in the docstrings somewhere to say that "method" is a required key for the turnaround configuration file?

That I can do easily or would you prefer if the config file is missing that key that I default to one or the other (my preference would be `scanspeed`) as default?

kmharrington:
I'm good with scanspeed as the default. Just something that accounts for it.

msilvafe:
> I'm good with scanspeed as the default. Just something that accounts for it.

Ok updated now. Look good for merge?

kmharrington (approved):
looks good! thanks

Issue #623, opened 2023-12-06
Obsdb Query returns repeated observations


msilvafe:
When querying the res_check + take_noise tags like this:
```obs_list = ctx.obsdb.query(tags=['res_check=1', 'take_noise=1'])```
I'm getting lots of repeated obs_ids returned in the obs_list:
```[IN]
for obs in obs_list:
    print(obs['obs_id'])

[OUT]
oper_1698600247_satp1_0000100
oper_1698600247_satp1_0000100
oper_1698600247_satp1_0000100
oper_1698600247_satp1_0000100
oper_1698600249_satp1_0000001
oper_1698600249_satp1_0000001
oper_1698600249_satp1_0000001
oper_1698600249_satp1_0000001
oper_1698600250_satp1_0100000
oper_1698600250_satp1_0100000
oper_1698600250_satp1_0100000
oper_1698600250_satp1_0100000
oper_1698600251_satp1_0001000
oper_1698600251_satp1_0001000
oper_1698600251_satp1_0001000
oper_1698600251_satp1_0001000
oper_1698600264_satp1_1000000
```
Per Matthew's suggestion I also tried querying like this: `query('res_check and take_noise', tags=['res_check', 'take_noise'])` and results were the same.

mhasself:
Additional info from quick investigation:

Getting tags for one of these obs from the obsdb shows each tag listed multiple times:
```
>>> ctx.obsdb.get('oper_1698600247_satp1_0000100', tags=True)['tags']
['oper', 'take_noise', 'res_check', 'oper', 'take_noise', 'res_check']
```

The book index.yaml only lists tags once though
```
tags:
- oper
- take_noise
- res_check
```

So that's a clue.

kmharrington:
I believe this was all closed by that PR?

PR #637, opened 2023-12-19
Metadata fixes

mhasself:
This PR brings one major functional change needed for proper functioning of the context system, and a few little bugfixes.

Major functional change is in how det_info is merged during metadata parsing with Context.  det_info now merges more like how other metadata does -- any fields are matched against fields with the same name in the existing det_info, and that is used to associate new data to each existing readout channel.  This should be backwards compatible (stuff that used to work will still work), but now lots of other stuff will work too (this is especially needed to associate stuff with det_id=NO_MATCH rows).

Other little fixes:
- CLI: `so-metadata metadata metadata.sqlite entries` -- mode to dump all rows in the table existed but didn't show up in help output.
- DetDb docs are updated to downplay use of DetDb in favor of det_info.  Not complete, but better.
- ObsDb: useless duplicate "tag" entries are now forbidden; existing ObsDbs with duplicate rows will not cause multiple entries to show up in queries on those tags.  Addresses #623 .
- Docs for ResultSet.from_friend.
- ManifestDb.inspect handles prefix=... arg in the same was as .match does; this is needed for getting an empty metadata shell when no dets are being loaded.

This PR does nothing to address the "ignore_missing" stuff -- that is coming next ...

### DIFF
--- a/docs/context.rst
+++ b/docs/context.rst
@@ -1001,10 +1001,15 @@ intended to carry precision results needed for mapping, such as
 calibration information or precise pointing and polarization angle
 data.
 
-Certain properties may change with time, for example if wiring or
-optics tube arrangements are adjusted from one season of observations
-to the next.  The ``DetDb`` is intended to support values that change
-with time.  **The timestamp support is still under development.**
+
+.. note::
+
+   DetDb is not planned for use in SO, because of the complexity of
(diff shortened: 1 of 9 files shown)


mhasself:
@kmharrington Would like to merge this week; let me know if I should wait.

kmharrington (approved):
The new docs look helpful. I admit I don't quite see the error this is fixing but I'll believe you.

PR #639, opened 2023-12-20
coords: minor fix to bilinear projection support code

mhasself:
The so3g "threads" should be treated as a list of RangesMatrix.


amaurea (approved):
Ok, so this compensates for wrap_ivals now returning a list instead of a RangesMatrix, and as well as fixing a bug in the handling of cuts in pmat (which I didn't see because I handle cuts differently in the maximum-likelihood mapmaker). Looks good!

PR #640, opened 2023-12-20
Bookbinding / Automation Fixes

kmharrington:
Next round of bookbinding fixes and setup for smurf calibration databases and hwp angles. Updates include:

1. Implementing a method of checking for duplicate ancillary data and removing it from bookbinding if found (and if we say it's ok to remove it). Associated updates to the imprinter cli to choose to rebind books with flags.
2. Update how we find edges in scanning to deal with slower scans and make it impossible for infinite loops to happen. I have checked this new method versus the old one on 24 hours worth of observations a from satp1 and the new frame edges were never more than 2 samples different from the old frame edges.
3. Change smurf caldb function to accept a context file per type of caldb (det-cal needs detsets but detsets can't use a context that assumes detsets already exist).
4. A small edit to g3thwp I found while testing the setup for make-hwp-solutions

### DIFF
--- a/sotodlib/hwp/g3thwp.py
+++ b/sotodlib/hwp/g3thwp.py
@@ -650,6 +650,7 @@ class G3tHWP():
     def _write_empty_solution_h5(self, tod, output=None, h5_address=None):
 
         logger.info('Writing empty solutions')
+        # metadata loader requires a dets axis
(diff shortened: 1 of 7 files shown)


ykyohei (approved):
Small edit on g3thwp.py looks good to me

PR #642, opened 2023-12-29
g3thwp: add force_quad to the argument of analyze

ykyohei:
I think we should have little more flexibility in the `force_quad`, because we would like to `force_quad` both forward and reverse direction depending on the observation conditions.

To have more fallibility in `force_quad`, I moved the `force_quad` from `hwp_config` to the argument of `analyze`.

### DIFF
--- a/docs/hwp.rst
+++ b/docs/hwp.rst
@@ -67,11 +67,6 @@ Here's an annotated example of yaml config file:
   # Search range of reference slot
   ref_range: 0.1
 
-  # force to quad value
-  # 0: use measured quad value (default)
-  # 1: positive rotation direction, -1: negative rotation direction
-  force_quad: 0
-
 Data Loading
 ------------
 
--- a/docs/hwp.rst
+++ b/docs/hwp.rst
@@ -67,6 +67,11 @@ Here's an annotated example of yaml config file:
   # Search range of reference slot
   ref_range: 0.1
 
+  # force to quad value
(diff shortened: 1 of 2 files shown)


yukisakurai (commented):
I agree to be flexible, but can we prioritize 'argument' over 'config' so that they coexist? Some of us have existing code, and this approach seems more flexible. 
At least we need to add instruction in readthedoc: https://sotodlib.readthedocs.io/en/latest/hwp.html#initialize
Also, just to be confirm, is pipeline yaml file managed for this change?

ykyohei:
I have updated the docs.
I think there is no point to have force_quad in config, because we do not force all the observations to have same quad value.
I deleted the `force_quad` from config and added argument in `analyze`.

I will also update the pipeline.yaml, but the modification of this branch does not mess the running pipeline.
The  force_quad is not activated in the current pipeline.yaml.
Having unused `force_quad` will affect `G3tHWP.configs`, but will not affect the hwp angle calculation process.

yukisakurai:
The problem with making force_quad only for argument is that the [`analyze`](https://github.com/simonsobs/sotodlib/blob/master/sotodlib/hwp/g3thwp.py#L739) is in `make_hwp_solution` (`write_solution_h5`), so the quad is not accessible from the outside. I assume a config file for a specific observation can be easily prepared and placed. If this point is not an issue with this PR, I'm fine.

yukisakurai (approved):

PR #643, opened 2023-12-30
g3thwp: correct duplication of hk data

ykyohei:
Sometimes house keeping data gets duplicated.
This issue has happened in satp3 in December, and we also had this issue in one of the UTokyo HWP testing

observation id `obs_ufm_mv5_1703041906` is one of the example.
This is the slack link when we found this issue: https://simonsobs.slack.com/archives/C034C6WQEHH/p1703098228029259

To handle this issue, I added a process to remove the duplication in hk data.
We had a similar process for IRIG data (`_irig_quality_check`), so I changed the order of processes to make it easy to separate the issues in data.

### DIFF
--- a/sotodlib/hwp/g3thwp.py
+++ b/sotodlib/hwp/g3thwp.py
@@ -280,10 +280,6 @@ class G3tHWP():
             out['irig_time'] = data['irig_synch_pulse_clock_time'+suffix][1]
             out['rising_edge_count'] = data['irig_synch_pulse_clock_counts'+suffix][1]
 
-        logger.info('IRIG timing quality check.')
-        out['irig_time'], out['rising_edge_count'] = self._irig_quality_check(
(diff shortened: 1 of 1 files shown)


yukisakurai (commented):
Thank you for adding good function. The code itself looks fine. Could you specify the HK data (duplicate data) to validate this code?

ykyohei:
satp3, obs_ufm_mv5_1703041906, unixtime from 1703041906.0 to 1703044901 is an example of HK data with this issue.
This hk duplication issue is going to be fixed, but good to have this correction. https://github.com/simonsobs/ocs/pull/366

yukisakurai:
Thanks, confirmed it works correctly through `make_hwp_solution`. Please marge it.
image

yukisakurai (approved):

PR #647, opened 2024-01-05
Detrend inplace when using mean/median options and some PEP8 fixes

skhrg:


mhasself (approved):
Thanks!

If you want bonus points, add a test that actually checks in_place is working for all methods (there are partial tests already; see test_tod_ops.py).

skhrg:
NOTE: Failed on `test_azss.test_fit` once. I suspect this is because `make_fake_azss_tod` highrolled onto something horrific. I think it should include a `np.random.seed(0)` or something to make the test predictable.
Not sure who to contact about this, happy to make the change myself but figure the azss people should at least confirm that its reasonable to do.

PR #648, opened 2024-01-05
Specify numpy RNG seed in azss test

skhrg:


msilvafe (approved):
Thanks for the fix. This is a fine solution although we also probably don't need any random noise for the purpose of this test anyway.

PR #649, opened 2024-01-08
Toast timer fix

keskitalo:
Fix issues importing `Timer` from TOAST.

### DIFF
--- a/sotodlib/toast/workflows/proc_noise_est.py
+++ b/sotodlib/toast/workflows/proc_noise_est.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2023-2023 Simons Observatory.
+# Copyright (c) 2023-2024 Simons Observatory.
 # Full license can be found in the top level "LICENSE" file.
 """Noise estimation for mapmaking.
 """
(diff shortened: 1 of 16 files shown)


sotodlib/toast/workflows/proc_noise_est.py
```diff
@@ -63,12 +63,6 @@ def noise_estimation(job, otherargs, runargs, data):
     # Configured operators for this job
     job_ops = job.operators
 
-    # If any operators are disabled, just return
-    all_enabled = job_ops.noise_estim.enabled and job_ops.noise_estim_fit.enabled
-    if not all_enabled:
-        log.info_rank("Noise estimation disabled", comm=data.comm.comm_world)
-        return
-
```
tskisner, line 71: This changes the behavior in a way that impacts the selection of the noise model for the mapmaking.  Some notes:

1.  The previously requested change [from the PR here](https://github.com/simonsobs/sotodlib/pull/632#discussion_r1429244730) was never made.  Can you re-enable the noise fitting by default (so that the behavior is controlled by just the enabling of the noise estimation)?
2.  In my experience the noise estimation is always (even with clever parameter choice) too oscillatory / noisy to use directly without first fitting to a model.  This is why both operators are enabled together.
3.  The mapmaking operator will look for the fit noise model (not the raw estimate).  If you change the behavior in this file then we need to change the logic in the `proc_mapmaker.py`.

Perhaps we could remove this change from this "timer fix" PR and discuss the desired behavior of noise model estimation and selection for a future PR?

tskisner (changes requested):
Let's remove changes to the noise estimation logic from this timer PR.

sotodlib/toast/workflows/proc_noise_est.py
keskitalo: Done. Now part of https://github.com/simonsobs/sotodlib/pull/652

tskisner (approved):

PR #650, opened 2024-01-10
make-position-match and finalize-focal-plane refactor

skhrg:
Couple of minor things left to do (do another round of cleanup, update docs, make plots easier to look at. etc.) but new logic is ready for review.


tskisner:
FYI, I just saw this same failure in another PR, and [fixed it by replacing the `cancel-workflow-action` ](https://github.com/simonsobs/sotodlib/pull/444/commits/302876787841aeb900eea3703890c0ec638b57a4) with the `concurrency` stanza [mentioned here](https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-concurrency-and-the-default-behavior).  This was the [recommendation by the author of that action](https://github.com/styfle/cancel-workflow-action) at the top of their README.  I'm going to merge that PR in a few minutes after the tests finish, and then you can pickup the change here by rebasing on master (or merging master, depending on your workflow).

skhrg:
@mhasself I think I am happy with this and given that people are using the outputs without (too many) complaints I'd like to get it merged.
I'll work on cleaning up the commit history but could you let me know if there are any changes you want me to make?

sotodlib/coords/optics.py
```diff
@@ -655,3 +562,147 @@ def SAT_focal_plane(aman, x=None, y=None, pol=None, rot=0, mapping_data=None):
...
+        mapping_data: Tuple of (x, theta) that can be interpolated to map the focal plane to the sky.
+                      Leave as None to use the default mapping.
+                      Only used by the SAT.
+
+        return_fp: If True also return the transformed focal plane in mm.
+
+    Returns:
+
+        xi: Detector elev on sky from physical optics in radians.
+            If aman is provided then will be wrapped as aman.focal_plane.xi.
```
mhasself, line 623: "elev"?

sotodlib/coords/optics.py
```diff
@@ -655,3 +562,147 @@ def SAT_focal_plane(aman, x=None, y=None, pol=None, rot=0, mapping_data=None):
...
+
+        eta: Detector xel on sky from physical optics in radians.
```
mhasself, line 625: "xel"?  I see "elev and xel" used all over this module... I don't know what that means, but it is unlikely to be what "xi" and "eta" usually mean.

sotodlib/site_pipeline/finalize_focal_plane.py
```diff
@@ -214,45 +409,224 @@ def _gamma_min(scale, shift, gamma):
...
+        if "wafer" not in aman.det_info and dm_name in aman:
+            dm_aman = aman[dm_name].copy()
+            aman.det_info.wrap("wafer", dm_aman)
+            if "det_id" not in aman.det_info:
+                aman.det_info.wrap(
+                    "det_id", aman.det_info.wafer.det_id, [(0, aman.dets)]
+                )
+        if "det_id" in aman.det_info:
+            aman.restrict("dets", aman.dets.vals[aman.det_info.det_id != ""])
+            aman.restrict("dets", aman.dets.vals[aman.det_info.det_id != "NO_MATCH"])
```
mhasself, line 458: LabelAxis accepts a mask now! :)

Also see numpy.isin.

sotodlib/site_pipeline/finalize_focal_plane.py
```diff
@@ -214,45 +409,224 @@ def _gamma_min(scale, shift, gamma):
...
+        elif "det_info" not in aman or "det_id" not in aman.det_info:
```
mhasself, line 459: If you want to trap this informatively, it needs to come first.  Currently it's equivalent to "else" (code will already be crashed if det_info not in aman).

mhasself (approved):
Yes, approved.

See a bunch of minor comments, but nothing substantive so don't let them hold you up.  I'll probably have a bunch of documentation requests when I sit down to use this ... let's all hope that never happens.

I would appreciate some rebase / squash before merge though.  Thanks!

sotodlib/coords/optics.py
skhrg: Sorry left over from the Simon script I adapted this from. They are the same (modulo the zemax rotation) its just radio astronomy lingo I think. Will swap them out.

PR #651, opened 2024-01-12
Update eval_angle to be robust against outliers and packet drops

17-sugiyama:
This pull request suggests modification in `eval_angle`.
Template subtraction is found to be weak against angle jitter in the conventional script. 
Template is evaluated by averaging the one cycle of the angle pattern.
Angle jitter affects to the template evaluation, thus results in jitters appearing every one cycle after the template subtraction.

The new function uses `scipy.stats.trim_mean()` to calculate averaged template without outliers.

supplemental slides (p2-4)
https://docs.google.com/presentation/d/1FyuvC_oyzi4VHhw9ofsP-fmVCFoXpyoEJe7Q_0TOR3E/edit#slide=id.g2ae72945b36_0_16

### DIFF
--- a/sotodlib/hwp/g3thwp.py
+++ b/sotodlib/hwp/g3thwp.py
@@ -965,28 +965,38 @@ class G3tHWP():
 
         # calculate template subtracted angle
         if 'fast_time' in solved1.keys():
+            template_subtraction_1 = True
             try:
                 self.eval_angle(solved1)
             except Exception as e:
                 logger.error(f"Exception '{e}' thrown while the template subtraction for the 1st encoder.")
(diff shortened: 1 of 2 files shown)


17-sugiyama:
The update [8e6f498](https://github.com/simonsobs/sotodlib/pull/651/commits/8e6f4985ce8630d8e17e2dd9f8f8e0dabf8cf9ca) makes template subtraction work after the packets filling process run.
`_fill_dropped_packets` fills the dropped packets, but does not rebuild the reference index. This results in a mismatch between the `fast_time` length and the template array length.
The modified script adjusts the template array length considering the packet drop.

PR #652, opened 2024-01-12
Support noise estimation without noise fitting

keskitalo:


tskisner (approved):
These changes seem fine, thanks for addressing this in a separate PR.  The doc test failure is unrelated and I believe fixed elsewhere.

PR #654, opened 2024-01-14
Adds site_pipeline update_det_match script

jlashner:
This PR makes a few changes to small changes to det-match, and adds a site_pipeline script to automate matching to solution files. 

Changes to det_match:
 - Fixes some of matthew's gripes, changes default `det_id` to 'NO_MATCH'.
 - Adds option to apply pointing information from `dst` (typically coming from solution files). This is useful to get some useful pointing info out when there is no corresponding pointing measurements that can be added.
 - Adds `matching_cost` attribute to Match class, useful for minimizing cost with respect to some input param such as `freq_offset_mhz`.
 - Fix issue with saving res-sets that are unnamed.

The automated update script will find create match files for each detset, saving them to `<result_dir>/matches`, and then separately, use those saved match files to populate a manifest db. This allows us to retain and analyze the matches that we are putting into the caldb. There are two separate manifest databases, `assignment.sqlite` contains just readout_id -> det_id mapping, and `det_match.sqlite` contains resonator information from the solution files along with the readout_id --> det_id mapping.

This version of the script assumes no pointing information, and will match to a solution file based just on resonator frequency and bias line. I find that this can be finicky sometimes and the number of off-by-one errors can be highly dependent on matching parameters such as `freq_offset_mhz`, so it is definitely not as accurate as when using pointing information. I have found that a decent way to find a good offset is to match for a range of freq-offsets, and find the freq-offset that minimizes the matching costs. Doing this with `offsets = np.arange(-4, 4, 0.3)` takes a couple of minutes per detset. This is not an issue since new tunes are infrequent, however we might want to think of how to find a better method for this or maybe pull freq-offsets from a config file.

An example context file and results directory can be found in `/so/home/jlashner/work/daily/20240109`. @mhasself if you could check things like the database schemas etc. and make sure everything is how you want it that would be great. I wasn't exactly sure, for instance, what the schema and keys should be in the `assignment` db.


sotodlib/site_pipeline/update_det_match.py
```diff
@@ -0,0 +1,374 @@
...
+        with open(dmap_cfg_file, 'r') as f:
+            self.wafer_map = yaml.safe_load(f)
+        
+        self.detset_db = None
+        self.detcal_db = None
+
+        self.detmap_cfg_path = os.path.join(
+            cfg.site_pipeline_root, 'shared/detmapping/wafer_map.yaml')
+        with open(self.detmap_cfg_path, 'r') as f:
+            self.detmap_cfg = yaml.safe_load(f)
```
mhasself, line 111: This is a duplicate, somehow, of lines 99-103.

It would be nice to be able to specify this file through the cfg arg.

sotodlib/site_pipeline/update_det_match.py
```diff
@@ -0,0 +1,374 @@
...
+            cfg.site_pipeline_root, 'shared/focalplane/ufm_to_fp.yaml')
+
+        for d in self.ctx['metadata']:
+            if d['name'] == cfg.detset_meta_name:
+                self.detset_db = core.metadata.ManifestDb(d['db'])
+            elif d['name'] == cfg.detcal_meta_name:
+                self.detcal_db = core.metadata.ManifestDb(d['db'])
+        if self.detset_db is None:
+            raise Exception(
+                f"Could not find detset metadtata entry with name: {cfg.detset_meta_name}")
```
mhasself, line 123: spelling "metadtata"

sotodlib/site_pipeline/update_det_match.py
```diff
@@ -0,0 +1,374 @@
...
+        if self.detcal_db is None:
+            raise Exception(
+                f"Could not find detset metadtata entry with name: {cfg.detcal_meta_name}")
```
mhasself, line 126: Duplicate of 121-123

sotodlib/site_pipeline/update_det_match.py
```diff
@@ -0,0 +1,374 @@
...
+
+    match_file = os.path.join(runner.cfg.results_path, f'matches/{detset}.h5')
+    with h5py.File(match_file, 'r') as f:
+        ra = np.array(f['merged'])
+
+    names = list(ra.dtype.names)
+    idx = names.index('readout_id')
+    names[idx] = 'dets:readout_id'
+    ra.dtype.names = tuple(names)
+    assignment = ra[['dets:readout_id', 'det_id']]
```
mhasself, line 306: In the `assignment` output, which is intended for det_info, the field must be called `dets:det_id`.  You'll find that with those two columns, it won't load as standard metadata anymore.  But it will load if you set `det_info: true` (and it will populate `det_info.det_id`).

mhasself (changes requested):
Sorry to be slow on this.  Couple comments including an important change to the "assignments" output.

Please also add an entry in site_pipeline.rst.  It should include an example of a config file.  And it should probably link the user to the site-pipeline-configs repo, since so much important stuff comes from there, though of course that's not public.

Very happy to have this in the pipeline!

jlashner:
Implemented Matthew's suggestions, and added a docs section to the `site_pipeline.rst`.

sotodlib/site_pipeline/update_det_match.py
```diff
@@ -163,15 +162,15 @@ def run_match_aman(runner: Runner, aman, detset, wafer_slot=None):
     stream_id = aman.det_info.stream_id[aman.det_info.detset == detset][0]
     sol_file = os.path.join(
         os.path.dirname(runner.detmap_cfg_path),
```
mhasself, line 164: `.detmap_cfg_path` appears to be gone now.  Yes, I ran the code...

jlashner: Oops... thanks for catching, I was avoiding re-creating the full match archive to save time but clearly that was a mistake, and it did not cover the match-making parts.

This should be fixed and has been tested fully on SAT1 data with this pipeline-configs branch: https://github.com/simonsobs/site-pipeline-configs/pull/32

mhasself (approved):
Hooray!

jlashner:
Since there's approval I will merge this, but we still need to configure the prefect worker, etc.

PR #658, opened 2024-01-18
Add negetive sign to account for rotation definition

skhrg:
See https://github.com/simonsobs/site-pipeline-configs/pull/21 for discussion


mhasself (changes requested):
This makes sense but explain it (however so briefly) in the docstring.

PR #659, opened 2024-01-19
ManifestDb HWP logic fix

kmharrington:
This fixes the logic for making and/or connecting to ManifestDbs with the HWP angles. Previously the behavior was creating a new database every time it was run. 

Also includes a fix to the Imprinter CLI to account better for an books we bind outside the docker image.

### DIFF
--- a/sotodlib/hwp/g3thwp.py
+++ b/sotodlib/hwp/g3thwp.py
@@ -535,7 +535,7 @@ class G3tHWP():
         solved: dict
           dict data from analyze
         output: str or None
-          output path + file name, overwrite config file
+          output path + file name, override config file
 
         Notes
         -----------
--- a/sotodlib/hwp/g3thwp.py
+++ b/sotodlib/hwp/g3thwp.py
@@ -535,7 +535,7 @@ class G3tHWP():
         solved: dict
           dict data from analyze
         output: str or None
-          output path + file name, overwirte config file
+          output path + file name, overwrite config file
 
         Notes
         -----------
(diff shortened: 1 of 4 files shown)


sotodlib/hwp/g3thwp.py
```diff
@@ -535,7 +535,7 @@ def write_solution(self, solved, output=None):
         solved: dict
           dict data from analyze
         output: str or None
-          output path + file name, overwirte config file
+          output path + file name, overwrite config file
```
mhasself, line 538: override

sotodlib/io/imprinter_cli.py
```diff
@@ -79,6 +82,13 @@ def check_failed_books(imprint:Imprinter):
             imprint.bind_book(
                 book, ignore_tags=ignore_tags, ancil_drop_duplicates=ancil_drop_duplicates
             )
+            if imprint.docker_output_root is not None:
```
mhasself, line 85: I feel like you need a comment somewhere explaining what's going on with this docker_output_root business.  The behavior here seems a bit more complex than "Overide imprinter output root?"  Maybe it's simpler than I'm thinking, which is I might deserve some reassuring comment.

mhasself (approved):
Looks fine ... couple requests for doc fixes but I'll understand if they're not a priority!

yukisakurai (approved):
Thank you, I confirmed it works normally for both satp1 and satp3. I had to delete the output files every time during debugging..., so this is very helpful!

PR #662, opened 2024-01-22
Flag fixes

msilvafe:
Currently the `det_cal` based flags have the opposite logic as the rest of flags (i.e. True for good channels and False for bad). Here I swapped that to match the rest of the module (i.e. True for bad channels and False for good). I also added an `edge_guard` argument to the `get_glitch_flags` function to exclude glitches detected at the ends of the data due to ringing of the HPF.


skhrg (approved):

kmharrington (approved):
one day I'd love to implement logic to treat specially treat the edge-guard area. But not having false positives is a good first step

PR #665, opened 2024-01-23
Jump preprocess and "slow" jumps

skhrg:
Adds preprocess modules for jump finding and fixing as well as support for slow jumps (aka a short unlock)

### DIFF
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -4,9 +4,12 @@ import sotodlib.core as core
 import sotodlib.tod_ops as tod_ops
 from sotodlib.hwp import hwp
 
-from sotodlib.core.flagman import (has_any_cuts, has_all_cut,
-                                   count_cuts,
-                                    sparse_to_ranges_matrix)
+from sotodlib.core.flagman import (
+    has_any_cuts,
+    has_all_cut,
+    count_cuts,
+    sparse_to_ranges_matrix,
+)
 
 from .core import _Preprocess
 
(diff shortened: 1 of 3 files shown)


sotodlib/preprocess/processes.py
```diff
@@ -4,9 +4,12 @@
 import sotodlib.tod_ops as tod_ops
 from sotodlib.hwp import hwp
 
-from sotodlib.core.flagman import (has_any_cuts, has_all_cut,
-                                   count_cuts,
-                                    sparse_to_ranges_matrix)
+from sotodlib.core.flagman import (
```
msilvafe, line 7: I think our PR best practices is to not include formatting changes in the same PR as substantive code changes.

sotodlib/preprocess/processes.py
```diff
@@ -107,86 +120,170 @@ class GlitchDetection(_Preprocess):
...
+            aman.signal,
+            proc_aman.jumps.jump_flag.mask(),
+            inplace=True,
+            heights=proc_aman.jumps.jump_heights,
+        )
+
+    def calc_and_save(self, aman, proc_aman):
+        jumps, heights = tod_ops.jumps.find_jumps(aman, merge=False, **self.calc_cfgs)
+        jump_aman = tod_ops.jumps.jump_aman(aman, jumps, heights)
+        aman.wrap(name, jump_aman)
```
msilvafe: Probably `name` needs to be `self.calc_cfgs['name']`

skhrg: Yeah caught that one this morning

sotodlib/preprocess/processes.py
skhrg: Sorry, sometimes I forget to turn off my `black` hook...

sotodlib/tod_ops/jumps.py
```diff
@@ -457,7 +463,7 @@ def twopi_jumps(
         atol = 3 * std_est(signal.astype(float))
         np.clip(atol, 1e-8, 1e-2)
 
-    _signal = _filter(signal, gaussian_width, tv_weight)
+    _signal = _filter(signal, 3, gaussian_width, tv_weight)
```
msilvafe, line 462: what exactly is this hardcoded in `medfilt = 3` needed for?

skhrg: That kills glitches in data without really effecting much else. It may not actually be needed but I was imagining a scenario where there is a glitch sufficiently close to to a jump such that it confuses things.

skhrg: Maybe what I really want is to just make that a kwarg with a sane default so that it can be turned off if desired.

kmharrington:
Looking at the setup here a bit. The implementation of the jumps preprocessing modules is pretty borked. But I'm messing around with making a better version. If we took that out I'd be interested in seeing the rest of this merged sooner rather than later assuming the code works.

PR #666, opened 2024-01-23
Failed file cache and relative paths in smurf_caldbs

jlashner:
This adds a few things to the `update_smurf_caldbs` script to:
- Cache failed obsids in the update_detcal function
- Adds option to write the h5path relative to the dbpath in both detset and detcal dbs. This makes it so that we can simply copy the h5 and sqlite files to a new location and don't have to re-root.
- Adds `root_dir` param to config, so you can update that once instead of changing the path for the h5files and dbfiles separately when testing in a new directory.
- Adds `--skip-detsets` and `--skip-detcal` args for testing

### DIFF
--- a/sotodlib/site_pipeline/update_smurf_caldbs.py
+++ b/sotodlib/site_pipeline/update_smurf_caldbs.py
@@ -10,14 +10,19 @@ Configuration file required::
     config = {
         'archive': {
             'detset': {
+                'root_dir': /path/to/detset/root,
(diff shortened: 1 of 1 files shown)


kmharrington (approved):
Ok, given this https://github.com/simonsobs/sotodlib/issues/670 we can put off using the policies for now. So approving but please don't merge until there's a matching update to `site-pipeline-configs`

PR #674, opened 2024-01-24
Fix bug when failed cache is not set

jlashner:
Katie found an error that causes smurf_caldbs pipeline script to fail when `root_dir` is set but not `failed_obsid_cache`.


kmharrington (approved):

kmharrington:
@jlashner can merge?

Issue #661, opened 2024-01-22
Remove WARNING statement from aman loading when `ignore_missing=True`


jlashner:
In site-pipeline scripts that build metadata based off of previous metadata, such as the obsdb -> detset -> det_cal -> det_match chain, it makes sense to use a single shared context file with all metadata entries and `ignore_missing=True`, and manually catching cases where the metadata you need is missing, rather than having a separate context file for each combination of required metadata.

This generally works well, except for whenever there is metadata missing (even with `ignore_missing=True`) it logs a warning statement: 
```
WARNING: sotodlib.core.metadata.loader: Failed to load metadata for spec={'db': '/so/home/jlashner/work/daily/2024-01-19/satp3-detmatch/results/det_match.sqlite', 'name': 'det_match'}; ignoring.
AxisManager(obs_info*, det_info*[dets], dets:LabelAxis(1525))
``` 
which clutters up logs. Would it be ok to change this to a debug statement when `ignore_missing=True` or add an additional kwarg to skip this warning?

mhasself:
I think I'm ok with that.  I sort of wonder if there's a way to use the `warnings` package, but have those warnings suppressed by default ... I guess that's not much different from burying it in log.debug.

Related: I'm just about done with a rewrite of how metadata are handled in the incomplete / missing cases.  (ignore_missing will still work, but you can also fine-grain target which metadata are ok to ignore.)

jlashner:
fine-grained targeting of expected metadata sounds perfect for this, so I can just wait for that!

mhasself:
Addressed in #675

PR #675, opened 2024-01-25
Metadata - on_missing

mhasself:
This updates the metadata handling system in Context to be more flexible in how missing data are treated.  This builds on the system started in #460 and the discussion there.

Main features:
- New `load_metadata` function can be used to load a result from a ManifestDb, starting from only an AxisManager and a metadata spec.
- Format of metadata entries in Context is modified (and features extended):
  - new `on_missing` = `trim`, `skip`, `fail` option
  - "name" is deprecated in favor of "unpack" and "label"
  - You can "unpack" into multiple places
- The on_missing for an entry can be overridden when calling Context get_obs or get_meta.

Also:
- Documentation of the metadata system is reorganized a fair bit.  This could still use some work, but it's better.
- More tests of metadata loading for incomplete data cases.
- Updates to `preprocess_tod` from @kmharrington so individual groups (data subsets) can be analyzed.

The logging might be too verbose still -- cf #661; let's talk about what most useful.  It might be that almost all `logger.warning` should be `logger.debug`.

### DIFF
--- a/sotodlib/core/metadata/loader.py
+++ b/sotodlib/core/metadata/loader.py
@@ -327,7 +327,7 @@ class SuperLoader:
           ignore_missing (bool): If True, don't fail when a metadata
             item can't be loaded, just try to proceed without it.
           ignore_missing_dets (bool): If True, don't fail with a metadata
-            entry _that is not used for detector selection_ does not contain
+            entry *that is not used for detector selection* does not contain
             information for any subset of detectors.
         Returns:
           In normal mode, an AxisManager containing the metadata
--- a/sotodlib/core/metadata/loader.py
+++ b/sotodlib/core/metadata/loader.py
@@ -304,7 +304,7 @@ class SuperLoader:
 
     def load(self, spec_list, request, det_info=None, free_tags=[],
              free_tag_fields=[], dest=None, check=False, det_info_scan=False,
-             ignore_missing=False):
+             ignore_missing=False, ignore_missing_dets=False):
         """Loads metadata objects and processes them into a single
         AxisManager.
 
@@ -326,7 +326,9 @@ class SuperLoader:
             directly update det_info.
           ignore_missing (bool): If True, don't fail when a metadata
             item can't be loaded, just try to proceed without it.
-
+          ignore_missing_dets (bool): If True, don't fail with a metadata
+            entry _that is not used for detector selection_ does not contain
+            information for any subset of detectors.
         Returns:
           In normal mode, an AxisManager containing the metadata
           (dest).  In check mode, a list of tuples (spec, exception).
@@ -344,6 +346,10 @@ class SuperLoader:
         # Augmented request -- note that dets:* restrictions from
         # request will be added back into this by check tags.
         aug_request = _filter_items('obs:', request, False)
+        # detector fields in request to know if data selection could be
+        # happening
+        det_request_fields = _filter_items('dets:', request.keys(), True)
+
         if self.obsdb is not None and 'obs:obs_id' in request:
             if dest is None:
                 dest = core.AxisManager()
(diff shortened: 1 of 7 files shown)


sotodlib/core/metadata/loader.py
```diff
@@ -452,22 +491,42 @@ def check_tags(det_info, aug_request, final=False):
...
+                    logger.warning(message + 'Discarding.')
+                    continue
+
             # Unpack it.
             try:
-                unpackers = Unpacker.decode(spec['name'], item)
-                for unpacker in unpackers:
-                    dest = unpacker.unpack(item, dest=dest)
+                dest = unpack_item(spec.unpack, item, dest=dest)
+                #unpacker = _Unpacker.from_spec(spec, target=item)
```
kmharrington, line 523: hanging commented lines?

sotodlib/core/metadata/loader.py
```diff
@@ -452,22 +491,42 @@ def check_tags(det_info, aug_request, final=False):
...
                     f'The decoded item {item} is not an AxisManager or '
                     f'other well-understood type.  Request was: {request}.')
 
+            # You have to check for detector loss here -- compare
+            # item.dets.vals to what's in det_info.
+            i0 = core.util.get_multi_index(
+                item.dets.vals, det_info['readout_id'])
+
+            n_dets_item = len(set(i0[i0>=0]))
+            if n_dets_item < n_dets:
```
kmharrington, line 508: Could we have this still fail if `on_missing == 'trim'` but `n_dets_item==0`? It doesn't really make sense to me to allow calculations to continue when we've cut all detectors and that also seems to be the most common failure mode folks are noticing. An error message that says which metadata is cutting all down to zero will be quite helpful.

mhasself: I don't know about this ... in that case it will already print a warning message that includes the metadata spec and the change in detector count.  The "computations to continue" should check, for themselves, that the data aren't useless for their purposes.

I don't like making little exceptions that hamstring general rules.  If we added this, then there would be *no way* to plow through with trim, if someone did want to do that.  Trim would sometimes trim, and sometimes fail... and that's an additional check (rather than one fewer check) that needs to be done in the calling code.

kmharrington: ya ok. I do think this will be the largest user-reported "error that's not an error" but suppose that just requires training for folks.

kmharrington:
@jlashner @msilvafe  any chance either of you are playing with manifests enough to test this out in the next day or so? I probably can't try running anything until Wednesday at the earliest

kmharrington (approved):
I have looked through this all enough to believe I like the general implementation idea and that it's a major improvement. I haven't been able to actually test it out and see how the different errors / warnings manifest. If thoughts on those can be "later posted issues" then I'm happy to approve.

Is it fully backwards compatible to the point where we wouldn't need to update site contexts on the merge? Several of the automated pieces are loading contexts so if those need to be updated / timed with updating sotodlib there then let me know?

mhasself:
> Is it fully backwards compatible to the point where we wouldn't need to update site contexts on the merge? Several of the automated pieces are loading contexts so if those need to be updated / timed with updating sotodlib there then let me know?

Thanks!  The only thing I would worry about, in terms of backwards compatibility, is the behavior of "ignore_missing".  That's not something I have ever used much; is it relied upon in site pipeline automation?

jlashner:
Yes, I currently use `ignore_missing` in the det-cal and det-match update scripts. If it still returns an axis-manager with existing metadata present, I think it should be ok. If not, we'll need to update those, but maybe this is as simple as just replacing it with `on_missing='skip'`?

mhasself:
> Yes, I currently use `ignore_missing` in the det-cal and det-match update scripts. If it still returns an axis-manager with existing metadata present, I think it should be ok. If not, we'll need to update those, but maybe this is as simple as just replacing it with `on_missing='skip'`?

I took another look at it and I think on_missing should work as it did before.  (Passing `on_missing='skip'` isn't the same; on_missing needs to be a dict.)

mhasself:
Ok I'm going to merge this, but try to activate it on site pipeline during business hours ...

sotodlib/core/metadata/loader.py
mhasself: Sorry, forgot to push this cleanup commit -- will include in my next PR.

Issue #669, opened 2024-01-24
Include reason HWP angles are zero in ManifestDb writeout for make-hwp-solutions


kmharrington:
Right now it looks like there are four difference reasons `_write_empty_solution_h5` could be called in `g3thwp`. And there's nothing in the hwp_solution info that says the empty solution was written. A field needs to be added to the database to indicate why this happened because otherwise there will continue to be a large overhead on investigating why there are zeros. (no hwp data would nicely mean "don't bother investigating" while "error in calculations" would indicate the need for follow up).

ykyohei:
This issue is addressed by https://github.com/simonsobs/sotodlib/pull/676.

PR #676, opened 2024-01-25
Include error indicator field in HWP ManifestDb

yukisakurai:
Added a field (`hwp_err`) in HWP ManifestDb to indicate when the HWP angle is all zeros, whether it is a calculation failure or just a lack of L2 HWP data.
Motivated by this issue: #669

### DIFF
--- a/sotodlib/hwp/g3thwp.py
+++ b/sotodlib/hwp/g3thwp.py
@@ -167,7 +167,6 @@ class G3tHWP():
         if not any(data):
             logger.info('HWP is not spinning in time range {' + str(
                 self._start) + ' - ' + str(self._end) + '}, data is empty')
-
         return data
 
     def load_file(self, file_list=None, instance=None):
@@ -189,7 +188,6 @@ class G3tHWP():
         dict
             {alias[i] : (time[i], data[i])}
         """
-
         if file_list is None and self._file_list is None:
             logger.error('Cannot find input g3 file')
             return {}
@@ -253,7 +251,6 @@ class G3tHWP():
             fields += [self._field_instance_sub + '_full.' + f if 'counter' in f
                        else self._field_instance_sub + '.' + f for f in self._field_list]
             alias += [a + '_2' for a in self._field_list]
-
         return fields, alias
 
     def _data_formatting(self, data, suffix=''):
@@ -375,7 +372,6 @@ class G3tHWP():
         hwp_rate = np.append(slow_hwp_rate, hwp_rate)[slow_idx]
 
         locked[np.where(hwp_rate == 0)] = False
-
         return {'locked': locked, 'stable': stable, 'hwp_rate': hwp_rate, 'slow_time': slow_time}
 
     def analyze(self, data, ratio=None, force_quad=None, mod2pi=True, fast=True):
@@ -422,7 +418,6 @@ class G3tHWP():
                 * the "approximate" HWP spin rate, with sign, in revs / second.
                 * Use placeholder value of 0 for cases when not "stable".
         """
-
         if not any(data):
             logger.info("no HWP field data")
 
@@ -456,7 +451,6 @@ class G3tHWP():
         out = self._slowdata_process(fast_time, d['irig_time'])
         out['fast_time'] = fast_time
         out['angle'] = angle
-
         return out
 
     def eval_angle(self, solved, poly_order=3):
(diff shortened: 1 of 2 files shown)


sotodlib/hwp/g3thwp.py
```diff
@@ -643,18 +643,20 @@ def _set_empty_axes(self, aman):
...
-    def _write_empty_solution_h5(self, tod, output=None, h5_address=None):
-
+    def _write_empty_solution_h5(self, tod, output=None, h5_address=None, err=False):
+        
         logger.info('Writing empty solutions')
         # metadata loader requires a dets axis
         aman = sotodlib.core.AxisManager(tod.dets, tod.samps)
         self._set_empty_axes(aman)
         aman.timestamps[:] = tod.timestamps
+        if err: aman.hwp_err[:] = np.ones(len(tod.timestamps))
```
ykyohei, line 657: I prefer something like
`aman.hwp_err[:] = np.full_like(tod.timestamps, err, dtype=bool)`

hwp_err is misleading as hwp angle uncertainty, so should we change to err_flag or err_msg?

But I think the error flag can be just a single bool value not an array.
Can we write a single bool value error flag to the h5 metadata file instead?
Or saving simple message like "No data", "data is too short", "Error in hwp angle calculation" would be more helpful?

sotodlib/hwp/g3thwp.py
```diff
@@ -643,18 +643,20 @@ def _set_empty_axes(self, aman):
...
         aman.wrap_new('hwp_rate', shape=('samps', ), dtype=np.float16)
         aman.wrap_new('eval', shape=('samps', ), dtype=bool)
+        aman.wrap_new('hwp_err', shape=('samps', ), dtype=bool)
 
         return
 
-    def _write_empty_solution_h5(self, tod, output=None, h5_address=None):
-
+    def _write_empty_solution_h5(self, tod, output=None, h5_address=None, err=False):
+
```
ykyohei, line 651: Please delete unnecessary spaces. There is another one.

ykyohei (changes requested):

yukisakurai:
Changed from boolean array to single logger message, `hwp_logger`.
The same stuff is happening with the `eval` flag, so I renamed it to `hwp_angle_ver2_flag` and changed to a single boolean.

sotodlib/site_pipeline/make_hwp_solutions.py
```diff
@@ -18,6 +18,7 @@
 from sotodlib.hwp.g3thwp import G3tHWP
 from sotodlib.site_pipeline import util
 default_logger = util.init_logger(__name__, 'make-hwp-solutions: ')
+print(sotodlib.hwp.g3thwp.__file__)
```
ykyohei, line 21: This line looks unnecessary.

sotodlib/hwp/g3thwp.py
```diff
@@ -737,15 +736,19 @@ def write_solution_h5(self, tod, output=None, h5_address=None):
         try:
             solved = self.analyze(data, mod2pi=False)
         except Exception as e:
-            logger.error(f"Exception '{e}' thrown while calculating HWP angle. Encoder signal might have too much noise.")
+            logger.error(f"Exception '{e}' thrown while calculating HWP angle. Angle calculation failed.")
+            self._write_solution_h5_logger = 'Angle calculation failed'
             self._write_empty_solution_h5(tod, output, h5_address)
             return
-
+
```
ykyohei, line 743: unnecessary space

sotodlib/hwp/g3thwp.py
```diff
@@ -723,11 +720,13 @@ def write_solution_h5(self, tod, output=None, h5_address=None):
             data = self.load_data(start, end)
         except Exception as e:
             logger.error(f"Exception '{e}' thrown while loading HWP data. The specified encoder field is missing.")
+            self._write_solution_h5_logger = 'HWP data too short'
```
ykyohei, line 723: This logger should be "Error in loading HWPdata"?

ykyohei (commented):
Thank you. Overall looks good to me. Made some minor comments.

sotodlib/hwp/g3thwp.py
yukisakurai: This is not an error, just a short data length and no need to debug, so I prefer to leave it.

ykyohei (approved):

PR #677, opened 2024-01-25
update_obsdbcols.py, a script to add and populate some columns to an existing obsdb

AERAdler:
Add stuff without remaking the obsdb or obsfiledb: config file should have obsdb, obsfiledb and a list of the columns to add to the obsdb. Will only try to update books listed in the obsfiledb.

### DIFF
--- a/sotodlib/site_pipeline/update_obsdb.py
+++ b/sotodlib/site_pipeline/update_obsdb.py
@@ -1,6 +1,7 @@
 """update_obsdb.py
 
 Create and/or update an obsdb and obsfiledb based on some books.
+Alternatively, update an obsdb based on an obsfiledb.
 The config file could be of the form:
 
 .. code-block:: yaml
@@ -16,7 +17,7 @@ The config file could be of the form:
       subtype: str
 
     obsdb: dummyobsdb.sqlite
-    obsfiledb: dummyobsfiledb.sqlite
+    obsfiledb: dummyobsfiledb.sqlite Needed if running with the --update_cols option
     lat_tube_list_file: path to yaml dict matching tubes and bands
     tolerate_stray_files: True
     skip_bad_books: True
@@ -28,7 +29,7 @@ The config file could be of the form:
 
 """
 
-from sotodlib.core.metadata import ObsDb
+from sotodlib.core.metadata import ObsDb, ObsFileDb
 from sotodlib.core import Context 
 from sotodlib.site_pipeline.check_book import main as checkbook
 from sotodlib.io import load_book
@@ -86,6 +87,7 @@ def main(config: str,
         booktype: Optional[str] = "both",
         verbosity: Optional[int] = 2,
         overwrite: Optional[bool] = False,
+        update_cols: Optional[bool] = False,
         logger=None):
 
     """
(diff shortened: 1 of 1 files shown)


sotodlib/site_pipeline/update_obsdbcols.py
```diff
@@ -0,0 +1,129 @@
...
+    tnow = time.time()
+    if recency is not None:
+        tback = tnow - recency*86400
+    else:
+        tback = 0 #Back to the UNIX Big Bang 
+    
+    obs_ids = obsfiledb.get_obs()
+    for obs_id in obs_ids:
+        if obsdb.get(obs_id) is None:
+            logger.info(f"{obs_id} in obsfiledb but not in obsdb, skipping")
```
kmharrington, line 98: there might be a point where we want to add new entries to the `obsdb` if we find them in the `obsfiledb`. is it possible to have this work here?

sotodlib/site_pipeline/update_obsdbcols.py
```diff
@@ -0,0 +1,129 @@
...
+            continue
+        if obsdb.get(obs_id)["start_time"]<tback:
+            logger.info(f"{obs_id} starts too far in the past, skipping")
+            continue
+
+        obs_files = obsfiledb.get_files(obs_id)
+        first_file_key = next(iter(obs_files))
+        book_path = os.path.dirname(obs_files[first_file_key][0][0])
+        if os.path.exists(os.path.join(book_path, "M_index.yaml")):
```
kmharrington, line 107: this should probably record an error somehow if it's not found?

AERAdler: Good catch

sotodlib/site_pipeline/update_obsdb.py
```diff
@@ -235,7 +235,12 @@ def main(config: str,
             if stream_ids is not None:
                 bookcartobsdb.add_obs_columns(["wafer_count int"])
                 very_clean["wafer_count"] = len(stream_ids)
-
+                bookcartobsdb.add_obs_columns(["ufms str"])
+                very_clean["ufms"] = (", ".join(stream_ids)).replace("ufm_","")
+            wafer_slots = index.pop("wafer_slots")
+            if wafer_slots is not None:
+                bookcartobsdb.add_obs_columns(["wafer_slots str"])
+                very_clean["wafer_slots"] = (', '.join([sl["wafer_slot"] for sl in wafer_slots]))
```
mhasself, line 243: Request for simply  ','.join rather than ', '.  That's gentler on users who are going to split this string, or who rely on whitespace being used to separate true columns in a text output...

sotodlib/site_pipeline/update_obsdb.py
```diff
@@ -235,7 +235,12 @@ def main(config: str,
             if stream_ids is not None:
                 bookcartobsdb.add_obs_columns(["wafer_count int"])
                 very_clean["wafer_count"] = len(stream_ids)
-
+                bookcartobsdb.add_obs_columns(["ufms str"])
+                very_clean["ufms"] = (",".join(stream_ids)).replace("ufm_","")
+            wafer_slots = index.pop("wafer_slots")
+            if wafer_slots is not None:
+                bookcartobsdb.add_obs_columns(["wafer_slots str"])
+                very_clean["wafer_slots"] = (",".join([sl["wafer_slot"] for sl in wafer_slots]))
```
mhasself, line 243: Thanks for this -- but how do odd computed things like this play with update_obsdbcols?

Would it make sense to have a function that does "collect all obsdb data for book _X_", and then call that function from both update_obsdb.py and update_obsdbcols.py?

I think you should also consider just putting all of this capability into update_obsdb.py.  There's a lot of duplicate logic + boilerplate already, so your new code won't add too many lines; it could be nice to have all the obsdb building / rebuilding in one place.

PR #680, opened 2024-01-26
G3thwp overflow glitch

ykyohei:
Found a hwp angle glitch due to the overflow of the 32 bit internal counter of the bbb in `obs_1704077057_satp3_1111111`, and implemented process to detect and correct this type of hwp angle glitch.

I note that there is another type of glitch due to noisy quad in this observation that is independent from this bug.
We will address in another PR.

Figure 25

### DIFF
--- a/sotodlib/hwp/g3thwp.py
+++ b/sotodlib/hwp/g3thwp.py
@@ -1063,7 +1063,7 @@ class G3tHWP():
             self._rising_edge = np.array([])
 
     def _process_counter_overflow_glitch(self):
-        """ Treat glitches due to 32 bits counter overflow """
+        """ Treat glitches due to 32 bit internal counter overflow """
         idx = np.where(np.diff(self._encd_clk)>2**32)[0] + 1
         for i in idx:
             self._encd_clk[i] -= 2**32
--- a/sotodlib/hwp/g3thwp.py
+++ b/sotodlib/hwp/g3thwp.py
@@ -809,6 +809,9 @@ class G3tHWP():
         # check IRIG timing quality
         self._irig_quality_check()
 
+        # check 32 bit internal counter overflow glitch
+        self._process_counter_overflow_glitch()
+
         # treat counter index reset due to agent reboot
         self._process_counter_index_reset()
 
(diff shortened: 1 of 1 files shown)


sotodlib/hwp/g3thwp.py
```diff
@@ -1059,12 +1062,21 @@ def _irig_quality_check(self):
             self._irig_time = np.array([])
             self._rising_edge = np.array([])
 
+    def _process_counter_overflow_glitch(self):
+        """ Treat glitches due to 32 bit internal counter overflow """
+        idx = np.where(np.diff(self._encd_clk)>2**32)[0] + 1
+        for i in idx:
+            self._encd_clk[i] -= 2**32
+        if len(idx) > 0:
```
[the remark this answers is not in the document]
ykyohei: Thank you for your suggestion. Corrected the order of warning and fixing.

sotodlib/hwp/g3thwp.py
```diff
@@ -1059,12 +1062,21 @@ def _irig_quality_check(self):
             self._irig_time = np.array([])
             self._rising_edge = np.array([])
 
+    def _process_counter_overflow_glitch(self):
+        """ Treat glitches due to 32 bit internal counter overflow """
+        idx = np.where((np.diff(self._encd_clk)>2**32) & (np.diff(self._encd_clk)<2**32+1e6))[0] + 1
```
mhasself, line 1067: This is surprising to me ... doesn't `_encd_clk` increase at each step?  So the overflow should be when diff is a large *negative* number?

Also, a typical overflow in a 32-bit truncated word would not have a diff as big as 2**32 ... it would only be 2**32 - 1, and it's often simply detected by comparing to 2**31.

So if what you put here is correct, maybe it needs more comments describing the issue that you are solving with this...

ykyohei: I was surprized to find this glitch too. 
I modified the detection criteria.
To be honest, I'm not exactly sure why this happens, but I found that the 32-bit jump can explain the glitch.

The bbb internal counter overflows roughly once in 35 mins (200MHz, 32bit). 
This jump doesn't happen most of the time.

yukisakurai:
I found that there are spikes in unix time of HWPEncoder_full, not only bbb internal clock.
We can guess that there was a temporary problem in the PC side where the HWPEncoder agent is running.
I confirmed this code itself is working correctly to solve.
So please put comments describing the issue in the code, as Matthew suggested.
Lets merge this after that.
counter_0
counter_1

yukisakurai (approved):

ykyohei:
Thank you for your confirmation.
I added the description of this glitch.
I'll merge this PR shortly.

PR #685, opened 2024-01-31
make_det_info_wafer: use site-pipeline wafer_map.yaml

mhasself:
Mostly cosmetic change, trying to reduce duplication of config info.

Instead of requiring a custom list of stream_id -> array_name in the config file, use wafer_map.yaml from shared config repo.  (You can override on command line though.)

Relies on "array_name" being added to that wafer_map config.

### DIFF
--- a/docs/site_pipeline.rst
+++ b/docs/site_pipeline.rst
@@ -135,26 +135,36 @@ resulting ManifestDbs should work for both level 2 and level 3 SMuRf data.
 
 make_det_info_wafer
 ```````````````````
-This script uses based array construction inputs to build detector IDs for a set of
-UFMs and save them in a ManifestDb / HDF5 file. The formatting of the ResultSet 
-saved in HDF5 file will map all this information into ``det_info.wafer`` when used 
-with a correctly formatted context file and a readout to detector id mapping.
-The detector info mapping created by this script will be stable as long as the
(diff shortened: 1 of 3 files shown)


kmharrington:
Question: Beyond the first cooldown with satp1, these wafer maps are now written into the books. Can we not use those? The idea was that the ones in the books would be changed if/when we ever changed the installed UFMs so that we could have a live record.

mhasself:
> Question: Beyond the first cooldown with satp1, these wafer maps are now written into the books. Can we not use those? The idea was that the ones in the books would be changed if/when we ever changed the installed UFMs so that we could have a live record.

Right but I'm not using the stream_id <-> wafer_slot part of it, what I need is stream_id <-> array_name.  And a list of all interesting stream_id.  In the long term I think a file like wafer_map.yaml will be a good, quick source that lists "all stream_id of interest".  Without that, you'd need to scan ... every book? to figure out the list of all used stream_ids.

kmharrington:
The issue is this is mixing permanent things (stream_id --> array name) and things that could change with time (what telescope it's in). Honestly at this point we have the rule that `ufm_XvY` is the stream id and `XvY` is the array name, so this little config setup seems like just extra things to keep updated.

mhasself:
As discussed elsewhere, this won't use wafer_map.yaml and instead we'll just give it a list of stream_ids.

PR #686, opened 2024-01-31
Remove overlapping operations from multi-ufm observations

kmharrington:
I'm amazed we only just found this to be an issue. But the logic was working out that operations for single UFMs that overlap multi-UFM observations were trying to be added to two books at once. This requires all level 2 observations going into obs books to be of type obs.


guanyilun (approved):
looks good to me! surprised that this hasn't been a problem till now

PR #688, opened 2024-02-01
1/f +white noise fitting and new highpass filter

msilvafe:
This adds the ability to fit the PSD to a noise model that includes white + low-f noise. And construct a highpass filter with an adjustable f_knee and rolloff slope/index.

We started using this with @susannaaz and @chervias for SAT mapmaking (the fixed filter part not the 1/f fits) which is why I wanted to try to get this merged but there's a few things we might consider adding/changing before merge (but wanted to see how much @kmharrington thinks this should hold us back):

1. Adding into a preprocess the 1/f parameters.
2. Changing the fits to be on the PSD instead of the ASD so that there's not the awkward sqrt in the numerator.


kmharrington:
Once upon a time I wrote a whole rant about 1/f fitting and how it gets messed up if the 1/f knee frequencies are too low (section 3.2 in this paper: https://arxiv.org/abs/2101.00034 ). If we're planning to use this on demodulated data a lot then it's probably worth implementing a less biased fitting algorithm. The one referenced in that paper fits in log space instead of linear space. 

At the time I wrote that I went through enough of the other CMB papers to conclude that 1/f knee frequency was defined in terms of PSD and not ASD. (That's what I meant when I had wn^2 in the equation definition in that paper). So fitting the PSD is correct.

msilvafe:
> Once upon a time I wrote a whole rant about 1/f fitting and how it gets messed up if the 1/f knee frequencies are too low (section 3.2 in this paper: https://arxiv.org/abs/2101.00034 ). If we're planning to use this on demodulated data a lot then it's probably worth implementing a less biased fitting algorithm. The one referenced in that paper fits in log space instead of linear space.
> 
> At the time I wrote that I went through enough of the other CMB papers to conclude that 1/f knee frequency was defined in terms of PSD and not ASD. (That's what I meant when I had wn^2 in the equation definition in that paper). So fitting the PSD is correct.

Cool thanks for sharing. Would you like us to implement this sort of PSD estimation using the log-binned periodogram instead of the standard welch method we've been using? And perform the low f_knee-fit bias correction you got from simulations in that paper? Or do you think just modifying the fitting to be of the PSD calculated from welch in log space is ok for now?

Immediately I just want the filter with a tunable knee and rolloff index so perhaps I can split this PR to get that in sooner while we work on the better PSD noise model fitting?

mhasself:
I have found that basic curve fitting of a simple PSD, to the 1/f + white noise model, works well enough _if you minimize the correct function_.  It's not in general correct to use a simple bin-by-bin least squares function, because the statistics of random variables in a PSD are not gaussian.  But they're the squares of gaussian distributed variables, so it can be saved.  I suspect this issue is what @kmharrington was running into in her analysis.

The correct analytic form is a bit odd, let me try to find it.

mhasself:
> The correct analytic form is a bit odd, let me try to find it.

It is remarkably hard to find this business stated clearly; I probably just don't have the right search terms.  But this blog totally gets it:
http://keatonb.github.io/archivers/powerspectrumfits

Look for the definition of `neglnlike`.

msilvafe:
> > The correct analytic form is a bit odd, let me try to find it.
> 
> It is remarkably hard to find this business stated clearly; I probably just don't have the right search terms. But this blog totally gets it: http://keatonb.github.io/archivers/powerspectrumfits
> 
> Look for the definition of `neglnlike`.

Awesome, thanks @mhasself!

sotodlib/tod_ops/fft_ops.py
```diff
@@ -293,3 +294,115 @@ def calc_wn(aman, pxx=None, freqs=None, low_f=5, high_f=10):
...
+        fidx = np.argmin(np.abs(10**np.polyval(pfit, np.log10(f)) - wnest))
+        p0 = [f[fidx], wnest, -pfit[0]]
+        res = minimize(neglnlike, p0, args=(f, p), method='Nelder-Mead')
+        try:
+            Hfun = ndt.Hessian(lambda params : neglnlike(params, f, p), full_output=True)
+            hessian_ndt, _ = Hfun(res['x'])
+            covout[i] = np.sqrt(np.diag(np.linalg.inv(hessian_ndt)))
+        except np.linalg.LinAlgError:
+            print(f'Cannot calculate Hessian for detector {aman.dets.vals[i]} skipping.')
+            covout[i] = [np.nan, np.nan, np.nan]
```
msilvafe, line 391: So the `Nelder-Mead` method in minimize doesn't return a Hessian or covariance matrix. So I've imported this `numdifftools` package to calculate it. Is it ok to add a new dependency like this @mhasself or @kmharrington?

msilvafe:
I ran fits on demodQ calculating PSD on a 10 min interval (max_samples and nperseg = 10*60*200 in fft_ops.calc_psd) from a CMB scan in SATp1 which still has plenty of 1/f noise from the big polarizing filter so f-knee fits aren't in the super low limit that @kmharrington brought up as being the most biased but I've attached the data and fits for the first 100 detectors, and I think things are looking pretty good:
image

sotodlib/tod_ops/fft_ops.py
mhasself: I do hate adding packages.  But go for it.  Please store the full 3x3 covar though ... or don't call it covar.

msilvafe:
Ok, I had to make some changes to preprocess to get this to work. But now it all seems to be running fine. I think this is ready to merge but I think @kmharrington should probably look at the changes I made to PSD + noise in preprocess. 

An example of the preprocess config file I tested with is at `compute11:/so/home/msilvafe/analysis_notebooks/preprocess_config_noise_fits.yaml` and I was testing in notebook at: `compute11:/so/home/msilvafe/analysis_notebooks/20250212_lowf_fits.ipynb`

The axis manager at the end has the following fields so you can look at a plot of the T/Q/U PSD and fits like this:

```
eix = np.argmin(np.abs(aman.psdQ.freqs-2))
plt.loglog(aman.psdQ.freqs[:eix], aman.psdQ.Pxx[0,:eix], alpha=0.4, label='Q data')
plt.loglog(aman.psdQ.freqs[:eix], 
           fft_ops.noise_model(aman.psdQ.freqs[:eix], aman.noise_fit_Q.fit[0]), label='Q fit')

plt.loglog(aman.psdU.freqs[:eix], aman.psdU.Pxx[0,:eix], alpha=0.4, label='U data')
plt.loglog(aman.psdU.freqs[:eix], 
           fft_ops.noise_model(aman.psdU.freqs[:eix], aman.noise_fit_U.fit[0]), label='U fit')

plt.loglog(aman.psdT.freqs[:eix], aman.psdT.Pxx[0,:eix], alpha=0.4, label='T data')
plt.loglog(aman.psdT.freqs[:eix],
           fft_ops.noise_model(aman.psdT.freqs[:eix], aman.noise_fit_T.fit[0]), label='T fit')

plt.legend()
plt.xlabel('Frequency [Hz]')
```


image

kmharrington (approved):
I love the look of those fits! I may one day have more thoughts on the processing implementation but looks pretty good for now.

PR #690, opened 2024-02-05
Add support for single detector maps

keskitalo:


sotodlib/toast/workflows/proc_mapmaker.py
```diff
@@ -143,35 +143,58 @@ def mapmaker(job, otherargs, runargs, data):
...
-                # Data object that only covers one observation
-                obs_data = data.select(obs_uid=obs.uid)
-                # Replace comm_world with the group communicator
-                obs_data._comm = new_comm
-                job_ops.mapmaker.name = f"{orig_name}_{obs.name}"
-                job_ops.mapmaker.reset_pix_dist = True
-                job_ops.mapmaker.apply(obs_data)
+        # See if user wants separate detector maps
+        if hasattr(otherargs, "detmaps") and otherargs.detmaps:
+            my_dets = data.all_local_detectors()
```
tskisner, line 148: This might cause problems if some detectors are cut in all observations.  You could replace this line with
```
my_dets = data.all_local_detectors(flagmask=defaults.det_mask_invalid)
```
to get the global detectors which are not cut by major issues.

sotodlib/toast/workflows/proc_mapmaker.py
```diff
@@ -143,35 +143,58 @@ def mapmaker(job, otherargs, runargs, data):
...
+            if data.comm.comm_world is None:
+                all_dets = my_dets
+            else:
+                from matplotlib.cbook import flatten
```
tskisner, line 152: Interesting, I had never seen this method!  Definitely useful for the typical use case of calling `gather` / `allgather` on python lists, resulting in a list of lists.  We could consider doing a namespace import as `toast.mpi.flatten` and use it more generally.

tskisner (approved):
This looks fine. One small improvement noted if you wish.

keskitalo:
This PR requires https://github.com/hpc4cmb/toast/pull/743 to be merged first.

PR #691, opened 2024-02-05
[WIP] QA metrics for preprocesstod

tristpinsm:
I fleshed out one way that recording QA metrics could work in `preprocesstod`. This is very much a work in progress and I haven't tested anything at all, but I thought it would be useful to get some feedback on whether or not this may be a good approach (and on all the ways I am misunderstanding the codebase!).

I tried to integrate it into the `preprocesstod` flow, while making it possible to run after the fact without loading the data. Not sure if this ends up being awkward in the end... One limitation may be that this results in recording a single `detset` at a time, because this is how `preprocesstod` does it, but otherwise it might be more efficient to load it all at once (?).

### DIFF
--- a/sotodlib/preprocess/core.py
+++ b/sotodlib/preprocess/core.py
@@ -30,7 +30,7 @@ class _Preprocess(object):
         self.calc_cfgs = step_cfgs.get("calc")
         self.save_cfgs = step_cfgs.get("save")
         self.select_cfgs = step_cfgs.get("select")
-    
+
     def process(self, aman, proc_aman):
         """ This function makes changes to the time ordered data AxisManager.
         Ex: calibrating or detrending the timestreams. This function will use
@@ -114,7 +114,26 @@ class _Preprocess(object):
         if self.select_cfgs is None:
             return meta
         raise NotImplementedError
-    
+
+    @classmethod
+    def gen_metric(cls, meta, proc_aman):
+        """ Generate a QA metric from the output of this process.
+
+        Arguments
+        ---------
+        meta : AxisManager
+            Metadata related to the specific observation
+        proc_aman : AxisManager
+            The output of the preprocessing pipeline.
+
+        Returns
+        -------
+        line : dict
+            InfluxDB line entry elements to be fed to
+            `site_pipeline.monitor.Monitor.record`
+        """
+        raise NotImplementedError
+
     @staticmethod
     def register(process_class):
         """Registers a new modules with the PIPELINE"""
@@ -126,7 +145,6 @@ class _Preprocess(object):
             raise ValueError(
                 f"Preprocess Module of name {name} is already Registered"
             )
-
 
 
 class Pipeline(list):
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -45,6 +45,7 @@ class DetBiasFlags(_Preprocess):
     .. autofunction:: sotodlib.tod_ops.flags.get_det_bias_flags
     """
     name = "det_bias_flags"
+    _influx_field = "det_bias_flags_frac"
 
     def calc_and_save(self, aman, proc_aman):
         msk = tod_ops.flags.get_det_bias_flags(aman, merge=False,
@@ -67,6 +68,43 @@ class DetBiasFlags(_Preprocess):
         keep = ~proc_aman.det_bias_flags.det_bias_flags
         meta.restrict("dets", meta.dets.vals[has_all_cut(keep)])
         return meta
+
+    @classmethod
+    def gen_metric(cls, meta, proc_aman):
+        """ Generate a QA metric from the output of this process.
+
+        Arguments
+        ---------
+        meta : AxisManager
+            The full metadata container.
+        proc_aman : AxisManager
+            The metadata containing just the output of this process.
+
+        Returns
+        -------
+        line : dict
+            InfluxDB line entry elements to be fed to
+            `site_pipeline.monitor.Monitor.record`
+        """
+        # For now just compute the fraction of samples that were flagged
+        # (I think in practice this will be 1 or 0...)
+        # TODO: come up with a more useful metric
+        frac_flagged = np.array([
+            np.dot(r.ranges(), [-1, 1]).sum() / meta.obs_info.n_samples
+            for r in proc_aman.det_bias_flags.det_bias_flags
+        ])
+        obs_time = [meta.obs_info.timestamp] * frac_flagged.size
+        # extract these tags for the metric
+        tag_keys = ["detset", "readout_id", "stream_id", "wafer_slot", "tel_tube", "det_id", "bandpass"]
+        tags = [{k: meta.det_info[k][d] for k in tag_keys if k in meta.det_info} for d in range(meta.dets.count)]
+        tags[0]["telescope"] = meta.obs_info.telescope
+        return {
+            "field": cls._influx_field,
+            "values": frac_flagged,
+            "timestamps": obs_time,
+            "tags": tags,
+        }
+
 
 class Trends(_Preprocess):
     """Calculate the trends in the data to look for unlocked detectors. All
--- a/sotodlib/qa/metrics.py
+++ b/sotodlib/qa/metrics.py
@@ -1,0 +1,141 @@
+from ..core import metadata
+
+
+class QAMetric(object):
+    """ Base class for quality assurance metrics to be recorded in Influx,
+    derived from processed metadata files.
+
+    The base class includes methods for recording a metric value to Influx,
+    checking if a given obs_id has been recorded for this metric, and fetching
+    a list of obs_id's that can be recorded.
+
+    Metrics are labelled by the observation ID they describe, an Influx 'measurement'
+    and 'field', and any number of additional tags. A measurement can have multiple
+    fields associated with it, and here we've tried to organize the metrics so that
+    a source of information (e.g. HWP angle solutions) is a measurement and relevant
+    quantities describing it (e.g. HWP success, mean rate, etc) are fields.
+
+    Subclasses should implement the `_process` and `_get_available_obs` methods to
+    access the source of data they are going to tap, as well as define the
+    `_influx_meas` and `_influx_field` attributes.
+    """
+
+    # these will be defined by child classes
+    _influx_meas = None  # measurement to record to
+    _influx_field = None  # the field to populate
+
+    def __init__(self, context, monitor, log="qa_metrics_log"):
+        """ A QA metric base class.
+
+        Arguments
+        ---------
+        context : core.Context
+            Context that includes all necessary metadata to generate metrics.
+        monitor : site_pipeline.monitor.Monitor
+            InfluxDB connection.
+        log : str
+            InfluxDB measurement that is used to log new entries.
+        """
+
+        self.context = context
+        self.monitor = monitor
+        self._influx_log = log
+
+    def exists(self, obs_id, tags={}):
+        """ Check if a metric exists for this obs_id in Influx.
+
+        Arguments
+        ---------
+        obs_id : str
+            The observation ID to check.
+        tags : dict (optional)
+            Further restrict to given tags.
+
+        Returns
+        -------
+        exists : bool
+            Whether it exists or not.
+        """
+        return self.monitor.check(self._influx_field, obs_id, tags, log=self._influx_log)
+
+    def get_existing_obs(self):
+        """ Get a list of observations already recorded to Influx.
+        """
+        # query influx log measurement for observations of this field
+        res = self.monitor.client.query(
+            f"select {self._influx_field}, observation from {self._influx_log}"
+        ).get_points()
+        return [r["observation"] for r in res]
+
+    def process_and_record(self, obs_id, meta=None):
+        """ Generate a metric for this obs_id and record it to InfluxDB.
+
+        Arguments
+        ---------
+        obs_id : str
+            The observation ID to process.
+        meta : AxisManager (optional)
+            The metadata for this observation ID. If not provided will read from file.
+        """
+        if meta is None:
+            meta = self.context.get_meta(obs_id, ignore_missing=True)
+        if meta.obs_info.obs_id != obs_id:
+            raise Exception(f"Metadata does not correspond to obs_id {obs_id}.")
+        line = self._process(meta)
+        log_tags = {"observation": obs_id}  # used to identify this entry
+        for t in line["tags"]:  # make sure this is also recorded in metric itself
+            t.update(log_tags)
+        self.monitor.record(**line, log=self._influx_log, measurement=self._influx_meas, log_tags=log_tags)
+        self.monitor.write()
+
+    def get_new_obs(self):
+        """ Get a list of available observations not yet recorded to InfluxDB.
+        """
+        # get a list of obs_id to update
+        avail_obs = set(self._get_available_obs())
+        exist_obs = set(self.get_existing_obs())
+        return avail_obs - exist_obs
+
+    def _process(self, meta):
+        """ Implement this to process an actual metric."""
+        raise NotImplementedError
+
+    def _get_available_obs(self):
+        """ Implement this for a given metric."""
+        raise NotImplementedError
+
+
+class PreprocessQA(QAMetric):
+    """ A metric derived from a preprocesstod process.
+    """
+
+    _influx_meas = "preprocesstod"
+
+    def __init__(self, context, monitor, process_name, **kwargs):
+        """ In addition to the context and monitor, pass the name of the
+        preprocess process to record. It should have a `gen_metric` method
+        implemented and `_influx_field` attribute.
+        """
+        super().__init__(context, monitor, **kwargs)
+
+        from sotodlib.preprocess import Pipeline
+        proc = Pipeline.PIPELINE.get(process_name, None)
+        if proc is None:
+            raise Exception(f"No preprocess process with name {process_name}")
+        self._pipe_proc = proc
+
+        # get the field name from the process
+        self._influx_field = self._pipe_proc._influx_field
+
+    def _process(self, meta):
+        return self._pipe_proc.gen_metric(meta, meta.preprocess)
+
+    def _get_available_obs(self):
+        # find preprocess manifest file
+        man_file = [p["db"] for p in self.context["metadata"] if p.get("name", "") == "preprocess"]
+        if len(man_file) == 0:
+            raise Exception(f"No preprocess metadata block in context {self.context.filename}.")
+
+        # load manifest and read available observations
+        man_db = metadata.ManifestDb.from_file(man_file[0])
+        return [o[0] for o in man_db.get_entries(["\"obs:obs_id\""]).asarray()]
--- a/sotodlib/qa/metrics.py
+++ b/sotodlib/qa/metrics.py
@@ -1,3 +1,6 @@
+import re
+import numpy as np
+
 from ..core import metadata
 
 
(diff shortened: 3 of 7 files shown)


kmharrington:
I have implementation questions but bigger picture first: Are we sure we want this so integrated with the existing `preprocess_tod.preprocess_tod` function?  The `if qa_only` behavior you've built in really just points me to "run this as a different function."  You're eventually going to need to add lots of logic to "check if entry is already in influx" etc. I think we might want to split these to "function that generate datasets" and "function that parses datasets and posts to influx," especially because we're very close to needing to MPI / otherwise parallelize the generation of the datasets. Am I missing something we gain by doing it this way?

sotodlib/preprocess/core.py
```diff
@@ -101,19 +109,39 @@ def select(self, meta):
...
+        meta : AxisManager
+            Metadata related to the specific observation
+        proc_aman : AxisManager
+            The output of the preprocessing pipeline.
+        """
+
+        if self.qa_cfgs is None or self.monitor is None:
+            return
+        metric = self._gen_qa(meta, proc_aman)
+        self.monitor.record(**metric, measurement=self._influx_meas)
```
kmharrington, line 139: Are you sure these lines won't change in the future / won't constrain posting different measurements? I think you're basically just pushing the definition of all the kwargs to the `_gen_qa` function so it's fine, but just double checking.

ahincks:
Nice to see a PR!

I have a similar question to Katie: my inclination has been to have it as a separate script, which I think we discussed, but perhaps there are compelling arguments to put it in `preprocess_tod`? I'm open to the best option.

Also, is the `qa_cfg` dictionary defined somewhere?

sotodlib/site_pipeline/preprocess_tod.py
```diff
@@ -109,6 +118,14 @@ def preprocess_tod(obs_id, configs, overwrite=False, logger=None):
     for group in groups:
         logger.info(f"Beginning run for {obs_id}:{group}")
 
+        if qa_only:  # avoid loading data
+            # TODO: using the dets={group_by:group} argument didn't work in testing...
```
kmharrington, line 122: needs fixes that are on this branch: https://github.com/simonsobs/sotodlib/pull/675

kmharrington:
> I think the benefit of doing it this way would be that it is generated in step with the `preprocess` results so that in normal operation you don't need to go back and scrape things and check if they were already entered. So like a model where the tasks record their metrics, vs scraping them after the fact.

One note here is that my goal for the preprocess modules/preprocess tod was for them to remain useful well beyond the first light pipeline and automated site analysis. To the point where I'd expect this to be run off-site more than on-site by the time we're done with SO. As long as "scraping after the fact" is easy, which I think it is, then I don't see a reason to really want to run QA posts exactly when we run the analysis. 

Another reason to split running preprocess_tod and recording QA is the grouping. We're often going to run preprocess_tod with small groups of detectors b/c of RAM limits but then you might just be wanting to send "one number per observation" or "few numbers per observation." Might be nice to not have RAM limits jump in and cause issues for how you want to present data. 

Related: do you have an influx db to test with? Looking at the code that sends an entry per group I'm not instantly sure it will work, would be nice to see what this ends up looking like.

tristpinsm:
Thanks for taking a look and for all the feedback! This is exactly what I was hoping for.

It sounds like making the metrics gathering a parallel pipeline is the way to go, so I'll start looking at that and come back with an alternative implementation.

> Related: do you have an influx db to test with? Looking at the code that sends an entry per group I'm not instantly sure it will work, would be nice to see what this ends up looking like.

I do not and it would be very helpful. That was going to be the next-ish step, and I think you suggested I ask Brian Koopman about it.

ahincks:
Thanks @kmharrington for your good advice/observations! And @tristpinsm the plan sounds good. (And Brian can help assess whether the end-to-end system will give you what you need to test out InfluxDB.)

tristpinsm:
OK, I revived my parallel attempt at a QA metric pipeline and filled it in a bit. Now this is really just a sketch, but maybe useful to take a look given the discussion about how to proceed.

kmharrington:
Oh sorry! I think you pulled out a lot more than I was thinking. I like having the QA calculation options as part of the preprocess modules. (so the stuff you had in here: https://github.com/simonsobs/sotodlib/tree/master/sotodlib/preprocess). I think it makes all kind of sense to have optional QA metric calculations in there. 

I just think we don't want to have the thing that sends data to influx running in `site_pipeline.preprocess_tod.preprocess_tod()`. I think it should be either a different function or a different script that runs after `preprocess_tod` has been completed for an observation.

tristpinsm:
Ah ok, I guess I'm a bit confused about the scope here then. If the task that collects and records metrics is going to be a separate pipeline from `preprocesstod`, should the code still be part of those structures? It wouldn't run in sequence with the `process` and `calc_and_save` methods, right? It could live there, but then it would be wrapped or called by some parallel pipeline.

And then, I was assuming there would be other sources of metrics that might need to be supported beyond `preprocess`, but maybe that is not the case?

ahincks:
> And then, I was assuming there would be other sources of metrics that might need to be supported beyond `preprocess`, but maybe that is not the case?

I'm not certain, but I think it could be! Good point.

kmharrington:
I think we may be confusing words for things and/or there are other misunderstandings for the structures as they exist. I've been calling all the code that lives in `sotodlib.preprocess` to be the "preprocess modules." I like the idea of adding 

https://github.com/simonsobs/sotodlib/blob/b03c141758d19fe2474a112fb05d4079f72a3488/sotodlib/site_pipeline/preprocess_tod.py#L121-L127

kmharrington:
(sorry pushing 10000 wrong buttons while trying to comment)

kmharrington:
Since your commit history is beautiful, I'd say to revert just back to here: https://github.com/simonsobs/sotodlib/pull/691/commits/582df9659196d6c3b06bc98bbafc9f70079a0024  and then it's just edits to what is implemented in this commit; https://github.com/simonsobs/sotodlib/pull/691/commits/b03c141758d19fe2474a112fb05d4079f72a3488

tristpinsm:
Alright, thanks for laying all that out so clearly :) I think I have a path forward on this.

ahincks:
If we want metrics that come from methods outside of `sotodlib.preprocess`, does that argue for a separate `record-qa` script? For instance, if we want metrics from `make-hwp-solutions`, or, e.g., a planet mapper, then it might make sense not to put this in `preprocess_tod`. What do you think?

tristpinsm:
> If we want metrics that come from methods outside of `sotodlib.preprocess`, does that argue for a separate `record-qa` script? For instance, if we want metrics from `make-hwp-solutions`, or, e.g., a planet mapper, then it might make sense not to put this in `preprocess_tod`. What do you think?

So more like the scheme that I have currently pushed to this branch?

My impression is there is sort of two options pulling in opposite directions here: on the one hand the desire to have this integrated with `preprocesstod` and on the other of having a more generic system for QA metrics in general. I have a kind of half baked version of both of these now, and I'm not sure how feasible it is to combine them.

One thing I could do is keep the calculation of the preprocess metrics in `preprocess.processes` but move all of the Influx-related code out to a more generic object like I have here. But then I'd like to know, is there any need to support recording these metrics as part of the `preprocesstod` script (i.e. immediately after generating the output of that script)? Because that sets some constraints due to having to integrate things into that `process` structure.

kmharrington:
hmm... right. Maybe that's an argument for taking the monitor connection things out of preprocess.core and putting them in site-pipeline somewhere. 

We still probably want to use the `_gen_qa` type functions inside of preprocess modules to generate the statistics that come out of preprocess-tod. But then have that plug in somewhere else which combines results from that and things like beam maps, etc.

kmharrington:
> But then I'd like to know, is there any need to support recording these metrics as part of the `preprocesstod` script (i.e. immediately after generating the output of that script)? Because that sets some constraints due to having to integrate things into that `process` structure.

The one nice thing about integrating the QA generation bit with preprocess modules will be that it will be more flexible to changes later on (at least, that was my goal). If someone wants to change the config file that manages the site's preprocess-tod that should also simultaneously fix the QA setup (ex: we stop calculating value X, the QA system also automatically stops looking for it to be there). But if you're already building a bunch of other stuff maybe this overhand isn't that bad? idk. 

I think a good step might be working through examples of things that don't come out of preprocess-tod, since I think most will come from there. How will other results be saved and then pushed to influx?

tristpinsm:
FWIW here are the changes I made based on the previous round of comments. In this scheme, there is a separate function in `preprocesstod` that runs through all the QA metrics generation, which themselves live in the preprocess module.

If we decide we want a more generic system, I can try and pull all the Influx steps out of `preprocesstod`/`preprocess` (as I have started in the `qa` module).

guanyilun:
Sorry to be slow in the game! I like how the monitoring is integrated with preprocess tod, but I do have one suggestion. It looks like currently each Preprocess module will initiate a new connection to influx db (is that right?). I think it's more robust to adopt a singleton pattern for database connection if possible. One way is to pass in a client instead of initiating a client when constructing each preprocess module. I think @kmharrington has some plan to turn pipeline into a class. If so I think the monitor client could be initiated when pipeline is constructed and then get passed down to all of its preprocess modules as an object.

tristpinsm:
Alright, sorry for the continued back-and-forth but I think we may be converging on something!

I've moved everything out of the `preprocess` module except for the calculation itself. There is also a single `Monitor` instance now, addressing @guanyilun 's comment.

This is definitely still a sketch, but let me know if the structure is more appropriate. Thanks for the feedback, and for bearing with me.

sotodlib/preprocess/processes.py
```diff
@@ -57,32 +60,50 @@ def select(self, meta):
...
+    @staticmethod
+    def gen_metric(cls, meta, proc_aman):
+        # For now just compute the fraction of samples that were flagged
+        # (I think in practice this will be 1 or 0...)
+        # TODO: come up with a more useful metric
+        frac_flagged = np.array([
+            np.dot(r.ranges(), [-1, 1]).sum() / meta.obs_info.n_samples
+            for r in proc_aman.det_bias_flags
+        ])
+        obs_time = [meta.obs_info.timestamp] * frac_flagged.size
```
tristpinsm, line 72: just noting that the only reason to pass the `meta` object is to extract this timestamp, but in the new scheme this could probably be done in the `QAMetric` method.

sotodlib/qa/metrics.py
```diff
@@ -0,0 +1,77 @@
...
+        ---------
+        obs_id : str
+            The observation ID to check.
+
+        Returns
+        -------
+        exists : bool
+            Whether it exists or not.
+        """
+        return self.monitor.check(self._influx_field, obs_id, log=self._influx_log)
```
tristpinsm, line 37: hm, this may not be unique enough, unless I log each measurement separately. Still figuring out how this is supposed to work.

sotodlib/preprocess/processes.py
msilvafe: I think this is a totally legitimate way to do this but just to point out as you're writing QA metrics from flags there is a method in the `RangesMatrix` class already called `get_stats()` which you could use. I think the only place it appears in sotodlib at the moment is in `flag_utils` here: https://github.com/simonsobs/sotodlib/blob/60a63dcef338dad7be4ad6d02d5fb348f4cc636c/sotodlib/flag_utils.py#L51

tristpinsm:
I've tested the basic `monitor` operations on a local InfluxDB server, and the next step will be to try this out at site with a test instance so that I can run the full thing with input data. I've asked Brian about setting that up.

But I think the general scheme is more or less in place now.

tristpinsm:
basic example graph in grafana: [link](https://site.simonsobs.org/grafana/explore?orgId=1&left=%7B%22datasource%22:%22b818c9bf-d9a0-4643-8ff5-0c38608b9fb7%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22datasource%22:%7B%22type%22:%22influxdb%22,%22uid%22:%22b818c9bf-d9a0-4643-8ff5-0c38608b9fb7%22%7D,%22resultFormat%22:%22time_series%22,%22orderByTime%22:%22ASC%22,%22tags%22:%5B%5D,%22groupBy%22:%5B%7B%22type%22:%22time%22,%22params%22:%5B%22$__interval%22%5D%7D,%7B%22type%22:%22tag%22,%22params%22:%5B%22tel_tube::tag%22%5D%7D,%7B%22type%22:%22fill%22,%22params%22:%5B%22null%22%5D%7D%5D,%22select%22:%5B%5B%7B%22type%22:%22field%22,%22params%22:%5B%22det_bias_flags_frac%22%5D%7D,%7B%22type%22:%22sum%22,%22params%22:%5B%5D%7D%5D%5D,%22policy%22:%22autogen%22,%22measurement%22:%22preprocesstod%22,%22alias%22:%22Total%20flagged%20detectors%20%28%5B%5Btag_tel_tube%5D%5D%29%22%7D%5D,%22range%22:%7B%22from%22:%22now-1h%22,%22to%22:%22now%22%7D%7D)

There is only one observation in there right now, but next step is to add more !

ahincks:
> basic example graph in grafana: [link](https://site.simonsobs.org/grafana/explore?orgId=1&left=%7B%22datasource%22:%22b818c9bf-d9a0-4643-8ff5-0c38608b9fb7%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22datasource%22:%7B%22type%22:%22influxdb%22,%22uid%22:%22b818c9bf-d9a0-4643-8ff5-0c38608b9fb7%22%7D,%22resultFormat%22:%22time_series%22,%22orderByTime%22:%22ASC%22,%22tags%22:%5B%5D,%22groupBy%22:%5B%7B%22type%22:%22time%22,%22params%22:%5B%22$__interval%22%5D%7D,%7B%22type%22:%22tag%22,%22params%22:%5B%22tel_tube::tag%22%5D%7D,%7B%22type%22:%22fill%22,%22params%22:%5B%22null%22%5D%7D%5D,%22select%22:%5B%5B%7B%22type%22:%22field%22,%22params%22:%5B%22det_bias_flags_frac%22%5D%7D,%7B%22type%22:%22sum%22,%22params%22:%5B%5D%7D%5D%5D,%22policy%22:%22autogen%22,%22measurement%22:%22preprocesstod%22,%22alias%22:%22Total%20flagged%20detectors%20%28%5B%5Btag_tel_tube%5D%5D%29%22%7D%5D,%22range%22:%7B%22from%22:%22now-1h%22,%22to%22:%22now%22%7D%7D)
> 
> There is only one observation in there right now, but next step is to add more !

It appears to have been deleted, but I did see it last night. Nice work.

tristpinsm:
I think this is ready for a final round of review now. I've tested the code on a parallel InfluxDB on site and made a few basic Grafana dashboards to try it out: https://site.simonsobs.org/grafana/dashboards/f/e5064171-2666-4407-b95f-cebc735a8104/dev.

I've also cleaned up the commit history and tried to document everything.

One thing to note is I've built this to support deriving metrics from metadata files, because these are the example cases I had to work with, but I realise now maybe this is constraining? I think this scheme can pretty easily be adapted for other sources though.

When the code is deemed ready, we could try to deploy it on site, and later move on to adding more metrics.

msilvafe:
> I think this is ready for a final round of review now. I've tested the code on a parallel InfluxDB on site and made a few basic Grafana dashboards to try it out: https://site.simonsobs.org/grafana/dashboards/f/e5064171-2666-4407-b95f-cebc735a8104/dev.
> 
> I've also cleaned up the commit history and tried to document everything.
> 
> One thing to note is I've built this to support deriving metrics from metadata files, because these are the example cases I had to work with, but I realise now maybe this is constraining? I think this scheme can pretty easily be adapted for other sources though.
> 
> When the code is deemed ready, we could try to deploy it on site, and later move on to adding more metrics.

@tristpinsm I'm used to viewing data in grafana over variable time windows. I like how you have it setup now indexed by obsid so you can get a snapshot of all metrics for a given observation, but I think that it would be useful to be able to visualize some of these metrics over time to see if something is systematically degrading over a number of observations. 

Also now that the structure, and config files and such a largely in the final state can you write a some docs into `docs/preprocess.rst` and probably make a new section called `docs/metrics.rst`

tristpinsm:
> @tristpinsm I'm used to viewing data in grafana over variable time windows. I like how you have it setup now indexed by obsid so you can get a snapshot of all metrics for a given observation, but I think that it would be useful to be able to visualize some of these metrics over time to see if something is systematically degrading over a number of observations.

Did you take a look at [this dashboard](https://site.simonsobs.org/grafana/d/b740304e-0c55-42bd-8a8e-6986a63bf291/qa-dev?orgId=1) ? I think that's what you're describing, no?
> Also now that the structure, and config files and such a largely in the final state can you write a some docs into `docs/preprocess.rst` and probably make a new section called `docs/metrics.rst`.

will do!

tristpinsm:
Just added a docs page.

msilvafe:
> Just added a docs page.

Thanks! I think a short doc string for each of these is useful (could point to @ykyohei's hwp docs if available):

Screenshot 2024-03-05 at 12 24 26 PM

Also, can you add in somewhere here an example of how to configure the preprocess metrics using the det_cal you tested so far?

msilvafe:
> Not sure I understand this. What's the det_cal? Do you mean how to configure the QA recording script?

Sorry, I meant your det_bias_flags example but ya in particular I was looking at an example in the docs for how to setup the script for recording metrics from one of the preprocess metadata fields.

tristpinsm:
> Sorry, I meant your det_bias_flags example but ya in particular I was looking at an example in the docs for how to setup the script for recording metrics from one of the preprocess metadata fields.

Do you mean in addition to what is under the "Metric-recording Script" section?

tristpinsm:
Just added docstrings to the HWP metrics and made some changes based on feedback from @ykyohei

msilvafe:
Looks good to me, will leave final approval review for merge to @kmharrington

sotodlib/preprocess/processes.py
```diff
@@ -68,6 +69,42 @@ def select(self, meta, proc_aman=None):
         meta.restrict("dets", meta.dets.vals[has_all_cut(keep)])
         return meta
 
+    @classmethod
```
kmharrington, line 72: this isn't actually a class method as it doesn't generate a class instance? Maybe you want this to be a static method?

kmharrington: Looking into python/code things and how you're using it. This is a bit of a non-standard use of a classmethod (at least how I see it the most). 

Are we ever going to be running multiple pipelines at once? If so the influx-name probably shouldn't be a class variable and should instead be defined through a config file. 

I think we want to include telescope in the `tags` because tel-tube is not unique between the SATs

kmharrington: Sorry I realize I'm mixing nomenclatures. tel-tube is from imprinter and is unique. But I think telescope (`meta.obs_info.telescope`) is another tag you should add to this data.

kmharrington (approved):
Approving because very interested in getting this on site as soon as possible! But I would add that telescope tag to things

sotodlib/preprocess/processes.py
tristpinsm: I'll add the telescope tag.

re `classmethod`: I'm using it here because I want to access a class attribute. I think this is a normal use case, but I'd be curious to know if this is discouraged in any references.

tristpinsm: That's a good question about running multiple pipelines... I thought putting the name in the class definition might be a good way of keeping it standard. Maybe if there are parallel pipelines they should be recording to sepaprate databases?

tristpinsm:
question for merging once I make those changes: Do you squash merge, or is it ok to rebase if I cleaned up the commit history?

PR #694, opened 2024-02-07
Remove negative sign in front of focal-plane theta in det-match

jlashner:
This is required in order for det-match solution pointing to be aligned with measurements, since this PR in site-pipeline-configs: https://github.com/simonsobs/site-pipeline-configs/pull/21/files


mhasself (approved):
I believe you.

PR #697, opened 2024-02-11
Small fix for the toast_so_convert script

chervias:
quick fix for the toast_so_convert script.


tskisner (approved):
Thanks for catching these typos!

PR #700, opened 2024-02-13
Move process registration to __init__.py in Preprocess.

tristpinsm:
I got a bit tripped up figuring out how to import registered processes, because you first had to import the `processes` module, then `PIPELINE` from `core` in order for it to be populated. I thought doing it this way made more sense, but let me know what you think.

Now you can `from sotodlib.preprocess import PIPELINE` to get the registered processes.


kmharrington (approved):
Oh dang, I guess I've never tried running this without having the module imported. Thanks for the catch

PR #706, opened 2024-02-14
Preprocess Upgrades

kmharrington:
classifying the pipeline and expanding utility options. Not complete or fully tested yet.

PR #708, opened 2024-02-15
Fix select in det_bias_cuts process.

msilvafe:
When we first put in this process we weren't returning ranges matrices so select was not using the has_all_cut function. I've added that here. Hopefully with a single line fix this is an easy review.


kmharrington (approved):

msilvafe:
This is fixed in PR #706 - Preprocess Updates so I'll close this and work on reviewing that.

PR #716, opened 2024-02-19
load_book: mods to work better on sim data

mhasself:
This basically makes it so 'primary' in data frames is not necessary. The returned AxisManagers will lack the primary and biases fields.

This was developed on sims so wait for me to test it on some real books ...


mhasself:
I've tested this on a few SATP1 and LAT books now, seems fine.

kmharrington (approved):
works for me

PR #718, opened 2024-02-19
Setting `source_flags` as optional.

iparask:
This PR fixes the following error:
```python
  0    0.38  3.85  4.93  8.59 Downsampling 1,1
  0    0.38  3.85  4.93  8.59 Active downsampling 1
  File "/scratch/gpfs/ip8725/tmp/mapmaker.py", line 286, in <module>
    obs = mapmaking.downsample_obs(obs, passinfo.downsample)
  File "/home/ip8725/.local/lib/python3.10/site-packages/sotodlib/mapmaking/utilities.py", line 513, in downsample_obs
    res.wrap(key, downsample_cut(getattr(obs, key), down), [(0,"dets"),(1,"samps")])
  File "/home/ip8725/.local/lib/python3.10/site-packages/sotodlib/core/axisman.py", line 375, in __getattr__
    return self[name]
  File "/home/ip8725/.local/lib/python3.10/site-packages/sotodlib/core/axisman.py", line 356, in __getitem__
    raise KeyError(name)
KeyError: 'source_flags'
```

### DIFF
--- a/sotodlib/mapmaking/utilities.py
+++ b/sotodlib/mapmaking/utilities.py
@@ -508,9 +508,16 @@ def downsample_obs(obs, down):
         bore.wrap(key, getattr(obs.boresight, key)[::down], [(0, "samps")])
     res.wrap("boresight", bore)
     res.wrap("signal", resample.resample_fft_simple(obs.signal, onsamp), [(0,"dets"),(1,"samps")])
+
     # The cuts
-    for key in ["glitch_flags", "source_flags"]:
+    cut_keys = ["glitch_flag"]
+
+    if "source_flags" in obs:
+        cut_keys.append("source_flags")
+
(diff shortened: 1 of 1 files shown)


chervias:
So wrapping these dummy flags for the mapmakers is fine in the meantime since I think the standard of flags is not yet 100% decided. Eventually, all flags will be added in the preprocessing of the tod, and at the mapmaker stage, the tod will be ready to map directly. That's how it's supposed to work in the SATs at least, so I think it will be the same for the LAT. The tods will have a FlagManager called flags, and inside that, there will be all sorts of flags like glitches, etc. But like I said in the meantime wrapping dummy flags when they are not in the tod is fine.

iparask:
Thank you @chervias I added a TODO comment to keep in mind.

Issue #693, opened 2024-02-06
Preprocess plotting


davidvng:
I'm working on integrating plotting into each step of preprocess and wanted to get comments from sotodlib folks (I've talked with @msilvafe about this but particularly wanted opinions from @kmharrington ). The idea is to add the option to automatically produce plots after each preprocess step, which would then be served to Teleview (or some other web service). The types of plots would depend on the step, specifically separating steps that have `calc` (`calc_and_save`) vs those that have `process`.

- After each `calc_and_save` step, make plots specific to that step's calculation (I have an example [here](https://github.com/simonsobs/sotodlib/compare/master...preprocess-plots) for `get_det_bias_flags`)
- After each `process` step, make an updated PSD plot. Then maybe make one at the end with all of them overplotted like in [this draft PR](https://github.com/simonsobs/sotodlib/pull/522).

This plotting would be an option to set to True in the preprocess config. The path and naming conventions are other topics of discussion. I was just going to follow the typical 5 digit timecode structure along with the `obs_id` and the plot type (ex: `/17047/obs_1704770693_satp1_1111111_bias_cuts.png`)

kmharrington:
I think the biggest first thing to decide on is "inside the tod-ops functions" or "inside the preprocess modules." For your first example, you aren't actually doing anything with the preprocess modules so it doesn't really have to do with the `calc_and_save` step.  That's really editing the plotting into the tod-ops functions directly to include a plot/save option.

I think I'm leaning toward we should do this all as an expansion to the preprocess modules and not inside the functions directly. That would continue to separate the "functions that do things to/with the TODs" from the "results/statistics from those functions." Basically it keeps all the IO out of `tod-ops`. ( TBH lots of overlap between plotting and generating statistics so there are a bunch of thoughts in this PR: https://github.com/simonsobs/sotodlib/pull/691 ) I think I would do this by editing the modules `calc_and_save` or `save` functions to look for specific plotting keywords and then building plots there. The `calc_and_save` function is responsible for calling the `save` function so you can be flexible about what goes between those two. And `calc_and_save` is called after `process` so it will catch changes made to there. 

The one downside to "inside preprocess modules" is it's not totally clear to me what the overhead looks like for somehow who "just wants to run this function and plot what things look like." To it's worth thinking if there's an easy way to make things flexible enough for that to be easy. 

[...]

PR #719, opened 2024-02-19
Preprocess plots

davidvng:
PR for preprocess plotting, described in #693.

Overall changes:
- Restructure `Preprocess` to implement `plot`, separated from `calc_and_save`.
- Made plotting functions for `DetBiasFlags` and `EstimateHWPSS` preprocess steps so far.
- Made new `obs_ops` module for observation-level operations. Added `get_sso` function.
- Made new preprocess step `SSOFootprint`
- Made new site-pipeline script `preprocess_obs.py` to use with `SSOFootprint` process that's run per-obs instead of per-wafer.

# Preprocess_tod
# 1. DetBiasFlags
Plots venn diagram of flagged dets found by `get_det_bias_flags` separated by the ranges that go into the RangesMatrix.
https://github.com/simonsobs/sotodlib/blob/4ee8e227648b58fe6c64a463d15ee97354705ca8/sotodlib/preprocess/preprocess_plot.py#L12-L13
image
Additionally, I'm making a function to plot a pie chart of the fractions of dets cut per flag type. This function should be able to take additional flag types as the process pipeline continues. An example after this first step:
image

# 2. EstimateHWPSS
Two plots are made here, one for 4f/2f counts and one for the HWPSS fit status (from #522)
https://github.com/simonsobs/sotodlib/blob/4ee8e227648b58fe6c64a463d15ee97354705ca8/sotodlib/preprocess/preprocess_plot.py#L57
image

https://github.com/simonsobs/sotodlib/blob/4ee8e227648b58fe6c64a463d15ee97354705ca8/sotodlib/preprocess/preprocess_plot.py#L125-L126
image

Config file being used to run `preprocess_tod.py` with `--obs-id=obs_1704770693_satp1_1111111`:
```
context_file: '/so/home/dnguyen/repos/scripts/contexts/satp1_scr3_context_20240125.yaml'

plot_dir: '/so/home/dnguyen/repos/scripts/preprocess/plots'

subobs:
  use: detset
  label: detset
  
archive:
  index: '/so/home/dnguyen/repos/scripts/preprocess/process_archive.sqlite'
  policy:
    type: 'simple'
    filename: '/so/home/dnguyen/repos/scripts/preprocess/preprocess_archive.h5'

process_pipe:
    - name: "det_bias_flags"
      calc:
        name: "det_bias_flags"
        rfrac_range: [0.05, 0.9]
        psat_range: [0, 20]
      save: True
      plot: True

    - name: "detrend"
      process:
        method: "median"

    - name: "estimate_hwpss"
      calc:
        hwpss_stats_name: "hwpss_stats"
      save: True
      plot: True
```

# Preprocess_obs
I made new site-pipeline script `preprocess_obs.py` for per-observation preprocessing. This was needed since the current `preprocess_tod.py` loads data grouped by dets. I added `obs_ops` which currently only has `sources.get_sso` which is used in the `SSOFootprint` process.

# 1. SSOFootprint
Finds nearby sources using `sotodlib.coords.planets.get_nearby_sources` and plots SSO footprint for each source.
https://github.com/simonsobs/sotodlib/blob/d0b1d491ef88860a0ca65d55bbd49a18b0e6f5d2/sotodlib/preprocess/preprocess_plot.py#L202
image
image

Config file being used to run `preprocess_obs.py` with `--obs-id=obs_1704938434_satp1_1111111`:
```
context_file: '/so/home/dnguyen/repos/scripts/contexts/satp1_scr3_context_20240125.yaml'

plot_dir: '/so/home/dnguyen/repos/scripts/preprocess/plots'

archive:
  index: '/so/home/dnguyen/repos/scripts/preprocess/process_obs_archive.sqlite'
  policy:
    type: 'simple'
    filename: '/so/home/dnguyen/repos/scripts/preprocess/preprocess_obs_archive.h5'

process_pipe:
    - name: "sso_footprint"
      calc:
        distance: 20
        nstep: 100
      save: True
      plot:
        wafer_offsets: {'ws0': [-2.5, -0.5],
                        'ws1': [-2.5, -13],
                        'ws2': [-13, -7],
                        'ws3': [-13, 5],
                        'ws4': [-2.5, 11.5],
                        'ws5': [8.5, 5],
                        'ws6': [8.5, -7]}
        focal_plane: '/so/home/msilvafe/shared_files/sat_hw_positions.npz'
```


sotodlib/hwp/hwp.py
```diff
@@ -11,7 +11,7 @@ def get_hwpss(aman, signal=None, hwp_angle=None, bin_signal=True, bins=360,
               lin_reg=True, modes=[1, 2, 3, 4, 6, 8], apply_prefilt=True,
               prefilt_cfg=None, prefilt_detrend='linear', flags=None,
               merge_stats=True, hwpss_stats_name='hwpss_stats',
-              merge_model=True, hwpss_model_name='hwpss_model'):
+              merge_model=True, hwpss_model_name='hwpss_model', **kwargs):
```
kmharrington, line 14: why do you need to add **kwargs here if there's no changes later?

kmharrington:
### Code structure thoughts

I really like the addition of `preprocess_plot` and the separation there. Could you move the import statements into the modules only if save_plot = True? So instead of `from . import preprocess_plot` do something like 

```
if self.calc_cfgs.get('save_plot', False):
            from .preprocess_plot import plot_4f_2f_counts
            plot_4f_2f_counts(aman, save_path=self.calc_cfgs['save_plot'])
            hwpss_stats = hwp.get_hwpss(aman, **self.calc_cfgs)
```

I think python as a whole caches the imports so that would be a little slower for the first obs_id we run on but then would be identical. (someone could check me on that).

### Plotting Thoughts

I don't think those detector bias plots gets use a whole lot of useful information. Raw TODs will always be that gross looking and we're always going to be running those plots on raw TODs. Is there a way you could design a single summary plot that just indicates how many detectors are cut by each criteria? Like a pie chart or something?

kmharrington:
Oh, I have a whole second thought. We could add another keyword to all process modules that is "plot" (so like, in here https://github.com/simonsobs/sotodlib/blob/79d60c72a748d2dcd7d46ea3f5e87145800a7072/sotodlib/preprocess/core.py#L28) and then setup the Pipeline object and/or Pipeline.run to know if it's plotting or not. 

That way there may be a way to use the same configuration file for "running and calculating on all observation IDs" and then "come back later and plot results from only some of them"? Worth thinking about it and seeing how much harder that would be to implement.

msilvafe:
> ### Plotting Thoughts
> I don't think those detector bias plots gets use a whole lot of useful information. Raw TODs will always be that gross looking and we're always going to be running those plots on raw TODs. Is there a way you could design a single summary plot that just indicates how many detectors are cut by each criteria? Like a pie chart or something?

Ya, I had thought that see what was being cut would be useful to see if we could visually identify if we started flagging a new population with these conditionals but with this being an early cut with lots of trending channels in it all in very different peak-peak scales it's not that informative. But I suppose we could just see this from the number of detectors (or fraction of detectors) but from each of the conditionals stated in the 6 plot titles. 

For the HWPSS plot (left one). I think you want to plot the unbinned signal with just points `'o'` otherwise you get lines connecting 2pi -> 0. Also instead of restricting the plot to angle < 3 I would just plot a ~10 seconds of data (or 2000 samples) this will give you plenty of sampling of the HWPSS but won't have all of the low frequency noise. For the binned signal I think having points with lines connects i.e. `'.-'` and then having caps on the errorbars i.e. `capsize` argument

sotodlib/hwp/hwp.py
davidvng: I needed to add this since these functions don't take `save_plot` which is within `**self.calc_cfgs` that's being passed

sotodlib/preprocess/preprocess_plot.py
```diff
@@ -0,0 +1,239 @@
...
+    modes : list of int
+        The HWPSS harmonic modes to extract.
+    filename : str
+        Full filename with direct path to plot output directory.
+    """
+    hwpss_ratsatp1 = {}
+    fig, axs = plt.subplots(3, 2, figsize=(15, 15))
+    for i, band in enumerate(['f090', 'f150']):
+        hwpss_ratsatp1[band] = {}
+        m90s = ((aman.det_cal.bg == 0) | (aman.det_cal.bg == 1) | (aman.det_cal.bg == 4) | (aman.det_cal.bg == 5) | (aman.det_cal.bg == 8) | (aman.det_cal.bg == 9))
```
msilvafe, line 81: Let's not do this...I think if det_info exists we should use `aman.det_info.bandpass == f090` for the 90GHz mask (and similar for 150) and if you want to use `amen.det_cal.bg` then you should use: `m90s = np.isin(man.det_cal.bg, [0,1,4,5,8,9])` for 90 GHz (and similar for 150)

sotodlib/preprocess/preprocess_plot.py
```diff
@@ -0,0 +1,239 @@
...
+        m150s = ((aman.det_cal.bg == 2) | (aman.det_cal.bg == 3) | (aman.det_cal.bg == 6) | (aman.det_cal.bg == 7) | (aman.det_cal.bg == 10) | (aman.det_cal.bg == 11))
+        m = [m90s, m150s]
+        hwpss_aman = aman.restrict('dets', aman.dets.vals[m[i]], in_place=False)
+        hwpss_aman.restrict('samps',(20*60*200, -100))
```
msilvafe, line 85: This samps restrict maybe shouldn't be in there...or the restrict range should be a tuple you can pass. This hardcoded value is a leftover kludge from an old SATp1 dataset that had the HWP spinning up for it's first 20 minutes...

sotodlib/preprocess/preprocess_plot.py
```diff
@@ -0,0 +1,239 @@
...
+                          label = f"median: {np.median(a_2f):.2e}")
+        axs[2, i].legend()
+        axs[2, i].set_xscale('log')
+        axs[2, i].set_xlabel('A_2f')
+        
+    for i in range(3):
+        axs[i, 0].set_ylabel('Counts')
+
+    obs_ts = aman.timestamps[0]
+    det = aman.dets.vals[0]
```
msilvafe, line 126: Same comment about obsid/readoutid

sotodlib/preprocess/preprocess_plot.py
```diff
@@ -0,0 +1,239 @@
...
+    Parameters
+    ----------
+    aman : AxisManager
+        Input axis manager.
+    planet_aman : AxisManager
+        Axis manager of results from obs_ops.sources.get_sso for a single planet.
+    sso : str
+        Name of planet.
+    wafer_offsets : dict
+        Dictionary of wafer offsets.
```
msilvafe, line 205: Add an example here of how to pass this.

sotodlib/preprocess/preprocess_plot.py
```diff
@@ -0,0 +1,239 @@
...
+        wafer_offsets = {'ws0': (-2.5, -0.5),
+                         'ws1': (-2.5, -13),
+                         'ws2': (-13, -7),
+                         'ws3': (-13, 5),
+                         'ws4': (-2.5, 11.5),
+                         'ws5': (8.5, 5),
+                         'ws6': (8.5, -7)}
+
+    # Get default focal plane from sotodlib
+    hw = np.load('/so/home/msilvafe/shared_files/sat_hw_positions.npz')
```
msilvafe, line 223: This file path needs to be an argument and if it's going to be used by the pipeline script needs to be stored somewhere more permanent than my user home directory.

sotodlib/preprocess/processes.py
```diff
@@ -47,10 +50,16 @@ class DetBiasFlags(_Preprocess):
     name = "det_bias_flags"
 
     def calc_and_save(self, aman, proc_aman):
-        msk = tod_ops.flags.get_det_bias_flags(aman, merge=False,
-                                               **self.calc_cfgs)
+        if self.plot_cfgs is None:
```
msilvafe, line 53: I think it's better to modify `flags.det_bias_flags` to always return an aman and if `full_output = True` have it wrap all of the submasks but also the main mask (called det_bias_flags). Here in the process let's just have it always run with full_output=True (not just when plot_cfgs is present) these two things should get rid of the two different mans you have here (dbc_aman and msk_aman)

sotodlib/preprocess/processes.py
```diff
@@ -585,6 +620,49 @@ class SubPolyf(_Preprocess):
...
+    name = 'sso_footprint'
+
+    def calc_and_save(self, aman, proc_aman):
+        ssos = planets.get_nearby_sources(tod=aman, distance=self.calc_cfgs.get("distance", 20))
+        if ssos:
+            sso_aman = core.AxisManager()
+            for sso in ssos:
+                planet = sso[0]
+                xi_p, eta_p = obs_ops.sources.get_sso(aman, planet, nstep=self.calc_cfgs.get("nstep", 100))
+                planet_aman = core.AxisManager()
```
msilvafe, line 638: This AxisManager should at least have a samps axis and when you wrap in xi_p/eta_p wrap them along the samps axis.

sotodlib/preprocess/processes.py
```diff
@@ -585,6 +620,49 @@ class SubPolyf(_Preprocess):
...
+            proc_aman.wrap("sso_footprint", sso_aman)
+
+    def plot(self, aman, proc_aman, filename):
+        if self.plot_cfgs is None:
+            return
+        if self.plot_cfgs:
+            from .preprocess_plot import plot_sso_footprint
+            filename = filename.replace('{ctime}', f'{str(aman.timestamps[0])[:5]}')
+            filename = filename.replace('{obsid}', aman.obs_info.obs_id)
+            for sso in proc_aman.sso_footprint.keys():
```
msilvafe, line 659: What you want here is `proc_aman.sso_footprint._assignment.keys()` then you shouldn't need the `sso == 'dets' or ssh == 'samps'` check.

msilvafe (changes requested):
This needs docs overviewing the difference between preprocess_obs and preprocess_tod and explaining the various plotting options with example configuration entries.

sotodlib/preprocess/processes.py
davidvng: The samps axis will be downsampled from `nstep` and this will downsample `proc_aman` too when wrap into it. Would this be an issue?

davidvng:
Addressed @msilvafe 's comments, @kmharrington do you also want to take a look before merging?

docs/preprocess.rst
```diff
@@ -125,7 +135,44 @@ This pipeline can be run through the functions saved in ``site_pipeline``. Each
 entry in "process_pipe" key will be used to generate a Preprocess module based
 on the name it is registered to. These entries will then be run in order through
 the processing pipe. The ``process`` function is always run before the
-``calc_and_save`` function for each module.
+``calc_and_save`` function for each module. The ``plot`` function can be run after
+``calc_and_save`` when ``plot: True`` for a module that supports it. Current modules
+with plotting: ``DetBiasFlags``, ``EstimateHWPSS``, ``SSOFootprint``.
+
+Example TOD Pipeline Configuration File
```
msilvafe, line 142: Maybe you should call this "Example Obs Pipeline Configuration File" in contrast to the "TOD Pipeline Configuration File" above. Now it looks like two entries with the same name in the tree
Screenshot 2024-03-13 at 6 22 26 PM

docs/preprocess.rst
```diff
@@ -125,7 +135,44 @@ This pipeline can be run through the functions saved in ``site_pipeline``. Each
 entry in "process_pipe" key will be used to generate a Preprocess module based
 on the name it is registered to. These entries will then be run in order through
 the processing pipe. The ``process`` function is always run before the
-``calc_and_save`` function for each module.
+``calc_and_save`` function for each module. The ``plot`` function can be run after
+``calc_and_save`` when ``plot: True`` for a module that supports it. Current modules
+with plotting: ``DetBiasFlags``, ``EstimateHWPSS``, ``SSOFootprint``.
```
msilvafe, line 140: Now every time you add a new module you need to remember to update this list here...I might just leave out that last sentence.

docs/preprocess.rst
```diff
@@ -125,7 +135,44 @@ This pipeline can be run through the functions saved in ``site_pipeline``. Each
...
+          save: True
+          plot:
+            wafer_offsets: {'ws0': [-2.5, -0.5],
+                            'ws1': [-2.5, -13],
+                            'ws2': [-13, -7],
+                            'ws3': [-13, 5],
+                            'ws4': [-2.5, 11.5],
+                            'ws5': [8.5, 5],
+                            'ws6': [8.5, -7]}
+            focal_plane: '/so/home/msilvafe/shared_files/sat_hw_positions.npz'
```
msilvafe, line 175: I still want to get this moved to somewhere more official and don't really want a path to somewhere in my user home on site computing in docs for a public repo.

docs/site_pipeline.rst
```diff
@@ -447,6 +447,19 @@ the preprocessing steps applied to them.
    :module: sotodlib.site_pipeline.preprocess_tod
    :func: get_parser
 
+preprocess-obs
+--------------
+This script is set up to run a preprocessing pipeline using the preprocess
+module. See details in :ref:`See details here<preprocess-module>` for how to
+build an obs preprocessing pipeline.
+
+This module is similar to ``preprocess_tod`` but removes grouping by detset so
```
msilvafe, line 456: I like this maybe add this sort of statement also up when you introduce tod/obs in the preprocess.rst.

sotodlib/preprocess/preprocess_plot.py
```diff
@@ -219,8 +222,11 @@ def plot_sso_footprint(aman, planet_aman, sso, wafer_offsets=None, filename='./s
                          'ws5': (8.5, 5),
                          'ws6': (8.5, -7)}
 
-    # Get default focal plane from sotodlib
-    hw = np.load('/so/home/msilvafe/shared_files/sat_hw_positions.npz')
+    if focal_plane is None:
+        # Get default focal plane from sotodlib
+        hw = np.load('/so/home/msilvafe/shared_files/sat_hw_positions.npz')
```
msilvafe, line 226: @skhrg does this exist somewhere office that we can point to by default? It's really just wafer outlines in xi-eta...we could even probably define this in a tiny file that lives in sotodlib potentially.

sotodlib/preprocess/processes.py
```diff
@@ -50,16 +50,8 @@ class DetBiasFlags(_Preprocess):
     name = "det_bias_flags"
 
     def calc_and_save(self, aman, proc_aman):
```
msilvafe, line 52: I'm planning to merge this branch (preprocess-full) in soon (this week hopefully), can you test merging that branch into this one and confirm you can still run through your examples? Won't hold anything up from this tho if you don't.

https://github.com/simonsobs/sotodlib/blob/250f257c7cef67ca3f5dfa3a2efa8cbf1fa017e1/sotodlib/preprocess/processes.py#L50-L55

sotodlib/preprocess/core.py
```diff
@@ -30,6 +31,7 @@ def __init__(self, step_cfgs):
         self.calc_cfgs = step_cfgs.get("calc")
```
msilvafe, line 31: For the doctoring of `_Preprocess` also add a mention to `plot` where it describes the list of keys that are matched to functions in the module.

msilvafe (approved):
Happy to have merged, but please clean up the couple remaining docs comments I added in.

sotodlib/preprocess/processes.py
msilvafe: You can create a new offset axis (called like `ds_samp`) and wrap them into that something like this:

```
planet_aman = core.AxisManager(core.OffsetAxis('ds_samps', count=len(downsampled_axis),
                                                          offset=aman.samps.offset, 
                                                          origin_tag=aman.samps.origin_tag)
```

docs/preprocess.rst
kmharrington: seconded

docs/preprocess.rst
kmharrington: always use fake paths for the documentation.

And I think this is something we could load from Saianeesh's setup? A future issue/upgrade, shouldn't prevent merging.

sotodlib/site_pipeline/preprocess_obs.py
```diff
@@ -0,0 +1,231 @@
...
+
+    # Update the index.
+    db_data = {'obs:obs_id': obs_id,
+                'dataset': dest_dataset}
+    
+    logger.info(f"Saving to database under {db_data}")
+    if len(db.inspect(db_data)) == 0:
+        db.add_entry(db_data, dest_file)
+
+def load_preprocess_det_select(obs_id, configs, context=None):
```
kmharrington, line 106: Is detector selection something that's necessary for observation level analysis?

kmharrington (changes requested):
Small change but I don't want it merged before it's changed. We shouldn't have hard coded paths in any of sotodlib. (Ping me on slack when it's fixed and I'll approve)

sotodlib/preprocess/processes.py
davidvng: I've changed `tod_ops.flags.get_det_bias_flags` to output an AxisManager following your suggestion in another comment, so merging this in will create two layers of 'det_bias_flags'.

davidvng:
@kmharrington @msilvafe ready to merge

kmharrington (approved):
Let's try it! Want to get it running at the site

PR #720, opened 2024-02-20
Fix demod mapmaking

mhasself:
Addresses problems in coords.demod.make_map identifed by @chervias.

I believe that what was in there before was correct only when the focal plane and celestial coordinates were well aligned; e.g. when looking due north or due south, or in source-scan coordinates.

There's a unit test now.


chervias (approved):
I tested it and it looks good to me

mhasself:
> I tested it and it looks good to me

Thanks!

PR #723, opened 2024-02-21
Updates to det-match and site-pipeline-script to create better solutions

jlashner:
A few updates to `det_match` and `update_det_match` that I added when creating new solutions and a new manifestdb for sat1 wafers.

Detmatch changes:
- PointingConfig class used to help with generating pointing from an optics model
- `ResSet.from_wafer_info_file` fn to load a resonator set with resonator design info based on Matthew's wafer_info file.
- In `ResSet.from_array`, adds option to ignore extra field, which lets you load older ResSet data after the Resonator interface has changed.
- Bug fixes in the `get_stats` fn.

update_det_match change:
- Begins to support different "solution-set" configurations. Adds some solution configuration that allows you to base matches off of a dir of npy ResSet files instead of kaiwen's solutions in site-pipeline-configs. I expect this to evolve as we begin to understand more about what we need from solutions and how long they remain applicable for (I can see us wanting to use different solutions for different cooldowns/tunes for example)

This is needed for the corresponding PR in site-pipeline-configs that creates the `update_det_match_config` file: https://github.com/simonsobs/site-pipeline-configs/pull/40

### DIFF
--- a/sotodlib/site_pipeline/update_det_match.py
+++ b/sotodlib/site_pipeline/update_det_match.py
@@ -72,7 +72,9 @@ class UpdateDetMatchesConfig:
     resonator_set_dir: Optional[str]
         If ``solution_type`` is 'resonator_set', this must be specified and
         contain the path to the resonator-set solutions. This directory must
-        have a res-set npy file for each stream_id that is expected in the matching.
+        have a res-set npy file for each stream_id that is expected in the
+        matching, formatted like ``<resonator_set_dir>/<stream_id>.npy``, which
+        contains the result from ``np.save(fname, match.merged.as_array())``.
     
     Attributes
     -------------
--- a/sotodlib/site_pipeline/update_det_match.py
+++ b/sotodlib/site_pipeline/update_det_match.py
@@ -63,6 +63,16 @@ class UpdateDetMatchesConfig:
     write_relpath: bool
         If True, will use the relative path to the h5 file (relative to the db
         path) when writing to the manifestdb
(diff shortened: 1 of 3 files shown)


sotodlib/site_pipeline/update_det_match.py
```diff
@@ -63,6 +63,16 @@ class UpdateDetMatchesConfig:
...
+    solution_type: str
+        Type of solutions to use. Must be one of ['kaiwen_handmade',
+        'resonator_set'].  If 'kaiwen_handmade', will use the handmade solutions
+        from Kaiwen pulled from the wafer_map file in the site-pipeline-configs.
+        If `resonator_set`, must also specify the ``resonator_set_dir`` to pull
+        solutions from.
+    resonator_set_dir: Optional[str]
+        If ``solution_type`` is 'resonator_set', this must be specified and
+        contain the path to the resonator-set solutions. This directory must
+        have a res-set npy file for each stream_id that is expected in the matching.
```
mhasself, line 75: Can you connect this explicitly to the name of the class in the code that would write out the npy file?  I.e. "as written by ResonatorSet.save()" or whatever.  Also, filename pattern is required here?

mhasself (approved):
Approving, but please address one inline docs comment. 

Generally I was reading this for documentation...  Consider whether additional site_pipeline.rst documentation is warranted.  Is the current det_match config file running at site substantially more complicated than the examples presented in the site_pipeline docs?  If so, please expand the docs.

PR #727, opened 2024-02-22
Adds bias wave metadata to metadata index

jlashner:
Looked into the code, and turns out the bookbinder just copies every npy file into the book, so I think that should not be a problem with bias waves: https://github.com/simonsobs/sotodlib/blob/468bd7f75eddb2b324a40a0c17917056221b21f4/sotodlib/io/bookbinder.py#L1157-L1160

The added lines just correctly point to the copied bias_waves file in M_index yaml file.


msilvafe (approved):
Thanks for the quick work on this.

PR #730, opened 2024-02-26
Compress h5 output of axisman.save

ykyohei:
Compress h5 output of axisman.save.
I chose gzip with default compression level as default. This is lossless compression and available with every installation of HDF5.
https://docs.h5py.org/en/stable/high/dataset.html#lossless-compression-filters

I tested this compression on satp3 hwp angle files. The data size reduces by factor of ~4.

Before
-rw-r--r-- 1 site-pipeline so       138G Feb 12 20:06 hwp_angle.h5

After
total 32G
-rw-r--r-- 1 ykyohei so 608K Feb 24 15:33 hwp_angle.sqlite
-rw-r--r-- 1 ykyohei so 2.2G Feb 24 05:17 hwp_angle_1701.h5
-rw-r--r-- 1 ykyohei so 1.2G Feb 24 06:16 hwp_angle_1702.h5
-rw-r--r-- 1 ykyohei so 3.9G Feb 24 08:32 hwp_angle_1703.h5
-rw-r--r-- 1 ykyohei so  27G Feb 24 13:08 hwp_angle_1704.h5
-rw-r--r-- 1 ykyohei so 1.4G Feb 24 14:57 hwp_angle_1707.h5
-rw-r--r-- 1 ykyohei so 159M Feb 24 15:33 hwp_angle_1708.h5

### DIFF
--- a/sotodlib/core/axisman.py
+++ b/sotodlib/core/axisman.py
@@ -821,7 +821,7 @@ class AxisManager:
             self._assignments.update(aman._assignments)
         return self
 
-    def save(self, dest, group=None, overwrite=False):
+    def save(self, dest, group=None, overwrite=False, compression=None):
         """Write this AxisManager data to an HDF5 group.  This is an
         experimental feature primarily intended to assist with
         debugging.  The schema is subject to change, and it's possible
@@ -835,6 +835,7 @@ class AxisManager:
             dest).
           overwrite (bool): If True, remove any existing thing at the
             specified address before writing there.
+          compression (str ot None): Compression filter to apply.
 
         Notes:
           If dest is a string, it is taken to be an HDF5 filename and
@@ -868,7 +869,7 @@ class AxisManager:
 
         """
         from .axisman_io import _save_axisman
-        return _save_axisman(self, dest, group=group, overwrite=overwrite)
+        return _save_axisman(self, dest, group=group, overwrite=overwrite, compression=compression)
 
     @classmethod
     def load(cls, src, group=None):
--- a/sotodlib/core/axisman.py
+++ b/sotodlib/core/axisman.py
@@ -835,7 +835,9 @@ class AxisManager:
             dest).
           overwrite (bool): If True, remove any existing thing at the
             specified address before writing there.
-          compression (str ot None): Compression filter to apply.
+          compression (str or None): Compression filter to apply. E.g.
+            'gzip'. This string is passed directly to HDF5 dataset
+            routines.
 
         Notes:
           If dest is a string, it is taken to be an HDF5 filename and
--- a/sotodlib/core/axisman_io.py
+++ b/sotodlib/core/axisman_io.py
@@ -123,7 +123,7 @@ def _safe_scalars(x):
     # Must be fine then!
     return x
 
-def _save_axisman(axisman, dest, group=None, overwrite=False):
+def _save_axisman(axisman, dest, group=None, overwrite=False, compression='gzip'):
     """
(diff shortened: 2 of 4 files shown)


skhrg:
Going to pop in here because I spent a bit of time messing around with this when I considered trying to force M2 to use h5 files for our TODs instead of fits. Given the size of these files have you played around with chunking at all? When chunked it will only decompress the chunks as you load them so that + a good chunk cache helps a lot with read performance. I think it would be relatively straightforward to add logic to `aman.save` to chunk along an axis.

kmharrington (commented):
Nice! I think my only question is if we want compression on by default? I make a huge number of small AxisManagers just for the convenience of the save/load features while doing analysis. I doubt it will make a difference to have compression always on for that sort of thing but curious what @mhasself  thinks.

mhasself:
I don't think we need chunking yet because our _datasets_ are not actually very large (these HWP files, for example, are only big because they have data for multiple observations ... but those are stored at different addresses and thus should be random-accessible already).  (Though maybe I don't understand HDF5 that well.)

I am inclined to leave compression  _disabled_ by default.  For realistic floaty data, gzip will not do very well, so it's just a waste of CPU.

Your savings in HWP data probably comes from it making the integer fields (such as flags) compress down to almost nothing.

sotodlib/core/axisman_io.py
```diff
@@ -220,21 +220,21 @@ def _save_axisman(axisman, dest, group=None, overwrite=False):
...
-                g.create_dataset(k, data=v)
+                g.create_dataset(k, data=v, compression=compression)
         elif item['encoding'] == 'csrarray':
             g = dest.create_group(item['name'])
             for k, v in flatten_csr_array(data).items():
-                g.create_dataset(k, data=v)
+                g.create_dataset(k, data=v, compression=compression)
         elif item['encoding'] == 'axisman':
             g = dest.create_group(item['name'])
             _save_axisman(data, g)
```
mhasself, line 240: Pass compression=compression here so that sub-AxisMans get same treatment.

sotodlib/core/axisman.py
```diff
@@ -835,6 +835,7 @@ def save(self, dest, group=None, overwrite=False):
             dest).
           overwrite (bool): If True, remove any existing thing at the
             specified address before writing there.
+          compression (str ot None): Compression filter to apply.
```
mhasself, line 838: str _or_ None.

Add "E.g. 'gzip'. This string is passed directly to HDF5 dataset routines."

mhasself (changes requested):
Thanks for looking into this.  A couple of changes inline.

Also, please add some tests for this -- see tests/test_core.py, TestAxisManager.test_500_io.

ykyohei:
Added one line inTestAxisManager.test_500_io and the test passed.

```
$ python3 test_core.py
............../so/home/ykyohei/tools/sotodlib/sotodlib/core/axisman_io.py:314: DeprecationWarning: `product` is deprecated as of NumPy 1.25.0, and will be removed in NumPy 2.0. Please use `prod` instead.
  axisman.wrap(item['name'], expand_RangesMatrix(rm_flat), assign)
........
----------------------------------------------------------------------
Ran 22 tests in 0.207s

OK
```

I also ran the entire test. There are some errors, but these are not because of this update.
I paste the output of entire test here.
[240226compress.txt](https://github.com/simonsobs/sotodlib/files/14412270/240226compress.txt)

mhasself (approved):
Cool, thanks!

PR #731, opened 2024-02-26
remove logger arg of the main function in site_pipeline scripts

guanyilun:
logger arg in main functions of each site_pipeline script is to provide a way to pass prefect logger into the script. Now we are using stream capture, so directly passing in logger is no longer required. Directly modifying function signature is also somewhat unstable. This PR is meant to go along with the incoming update in prefect wrapper: https://github.com/simonsobs/so-workflow/pull/33

### DIFF
--- a/sotodlib/site_pipeline/analyze_bright_ptsrc.py
+++ b/sotodlib/site_pipeline/analyze_bright_ptsrc.py
@@ -25,6 +25,8 @@ matplotlib.use('agg')
 
 opj = os.path.join
 
+logger = util.init_logger(__name__)
+
 
 def get_xieta_src_centered(ctime, az, el, data, sso_name, threshold):
     """Create a planet centered coordinate system for single detector data
@@ -521,7 +523,6 @@ def main(
     representative_dets,
     test_mode=False,
     plot_results=False,
-    logger=None,
 ):
     """
     Initiate an MPI environment and fit a simulation in the time domain
@@ -539,8 +540,6 @@ def main(
 
     kwargs['img_dir'] = img_dir
 
-    if logger is None:
-        logger = util.init_logger(__name__)
     t1 = time.time()
 
     comm = MPI.COMM_WORLD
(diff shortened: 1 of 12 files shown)


kmharrington:
Looking at the defaults for imprinter, this should be fine for update_book_plan, but I think I'd prefer to initialize the site-pipeline logger inside the site-pipeline scripts and send them to imprinter? doesn't have to come from prefect though.

guanyilun:
> Looking at the defaults for imprinter, this should be fine for update_book_plan, but I think I'd prefer to initialize the site-pipeline logger inside the site-pipeline scripts and send them to imprinter? doesn't have to come from prefect though.

ok, I have added that back now!

mhasself (approved):
Looks fine to me; thanks.

jlashner:
Currently it looks like if init_logger is called multiple times with the same name it will register the same handlers multiple times... can we add a check for this to prevent duplicate logs?

mhasself:
> Currently it looks like if init_logger is called multiple times with the same name it will register the same handlers multiple times... can we add a check for this to prevent duplicate logs?

Good point.

Another problem is that simply importing a site_pipeline script will cause the logger to print to stdout, if `init_logger` is called on import (rather than, say, in the main function).

I think it might make more sense for the "scripts" in site_pipeline to declare their loggers in the usual library-like way at the top:
```
logger = logging.getLogger(__name__)
```
and then in the _main_ function, call `init_logger` or similar to add a handler that makes the local logger verbose.
```
def main(...):
    ...
    util.init_script_logger(logger, verbosity=verbosity)
```

This is a bit of work + testing so not sure if it goes in this PR.

guanyilun:
> Currently it looks like if init_logger is called multiple times with the same name it will register the same handlers multiple times... can we add a check for this to prevent duplicate logs?

Is this a satisfactory solution to that?
https://github.com/simonsobs/sotodlib/pull/731/commits/4bc8b97a203526e2705661ce71779ea8153e2bbf

jlashner:
I think this should be fine! this is typically what I do.

jlashner (approved):

Issue #689, opened 2024-02-03
context.get_meta can interrupt site_pipeline?


ykyohei:
I knew that keep opening the metadata h5 file interrupts `site_pipeline`, but I was assuming `get_meta` does not.
However, I observed that the `get_meta` sometimes interrupts `make_hwp_solution.py` during the process of re-producing the `hwp_angles`.
I think this is because `aman.save` opens h5 file short period of time.

This happened while I was watching the log of `make_hwp_solution.py`  in real time, so I'm relatively confident that something other than `get_meta` did not cause the interruption.

I did following in the jupyter-lab.
```
ctx = core.Context("/so/home/ykyohei/scratch/ctx_satp3.yaml")
aman = ctx.get_meta('obs_1704138534_satp3_1111111')
```

Following is the log of `make_hwp_solution.py`.
The error looked synchronized with the execution of the `get_meta`.
```
10412.115: Calculating Angles for obs_1704161020_satp3_1111111 (INFO)
WARNING: sotodlib.core.context: Key "metadata" not present in context.
WARNING: sotodlib.hwp.g3thwp: flipping quad is corrected by mean value
ERROR: sotodlib.hwp.g3thwp: Exception '[Errno 11] Unable to open file (unable to lock file, errno = 11, error message = 'Resource temporarily unavailable')' thrown while the template subtraction.
WARNING: sotodlib.hwp.g3thwp: Packet drop exists
WARNING: sotodlib.hwp.g3thwp: 1 dropped packets are found, performing fill process
WARNING: sotodlib.hwp.g3thwp: Offcentering calculation is only available when two encoders are operating. Skipped.
Traceback (most recent call last):
  File "/so/home/ykyohei/tools/sotodlib/sotodlib/hwp/g3thwp.py", line 994, in write_solution_h5
    aman.save(output, h5_address, overwrite=True)
  File "/so/home/ykyohei/tools/sotodlib/sotodlib/core/axisman.py", line 871, in save
    return _save_axisman(self, dest, group=group, overwrite=overwrite)
  File "/so/home/ykyohei/tools/sotodlib/sotodlib/core/axisman_io.py", line 191, in _save_axisman
    file_to_close = h5py.File(dest, 'a')
  File "/so/shared-software/miniconda3/envs/so-base-py310/lib/python3.10/site-packages/h5py/_hl/files.py", line 567, in __init__
    fid = make_fid(name, mode, userblock_size, fapl, fcpl, swmr=swmr)
[...]

PR #732, opened 2024-02-27
Split hwp h5 files

ykyohei:
To address this issue https://github.com/simonsobs/sotodlib/issues/689
- Split h5 files by first 4 digits of Unix time. This might be temporary solution.
- Make make_hwp_solutions.py robust to temporary IO blocking.

We update the version of `hwp_angles` with this PR. Identifier: `satpX_hwp_angles_240301m`
Documentation: https://simonsobs.atlassian.net/wiki/spaces/PRO/pages/306708481/hwp+angles+Metadata+Version+Log#240301

### DIFF
--- a/sotodlib/hwp/g3thwp.py
+++ b/sotodlib/hwp/g3thwp.py
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 
 import os
+import time
 import numpy as np
 import scipy.interpolate
 import h5py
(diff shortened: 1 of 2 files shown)


guanyilun (approved):
I vote for getting this merged now. The changes look simple enough. I ran into IO blocking issue several times in automated runs from prefect, so I need this fix here to get things going. In the long run I think we should look into single-writer-multiple-reader pattern as a solution to IO Blocking, but happy to have something that works for now!

PR #737, opened 2024-03-01
Keeping metadata shape even with preprocess select=True.

msilvafe:
This PR amounts to adding a full samps -- `fsamps` and full dets -- `fdets` which we fill with a nominal value at the beginning before calculating anything in preprocess. Then as preprocess proceeds and the dets or samps axis gets restricted the data is sliced into the full dets and full samps so that it's not further restricted as the pipeline proceeds. This amounts to keeping all of the calculated values and flags at the shape the data was when the calculation was run.


kmharrington:
only comment is I'd put it in `preprocess.core` instead of in `preprocess.preprocess` since it's more of a utility

PR #739, opened 2024-03-05
Convert example preprocess configs to valid yaml

BrianJKoopman:
Title basically says it all. The yaml will be more useful for users writing a preprocess config file.

I added missing classes to the `preprocess.rst` file when I found missing ones with example config blocks in their docstrings.

I also indented the yaml with two spaces, as that seemed like convention in the head of the preprocess configs I've been working with.


kmharrington (approved):
thx

Issue #636, opened 2023-12-19
Restrict `in_place=False` "not working"? 


jlashner:
This might be expected behavior, however it stumped me for a couple of hours today so I wanted to check if it's a bug / make the behavior known...

But it looks like when you restrict an axismanager with `in_place=False`, the new axis-manager that is returned does not have a deep-copy of all of the fields, allowing things like this to happen:
``` python3
obs_id = 'obs_1697164370_satp1_1111111'
am = ctx.get_obs(obs_id, dets=['sch_ufm_mv7_1696796495_4_003'], samples=(0, 100))
am2 = am.restrict('samps', (0, 10), in_place=False)
am2.signal += 20
plt.plot(am.signal[0])
```
![image](https://github.com/simonsobs/sotodlib/assets/4718487/003cbe98-cc4c-48b4-82ba-bb17323f0f28)

It makes sense why this is happening but it seems very unintuitive to me. Is this intended?

kmharrington:
yeah I feel like this is the wrong sort of behavior we want. If in_place is false we should make actual copies.

PR #741, opened 2024-03-05
Make copy of slices because slices are pointers

kmharrington:
Slices need to be copied when restricting with `in_place=False`. Also added a test to catch this error and a comment to remind us not to change the nested in_places. 

Closes https://github.com/simonsobs/sotodlib/issues/636


mhasself (approved):
Thanks!  I feel like we need to document the implications of in_place=True a little more clearly somewhere ... but not for today.

PR #744, opened 2024-03-05
Change default nperseg behaviour

### LINKED ISSUE #724: Change PSD default window length
Right now we just default to all of `scipy.welch`'s defaults but 256 data points gives really useless PSDs for the types of data we work with. Let's use the timestamps in the axismanager to give a slightly better default when one isn't given.

skhrg:
Closes #724 (my first issue!)

### DIFF
--- a/sotodlib/tod_ops/fft_ops.py
+++ b/sotodlib/tod_ops/fft_ops.py
@@ -196,28 +196,34 @@ def calc_psd(
     timestamps=None, 
     max_samples=2**18,
     prefer='center',
+    freq_spacing=None,
     merge=False, 
     overwrite=True, 
     **kwargs
 ):
-    """Calculates the power spectrum density of an input signal using signal.welch()
-    Data defaults to aman.signal and times defaults to aman.timestamps
+    """Calculates the power spectrum density of an input signal using signal.welch().
+    Data defaults to aman.signal and times defaults to aman.timestamps.
+    By default the nperseg will be set to power of 2 closest to the 1/50th of
+    the samples used, this can be overridden by providing nperseg or freq_spacing.
         
     Arguments:
         aman (AxisManager): with (dets, samps) OR (channels, samps)axes.
-        signal (float ndarray): data signal to pass to scipy.signal.welch()
-        timestamps (float ndarray): timestamps associated with the data signal         
-        max_samples (int): maximum samples along sample axis to send to welch
+        signal (float ndarray): data signal to pass to scipy.signal.welch().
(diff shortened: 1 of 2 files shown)


sotodlib/tod_ops/fft_ops.py
```diff
@@ -196,12 +196,15 @@ def calc_psd(
...
     max_samples=2**18,
     prefer='center',
+    freq_spacing=None,
     merge=False, 
     overwrite=True, 
     **kwargs
 ):
     """Calculates the power spectrum density of an input signal using signal.welch()
     Data defaults to aman.signal and times defaults to aman.timestamps
+    By default the nperseg will be set to the base 2 number closest to the 1/50th of
```
mhasself, line 206: End previous sentence ^^^  with a period.

"base 2 number" -> "power of 2"

sotodlib/tod_ops/fft_ops.py
```diff
@@ -211,6 +214,9 @@ def calc_psd(
         prefer (str): One of ['left', 'right', 'center'], indicating what
             part of the array we would like to send to welch if cuts are
             required
+        freq_spacing (float): The approximate desired frequency spacing of the PSD.
+                              If None the default nperseg of ~1/50th the signal length is used.
+                              If an nperseg is explicitly passed then that will be used.
```
mhasself, line 219: Please repair formatting to match other args.

sotodlib/tod_ops/fft_ops.py
```diff
@@ -240,13 +246,15 @@ def calc_psd(
             raise ValueError(f"Invalid choise prefer='{prefer}'")
         start = offset
         stop = offset + max_samples
-    
+    fs = 1/np.nanmedian(np.diff(timestamps[start:stop]))
+    if "npserseg" not in kwargs:
```
mhasself, line 250: npserseg is a typo.  The punishment for this is usually to make a unit test!

mhasself (changes requested):
So I guess there's no point in making the "50" a parameter for the function, because more likely a user will have a particular frequency resolution in mind?  That's fine.

See requests inline.

skhrg:
Requesting permission to fix all the PEP8 issues in the `fft_ops` file in this PR as well.
Also on unit tests, if someone makes a list of ones we want in an issue (or an issue per) I am happy to start banging those out on days like this where site computing is betraying me.

mhasself (approved):
Fine by me!

I guess it's ok to pep8 the file just before merge... . Make sure it's a dedicated commit, no functional changes, and the commit message is clear that it's "reformatting only".

PR #746, opened 2024-03-06
ensure pid_direction to be 1 or -1 or 0

ykyohei:
Fixed issue with getting nan value for pid_direction.

This issue resulted in a ValueError when loading hwp metadata.

ValueError: The field 'pid_direction' does not share axis 'dets'; pid_direction is not identical across all items pass other_fields='drop' or 'first' or else remove this field from the targets.


yukisakurai (approved):
Looks good, it worked correctly on `obs_1700290002_satp1_1111111`.
(Apologies that I can't contribute much lately due to duty works.)

PR #747, opened 2024-03-06
Improve AxisManager concatenation (for metadata loading)

mhasself:
When concatenating AxisManagers (during metadata loading):
- Non-dets vectors that contain nans do not automatically trigger a concatenation failure.
- You can use metadata spec entry like `drop_fields: ["one", "two", "other_*"]` to discard things before concatenation. 
- We don't concatenate single items any more, we just return them.

### DIFF
--- a/docs/context.rst
+++ b/docs/context.rst
@@ -114,9 +114,9 @@ annotated example.
   # A list of metadata products that should be loaded along with the
   # TOD.  Each entry in the list points to a metadata database (an
   # sqlite file) and specifies the name under which that information
-  # should be associated in the loaded data structure.  In some
-  # cases a "loader" name is also given, but usually this is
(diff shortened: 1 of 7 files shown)


msilvafe (approved):
Thanks for the update, looking forward to the HWP metadata loading speedup from this :)

In the `Context/Annotated Exmple` section of the docs can you add in an example of how to use `drop_fields` and `load_fields`? Or maybe you have another place you think it would fit well but I think the core of it pretty clearly outlined in the docstring you added to `AxisManager.load` but I think an example of using it at the higher level would be helpful. Also in the docstring of the `MetadataSpec` class you call out the usage of `'.'` for nested fields and `'*'` for wildcard perhaps you can give an example there as well like you do in `AxisManager.load` .

PR #749, opened 2024-03-06
Generalize passing psd and fit cfgs to Noise operation

BrianJKoopman:
I tested this with this `process_pipe`:
```
process_pipe:
    - name : "psd"
      signal: "signal" # optional
      wrap: "psd" # optional
      process:
        psd_cfgs: # optional, kwargs to scipy.welch
          nperseg: 1024
        wrap_name: "psd" # optional
      calc: True
      save: True
    - name: "noise"
      calc:
        low_f: 5
        high_f: 10
      save: True
      select:
        max_noise: 2000
```

The changes below were mostly in response to each failed run of the pipeline. I think this should probably be tested with a more complicated "noise" config before merging though.

### DIFF
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -341,7 +341,10 @@ class Noise(_Preprocess):
        select:
          max_noise: 2000
 
-    .. autofunction:: sotodlib.tod_ops.fft_ops.calc_wn
+    If ``fit: True`` this operation will run
+    :func:`sotodlib.tod_ops.fft_ops.fit_noise_model`, else it will run
+    :func:`sotodlib.tod_ops.fft_ops.calc_wn`.
+
     """
     name = "noise"
 
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -370,10 +370,11 @@ class Noise(_Preprocess):
                                          **self.calc_cfgs)
             noise = core.AxisManager(aman.dets)
             noise.wrap("white_noise", wn, [(0,"dets")])
-            if self.calc_cfgs.get('wrap_name') is None:
-                aman.wrap("noise", noise)
-            else:
-                aman.wrap(self.calc_cfgs['wrap_name'], noise)
+
+        if self.calc_cfgs.get('wrap_name') is None:
+            aman.wrap("noise", noise)
+        else:
+            aman.wrap(self.calc_cfgs['wrap_name'], noise)
 
         self.save(proc_aman, noise)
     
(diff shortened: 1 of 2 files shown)


sotodlib/preprocess/processes.py
```diff
@@ -330,50 +330,66 @@ class Noise(_Preprocess):
...
+                                                    **self.calc_cfgs)
         else:
             wn = tod_ops.fft_ops.calc_wn(aman, pxx=psd.Pxx,
                                          freqs=psd.freqs,
-                                         **self.calc_cfgs['noise_args'])
+                                         **self.calc_cfgs)
             noise = core.AxisManager(aman.dets)
             noise.wrap("white_noise", wn, [(0,"dets")])
-            if self.calc_cfgs['wrap_name'] is None:
+            if self.calc_cfgs.get('wrap_name') is None:
```
msilvafe, line 373: This if-else block is indented once to much. It should get wrapped whether or not fit = True.

sotodlib/preprocess/processes.py
```diff
@@ -330,50 +330,66 @@ class Noise(_Preprocess):
...
+
+     - name: "noise"
+       calc:
+         low_f: 5
+         high_f: 10
+       save: True
+       select:
+         max_noise: 2000
+
     .. autofunction:: sotodlib.tod_ops.fft_ops.calc_wn
```
msilvafe, line 344: This function is only called if fit = False, if True, fit_noise_model is called.

msilvafe (changes requested):
A couple changes then ready to merge I think.

sotodlib/preprocess/processes.py
BrianJKoopman: Fixed in [703905e](https://github.com/simonsobs/sotodlib/pull/749/commits/703905e3bcb1b32f1e428ed739071455e27b3f98).

sotodlib/preprocess/processes.py
BrianJKoopman: Want me to add the `fit_noise_model` autofunction here too then?

BrianJKoopman: After talking about this at lunch, I'm planning on linking to where both of these functions are documented instead. Will get to that later this afternoon.

BrianJKoopman: Alright, linked to this in [da22eb0](https://github.com/simonsobs/sotodlib/pull/749/commits/da22eb0502d83efea8e7dcbbd718e5c5b77478bf). I had to add an automodule block to the `tod_ops` docs page.

msilvafe (approved):
Looks good, thanks @BrianJKoopman

PR #750, opened 2024-03-06
pointing quickfit for time domain fitting

jlashner:
Adds some det-pointing fitting functions developed during the hack session. 

The `pointing_quickfit` function fits pointing for a bright source (such as the moon) for a detector. It has a few differences with respect to the full point source fits, including:
- downsampling signal between filtering and fitting
- only fitting xi and eta, with other beam params fixed
- Use scipy.optimize.minimize instead of curve-fit
- model is filtered at each iteration

The function takes 1-2 sec per detector, and results agree nicely with tomoki's map-based pointing (see below for ws:
image

Some missing dets can hopefully be recovered  using jump-fixes and flags dbs.

A few additional added functions:
- `quickfit_and_save_pointing` runs the fit on specified channels in an axis-manager, and saves the fitted xi, eta to a result set
- `plot_quickfit_res` makes a plot based on the result of the fn, for example:
image

### DIFF
--- a/sotodlib/site_pipeline/analyze_bright_ptsrc.py
+++ b/sotodlib/site_pipeline/analyze_bright_ptsrc.py
@@ -10,16 +10,19 @@ from sotodlib.io.metadata import write_dataset, read_dataset
 from sotodlib.tod_ops.filters import high_pass_sine2, low_pass_sine2, fourier_filter
 
 from astropy import units as u
+from copy import deepcopy
 import numpy as np
 import scipy.signal
 import time
 import ephem
 import os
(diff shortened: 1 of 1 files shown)


samdayweiss (approved):
This looks good Jack, thanks for sharing. I like the status bar addition in quickfit_and_save_pointing :)

mhasself (approved):
Thanks for the nice contributions.  I want everyone in SO to use dataclass more!

I'm approving (there are lots more changes coming to this module) but I think an important thing you could add, to help folks understand and use even in the short term, is to declare the _units_ for all your dataclass fields and function parameters.

Issue #712, opened 2024-02-18
Terminate Preprocess Pipeline when n_dets = 0


msilvafe:
When running a preprocess pipeline and the select method restricts so that the axis manager has no detectors left the next step in the pipeline will fail. This happens occasionally with the detector bias cuts due to tuning in very bad weather or failed tuning. We need to catch this case and make the pipeline exit gracefully (i.e. appropriately update databases with a no_data, or no_detectors all steps after the select) so that we can batch run on all observations with the `preprocess_tod.py` site-pipeline script.

kmharrington:
Yes. I suggest implementing some sort of results dictionary to `pipeline.run()` for status things. So that we can do something like

```
proc_aman, status = pipe.run(aman)
if not status['success']:
    do something about status['message'] 
    ... etc
```

msilvafe:
Addressed in #751

PR #751, opened 2024-03-07
Preprocess Full

kmharrington:
Using to AxisManagers to solve the preprocess selection issue. Should also solve #712  but not tested yet.

This PR adds another AxisManager inside of the preprocess Pipeline that is used to track the original size of the TOD AxisManager. At every step any new fields in `proc_aman` are added to the `full` AxisManager after they are expanded in size to match the original shape of the TOD AxisManager.  This expansion assumes that each top-level entry in `proc_aman` is an AxisManager (true for all existing processes, so now just required) and it was complicated enough to start test cases to make sure it works the way we'd like.

### DIFF
--- a/sotodlib/preprocess/core.py
+++ b/sotodlib/preprocess/core.py
@@ -218,6 +218,11 @@ class Pipeline(list):
         """
         if proc_aman is None:
             proc_aman = core.AxisManager( aman.dets, aman.samps)
+            fman = core.AxisManager(core.LabelAxis(name='fdets', vals=aman.dets.vals), 
+                                    core.OffsetAxis('fsamps', count=aman.samps.count,
+                                                    offset=aman.samps.offset, 
+                                                    origin_tag=aman.samps.origin_tag))
+            proc_aman.merge(fman)
             run_calc = True
         else:
             if aman.dets.count != proc_aman.dets.count or not np.all(aman.dets.vals == proc_aman.dets.vals):
--- a/sotodlib/preprocess/core.py
+++ b/sotodlib/preprocess/core.py
@@ -186,19 +186,22 @@ def expand_proc_aman(calc_aman, proc_aman, flag_fill_val=False,
         else:
             raise ValueError('Data type in axis manager not supported '+
                             'in expand_proc_aman')
-        
-        out_aman.wrap_new('f'+fld, faxes, cls=np.full, fill_value=fill_value)
-        out_aman.wrap(fld, calc_aman[fld], list(enumerate(axes)))
-        d = {'dets': fmdets, 'samps': fmsamps}
-        slices = [d.get(a, slice(None)) for a in axes]
-        if isinstance(calc_aman[fld], RangesMatrix):
-            out_aman['f'+fld][tuple(slices)]=calc_aman[fld].mask()
-            out_aman['f'+fld] = RangesMatrix.from_mask(out_aman['f'+fld])
+
+        if np.any(np.isin(axes, ['dets', 'samps'])):
+            out_aman.wrap_new('f'+fld, faxes, cls=np.full, fill_value=fill_value)
+            out_aman.wrap(fld, calc_aman[fld], list(enumerate(axes)))
+            d = {'dets': fmdets, 'samps': fmsamps}
+            slices = [d.get(a, slice(None)) for a in axes]
+            if isinstance(calc_aman[fld], RangesMatrix):
+                out_aman['f'+fld][tuple(slices)]=calc_aman[fld].mask()
+                out_aman['f'+fld] = RangesMatrix.from_mask(out_aman['f'+fld])
+            else:
+                out_aman['f'+fld][tuple(slices)]=calc_aman[fld]
         else:
-            out_aman['f'+fld][tuple(slices)]=calc_aman[fld]
+            out_aman.wrap(fld, calc_aman[fld], list(enumerate(axes)))
     return out_aman
 
-def collape_proc_aman(proc_aman):
+def collapse_proc_aman(proc_aman):
     """
     Replaces dets, samps axes + fields aligned with them with the data
     in fdets, fsamps axes. This is intended to go at the end of the
@@ -224,10 +227,19 @@ def collape_proc_aman(proc_aman):
                                   if not(x in ['fdets','fsamps','dets','samps'])])
     out_aman.merge(dets_samps)
     for fld, axes in zip(assn.keys(), assn.values()):
-        if np.any(np.isin(axes, ['dets', 'samps'])):
-            continue
-        naxes = [a.strip('f') if a in ['fdets', 'fsamps'] else a for a in axes]
-        out_aman.wrap(fld[1:], proc_aman[fld], list(enumerate(naxes)))
+        assn2 = proc_aman[fld]._assignments
+        fld_aman = core.AxisManager(*[proc_aman[x] for x in proc_aman[fld]._axes \
+                                if not(x in ['fdets','fsamps','dets','samps'])])
+        fld_aman.merge(dets_samps)
+        for fld2, axes2 in zip(assn2.keys(), assn2.values()):
+            if np.any(np.isin(axes2, ['dets', 'samps'])):
+                continue
+            if np.any(np.isin(axes2, ['fdets', 'fsamps'])):
+                naxes = [a.strip('f') if a in ['fdets', 'fsamps'] else a for a in axes2]
+                fld_aman.wrap(fld2[1:], proc_aman[fld][fld2], list(enumerate(naxes)))
+            else:
+                fld_aman.wrap(fld2, proc_aman[fld][fld2], list(enumerate(axes2)))
+        out_aman.wrap(fld, fld_aman)
     return out_aman
 
 class Pipeline(list):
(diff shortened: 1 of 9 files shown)


sotodlib/site_pipeline/preprocess_tod.py
```diff
@@ -118,7 +118,9 @@ def preprocess_tod(
         logger.info(f"Beginning run for {obs_id}:{group}")
 
         aman = context.get_obs(obs_id, dets={group_by:group})
-        proc_aman = pipe.run(aman)
+        proc_aman, success = pipe.run(aman)
+        if success != 'end':
```
msilvafe, line 122: @kmharrington I added this here because if it tried to save the `proc_aman` to h5 for a group that failed then the loading failed because some of the subfields under the preprocess axis manager were missing for that group (wafer slot). This instead just doesn't save that wafer to h5 at all then if you use `on_missing: trim` for the preprocess section in your context it loads fine just skipping that wafer in the load.

sotodlib/site_pipeline/preprocess_tod.py
```diff
@@ -152,12 +154,12 @@ def load_preprocess_det_select(obs_id, configs, context=None):
     """
     configs, context = _get_preprocess_context(configs, context)
     
-    pipe = Pipeline(configs["process_pipe"], logger=logger)
     meta = context.get_meta(obs_id)
-
-    for process in pipe:
-        logger.info(f"Selecting On {process.name}")
-        process.select(meta)
+    proc_aman = meta["preprocess"]
```
msilvafe, line 158: @kmharrington this is how I chose to select on the final process valid flags. I would prefer to get `assn` from `configs["process_pipe"]` however the name that is wrapped in is an optional argument in each process + not always defined consistently and using the process.name doesn't work because that's not unique (i.e. you can call the same process with the same name multiple times in the same pipe) so I used the order in aman.preprocess._assignments because the order is the order they're wrapped in the pipe.

sotodlib/site_pipeline/preprocess_tod.py
```diff
@@ -152,12 +154,12 @@ def load_preprocess_det_select(obs_id, configs, context=None):
...
+    assn = [a for a in proc_aman._assignments][-1]
+    logger.info(f"Cutting on the last process: {assn}")
+    keep = core.flagman.has_all_cut(proc_aman[assn].valid)
+    meta.restrict('dets', meta.dets.vals[keep])
     return meta
 
 def load_preprocess_tod(obs_id, configs="preprocess_configs.yaml", context=None ):
```
msilvafe, line 165: Currently this cannot load a single wafer, perhaps we want to update that but maybe not the purview of this PR.

sotodlib/preprocess/processes.py
```diff
@@ -518,12 +518,21 @@ class SubtractHWPSS(_Preprocess):
...
+        super().__init__(step_cfgs)
+
     def process(self, aman, proc_aman):
-        hwp.subtract_hwpss(
-            aman,
-            hwpss_template = aman[self.process_cfgs["hwpss_extract"]],
-            subtract_name = self.process_cfgs["subtract_name"]
-        )
+        if not(proc_aman[self.hwpss_stats] is None):
+            modes = [int(m[1:]) for m in proc_aman[self.hwpss_stats].modes.vals[::2]]
```
msilvafe, line 528: I had to add this because `hwpss_model` was used before which is wrapped into `aman` during the `calc` step of  `EstimateHWPSS` however when running `load_preprocess_tod` this fails because that relies only on what's in `proc_aman`/what's in the `process` step.

msilvafe (approved):
I put a few inline comments so you can be clear on why I added a few things during my testing after your final commit and decide if you want to discuss or change them before we merge.

The testing I did was all in notebook: `/so/home/msilvafe/analysis_notebooks/20240314_Test_Preprocess_Updates.ipynb` on site computing. I tested the following:

1. In jupyter I loaded an axis manager and ran the `pipe.run` function and inspected the output `proc_aman` and `success` flag on an obs that we know has no failures--our "nominal obs" 'obs_1704770693_satp1_1111111'
2. Test the same process but on an obs that we know fails: `'obs_1706657436_satp1_1111011:ws2'` and confirm that the `success` flag makes sense.
3. I test built a manifestdB using the `preprocess_tod` `site_pipeline` script that includes both of the two obs's
4. I checked that I could load both obs using `load_preprocess_tod` function and that the resulting processed axis manager and associated metadata made sense
5. I tested that I could load the metadata from the preprocess dB using `ctx.get_meta` and that it all made sense.

sotodlib/site_pipeline/preprocess_tod.py
kmharrington: Honestly this looks pretty clunky and possible prone to failure. Pipelines are lists, why not just run `pipe[-1].select`?

msilvafe: I'll try it.

sotodlib/preprocess/core.py
```diff
@@ -146,6 +148,122 @@ def register(process_class):
...
+        )
+    if 'samps' in new._axes:
+        _, fs_samps, ns_samps = full.samps.intersection(
+            new.samps, 
+            return_slices=True
+        )
+    else:
+        fs_samps = slice(None)
+
+    out = full.copy(axes_only=True)
```
kmharrington, line 205: were you going to make this a copy of only the axes that are in new? That should be pretty easy.

sotodlib/site_pipeline/preprocess_tod.py
kmharrington: agree, beyond the scope of this PR

sotodlib/site_pipeline/preprocess_tod.py
msilvafe: Ok @kmharrington I've fixed this in [c702f74](https://github.com/simonsobs/sotodlib/pull/751/commits/c702f744b852b63843840255c6c6baca94e5c401)

sotodlib/preprocess/core.py
msilvafe: Fixed this in [6fa2bd8](https://github.com/simonsobs/sotodlib/pull/751/commits/6fa2bd8da9bbfb04253883e03f4ae07884e501bf) and tested with updates. I think g2g now.

sotodlib/preprocess/core.py
```diff
@@ -146,6 +148,126 @@ def register(process_class):
...
+        fs_samps = slice(None)
+
+    out = core.AxisManager()
+    for k, v in full._axes.items():
+        if k in list(new._axes.keys())+['dets','samps']:
+            out._axes[k] = v 
+
+    for a in new._axes:
+        if a not in out:
+            out.add_axis( new[a] )
```
kmharrington, line 212: wouldn't this all just be this?:
````
for a in new._axes:
    if a in full._axes:
        out.add_axis(full[a])
    else:
        out.add_axis(new[a])
````

File: sotodlib/preprocess/core.py (hunk 5 above)
Comment (kmharrington) [reply sotodlib/preprocess/core.py]: added a suggestion b/c I don't think we need to explicitly pull out dets/samps?

File: sotodlib/site_pipeline/preprocess_tod.py (hunk 2 above)
Comment (kmharrington) [reply sotodlib/site_pipeline/preprocess_tod.py]: ya looks good!

File: sotodlib/preprocess/core.py (hunk 6 above)
Comment (msilvafe) [reply sotodlib/preprocess/core.py:212]: Hey @kmharrington, yes I tried this at first then ran into an error because the `valid` flag needs to have `dets` and `samps` but some of the processes don't have both of those axes so I thought it was easier to just force `out` to always have `dets` and `samps`. Below is the traceback when I run with your suggestion for reference: 

```
ValueError                                Traceback (most recent call last)
File <timed exec>:1

File ~/.local/lib/python3.10/site-packages/sotodlib/preprocess/core.py:379, in Pipeline.run(self, aman, proc_aman, select)
    377 if run_calc:
    378     process.calc_and_save(aman, proc_aman)
--> 379     update_full_aman( proc_aman, full, self.wrap_valid)
    381 if select:
    382     process.select(aman, proc_aman)

File ~/.local/lib/python3.10/site-packages/sotodlib/preprocess/core.py:268, in update_full_aman(proc_aman, full, wrap_valid)
    264 if fld not in full._fields:
    265     assert isinstance(proc_aman[fld], core.AxisManager)
    266     full.wrap( 
    267         fld,
--> 268         _expand( proc_aman[fld], full, wrap_valid=wrap_valid)
    269     )

File ~/.local/lib/python3.10/site-packages/sotodlib/preprocess/core.py:247, in _expand(new, full, wrap_valid)
    242     v = Ranges.from_mask(m)
    244     valid = RangesMatrix( 
    245         [v if i in fs_dets else x for i in range(full.dets.count)]
    246     )
--> 247     out.wrap('valid',valid,[(0,'dets'),(1,'samps')])
    248 return out

File ~/.local/lib/python3.10/site-packages/sotodlib/core/axisman.py:605, in AxisManager.wrap(self, name, data, axis_map, accept_new, accept_merge)
    602 if not isinstance(axis, AxisInterface):
    603     # So it better be a string label... that we've heard of.
    604     if axis not in self._axes:
--> 605         raise ValueError("Axis assignment refers to unknown "
    606                          "axis '%s'." % axis)
    607     axis = self._axes[axis]
    608 axis = axis.resolve(data, index)

ValueError: Axis assignment refers to unknown axis 'samps'.
```

kmharrington: ahh gotcha. makes sense!

PR #752, opened 2024-03-07
Implement transfer function workflow with splits

tskisner:
This is currently a work in progress.  Immediate focus is the `toast_so_sat` workflow for producing maps of splits.


keskitalo (approved):
Solid work! It will probably require tweaking once the details of ISO data reduction are established but this is everything we need right now.

PR #753, opened 2024-03-08
Add main function to update_det_match

jlashner:
Adds main function to `update_det_match.py` so it can be run through prefect.


guanyilun (approved):
looking good to me!

PR #754, opened 2024-03-08
Fix search on readthedocs website

BrianJKoopman:
This removes `html_theme_path` in order to fix search.

Search works locally, but not on the readthedocs website. This was suggested as a possible fix in [1].

Will test in the docs build on this PR.

[1] - https://stackoverflow.com/a/77358365


BrianJKoopman:
This indeed did the trick! Search is working on the [build for this PR](https://sotodlib--754.org.readthedocs.build/en/754/).

mhasself (approved):
Thanks!

PR #758, opened 2024-03-11
Quick map-to-tod function for SAT (for the purpose of transfer function)

tterasaki:
We need a function to generated TOD from signal only map to validate our preprocess functions and calculate transfer function.

Basic code already exists in `coords.P.from_map`.
But we should again consider the coordinates for w/ HWP configuration as [Carlos and Matthew's PR](https://github.com/simonsobs/sotodlib/pull/720).
(BTW, thank you so much Carlos and Matthew for fixing the bug!)

The new function is shown below.
It receives an axismanager with `timestamps`, `focal_plane`, and `boresight` and input T,Q,U maps. And it returns timestreams.
https://github.com/simonsobs/sotodlib/blob/be2ee3e4ba775101be40874de5f310939f5e2c81/sotodlib/coords/demod.py#L145-L182

Please see the log book
https://simonsobs.atlassian.net/wiki/spaces/PRO/pages/337772548/Transfer+Function
or
`computeXX: /so/home/tterasaki/code_dev/transfer_func/demo_transfer_func.ipynb`

### DIFF
--- a/docs/coords.rst
+++ b/docs/coords.rst
@@ -274,6 +274,12 @@ Planet Mapmaking support
 .. automodule:: sotodlib.coords.planets
    :members:
 
+Mapmaking for HWP data
+----------------------
+
+.. automodule:: sotodlib.coords.demod
+   :members:
+
 Local coordinate systems
 ------------------------
 
--- a/sotodlib/coords/demod.py
+++ b/sotodlib/coords/demod.py
@@ -142,15 +142,11 @@ def make_map(tod,
              'weight': wTQU}
     return output
 
-def from_map(tod, signal_map, P=None, wcs_kernel=None,
-             res=0.1 * coords.DEG, cuts=None, flip_gamma=True, wrap=False, pre_demod=False):
-    if P is None:
-        if wcs_kernel is None:
-            wcs_kernel = coords.get_wcs_kernel('car', 0, 0, res)
-        P = coords.P.for_tod(
-            tod=tod, wcs_kernel=wcs_kernel, cuts=cuts, comps='QU', hwp=flip_gamma)
-    
+def from_map(tod, signal_map, cuts=None, flip_gamma=True, wrap=False, pre_demod=False):
     Tmap, Qmap, Umap = signal_map
+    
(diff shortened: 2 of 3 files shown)


msilvafe:
I'm not sure what this is addressing @tterasaki. Between what @chervias has been doing and our fn's in coords don't we already have this?

tterasaki:
Oh, if it is True, it is unfortunate. (But I believe not totally waste as I learned a lot ) Can you specify where it is?

chervias:
Hi @tterasaki, the purpose of this PR is that you can simulate some TQU map (so you know exactly what you put in), transform that into a modulated timestream with the HWP angles, and then use that timestream to map it through the SAT mapmaker. That way you will have a filtered map for TF estimation. Is that correct?
The way that I thought we would do this is by using TOAST. We would load a real observation, empty the signal and fill the signal with a simulated map scanned in TOAST (using the same metadata, same HWP, same pointing, same everything). Then we run that through an SAT mapmaker to obtain a filtered map. Could you please show your approach in the FLP call? Probably we can use both approaches to cross-check our results.

mhasself:
It's very useful to have this function, thanks.

Can you make a unit test?  Add to tests/test_demod_map.py.  I think we just want to see a round-trip (map -> TOD -> map).  Doesn't have to be exact; just verify sign conventions and so on.

msilvafe:
> Oh, if it is True, it is unfortunate. (But I believe not totally waste as I learned a lot ) Can you specify where it is?

I don't think it's wasted time. I also didn't actually process the benefit of this at first glance. I think I basically understand now that this is a super stripped down toolset that allows us to estimate a transfer function without the overhead of toast to do map->tod or the BB-tools for PS-estimation (SOOPERCOOL/BB-Master) just sotodlib, pspy, and camb. That does seem really valuable for quick turn checks. This is awesome, thx (and sorry for the noise in my earlier question)!

tterasaki:
I think it is ready to go.

mhasself (approved):
Thanks!

I added this submodule to the docs rendering ... will need a bit of formatting (but so does everything else in coords, so let's leave it for another day).

PR #759, opened 2024-03-11
depth-1 mapmaking script, part of so_site_pipeline

chervias:
Depth-1 mapmaker script that was in pwg_scripts. Part of so_site_pipeline. Uses the same format as make_ml_map. Tested on the deep56 toast sims that Tanay created in simons1.


chervias:
Hi, it seems we need this now with the incoming LAT data.

iparask:
@chervias, @mhasself can we get this PR over the finish line? It would be very useful to have this released since it is already used to create depth-1 maps.

sotodlib/coords/helpers.py
```diff
@@ -796,3 +792,17 @@ def get_deflected_sightline(aman, wobble_meta, site='so', weather='typical'):
     sight.Q = sight.Q * ~deflq
     return sight
 
+def normalize_geometry(shape, wcs):
```
mhasself:
```suggestion
def normalize_geometry(shape, wcs):
    """Analyze (shape, wcs) and return a pixel-compatible (shape, wcs)
    that puts CRVAL as near as possible to the center of the map. This is
    needed for so3g routines in cylindrical projections, or the map may be
    clipped.
    
    """     
```

sotodlib/site_pipeline/make_depth1_map.py
```diff
@@ -0,0 +1,573 @@
...
+        map_tods = existing_map.tods if existing_map else []
+
+        total_tods = np.sum([map_tod.wafer_count for map_tod in map_tods])
+
+        if total_tods < len(inds_to_use):
+            return True
+    return False
+
+def commit_depth1_tods(map_name:str, obslist: Dict[Tuple[int, str, str], List[Tuple[str, str, str, int]]], obs_infos: List[Tuple[str, float, float, float, int, str, str, str, str, str, str, int, str, float, float, float, float, float,  float, float, str, str]],
+                       band: str, inds: List[int], mapcat_settings: Dict[str, str]) -> List[TODDepthOneTable]:
```
mhasself: Fix or remove the obs_infos type annotation.

sotodlib/site_pipeline/utils/constants.py
```diff
@@ -0,0 +1,41 @@
+DEPTH1MAPMAKER_DEFAULTS = {
```
mhasself: Move both these dicts to depth1_utils.

sotodlib/site_pipeline/utils/exceptions.py
```diff
@@ -0,0 +1,6 @@
+from sotodlib.core import metadata
```
mhasself: We should probably just put these in depth1_utils, too.

mhasself (changes requested):
I have not reviewed the detailed function here, but found one probable typo (dtype_tod).

I request that the reorganization be limited a bit -- I don't think the move to lots-of-little-files is warranted here.  `depth1_utils` is fine, but the other new utils files seem to also basically be places to put single-purpose utils for depth1 maps... i.e. depth1_utils.

Going forward (i.e. I will not hold this PR up for this) -- I think we need better documentation on the support functions, e.g. the footprint stuff and tricky coordinate handling.  The point of support functions is that they do some specific job well, but if it's not clear what the definition of that job is, nor what the idea of well should be, then they become weird black boxes.  Many of the support functions are amenable to unit testing, and I'd like to see such tests.

sotodlib/site_pipeline/make_depth1_map.py
```diff
@@ -0,0 +1,438 @@
...
+from sotodlib.preprocess import preprocess_util as pp_util
+
+from sotodlib.site_pipeline.utils.config import _get_config
+from sotodlib.site_pipeline.utils.constants import DEPTH1MAPMAKER_DEFAULTS
+from sotodlib.site_pipeline.utils.exceptions import DataMissing, LoaderError
+from sotodlib.site_pipeline.utils.depth1_utils import (
+    calibrate_obs, create_mapmaker_config, find_footprint, read_tods)
+from sotodlib.site_pipeline.utils.io import write_depth1_map
+from sotodlib.site_pipeline.utils.mapcat import (
+    commit_depth1_map, commit_depth1_tods, map_to_calculate)
```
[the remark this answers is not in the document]
iparask: I prefer explicit imports because I know what is imported immediately especially when a few functions from a submodule are used. It also does not overload python's namespace. I can change it if you prefer.

iparask:
> I request that the reorganization be limited a bit -- I don't think the move to lots-of-little-files is warranted here. `depth1_utils` is fine, but the other new utils files seem to also basically be places to put single-purpose utils for depth1 maps... i.e. depth1_utils.

I will consolidate them, but I would like to leave the mapcat integration in a different file. The reason is the atomic maps mapmaker is integrating with mapcat and it would be best to have all the interactions together in case something changes there.

Small files are my preference because it makes it easier to manage in terms of development, but I can consolidate them if you prefer.

chervias:
> > I find especially that it's important to explain how such functions handle edge cases. For example, if the input TODs end up exceeding a full sky's worth of RA bounds, what happens? This can be done later, but we all know it's better to do it now.
> 
> @chervias if you can reply on this?
> 
> > Please add make_depth1_map to site_pipeline.rst (with an argparse block), and check the docs render sensibly.
> 
> Sounds good, will do

ok @mhasself I added extensive documentation on edge cases in [10ec8e2](https://github.com/simonsobs/sotodlib/pull/759/commits/10ec8e2007c7820abd2ac9639e8dd44fc3e22013).

PR #766, opened 2024-03-13
Use correct default name

keskitalo:


tskisner (approved):
Thanks for catching that!

Issue #763, opened 2024-03-13
Raise error in making projection object when NaN is in tod.focal_plane


tterasaki:
I (we) have wasted a lot in debugging when forgetting restricting detectors with non-NaN pointing result.

skhrg:
Spent a hot second debugging the same thing with @tdalford  the other day.

Worth noting that `aman.det_info.wafer.type == "OPTC"` should get you a mask that removes all the nan pointing (since only non optical dets should have them if you are using the full focal plane). So maybe something that tries to apply that by default would be useful?

msilvafe:
We had discussed having the focal plane manifestdB load in the flags you want to use to restrict to just the optical detectors in addition to the focalplane axis manager as is can we do that and then default to having the mapping functions use that set of flags?

skhrg:
Yeah I think that makes sense. To be clear though do we want this flag to make its way into `aman.flags` or just live in `aman.focal_plane`?

msilvafe:
I think `aman.focal_plane` is fine for now. Eventually it would be good to collect all flags into `aman.flags` (there are many in `aman.preprocess`) but I feel like maybe we want that to be a modification of the loader that checks for flags in the metadata and loads them into flags.

msilvafe:
I also like that solution. Maybe more.

mhasself:
That solution is a great mitigation for now (maybe test gamma instead, though ... sometimes gamma is the only nan, but that's enough to segfault).

I could imagine it being convenient to leave fp.* as is, and suppress unpointed dets using det_weights.  But that is a bit more work to add in.

davidvng:
Hi @mhasself , I'm interested in your suggestion of suppressing unpointed dets using det_weights. Could you give more details on what you're thinking? I've ran into this issue when trying to run https://github.com/simonsobs/sotodlib/blob/cb49a57d587f9bfed19a13026cd73df816f3eae5/sotodlib/coords/planets.py#L425 which eventually calls https://github.com/simonsobs/sotodlib/blob/cb49a57d587f9bfed19a13026cd73df816f3eae5/sotodlib/coords/helpers.py#L283 which fails to converge due to the NaNs

mhasself:
> which fails to converge due to the NaNs

Because there can be NaNs in focal plane, any method that "does stuff" with focal_plane.{xi,eta,gamma} has to be robust against that.  Either by raising exception, or just omitting those dets when appropriate.

[...]

PR #775, opened 2024-03-21
added focal_plane nan checker in coords.pmat.for_tod

tterasaki:
https://github.com/simonsobs/sotodlib/issues/763

### DIFF
--- a/sotodlib/coords/pmat.py
+++ b/sotodlib/coords/pmat.py
@@ -189,6 +189,9 @@ class P:
         # Set up the detectors in the focalplane
         fp = _valid_arg(focal_plane, 'focal_plane', src=tod)
         xi, eta, gamma  = fp.xi, fp.eta, fp.get('gamma')
+        if not np.all(~np.isnan([xi, eta, gamma])):
+            raise ValueError('np.nan is included in focal_plane')
+        
         assert (gamma is not None or not hwp)
         if hwp:
             gamma = -gamma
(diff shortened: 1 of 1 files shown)


sotodlib/coords/pmat.py
```diff
@@ -189,6 +189,9 @@ def for_tod(cls, tod, sight=None, geom=None, comps='T',
         # Set up the detectors in the focalplane
         fp = _valid_arg(focal_plane, 'focal_plane', src=tod)
         xi, eta, gamma  = fp.xi, fp.eta, fp.get('gamma')
+        if not np.all(~np.isnan([xi, eta, gamma])):
+            raise ValueError('np.nan is included in focal_plane')
+
```
mhasself, line 194: gamma is allowed to be None here, which I think will cause the array promotion of [xi, eta, gamma] to fail.

Simpler logic uses `if np.any(np.isnan(...))`.

mhasself, line 194: gamma is allowed to be None here, which I think will cause the array promotion of [xi, eta, gamma] to fail.

Simpler logic uses `if np.any(np.isnan(...))`.

mhasself (changes requested):

tterasaki:
Thanks Matthew and sorry for the late reply.
I think it is ready to merge.

mhasself:
> Thanks Matthew and sorry for the late reply. I think it is ready to merge.

No prob -- but see my comment about "gamma".  It can be None.  This code will choke on that.

kmharrington:
When will this converge for merging? Enough other people are hitting this issue and solving it (ex: https://github.com/simonsobs/sotodlib/pull/816 ) that we need to have the nice version in master

mhasself:
I am sorry this is so much fun, but I think you need to also add this:
```
        if gamma is not None and np.any(np.isnan(gamma)):
            raise ValueError('np.nan is included in gamma')
```

mhasself (approved):
To infinity and beyond!

PR #781, opened 2024-03-27
Jump speedups

skhrg:
Slowest routine is now ~7x faster.

~~Needs to have updated docs (including a note on getting better performance) and some correctness checks done before merging.~~

Some profiles here for those interested: `/so/home/saianeesh/data/profiles`

I also have some ideas that should increase accuracy but lets get these speed improvements in the wild first.


kmharrington (approved):
jumps go vroooom

PR #783, opened 2024-03-28
Add pointing_model script to sotodlib.coords

elle-shaw:
This includes the naive version of apply_pointing_model that will convert tod.ancil encoder data to tod.boresight pointing. This includes the proper definitions of roll for LAT and SAT.

### DIFF
--- a/sotodlib/coords/pointing_model.py
+++ b/sotodlib/coords/pointing_model.py
@@ -1,0 +1,35 @@
+import numpy as np
+from .. import core
+from .helpers import _valid_arg
+
+DEG = np.pi / 180
+
+
+def apply_pointing_model(tod):
+    """Redefines tod.boresight using pointing model parameters.
+
+    Args:
+        tod: AxisManager
+
+    Returns:
+        tod.boresight will be updated.
+    """
+
+    _a = _valid_arg("ancil", src=tod)
+    telescope = _valid_arg("obs_info", src=tod)["telescope"]
+    # Start with a fresh axismanager
+    _b = core.AxisManager(core.OffsetAxis("samps", len(tod.timestamps)))
+
+    if "lat" in telescope:
+        # Returns in Radians
+        _b.wrap("az", _a.az_enc * DEG, [(0, "samps")])
+        _b.wrap("el", _a.el_enc * DEG, [(0, "samps")])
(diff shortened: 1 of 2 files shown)


sotodlib/coords/pointing_model.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def apply_pointing_model(tod):
+    """Redefines tod.boresight using pointing model parameters.
+
+    Args:
+        tod: AxisManager
+
+    Returns:
+        tod.boresight will be updated.
```
mhasself, line 15: This doesn't return anything, currently.  Just None.  Mention the effect of the function (to modify tod.boresight) in the main description, not as a return parameter.

Or better, you could make it return the _b AxisMan that you create.  That would be useful, because we could imagine this (in the future) being something that computes the new boresight but does not _necessarily_ load it into tod yet.

mhasself (changes requested):
Thanks for the good start on this.  One comment about the docstring / return val.

Can you fix the angle convention also in load_book?  I think that's just a bug; let's take care of it.

sotodlib/coords/pointing_model.py
```diff
@@ -6,30 +6,31 @@
...
         # Returns in radians
-        _b.wrap("az", _a.az_enc * DEG, [(0, "samps")])
-        _b.wrap("el", _a.el_enc * DEG, [(0, "samps")])
-        _b.wrap("roll", -1 * _a.boresight_enc * DEG, [(0, "samps")])
+        _boresight.wrap("az", _ancil.az_enc * DEG, [(0, "samps")])
+        _boresight.wrap("el", _ancil.el_enc * DEG, [(0, "samps")])
+        _boresight.wrap("roll", -1 * _ancil.boresight_enc * DEG, [(0, "samps")])
 
-    tod["boresight"] = _b
+    return _boresight
```
mhasself, line 36: How about you do both of these things?  Save it, and return it. (We might make saving optional later, but it should be the default behavior, for now.)

mhasself (changes requested):
Thanks -- almost there :)

sotodlib/coords/pointing_model.py
elle-shaw: ok, I'll do that.

mhasself (approved):
Thanks!

PR #784, opened 2024-03-28
Allow labelaxis to be restricted by a mask

skhrg:
I keep getting annoyed that I can't do this and don't think I am alone.
In and ideal world maybe we want all the np fancy indexing but I think this is by far the most common use case and will be nice to have.


kmharrington (approved):
yes please

PR #785, opened 2024-03-28
Block Reduce flag

skhrg:
Been finding a version of this useful for lat stuff.
Honestly have not fully tested this more formal version but would like to get some comments on the structure before I get too far.


kmharrington:
seems pretty useful to me.

PR #787, opened 2024-03-29
Level-3 HK database and loading framework

jlashner:
This is a draft PR with an initial level-3 housekeeping indexing and loading framework.

Docs will come shortly, but this can currently be used and tested on satp1 data, since I have a satp1 index that should be up to date as of 03/28.

First create a hk.yaml file to use as configuration, such as:
```
hk_root: /so/data/satp1/hk
hk_db: /so/home/jlashner/work/daily/2024-03-28/l3-hk/satp1_hk.db
aliases:
    fp_temp: cryo-ls372-lsa21yc.temperatures.Channel_02_T
```
Then in python, you can load data the last week's worth of focal-plane temps as follows
```
t1 = time.time()
t0 = t1 - 7*24*3600
cfg = hkdb.HkConfig.from_yaml('hk.yaml')
ls = hkdb.LoadSpec(
    cfg=cfg,
    fields=['cryo-ls372-lsa21yc.*.*'],
    start=t0, end=t1,
)
res = hkdb.load_hk(ls, show_pb=True)
plt.plot(*res.fp_temp)

# The above is the same as below, attributes in the result are set based on aliases.
# plt.plot(*res.data['cryo-ls372-lsa21yc.temperatures.Channel_02_T'])
```

### DIFF
--- a/docs/hkdb.rst
+++ b/docs/hkdb.rst
@@ -1,0 +1,102 @@
+
+==============================
+Loading L3 Houskeeping Data
+==============================
+
+The ``sotodlib.io.hkdb`` module provides methods for indexing, and quickly
+loading housekeeping data from an archive.
+The databases implemented store the following information:
+
+- List of all housekeeping files, with start and stop times
+- List of every frame in the file, with the start and stop times of each frame,
+  agent instance-id, and the byte-offset of the frame in the file.
+
+To index and load housekeeping information, you will need to create a
+``HkConfig`` object, documented in the API section below. For example, the following yaml file can
+be loaded in directly ``HkConfig.from_yaml``:
+
+.. code-block:: yaml
+
+    hk_root: /so/data/satp1/hk
+    hk_db: satp1_hk.db
+    aliases:
+        fp_temp: cryo-ls372-lsa21yc.temperatures.Channel_02_T
+
+In addition to the hk-root and database paths, an "alias" dictionary can
+be defined here, to provide a mapping from a human-readable name to the
+field-id of the housekeeping data, where field-ids are always
+of the form ``<agent-instance>.<feed>.<field>``.
+Aliases defined here can be used to more easily load and access these fields.
+
+Loading Data
+---------------
+To load data from the hk archive, you can pass a ``LoadSpec`` object to the
+``load_hk`` function, which defines the time range and fields to load.
+
+For example, the code below will load the focal-plane temp for the last week:
+
+.. code-block:: python
+
+    from sotodlib.io import hkdb
+    cfg = hkdb.HkConfig.from_yaml('satp1_hk.yaml')
+    t1 = time.time()
+    t0 = t1 - 24 * 3600 * 7
+    lspec = hkdb.LoadSpec(
+        cfg=cfg, start=t0, end=t1,
+        fields=['fp_temp'],
+    )   
+    result = hkdb.load_hk(lspec, show_pb=True)
+
+The result object will be an HkResult object, where ``result.data`` contains
+the mapping from field-id to tuple where the first element is t, 
+where the first entry are the timestamps, and the second entry are the data
+values.  If any of the requested fields have an alias defined in the ``cfg``
+object, that alias will be set as an attribute in the HkResult object:
+
+.. code-block:: python
+
+    >> print(result.data['cryo-ls372-lsa21yc.temperatures.Channel_02_T'].shape)
+    (2, 4160195)
+    >> print(result.fp_temp)  # same exact thing as above
+    (2, 4160195)
+    >> plt.plot(*result.fp_temp)  # plots focal-plane temp
+
+The ``fields`` argument to ``LoadSpec`` is a list of field-specifications, which
+can either be:
+
+- An alias in the HkConfig
+- A field-id of the form ``<agent-instance>.<feed>.<field>``
+
+Field-ids specifications will also allow wildcards as the feed or field, to
+allow one to easily load all the fields of an agent or a feed.
+For example, you can run:
+
+.. code-block:: python
+
+    from sotodlib.io import hkdb
+    cfg = hkdb.HkConfig.from_yaml('satp1_hk.yaml')
+    t1 = time.time()
+    t0 = t1 - 24 * 3600 * 7
+    lspec = hkdb.LoadSpec(
+        cfg=cfg, start=t0, end=t1,
+        fields=['cryo-ls372-lsa21yc.*.*'],
+    )   
+    result = hkdb.load_hk(lspec, show_pb=True)
+    print(result.data.keys())
+
+    >> dict_keys(['cryo-ls372-lsa21yc.temperatures.Channel_02_R', 
+                  'cryo-ls372-lsa21yc.temperatures.Channel_02_T', 
+                  'cryo-ls372-lsa21yc.temperatures.sample_heater_out'])
+
+    # Note that the fp_temp alias will still be found, and assigned to the
+    # correct field.
+    print(result.fp_temp.shape)
+    >> (2, 4157762)
+
+
+API
+-------
+.. automodule:: sotodlib.io.hkdb
+   :members: HkConfig, HkDb, LoadSpec, load_hk, HkResult
+   :undoc-members:
+   :show-inheritance:
--- a/docs/hkdb.rst
+++ b/docs/hkdb.rst
@@ -12,8 +12,8 @@ The databases implemented store the following information:
   agent instance-id, and the byte-offset of the frame in the file.
 
 To index and load housekeeping information, you will need to create a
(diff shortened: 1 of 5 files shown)


sotodlib/io/hkdb.py
```diff
@@ -0,0 +1,366 @@
...
+        fs = []
+        for f in self.fields:
+            if f in self.cfg.aliases:
+                fs.append(Field.from_str(self.cfg.aliases[f]))
+            else:
+                fs.append(Field.from_str(f))
+        self.fields = fs
+
+
+class HkResult:
```
kmharrington, line 285: can this just be an axis manager?

kmharrington:
SQLAlchemy has a lot written up on how to use sessions. The starting them inside of functions like what's done here isn't recommended. I'm having multiple issues with that setup inside of the rest of data-packaging.  [See the thingOne, thingTwo "don't do this" section.]( https://docs.sqlalchemy.org/en/20/orm/session_basics.html#when-do-i-construct-a-session-when-do-i-commit-it-and-when-do-i-close-it ) The imprinter/other HKdb setup is closer to what they recommend (although not completely there). At least some of the g3tsmurf issues comes from the rollbacks instead of throwing errors and some of the readout_id querying generating their own sessions.
 
For that reason, I'm really not of fan of making these "functions passing around a database that sometimes has a session" instead of a "class with a stable session." But even if it stays functions instead of classes:  we don't want that setup in `index_all` where that function begins a session and then calls a function that begins a session. At an absolute minimum, nesting functions that make sessions is going to be quite confusing. Also, [I'm pretty sure that a transaction is issued every time we start a session](https://docs.sqlalchemy.org/en/20/orm/session_transaction.html) so that's going to hit the NFS mount way more than alternative setups.

sotodlib/io/hkdb.py
jlashner: I don't think this quite fits into an AxisManager so easily... because hk fields aren't cosampled, most of the benefits of AxisManager aren't really applicable, and I'm not sure how axis-managers handle long field names that aren't proper python identifiers, such as `cryo-ls372-lsa21yc.temperatures.Channel_02_T`.

I think it would be better to leave this as is, and have a separate function that transforms this into an AxisManager when re-sampling hk to be aligned with the 'samps' axis.

kmharrington: Sanah did the setup for getting axismanagers where the fields are co-sampled and then co-sampling. I don't see a reason not to use it?

jlashner: I really dislike the idea of making this load fn dependent on all of the hk -> axis-manager conversion stuff that is in hk_utils...

I think there is enough subtlety in how fields are grouped and how the cosampling is done, that it should happen separately and not as a part of the loading function.

jlashner:
Thanks katie for pointing me to those sqlalchemy pages, I think I addressed your concern by allowing for passing in of a session from the outer function, to avoid nesting.

I'm going to open this up as being "ready for review" since I've added docs, and I have satp1 and satp3 dbs accessible here:
```
- /so/home/jlashner/work/daily/2024-03-28/l3-hk/satp1_hk.db
- /so/home/jlashner/work/daily/2024-03-28/l3-hk/satp3_hk.db
```

sanahabhimani:
Testing this in prep to integrate into `hk_utils`--- I was playing around with `LoadSpec`. My config file has 2 fields/aliases I input. Currently, if I want to load data only from the fields in my config file, I set `fields = []`. Is it worth having `fields=None` as a default argument for  a case like this?

jlashner:
@kmharrington and @mhasself I think I addressed all of your comments from the data-packaging call last week, by:

- adding size, mod_time, index_status to HkFiles table
- adding lookback time option to the cfg object to limit scanning entire hk archive
- adding caching of failed files (through the index_status)

I also added saving / loading of the HkResults object because I find that really useful.

I don't think this needs a separate script for updating, since I believe prefect can just use the `update_index_all` as an entrypoint, but I'll check with Yulin.

@sanahabhimani I do get where you're coming from, but when building and testing I was sort of envisioning the hkcfg files having many aliases available, and for it being kind of rare to need to load all of them at once. So I'm not too keen on making this the default behavior, if that makes sense. Thanks for the suggestion though!

sanahabhimani:
I can push updates to hk_utils with this new framework to this branch if that's the cleanest way to go.

sotodlib/io/hkdb.py
```diff
@@ -0,0 +1,425 @@
...
+    ------
+    cfg : Union[HKConfig, str]
+        Configuration object or path to configuration file
+    """
+    def __init__(self, cfg: Union[HkConfig, str]):
+        if isinstance(cfg, str):
+            cfg = HkConfig.from_yaml(cfg)
+        self.cfg = cfg
+
+        self.engine = db.create_engine(f"sqlite:///{cfg.hk_db}", echo=cfg.echo_db)
```
mhasself, line 131: You could generalize this, for non-sqlite options, by accepting engine as a kwarg.

That would facilitate experimentation with alternative databases.

sotodlib/io/hkdb.py
```diff
@@ -0,0 +1,425 @@
...
+            os.path.join(sdir, f) 
+            for f in os.listdir(sdir)
+            if f.endswith('.g3')
+        ])
+
+    existing_files = [path for path, in session.query(HkFile.path).all()]
+    new_files = sorted(list(set(all_files) - set(existing_files)))
+
+    files = []
+    print(f"Adding {len(new_files)} new files to index...")
```
mhasself, line 163: Use logger instead of print for library-level code.

sotodlib/io/hkdb.py
```diff
@@ -0,0 +1,425 @@
...
+    "If true, shows progress bar when indexing"
+
+    aliases: Dict[str, str] = field(default_factory=dict)
+    """
+    Aliases for hk fields. In this dict, the key is the alias name, and the
+    value is the field descriptor, in the format of ``agent.feed.field``.
+    For example::
+
+        {
+            'fp_temp': 'cryo-ls372-lsa21yc.temperatures.Channel_02_T',
```
mhasself, line 48: Perhaps warn the user that aliases should be valid python attribute names, since you setattr with these, later.  (I.e. they should not use 'fp-temp'.)

sotodlib/io/hkdb.py
```diff
@@ -0,0 +1,425 @@
...
+
+    @classmethod
+    def load(cls, path):
+        d = np.load(path, allow_pickle=True)
+        aliases = d['aliases'].item()
+        data = {k: d[k] for k in d.files if k != 'aliases'}
+        return cls(data, aliases=aliases)
+
+
+def load_hk(load_spec: LoadSpec, show_pb=False):
```
mhasself, line 362: Since this is a main entry point, it might be helpful to also accept "load_spec" as a dict that gets auto-converted to LoadSpec.  Also consider engine=... arg to pass through to HkDb.

docs/hkdb.rst
```diff
@@ -0,0 +1,102 @@
...
+following yaml file can be loaded in directly with ``HkConfig.from_yaml``:
+
+.. code-block:: yaml
+
+    hk_root: /so/data/satp1/hk
+    hk_db: satp1_hk.db
+    aliases:
+        fp_temp: cryo-ls372-lsa21yc.temperatures.Channel_02_T
+
+In addition to the hk-root and database paths, an "alias" dictionary can
```
mhasself, line 25: Mention that aliases are pertinent to _load_ only, and don't affect how info is stored in the database.

Might be helpful to have an example where a common database configuration is loaded from file, but user modifies (in script) the aliases block to suit their current purpose.  (I.e. emphasize that aliases are a dynamic element to tune for each use case, whereas a database instance is a general storage tool useful for many purposes.)

mhasself (approved):
Minor suggestions, but I won't insist.

I get that this is sort of a "minimum working implementation", but there are lots of obvious extensions / conveniences one might pursue (e.g. more command line args for update_hkdb.py).

Also -- site_pipeline.rst should an entry for update-hkdb.

sotodlib/io/hkdb.py
jlashner: This sounds good to me, but instead of passing through an engine I'd rather have the `db_url` settable via the cfg, either as a string or as a [sqlalchemy.url](https://docs.sqlalchemy.org/en/20/core/engines.html#creating-urls-programmatically) object, so it is more straightforward to set via the cfg file, and we don't have to worry about passthrough args. Are there any reasons you think its better for the engine object to be passed in over the URL?

mhasself: Oh, ya, that sounds fine.

jlashner:
Made a few changes addressing Matthew's suggestions. In particular, expanded HkConfig so that it supports generic database URLs, so it can support more than sqlite databases.

This is currently tested and running from a postgres db on storage1.

PR #790, opened 2024-04-02
More preprocess metrics

tristpinsm:
I've added some more metrics for the current `preprocesstod` output.

I was also thinking a bit about the volume of data this will be feeding to InfluxDB, and whether it will remain a sustainable rate. A very rough estimate could be that an Influx line is up to 1 kB (1000 char, no compression), and here I am saving one line per detector per obs_id. If there are order tens of thousands of detectors for the full observatory, that would come to tens of MB per obs_id. Is that sensible?

### DIFF
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -11,7 +11,7 @@ from sotodlib.core.flagman import (has_any_cuts, has_all_cut,
                                    count_cuts,
                                     sparse_to_ranges_matrix)
 
-from .core import _Preprocess
+from .core import _Preprocess, _FracFlaggedMixIn
 from .. import flag_utils
 
 
@@ -41,7 +41,7 @@ class Detrend(_Preprocess):
         tod_ops.detrend_tod(aman, signal_name=self.signal,
                             **self.process_cfgs)
 
-class DetBiasFlags(_Preprocess):
+class DetBiasFlags(_FracFlaggedMixIn, _Preprocess):
     """
     Derive poorly biased detectors from IV and Bias Step data. Save results
     in proc_aman under the "det_bias_cuts" field. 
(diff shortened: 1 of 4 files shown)


msilvafe:
> I've added some more metrics for the current `preprocesstod` output.
> 
> I was also thinking a bit about the volume of data this will be feeding to InfluxDB, and whether it will remain a sustainable rate. A very rough estimate could be that an Influx line is up to 1 kB (1000 char, no compression), and here I am saving one line per detector per obs_id. If there are order tens of thousands of detectors for the full observatory, that would come to tens of MB per obs_id. Is that sensible?

I don't know that we need per-detector-per-obs values published here. I was assuming these metrics would be boiled down to per-obs numbers. If you think that's infeasible for some metrics perhaps you can use quartiles so you have ~4 numbers per obs but I agree that ~10,000 per-obs-per-metric seems like too much.

tristpinsm:
> I don't know that we need per-detector-per-obs values published here. I was assuming these metrics would be boiled down to per-obs numbers. If you think that's infeasible for some metrics perhaps you can use quartiles so you have ~4 numbers per obs but I agree that ~10,000 per-obs-per-metric seems like too much.

Yeah, doing quartiles is a good idea... So there is no need for per-detector metrics in general you think?

I also encountered a related issue: So far I've been tagging metrics with the observation ID, which directly maps to the timestamp. This results in time series that are all length-1, which is not really what Influx is designed to handle. It seemed convenient to tag the observation ID, but may be better not to, and instead map times to obs_id in grafana or something. Any thoughts?

msilvafe:
> > I don't know that we need per-detector-per-obs values published here. I was assuming these metrics would be boiled down to per-obs numbers. If you think that's infeasible for some metrics perhaps you can use quartiles so you have ~4 numbers per obs but I agree that ~10,000 per-obs-per-metric seems like too much.
> 
> Yeah, doing quartiles is a good idea... So there is no need for per-detector metrics in general you think?
I haven't thought about every metric that would be useful so I think we would have to make a good case for a per-detector metric. In general this sort of Grafana monitoring should be a very high level check for data _stability_ i.e. checking for if some quality metric is changing significantly over time. It should serve to raise a flag for us that we should look closer into the data through the preprocess-plotting that David is working on and dedicated follow up checks on the data. 

> I also encountered a related issue: So far I've been tagging metrics with the observation ID, which directly maps to the timestamp. This results in time series that are all length-1, which is not really what Influx is designed to handle. It seemed convenient to tag the observation ID, but may be better not to, and instead map times to obs_id in grafana or something. Any thoughts?
Ya good point. What is your alternative suggestion? Perhaps using the obs_id timestamp instead of the obs_id itself?

tristpinsm:
OK, so I've modified the metrics now to save percentiles and mean across detectors, for every detset. I also removed the `obs_id` tag so that these better behave as time series in Influx.

tristpinsm:
@kmharrington Sorry to bug you, but I think this is ready for review whenever you have a moment.  I also updated the test grafana page here: https://site.simonsobs.org/grafana/d/b740304e-0c55-42bd-8a8e-6986a63bf291/qa-dev?orgId=1.

tristpinsm:
max has sent me some feedback in slack, so actually I should implement that first...

tristpinsm:
After discussing with Max, I made some changes:

- calculate statistics (mean, percentiles) for preprocess output across detectors grouped by wafer_slot and bandpass
- record the amplitude of the HWPSS modes
- calculate statistics only across samples deemed valid by preprocess
- make some parameters configurable: extra tags, choice of percentiles, the name of the dataset that contains the flags
- added a metric that counts the number of valid detectors at the end of preprocess by checking the fraction of valid samples is above a configurable threshold, in a configurable dataset (glitches is the default as it is at the end of the pipeline)

We also discussed some ideas that I didn't implement in the end:
- make the conditions for the detector groups used to calculate statistics configurable: this would add some complexity to the code and it sounded like wafer_slot/bandpass is a pretty standard grouping. But we could also add this later if desired
- record the number of flags and total number of samples instead of a fraction: this seemed straighforward, but if the total number of **valid** samples (which is now being used in the denominator of the fraction) can be different per detector, then it's harder to think of a sensible way of calculating the statistics for these two numbers separately

sotodlib/preprocess/processes.py
```diff
@@ -549,9 +550,21 @@ def gen_metric(cls, meta, proc_aman):
                 # mask those that were not set
                 nonzero = np.any(coeff != 0.0, axis=1)
 
+                # calculate amplitude of each mode
+                mode_labels = list(proc_aman.hwpss_stats.modes.vals)
+                num_re = re.compile("^[SC]([\d])$")
```
msilvafe, line 550: Is this limited to a single digit number between 0-9? I think we could have >=10f.

sotodlib/preprocess/processes.py
```diff
@@ -538,6 +514,86 @@ def plot(self, aman, proc_aman, filename):
...
+                )[0]
+
+                # get the coefficients for every detector
+                coeff = proc_aman.hwpss_stats.coeffs[subset]
+                # mask those that were not set
+                nonzero = np.any(coeff != 0.0, axis=1)
+
+                # calculate amplitude of each mode
+                mode_labels = list(proc_aman.hwpss_stats.modes.vals)
+                num_re = re.compile("^[SC]([\d])$")
```
msilvafe, line 550: Is this limited to a single digit number between 0-9? I think we could have >=10f.

msilvafe (approved):
I just have one comment about the HWPSS amplitudes but other than that I think this looks good to merge and try to start running regularly.

sotodlib/preprocess/processes.py
tristpinsm: Indeed I was only anticipating up to one digit here. I'll change that. Good catch!

PR #794, opened 2024-04-04
update obsdb columns or add entries from obsfiledb

AERAdler:
Monolithic as per Matthew's request. see closed PR#677 for how the script worked when separate. Now it's just a matter of using an --update_cols argument to load books from the ones listed in the obsfiledb and fill out columns or even whole entries in the obsdb.

### DIFF
--- a/sotodlib/site_pipeline/update_obsdb.py
+++ b/sotodlib/site_pipeline/update_obsdb.py
@@ -1,6 +1,7 @@
 """update_obsdb.py
 
 Create and/or update an obsdb and obsfiledb based on some books.
+Alternatively, update an obsdb based on an obsfiledb.
 The config file could be of the form:
 
 .. code-block:: yaml
@@ -16,7 +17,7 @@ The config file could be of the form:
       subtype: str
 
     obsdb: dummyobsdb.sqlite
-    obsfiledb: dummyobsfiledb.sqlite
+    obsfiledb: dummyobsfiledb.sqlite Needed if running with the --update_cols option
     lat_tube_list_file: path to yaml dict matching tubes and bands
     tolerate_stray_files: True
     skip_bad_books: True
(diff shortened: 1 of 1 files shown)


sotodlib/site_pipeline/update_obsdb.py
```diff
@@ -298,7 +338,9 @@ def get_parser(parser=None):
     parser.add_argument("--booktype", default="both", type=str,
         help="Select book type to look for: obs, oper, both(default)")
     parser.add_argument("--overwrite", action="store_true",
-        help="If true, writes over existing entries")
+        help="If True, overwrite existing entries")
+    parser.add_argument("--update_cols", action="store_true",
```
mhasself, line 342: We use form "--update-cols" instead of "--update_cols", in site_pipeline.

mhasself (changes requested):
Thanks.  I'm a bit confused about why the thing called "update_columns" seems so intrinsically tied to obsfiledb.  Is this mostly because one needs a way to get "all the obs" rescanned?

It seems that the basic behavior here is that when you "update-obsdb" you
- generate a list of books to operate on
- update the obsdb for each book, which includes (a) creating the entry if it doesn't exist and (b) updating the entry to the latest schema if it does exist.

If that's right then I think you should make that more clear in the docs, and also rename "update_columns" to reflect what it does more directly, which is "rescan-from-obsfiledb" or something like that.

Is there a way to "update" or "rerun" or "add" just a single obs, provided by path?  I think that's an important capability to have.  For testing of the "update columns" aspect of this.

AERAdler:
It looks for the locations of the books to be rescanned in the obsfiledb. 
That also means you can skip calls to checkbook, because you don't need to build the obsfiledb. 
I'll work on the comments you made.

PR #800, opened 2024-04-09
Preprocess update delay

msilvafe:
Add `update_delay` parameter to restrict lookback time when running through prefect.


guanyilun (approved):
Tested that this is working well. Thanks!

PR #804, opened 2024-04-10
Add ability to run on multiple subgroups

msilvafe:
This update allows us to run `site_pipeline.preprocess_tod` on multiple sub groups for example a config file that specifies:

```
subobs:
  use: ['wafer_slot','wafer.bandpass']
```

Will process the data in len(wafer_slot) x len(wafer.bandpass) subsets of detectors at a time. The logs for a single subset look like this:

```
108.230: Beginning run for obs_1704770693_satp1_1111111:['ws0', 'f090'] (INFO)
 117.809: Running flag_turnarounds (INFO)
Ranges(n=479864:rngs=28)
 119.300: Saving data to /so/home/msilvafe/shared_files/preprocess/satp1_test/preprocess_archive.h5:obs_1704770693_satp1_1111111_wafer_slot_ws0_wafer.bandpass_f090 (INFO)
 119.335: Saving to database under {'obs:obs_id': 'obs_1704770693_satp1_1111111', 'dataset': 'obs_1704770693_satp1_1111111_wafer_slot_ws0_wafer.bandpass_f090', 'dets:wafer_slot': 'ws0', 'dets:wafer.bandpass': 'f090'} (INFO)
```

And the manifestdb entry returned from `core.metadata.ManifestDb(archive).inspect()` for a single entry looks like:

```
{'_id': 2, 'filename': '/so/home/msilvafe/shared_files/preprocess/satp1_test/preprocess_archive.h5', 'obs:obs_id': 'obs_1704770693_satp1_1111111', 'dets:wafer_slot': 'ws0', 'dets:wafer.bandpass': 'f090', 'dataset': 'obs_1704770693_satp1_1111111_wafer_slot_ws0_wafer.bandpass_f090'}
```

### DIFF
--- a/sotodlib/site_pipeline/preprocess_tod.py
+++ b/sotodlib/site_pipeline/preprocess_tod.py
@@ -42,16 +42,19 @@ def _get_preprocess_context(configs, context=None):
     return configs, context
 
 def _get_groups(obs_id, configs, context):
-    group_by = configs['subobs'].get('use', 'detset')
-    if group_by.startswith('dets:'):
-        group_by = group_by.split(':',1)[1]
+    group_by = np.atleast_1d(configs['subobs'].get('use', 'detset'))
+    for i, gb in enumerate(group_by):
+        if gb.startswith('dets:'):
+            group_by[i] = gb.split(':',1)[1]
 
-    if group_by == 'detset':
-        groups = context.obsfiledb.get_detsets(obs_id)
-    else:
-        det_info = context.get_det_info(obs_id)
-        groups = det_info.subset(keys=[group_by]).distinct()[group_by]
-    return group_by, list(groups)
+        if (gb == 'detset') and (len(group_by) == 1):
+            groups = context.obsfiledb.get_detsets(obs_id)
+            return group_by, [[g] for g in groups]
+        
+    det_info = context.get_det_info(obs_id)
+    rs = det_info.subset(keys=group_by).distinct()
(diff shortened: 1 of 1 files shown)


sotodlib/site_pipeline/preprocess_tod.py
```diff
@@ -42,16 +42,19 @@ def _get_preprocess_context(configs, context=None):
     return configs, context
 
 def _get_groups(obs_id, configs, context):
-    group_by = configs['subobs'].get('use', 'detset')
-    if group_by.startswith('dets:'):
-        group_by = group_by.split(':',1)[1]
+    group_by = np.atleast_1d(configs['subobs'].get('use', 'detset'))
```
kmharrington, line 45: will this fail if "use" is None? TBH I'm fine with it not working if use is None but maybe want a more useful error message than I think this will make.

kmharrington:
Have you tested this one all the previous "standard" options? Thinking mainly, `detsets` and then a single group off of det info like wafer slot. And then the inline comment.

msilvafe:
> Have you tested this one all the previous "standard" options? Thinking mainly, `detsets` and then a single group off of det info like wafer slot. And then the inline comment.

Ok I tested with:

```
subobs:
  # use: 'wafer_slot'
```

Which I think covers the case of `use = None` and I also tested with: 

```
subobs:
  use: 'wafer_slot'
```

Both work now...after my last commit ;)

kmharrington (approved):
I hear all these tests have happened

PR #806, opened 2024-04-12
Add meta and det_info restriction.

msilvafe:
This adds the ability to restrict by passing a dets dictionary or meta AxisManager as can be done in ctx.get_obs and ctx.get_meta when running `site_pipline.preprocess_tod.load_preprocess_tod`

I tested this by running it as is now (where all wafers/dets are loaded at once) and then restricting after by 

```
aman.restrict('dets', aman.dets.vals[(aman.det_info.wafer_slot == 'ws0') & (aman.det_info.wafer.bandpass == 'f090')])
```

And confirmed that I had the same number of detectors at the end as if I ran with this new branch using:

```
aman = pt.load_preprocess_tod(oid, configs="/so/home/msilvafe/repos/site-pipeline-configs/satp1/preprocess_config_cmb.yaml",
                       dets={'wafer_slot':'ws0', 'wafer.bandpass':'f090'})
```

### DIFF
--- a/sotodlib/site_pipeline/preprocess_tod.py
+++ b/sotodlib/site_pipeline/preprocess_tod.py
@@ -149,7 +149,8 @@ def preprocess_tod(
         if len(db.inspect(db_data)) == 0:
             db.add_entry(db_data, dest_file)
 
-def load_preprocess_det_select(obs_id, configs, context=None):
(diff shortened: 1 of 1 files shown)


kmharrington (approved):
nice!

Issue #776, opened 2024-03-21
Include sotodlib version in hwp_angle metadata


ykyohei:
Adding `sotodlib.__version__` to the metadata is helpful because metadata identifiers do not necessarily reflect the version of the sotodlib, and there may be minor inconsistencies in an hwp metadata.
For example, debugging may only be applied to observations after a certain time.

mhasself:
> Whatever soconda does ... you don't get a version code. I don't know why that is.

Oh --- this is probably because of a shallow clone (depth=1 or whatever).  That would be missing the tag information required to construct the version code.

tskisner:
Right, currently we are just installing sotodlib master, but I think what I will do going forward (unless there are objections) are to make an sotodlib tag whenever we are going to update the stack and then use that tag for the install.  Would that have any side-effects?

mhasself:
> Right, currently we are just installing sotodlib master, but I think what I will do going forward (unless there are objections) are to make an sotodlib tag whenever we are going to update the stack and then use that tag for the install. Would that have any side-effects?

Seems fine to me!

iparask:
The tags will also help to quickly reproduce environment variables. I'm in favor of such a side-effect!

PR #809, opened 2024-04-15
Hwp metadata update

ykyohei:
This PR updates hwp_angles metadata and implement apply_hwp_angle_model

- Add draft of apply_hwp_angle_model https://github.com/simonsobs/sotodlib/discussions/735
  - We still need to implemenet hwp_angle_model metadata
  - Previously make_hwp_solution was correcting relative hwp rotation direction between observations using configured method, but removed this correction from make_hwp_solution to implement more robust correction in apply_hwp_angle_model
- Add sotodlib version to hwp_angles metadata https://github.com/simonsobs/sotodlib/issues/776
- Split hwp_angles metadata more https://simonsobs.slack.com/archives/C05K7CSFTPX/p1712183463493169
  - This is necessary for more efficient hwp metadata data transfer
  - Save fast/slow data into different HDF5 files.
  - Split HDF5 file by first 4 digits of unixtime -> 5 digits for unixtime.

### DIFF
--- a/sotodlib/hwp/g3thwp.py
+++ b/sotodlib/hwp/g3thwp.py
@@ -796,7 +796,7 @@ class G3tHWP():
         result = (interp > 0.999)
         return result
 
-    def set_data(self, tod, load_h5=True):
+    def set_data(self, tod, h5_filename=None):
         """
         Output HWP hk data as AxisManager format. The results are stored in HDF5 files.
         We save the copy of raw hwp encoder hk data into HDF5 file, to save time for
@@ -806,8 +806,8 @@ class G3tHWP():
         ----
         tod: AxisManager
 
-        load_h5:
-            If true, try to load raw encoder data from hdf5 file
+        h5_filename:
+            If this is not None, try to load raw encoder data from hdf5 file
 
         Notes
         -----
@@ -843,10 +843,11 @@ class G3tHWP():
 
         self.data = {}
         try:
-            if load_h5:
+            if h5_filename is not None:
                 logger.info('Loading raw encoder data from h5')
                 try:
-                    self.data = self._load_raw_axes(aman, output, h5_address)
+                    obs_id = tod.obs_info['obs_id']
+                    self.data = self._load_raw_axes(aman, h5_filename, obs_id)
                 except Exception as e:
                     logger.error(f"Exception '{e}' thrown while loading HWP data from h5. Attempt to load from hk.")
                     self.data = self.load_data(start, end)
--- a/sotodlib/hwp/g3thwp.py
+++ b/sotodlib/hwp/g3thwp.py
@@ -805,11 +805,9 @@ class G3tHWP():
         result = (interp > 0.999)
         return result
 
-    def write_solution_h5(self, tod, output=None, h5_address=None, load_h5=True):
+    def set_data(self, tod, load_h5=True):
         """
-        Output HWP angle, flags, metadata as AxisManager format
-        The results are stored in HDF5 files. Since HWP angle solution HDF5 files are large,
-        we automatically split into the new output files.
+        Output HWP hk data as AxisManager format. The results are stored in HDF5 files.
         We save the copy of raw hwp encoder hk data into HDF5 file, to save time for
         re-calculating the hwp angle solutions.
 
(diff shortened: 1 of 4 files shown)


17-sugiyama:
It seems good for me!
In [this line](https://github.com/simonsobs/sotodlib/blob/eab5872ee813393f3368c5f9e2052cae11c72111/sotodlib/hwp/hwp_angle_model.py#L62), the different angle offset is defined for two bands. I think we can remove the band dependency if we don't apply it.

17-sugiyama (approved):

ykyohei:
I make this PR ready for review

- The new hwp metadata is available here
/so/home/ykyohei/metadata/satp3/manifests/hwp_angles/satp3_hwp_angles_240430m
- apply_hwp_angle_model return single approximate hwp_angle without passband dependence
- Junna checked the consistency between multiple rotation direction estimations and the benefit of using multiple estimations
https://simonsobs.atlassian.net/wiki/spaces/PRO/pages/257392641/Daily+analysis#2024%2F4%2F26
- This PR should be merged with https://github.com/simonsobs/site-pipeline-configs/pull/67

sotodlib/hwp/g3thwp.py
```diff
@@ -905,87 +958,35 @@ def write_solution_h5(self, tod, output=None, h5_address=None, load_h5=True):
...
+            if not 'counter' + suffix in self.data.keys():
                 logger.warning('No HWP data in the specified timestamps.')
                 self._write_solution_h5_logger = 'No HWP data'
                 continue
 
             # version 1
             # calculate HWP angle
             try:
-                solved = solved | self.analyze(data, mod2pi=False, suffix=suffix)
+                solved = solved | self.analyze(self.data, mod2pi=False, suffix=suffix)
```
mhasself, line 989: This is a very strange line... the operator `|` isn't defined for dicts.  Am I missing something?

mhasself: You probably mean "or", right?  Has this code been tested?

ykyohei: This is merging two dictionaries for two encoders. This way is supported from python 3.9.0
https://stackoverflow.com/questions/38987/how-do-i-merge-two-dictionaries-in-a-single-expression-in-python

ykyohei: And yes, this is tested. We have minor modification in this PR, but the this code (operator `|`) has been used from previous version of metadata 240301m.

mhasself: Ah, I have not seen that... and was checking on python 3.8.

We are allowed to require py3.9 now but I will note you could have simply done
```
solved.update(self.analyze(self.data, mod2pi=False, suffix=suffix))
```
... and that would have been python2 compatible!

Thanks.

ykyohei:
@mhasself Please let me confirm. Is it okay to moving on merging this and updating the site-pipeline with Yilun? 
I guess we need a review for apply_hwp_angle_model.py.

mhasself (changes requested):
Thanks for all the work on this and sorry for the delay in responding.

See a few comments inline.  After those, I think we can merge this.

However, I want to point out that the API for the function will likely change a bit, in the near future.  `apply_hwp_angle_model(tod)` will be fine, and fairly portable... but the other arguments will change.  And the parameters themselves, which are currently hard-coded, will become metadata products.  But that can wait.

sotodlib/hwp/hwp_angle_model.py
```diff
@@ -0,0 +1,75 @@
...
+        'satp3': -2.29,
+    }
+
+    offset = encoder_offset + \
+        encoder_assembly_offset[telescope][hwp.primary_encoder] + \
+        ahwp_offset[telescope]
+
+    # apply correction
+    tod.hwp_angle = np.mod(
+        sign*tod.hwp_angle + np.deg2rad(offset), 2*np.pi)
```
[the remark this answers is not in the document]
ykyohei: I found `accept_new` and `accept_merge` option of aman.wrap are not doing anything, so checked it by _assignments.
https://github.com/simonsobs/sotodlib/blob/12b4b91265269bb38d1272c44e86d341a322806f/sotodlib/core/axisman.py#L555-L556

sotodlib/hwp/hwp_angle_model.py
```diff
@@ -0,0 +1,75 @@
...
+        'satp3': {'pid': -1, 'offcenter':  1, },
+    }
+
+    # check available sign estimation methods
+    available_signs = []
+
+    methods = ['pid', 'offcenter']
+    for method in methods:
+        _sign = getattr(hwp, method + '_direction')
+        if _sign != 0:
```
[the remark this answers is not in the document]
ykyohei: None does not happen.

ykyohei:
@mhasself Addressed your suggestions and tested that `apply_hwp_angle_model` works with new metadata.

mhasself (approved):
Thanks for the adjustments!

PR #810, opened 2024-04-17
AxisManager metadata does not need dets axis

mhasself:
This should make it easier to produce metadata products such as hwp_angles or pointing/angle correction models.

Without this fix, the hwp_angle product has needed to store the complete "dets" axis, which is awkward.  And there was no way to store a simple dict-like structure that might be applicable to many obs.

### DIFF
--- a/docs/context.rst
+++ b/docs/context.rst
@@ -414,10 +414,87 @@ syntax::
         unpack: 'special&my_special_vector'
 
 
+Example 3
+---------
+
+In this example we store a calibration table, with a separate number
+for each passband value.
+
+.. code-block:: python
+
+    from sotodlib import core
+    import sotodlib.io.metadata as io_meta
+
+    # This cal result has one number per wafer x passband.
+    data = [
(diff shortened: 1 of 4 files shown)


docs/context.rst
```diff
@@ -414,10 +414,87 @@ syntax::
...
+
+    # Model data (t0, t1, scale, offset)
+    data = [
+        (1700000000., 1800000000., 1.00, 4.50),
+        (1800000000., 1900000000., 0.98, 4.42),
+    ]
+
+    # Get a new database ready
+    db = core.metadata.ManifestScheme(). \
+        add_range_match('obs:timestamp'). \
```
msilvafe, line 474: Can range_match have no end time (i.e. t1=None) so that you can store a cal product that is valid from the time the product was updated until the present time?

mhasself: No.  Just put in some high place-holder (4e9).

kmharrington (approved):
Thanks, I've also found cases where I didn't want to save dets so this will be broadly useful

PR #811, opened 2024-04-17
Better outlier rejection, switch transform library, and QOL updates

skhrg:
Helps deal with noisy data when there is a relatively large initial offset.


mhasself (commented):
I think "Do a rigid alignment before outlier rejection" might not capture all the changes here.

Maybe rebase + squash this a little, and word the commit(s) descriptively ... and I won't hold this up any longer.

skhrg:
Moved the similar commits together and squashed with a longer commit message.
Or do you want me to squash all into one giga commit?

mhasself (approved):
Look at me not holding this up any more!

PR #812, opened 2024-04-17
Updates to preprocess_obs and SSOFootprint

davidvng:
- Fixed `update_full_aman` (and `_expand`) in preprocess core to work with aman with no signal (no dets broke the function)
- Updates to `preprocess_obs` to add Prefect capability that was added to `preprocess_tod`
- Added `tags` argument to allow querying planet scans
- Updated SSOFootprint to allow provided source list in config file (as opposed to finding nearby sources)

@guanyilun after I test run this on multiple obs and after this gets merged, I'll PR to site-pipeline-configs to get this running on Prefect.

### DIFF
--- a/sotodlib/site_pipeline/preprocess_obs.py
+++ b/sotodlib/site_pipeline/preprocess_obs.py
@@ -166,7 +166,7 @@ def get_parser(parser=None):
     )
     parser.add_argument(
         '--tags',
-        help="Observation tags. Ex: --tags='jupiter=1'",
+        help="Observation tags. Ex: --tags 'jupiter' 'setting'",
         nargs='*',
         type=str
     )
--- a/sotodlib/site_pipeline/preprocess_obs.py
+++ b/sotodlib/site_pipeline/preprocess_obs.py
@@ -1,7 +1,10 @@
 import os
 import yaml
+import time
 import numpy as np
 import argparse
+import traceback
+from typing import Optional
 
 from sotodlib import core
 import sotodlib.site_pipeline.util as sp_util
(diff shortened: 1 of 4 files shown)


davidvng:
@kmharrington @msilvafe this should be ready for review, below is the config file I used and the command line.
```
context_file: '/so/home/dnguyen/repos/scripts/preprocess/use_this.yaml'

plot_dir: '/so/home/dnguyen/repos/scripts/preprocess/plots'

archive:
  index: '/so/home/dnguyen/repos/scripts/preprocess/process_obs_archive.sqlite'
  policy:
    type: 'simple'
    filename: '/so/home/dnguyen/repos/scripts/preprocess/preprocess_obs_archive.h5'

process_pipe:
    - name: "sso_footprint"
      calc:
        source_list: ['jupiter'] # remove to find nearby sources
        distance: 20
        nstep: 100
      save: True
      plot:
        wafer_offsets: {'ws0': [-2.5, -0.5],
                        'ws1': [-2.5, -13],
                        'ws2': [-13, -7],
                        'ws3': [-13, 5],
                        'ws4': [-2.5, 11.5],
                        'ws5': [8.5, 5],
                        'ws6': [8.5, -7]}
        focal_plane: '/so/home/msilvafe/shared_files/sat_hw_positions.npz'
```
command line: `python3 preprocess_obs.py /so/home/dnguyen/repos/scripts/preprocess/preprocess_obs_config.yaml --min-ctime=1710000000 --query="type == 'obs'" --tags='jupiter=1'`
note that I had to use my own `use_this.yaml` context file in which I modified `on_missing: trim`
```
- db: '{manifestdir}/det_cal/satp1_det_cal_240312m/det_cal.sqlite'
    label: det_cal
    unpack: det_cal
    on_missing: trim
```

sotodlib/site_pipeline/preprocess_obs.py
```diff
@@ -184,7 +203,7 @@ def main(
         if tot_query=="":
             tot_query="1"
     
-    obs_list = context.obsdb.query(tot_query)
+    obs_list = context.obsdb.query(tot_query, tags=tags)
```
msilvafe, line 206: Have you tested passing a list of tags i.e. `jupiter, saturn, moon`? And perhaps its more intuitive to just pass the SSO name to the argparse instead of `<sso_name>=1` and deal with the `=1` bit when you pass the tags to `obsdb.query`? Also I think this is case-sensitive can you add a `<str>.lower()` (I think it has to be all lower case) so this doesn't break if someone passing `Jupiter` instead of `jupiter`?

msilvafe (approved):
Looks good, just one inline comment to address. How have you tested this so far?

PR #814, opened 2024-04-19
Make metadata for important house keeping data

tterasaki:
I would like to request to make level3 metadata databases for important housekeeping data, such as pwv, mK-stage temperature and wiregrid. I followed the make_hwp_solutions' way to make metadata and sqlite.

### DIFF
--- a/sotodlib/site_pipeline/make_cosamp_hk.py
+++ b/sotodlib/site_pipeline/make_cosamp_hk.py
@@ -7,7 +7,7 @@ from typing import Optional
 from sotodlib import core
 from sotodlib.io import hk_utils
 from sotodlib.site_pipeline import util
-logger = util.init_logger('hk2meta', 'hk2meta: ')
+logger = util.init_logger('make_cosamp_hk', 'make_cosamp_hk: ')
 
 def get_parser(parser=None):
     if parser is None:
@@ -137,7 +137,7 @@ def main(context: str,
     logger.info(f'tot_query: {tot_query}')
     obs_list= ctx.obsdb.query(tot_query, query_tags)
     
-    # make obs_id
+    # output cosampled data for each obs_id
     for obs in obs_list:
         obs_id = obs['obs_id']
         try:
--- a/sotodlib/site_pipeline/make_cosamp_hk.py
+++ b/sotodlib/site_pipeline/make_cosamp_hk.py
@@ -11,10 +11,7 @@ logger = util.init_logger('make_cosamp_hk', 'make_cosamp_hk: ')
 
 def get_parser(parser=None):
     if parser is None:
-        parser = argparse.ArgumentParser()
-    parser.add_argument(
-        'context',
-        help="Path to context yaml file")
+        parser = argparse.ArgumentParser()    
     parser.add_argument(
         'config',
         help="configuration yaml file.")
@@ -50,10 +47,14 @@ def get_parser(parser=None):
         '--obs_id', type=str,
         help="obs_id of particular observation if we want to run on just one",
     )
+    parser.add_argument(
+        '--update-delay',
+        help="Number of days (unit is days) in the past to start observation list.",
+        type=int
+    )
     return parser
     
-def main(context: str,
-        config: str,
+def main(config: str,
         output_dir: Optional[str] = None,
         verbose: Optional[int] = 2,
         overwrite: Optional[bool] = False,
@@ -65,8 +66,6 @@ def main(context: str,
         ):
     """
     Args:
-        context: str
-            File path to the context file.
         config: str
             Path to the configuration file containing processing parameters.
         output_dir: str or None
@@ -99,7 +98,8 @@ def main(context: str,
     
     logger.info(f'min_ctime: {min_ctime}')
     # load context
-    ctx = core.Context(context)
+    context_file = hk2meta_config['context_file']
+    ctx = core.Context(context_file)
     
     # load configuration file
     with open(config) as f:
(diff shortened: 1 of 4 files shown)


tterasaki:
It is ready for review. See the directory for the code example.
`computeXX:/so/home/tterasaki/code_dev/hk_to_level3/satp1_example_240424`
`computeXX:/so/home/tterasaki/code_dev/hk_to_level3/satp3_example_240422`

sotodlib/site_pipeline/make_cosamp_hk.py
```diff
@@ -0,0 +1,212 @@
...
+                                          det_cosampled=True, det_aman=aman)
+            
+            hkman = core.AxisManager(meta.dets, aman.samps)
+            hkman.wrap('timestamps', aman.timestamps, [(0, 'samps')])
+
+            full_fields = hk2meta_config['fields']
+            aliases = hk2meta_config['aliases']
+            data_fields = [full_field.split('.')[1] for full_field in full_fields]
+            num_different_field = 0
+            for i, (full_field, data_field, alias) in enumerate(zip(full_fields, data_fields, aliases)):
```
msilvafe, line 192: Seems like `full_field` isn't used in this loop.

sotodlib/site_pipeline/make_cosamp_hk.py
```diff
@@ -0,0 +1,212 @@
...
+            logger.info(f"saved: {obs_id}")
+            
+        except Exception as e:
+            logger.error(f"Exception '{e}' thrown while processing {obs_id}")
+            continue
+
+if __name__ == '__main__':
+    parser = get_parser(parser=None)
+    args = parser.parse_args()
+    main(**vars(args))
```
msilvafe, line 212: Can replace this with `util.main_launcher(main, get_parser)`

sotodlib/site_pipeline/make_cosamp_hk.py
```diff
@@ -0,0 +1,212 @@
...
+            num_different_field = 0
+            for i, (full_field, data_field, alias) in enumerate(zip(full_fields, data_fields, aliases)):    
+                if i > 0:
+                    num_different_field += int(data_fields[i-1] != data_field)
+                i_in_same_field = i - num_different_field
+                hkman.wrap(alias, _hkman[data_field][data_field][i_in_same_field], [(0, 'samps')])
+            
+            h5_filename = f"{output_prefix}_{obs_id.split('_')[1][:4]}.h5"
+            output_filename = os.path.join(output_dir, h5_filename)
+            hkman.save(dest=output_filename, group=obs_id, overwrite=overwrite, compression='gzip')
```
msilvafe, line 200: Do you need to gzip the data (i.e. do you think it's large enough for us to care)? I think that adds some extra io time

msilvafe (approved):
I put a few minor comments in line that I hope are easy to address but nothing blocking so I approved here.

One large question I have is how we plan to use this effectively. Do you envision we have 1 ManifestdB and associated config file, query, etc for each usecase (i.e. wiregrid, pwv, etc.)? Or one config file for a set of fields that would be generally useful for all observation types? I think I could find a usecase for both but just thinking about how (or if) we will turn this on in an automated way through prefect and log the different types of versions via the metadata versions log like those kept here: https://simonsobs.atlassian.net/wiki/spaces/PRO/pages/306249767/Site+Pipeline+Metadata+Logs and managed by @ahincks or if this is something you think will more likely be used on a per-user basis and folks will just keep custom built ManifestdB's in their user spaces.

tterasaki:
Thank you, Max. I will reflect your comment later.
In my personal opinion, people should always aware of what is PWV and mK-stage. So I want to keep the pwv and mk-stage always loaded with detector data.
I want to hear from @mhasself and @kmharrington. 
And for the wiregrid calibration, I think the cosampled data is almost necessary.

msilvafe:
> Thank you, Max. I will reflect your comment later. In my personal opinion, people should always aware of what is PWV and mK-stage. So I want to keep the pwv and mk-stage always loaded with detector data. I want to hear from @mhasself and @kmharrington. And for the wiregrid calibration, I think the cosampled data is almost necessary.

Makes sense to me to have a specific set of fields that we log in this manifest for each obs. But I don’t think that the wiregrid info is needed for every obs so my suggestion would be to have at least 2 manifests one that queries all obs and includes pwv, mK_temp and one that queries just wiregrid tagged cal observations and includes the relevant wg fields.

sotodlib/site_pipeline/make_cosamp_hk.py
```diff
@@ -0,0 +1,212 @@
...
+import argparse
+import logging
+from typing import Optional
+
+from sotodlib import core
+from sotodlib.io import hk_utils
+from sotodlib.site_pipeline import util
+logger = util.init_logger('make_cosamp_hk', 'make_cosamp_hk: ')
+
+def get_parser(parser=None):
```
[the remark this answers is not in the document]
tterasaki: Thank you, Max. But I am a bit confused about prefect.
Is it ok just to add this '--update-delay' in the parser? Any modification for the main function is not necessary? It looks like so in the preprocess_tod.py.

sotodlib/site_pipeline/make_cosamp_hk.py
tterasaki: Ah, good point. That should be fine. 
Actually I just copy and pasted from update-hwp-angle, but our case should be smaller than them.

sotodlib/site_pipeline/make_cosamp_hk.py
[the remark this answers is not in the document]
msilvafe: Ya you can just add that argument to the parser and then in your main function add something like this to update the `min-ctime` variable in main accordingly: https://github.com/simonsobs/sotodlib/blob/071f2efc2dcd6cf33898a09e235fb004ac4e97b8/sotodlib/site_pipeline/preprocess_tod.py#L257-L260

sotodlib/site_pipeline/make_cosamp_hk.py
```diff
@@ -0,0 +1,212 @@
...
+            meta = ctx.get_meta(obs_id)
+            aman = ctx.get_obs(obs_id, dets=[])
+            
+            
+            
+            _hkman = hk_utils.get_detcosamp_hkaman(aman, 
+                                      fields = hk2meta_config['fields'],
+                                      data_dir = hk2meta_config['input_dir'])
+            
+            dt_samp_hk = hk2meta_config['dt_samp_hk']
```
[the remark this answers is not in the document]
tterasaki: I see. And I realized it should be fine that the dt_samp_hk, which is just used for a buffer of interpolation, is hard-coded. Removing that from the requirement in config files.

tterasaki:
Thank you, Max. I am merging it into master.

PR #817, opened 2024-04-22
Generalize outside data paths

### LINKED ISSUE #813: Planet data download from astropy
Function [`_get_astrometric`](https://github.com/simonsobs/sotodlib/blob/master/sotodlib/coords/planets.py#L239) tries to download planet data. This download does not work on HPC systems like Della and Tiger, where the compute nodes are not connected to the internet. The same will be true for Tiger 3.

In addition, the data are downloaded in the user's home directory by default, which requires every user to have a copy of this data. 

Astropy has a plan to provide configurable paths for caching ([Issue here](https://github.com/astropy/astropy/issues/10738)), but it does not seem to have a lot of traction.

Proposal: `sotodlib` checks first to a specific path, trying to load the necessary data, and falls back in trying to download them if they do not exist while issuing a warning

iparask:
This PR closes #813.

It introduces an environment variable for data that are downloaded. This variable is a serialized JSON.

This is a breaking change since it will raise a runtime error when the environment variable does not have the expected entry.

We may want to merge it along with an addition to soconda module file that set the variable.


kmharrington:
@jlashner  you had a different environmental variable name for site pipeline things?

iparask:
@kmharrington, @jlashner, the more we add to this env variable the better I think.

jlashner:
In certain scripts I allow the usage of the env var `SITE_PIPELINE_CONFIG_DIR` to specify the config dir root, which is useful during local testing, but that seems pretty unrelated to what you're adding! I don't know if they need to be merged right now.

mhasself:
Thanks for contribution -- but please undo the major reformatting.

It is long-standing policy that:
- bulk reformatting should be done in separate commits, and preferably separate PRs, from substantive changes.
- bulk reformatting should only be done when necessary.

I'll have some comments on the substance too but might as well get this part handled!

mhasself (changes requested):
On the substance --

1) the decoding you added in _get_astrometric can be made more general.  Move it into new submodule, `core.resources`, and have _get_astrometric simply call that('de421.bsp') to get a path.
2) I would like to leave a hard-coded dictionary in sotodlib, to name all the useful files (which will expand beyond this 1 ephemeris) and provide standard places to grab them from.  The soconda installer can use that to populate the cache.  (This was the original intention behind RESOURCE_URLS declared at top-level in planets.py)  I don't want to require that esoteric envvars be required by default.  Most installations of sotodlib do not use soconda.
3) The envvar can then be _entirely optional_, meaning that if "de421.bsp" is not listed in the decoded SOTODLIB_RESOURCES, the code will instead grab the spec from the core.resources.RESOURCE_DEFAULTS dict and use that.

Makes sense?

iparask:
@mhasself, I think it is ready for a new review.

sotodlib/coords/planets.py
```diff
@@ -270,8 +263,9 @@ def _get_astrometric(source_name, timestamp, site='_default'):
             pass
     else:
         options = list(planets.names().values())
-        raise ValueError(f'Failed to find a match for "{source_name}" in '
-                         f'ephemeris: {options}')
+        raise ValueError(
+            f'Failed to find a match for "{source_name}" in ' f"ephemeris: {options}"
```
mhasself, line 267: Please reformat this mangled string

sotodlib/coords/planets.py
```diff
@@ -739,4 +734,4 @@ class MmTimer(coords.Timer):
             'solved': map1b,
             'P': P,
             'det_weights': det_weights,
-            }
+            }
```
mhasself, line 737: Restore newline at eof.

sotodlib/core/resources.py
```diff
@@ -0,0 +1,62 @@
...
+def get_local_file(filename: str, cache: bool = True) -> str:
+    """
+    This function is responsible to check if a resource file exists either in a
+    predefined cache or in a user specified path. If not, it tries to download
+    from an ftp path.
+
+    Args:
+      filename: The name of the file to grab
+      cache: A path in the user's local folder to check for existing files and
+             cache them after a download
```
mhasself, line 21: This is not accurate.

sotodlib/core/resources.py
```diff
@@ -0,0 +1,62 @@
...
+
+    if de_url is None:
+        de_url = RESOURCE_DEFAULTS.get(filename, None)
+
+    if de_url is None:
+        raise RuntimeError(
+            f"File {filename} does not exist in default resources neither in "
+            + "$SOTODLIB_RESOURCES. Please set $SOTODLIB_RESOURCES with a "
+            + "json where the key is the resource name and value either a "
+            + "local path or an ftp path."
```
mhasself, line 43: All resources should be listed in RESOURCE_DEFAULTS, so that should be the primary instruction to the use in this (helpful) error message.

sotodlib/core/resources.py
```diff
@@ -0,0 +1,62 @@
...
+
+        # Set the target path for the file and create it if it does not exist
+        target_path = local_cache if cache else "/tmp/"
+        os.makedirs(name=target_path, exist_ok=True)
+        target_file = os.path.join(target_path, filename)
+        _, headers = urlretrieve(de_url, target_file)
+
+    elif de_url.startswith("file://"):
+        target_file = de_url[7:]
+
```
mhasself, line 61: Add `else` block to handle malformed de_url.

sotodlib/core/resources.py
```diff
@@ -0,0 +1,62 @@
...
+RESOURCE_DEFAULTS = {
+    "de421.bsp": "ftp://ssd.jpl.nasa.gov/pub/eph/planets/bsp/de421.bsp",
+}
+
+
+def get_local_file(filename: str, cache: bool = True) -> str:
+    """
+    This function is responsible to check if a resource file exists either in a
+    predefined cache or in a user specified path. If not, it tries to download
+    from an ftp path.
```
mhasself, line 16: In docstring, explain the path search order.  Provide an example of a well-formed SOTODLIB_RESOURCES var.

mhasself (changes requested):
Thank -- couple comments on the implementation, mostly documentation.

iparask:
@mhasself can you take another look? Thank you!

mhasself (approved):
Thanks!

PR #818, opened 2024-04-26
Rework preprocess for parallelization.

msilvafe:
Working to make the preprocess_tod script able to run observations in parallel. Basically this does:
1. Collect the `run_list` as we did before.
2. Subdivide the `run_list` into smaller lists that are at most `write_block` long called `prun_list` (parallel run list)
3. Then all of the obs in `prun_list` can feed to a parrallel pool via the `pool.async()` call with passes a different obs_id, group_list, and h5 filepath (through the config dictionary `archive.policy.filename` entry using swap_archive) for each obs in the list
4. I've modified the `preprocess_tod` function so that it now saves the h5 files (since they're all different h5 files) but does not write to the dB so that there aren't multiple writers to the sqlite db (since there's no lock for that). And I've moved the error handling into the function so now if we catch any errors they get returned along with the information to write into the database. We collect a tuple which is `(error_info, [db_data, dest_file])` if `error_info = None` and a similar structure if error_info is not None but with db_data, dest_file filled with info about the error. The asyc_apply pool collects all of these output tuples into a list.
5. Once the parallel bit gets through everything in prun_list then the outputs are looped over and we 1) move the temp h5 file to the destination specified in the config file 2) write/update the sqlite database entry 3) delete the temporary files.
6. Then we continue looping over the run_list in chunks size prun_list in this way until they're all done.

To test I ran the preprocess in master serially on two observations and ran the same two observations through this branch in parallel and confirmed that when I loaded the metadata and plotted them against each other (field I looked at was fitted hwpss amplitudes) they are the same.


sotodlib/site_pipeline/preprocess_tod.py
```diff
@@ -56,13 +59,39 @@ def _get_groups(obs_id, configs, context):
...
+            scheme=scheme
+        )
+    return db
+
+def swap_archive(config, fpath):
+    tc = copy.deepcopy(config)
+    tc['archive']['policy']['filename'] = os.path.join(os.path.dirname(tc['archive']['policy']['filename']), fpath)
+    dname = os.path.dirname(tc['archive']['policy']['filename'])
+    if not(os.path.exists(dname)):
+        os.mkdir(dname)
```
guanyilun, line 86: might be safer to use makedirs in case fpath is more than one level deep.

sotodlib/site_pipeline/preprocess_tod.py
```diff
@@ -292,18 +339,44 @@ def main(
...
+                                              'overwrite': overwrite,
+                                              'run_parallel': True}) for pr in prun_list]
+        outputs = [r.get() for r in async_out]
+        db = _get_preprocess_db(configs, group_by)
+        dest_file = configs['archive']['policy']['filename']
+        all_files = []
+        for err, output in outputs:
+            if err is None:
+                for db_data, src_file in output:
+                    with h5py.File(dest_file,'w') as f_dest, h5py.File(src_file,'r') as f_src:
```
guanyilun, line 364: does `dest_file` need to open and close once every loop?

sotodlib/site_pipeline/preprocess_tod.py
```diff
@@ -80,7 +109,8 @@ def preprocess_tod(
     logger: logging instance
         the logger to print to
```
guanyilun, line 110: missing docstring for run_parallel

sotodlib/site_pipeline/preprocess_tod.py
```diff
@@ -241,6 +270,18 @@ def get_parser(parser=None):
...
     )
+    parser.add_argument(
+        '--write-block',
+        help="How many obs before writing to db.",
+        type=int,
+        default=10
+    )
+    parser.add_argument(
+        '--nthreads',
+        help="Number of parallel threads to run on.",
```
guanyilun, line 281: multiprocessing uses parallel processes instead of parallel threads to avoid GIL

guanyilun (changes requested):
Looks good to me! Left some minor comments

sotodlib/site_pipeline/preprocess_tod.py
```diff
@@ -292,18 +343,61 @@ def main(
...
+    # Setup multiprocessing pool.
+    pool = get_context("spawn").Pool(processes=nproc) 
+
+    # Expects archive policy filename to be <path>/<filename>.h5 and then this adds
+    # <path>/<filename>_<xxx>.h5 where xxx is a number that increments up from 0 
+    # whenever the file size exceeds 10 GB.
+    nfile = 0
+    dest_file = os.path.splitext(configs['archive']['policy']['filename'])[0]+'_'+str(nfile).zfill(3)+'.h5'
+    if not(os.path.exists(dest_file)):
+            os.makedirs(dest_file)
```
kmharrington, line 355: these makedirs are causing me issues because they're making directories and not h5 files. Maybe need to change it to something like 
```
nfile = 0
    dest_file = os.path.splitext(configs['archive']['policy']['filename'])[0]+'_'+str(nfile).zfill(3)+'.h5'
    folder = os.path.split(folder)[0]
    if not (os.path.exists(folder)):
        os.makedirs(folder)
    
    while os.path.exists(dest_file) and os.path.getsize(dest_file) > 10e9:
        nfile += 1
        dest_file = os.path.splitext(dest_file)[0]+str(nfile).zfill(3)+'.h5'
```

both here and below

sotodlib/site_pipeline/preprocess_tod.py
```diff
@@ -320,18 +371,66 @@ def main(
...
+        if not(os.path.exists(folder)):
+            os.makedirs(folder)
+        if os.path.exists(dest_file) and os.path.getsize(dest_file) >= 10e9:
+            nfile += 1
+            dest_file = os.path.splitext(dest_file) + '_'+str(nfile).zfill(3)+'.h5'
+
+        h5_path = os.path.relpath(dest_file,
+                        start=os.path.dirname(configs['archive']['index']))
+
+        with h5py.File(dest_file,'w') as f_dest:
```
kmharrington, line 413: pretty sure this 'w' needs to be 'a'

msilvafe:
I've got this working now, including moving around h5 datasets and deleting the temp h5 files and populating the sqlite db as the jobs are completed in the multiprocessing pool. I left in this dummy function that I can add into the multiprocessing pool to speedup testing for all of the step in and after the multiprocessing pool.

guanyilun (approved):
Since you have tested this without problem, happy to let this through! My slight confusion is the appearance of both multiprocessing and concurrent. My naive understanding is they are different implementations of similar functionality. Do they work well together?

msilvafe:
> Since you have tested this without problem, happy to let this through! My slight confusion is the appearance of both multiprocessing and concurrent. My naive understanding is they are different implementations of similar functionality. Do they work well together?

Hi @guanyilun [the ProcessPoolExecutor in concurrent.futures is just a wrapper around the multiprocessing pool](https://docs.python.org/3/library/concurrent.futures.html#processpoolexecutor) so I use the `multiprocessing` package to set the pool to launch in `spawn` instead of `fork` mode to get around the issues we were having with the python logger deadlocking the different processes but `ProcessPoolExecutor` has this nice `as_completed` method that allows me to write out outputs from he multiprocessing pool and do things with them (write to the sqlite dB and move around and delete h5 files) without using the `write-block` implementation I had before.

PR #821, opened 2024-04-29
Functions for detector splits

chervias:
Functions for detector splits. First we start with a relative detector function, then we can add an absolute splits, etc.

### DIFF
--- a/docs/obs_ops.rst
+++ b/docs/obs_ops.rst
@@ -14,3 +14,16 @@ Module containing functions for making a SSO footprint.
 
 .. automodule:: sotodlib.obs_ops.sources
    :members:
+   
+obs_ops.splits
+==============
+
+Module containing functions for making splits of the data
+
+.. automodule:: sotodlib.obs_ops.splits
+    :members:
+
+Reference
+---------
+
+.. autofunction:: sotodlib.obs_ops.splits.det_splits_relative
--- a/docs/obs_ops.rst
+++ b/docs/obs_ops.rst
@@ -22,8 +22,3 @@ Module containing functions for making splits of the data
 
 .. automodule:: sotodlib.obs_ops.splits
     :members:
-
-Reference
----------
-
-.. autofunction:: sotodlib.obs_ops.splits.det_splits_relative
--- a/sotodlib/obs_ops/splits.py
+++ b/sotodlib/obs_ops/splits.py
@@ -20,8 +20,6 @@ def det_splits_relative(aman, det_left_right=False, det_upper_lower=False, det_i
 
     Returns
     -------
-    aman : AxisManager
-        Updated aman with the new flags
     fm: FlagManager with the requested flags
     """
 
@@ -55,10 +53,12 @@ def det_splits_relative(aman, det_left_right=False, det_upper_lower=False, det_i
 
     if not(wrap):
         if wrap == True:
-            aman.move('det_flags', None)
+            if 'det_flags' in aman._fields:
+                aman.move('det_flags', None)
             aman.wrap('det_flags', fm)
         else:
-            aman.move(wrap, None)
+            if wrap in aman._fields:
+                aman.move(wrap, None)
             aman.wrap(wrap, fm)
     return fm
 
--- a/sotodlib/obs_ops/splits.py
+++ b/sotodlib/obs_ops/splits.py
@@ -53,9 +53,10 @@ def det_splits_relative(aman, det_left_right=False, det_upper_lower=False, det_i
         mask = radii > radius_median
         fm.wrap_dets('det_out', np.logical_not(mask))
 
-    if wrap is not None and wrap is not False:
+    if not(wrap):
         if wrap == True:
             aman.wrap('det_flags', fm)
         else:
             aman.wrap(wrap, fm)
-    return aman, fm
+    return fm
+
--- a/sotodlib/obs_ops/splits.py
+++ b/sotodlib/obs_ops/splits.py
@@ -1,7 +1,7 @@
 import numpy as np
 from sotodlib.core import FlagManager
 
-def det_splits_relative(aman, det_left_right=False, det_upper_lower=False, det_in_out=False):
+def det_splits_relative(aman, det_left_right=False, det_upper_lower=False, det_in_out=False, wrap=None):
     """
     Function for adding relative detector splits to aman. A new FlagManager called det_flags will be created and the flags put there.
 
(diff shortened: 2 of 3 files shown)


sotodlib/obs_ops/splits.py
```diff
@@ -0,0 +1,55 @@
+import numpy as np
+from sotodlib.core import FlagManager
+
+def det_splits_relative(aman, det_left_right=False, det_upper_lower=False, det_in_out=False):
```
msilvafe, line 4: Add an argument called `wrap` which defaults to `None` and if None or False it does not wrap the flags into the input axis manager. If `wrap` is a True it wraps into some default field name `det_flags` and if `wrap` is a string then it wraps the  flags into that field name.

sotodlib/obs_ops/splits.py
```diff
@@ -0,0 +1,55 @@
...
+    Returns
+    -------
+    aman : AxisManager
+        Updated aman with the new flags
+    """
+
+    try:
+        aman.det_flags
+    except KeyError:
+        aman.wrap('det_flags', FlagManager.for_tod(aman))
```
msilvafe, line 28: Create a new axis manager called `det_flags` or `dflag_aman` and only wrap it if the wrap argument is not False/None.

sotodlib/obs_ops/splits.py
```diff
@@ -0,0 +1,55 @@
...
+    if det_in_out:
+        xi_center = np.min(xi) + 0.5 * (np.max(xi) - np.min(xi))
+        eta_center = np.min(eta) + 0.5 * (np.max(eta) - np.min(eta))
+        radii = np.sqrt((xi_center-xi)**2 + (eta_center-eta)**2)
+        radius_median = np.median(radii)
+        mask = radii <= radius_median
+        aman.det_flags.wrap_dets('det_in', np.logical_not(mask))
+        mask = radii > radius_median
+        aman.det_flags.wrap_dets('det_out', np.logical_not(mask))
+    return aman
```
msilvafe, line 55: Also return the `det_flags` or `dflag_aman` axis manager and only wrap it into `aman` if `wrap` is not None/False.

msilvafe (changes requested):
Just a few comments to make this a bit more compatible with the structure used in other functions and to be more compatible with the structure of the preprocess module. Perhaps @kmharrington or @mhasself has opinions about whether this should go in a new `obs_ops` module or if it should go in `tod_ops.flags` I suppose since these are returning boolean masks that are shape `dets` not RangesMatrices shape `dets x samps` it may make more sense to delineate that way and have `dets` type masks live here in `obs_ops`

chervias:
If I try to run again the function and a wrap is attempted for a field that already exists this will fail, but I think it should overwrite.

sotodlib/obs_ops/splits.py
```diff
@@ -0,0 +1,61 @@
...
+        xi_center = np.min(xi) + 0.5 * (np.max(xi) - np.min(xi))
+        eta_center = np.min(eta) + 0.5 * (np.max(eta) - np.min(eta))
+        radii = np.sqrt((xi_center-xi)**2 + (eta_center-eta)**2)
+        radius_median = np.median(radii)
+        mask = radii <= radius_median
+        fm.wrap_dets('det_in', np.logical_not(mask))
+        mask = radii > radius_median
+        fm.wrap_dets('det_out', np.logical_not(mask))
+
+    if wrap is not None and wrap is not False:
```
msilvafe, line 56: This can just be `if not(wrap)` if `wrap` is `None` or `False` then `not(wrap)` will be True.

sotodlib/obs_ops/splits.py
```diff
@@ -0,0 +1,61 @@
...
+        if wrap == True:
+            aman.wrap('det_flags', fm)
+        else:
+            aman.wrap(wrap, fm)
+    return aman, fm
```
msilvafe, line 61: Just return `fm` don't need to return `aman`

msilvafe (approved):
This looks good, the issue you brough up about overwriting I addressed in PR #824 and I don't think should be tied to this PR.

Actually before merge can you add some docs into `/docs` perhaps add an `obs_ops.rst` and describe the purpose of the splits and how to use there?

chervias:
done. So for example if I wrap a left/right split, and later I decide I actually want a left/right and a upper/lower split, when I run this function again it will fail. I think it should by default overwrite. Do you agree? We can add that later once there is a clean way of overwriting fields to a aman

chervias:
sorry I didn't see docs stuff, I'll do it before merging

msilvafe:
> done. So for example if I wrap a left/right split, and later I decide I actually want a left/right and a upper/lower split, when I run this function again it will fail. I think it should by default overwrite. Do you agree? We can add that later once there is a clean way of overwriting fields to a aman

Yes for now outside of this function you will need to run `aman.move('det_flags', None)` and can move overwrite in here if needed when that PR is merged.

docs/obs_ops.rst
```diff
@@ -13,4 +13,17 @@ obs_ops.sources
...
+
+Module containing functions for making splits of the data
+
+.. automodule:: sotodlib.obs_ops.splits
+    :members:
+
+Reference
+---------
+
+.. autofunction:: sotodlib.obs_ops.splits.det_splits_relative
```
msilvafe, line 29: You can delete these lines, the line above that calls `:members:` means that it will render the docstrings for all of the functions within the `obs_ops.splits` module so currently you're duplicating the docstring for that fn in two places:
Screenshot 2024-05-01 at 10 21 02 AM

sotodlib/obs_ops/splits.py
```diff
@@ -55,8 +55,10 @@ def det_splits_relative(aman, det_left_right=False, det_upper_lower=False, det_i
 
     if not(wrap):
         if wrap == True:
+            aman.move('det_flags', None)
```
msilvafe, line 58: I think that if `det_flags` doesn't exist in aman already this will fail so may need a check like:

```
if name in aman._fields:
    aman.move(name, None)
```

msilvafe (approved):
Looks good, just a couple comments but won't make them blocking. So you know, to test the docs build I create a virtual environment, install all of the dependencies, and then build the html. 

```
$ virtualenv test
$ source test/bin/activate
$ pip install -r docs/requirements.txt
$ cd docs
$ make html
```

Then I just open the html files written to `sotodlib/docs/_build/html` and inspect the relevant pages to make sure all the formatting looks reasonable.

msilvafe (approved):
Looks good, thanks Carlos.

PR #822, opened 2024-04-30
Deal with overlapping observations

kmharrington:
We ran into an issue today where one UFM streamed for ~3 hours while all the others were being turned off and on. This was an exact issue we'd said we "wouldn't deal with" but we need the drone data and the update functions were going to keep throwing errors about it. We can't automate a fix but this set of code makes it easier to manually fix these events. 

(1) the error is now a useful printout instead of just a uniqueness database constraint error
(2) we have a helper function to isolate the offending time range and make it easy to choose which book the overlapped observation goes in to. 
```
rsets = utils.find_overlaps(imprint, 'obs_ufm_mv9_1714406208', min_ctime, max_ctime)
imprint.register_book( rsets[0], commit=True)
```
(3) the update function now removes already registered level 2 observations from the search for new books to register (something I'd been thinking we should to for awhile and I'm pleasantly surprised at how easy it actually was to implement) 

------------------------
This PR now also includes a bunch of fixes for dealing with the timing system dropping out and improvements to the Imprinter CLI. For the timing system issues, books made with Low Precision timing should not be used for science analysis and so we want to clear them out but make sure they don't accidentally end up in obsdb queries that are expected to be useful. To achieve this we have the following changes:

* timing checks at level 2 now checks both counters are incrementing instead of just reading the "do you have counters" field.
* if a level 2 observation has low precision timing the tags are changed from `obs,cmb,things` to `obs,bad,bad_cmb,things,timing_issues`. Tags propagate all the way to the obsdb so that prevents the the books from being automatically categorized as cmb or cal. Similar things are done with operations
* update g3tsmurf throws errors if it finds low precision timing in complete observations
* books don't get bound automatically if low precision timing is there. they have to get flagged by hand. The by-hand time stamps are just using the linearized frame times

------------------------

Imprinter CLI is now much easier to use. Once installed on a system and with the correct environment variables set, we can just to `so-data-package platform failed` to cycle through and fix/catagorized failed book. Now we need to expand out the operations it can do beyond just fixing failed books

------------------------

Also added half the fix for ACU data missing in the drone calibration books. The bookbinder will now always expect to find HK data.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,6 +24,7 @@ setup_opts["entry_points"] = {
         "so_hardware_info = sotodlib.scripts.hardware_info:main",
         "so-metadata = sotodlib.core.metadata.cli:main",
         "so-site-pipeline = sotodlib.site_pipeline.cli:main",
+        "so-data-package = sotodlib.io.imprinter_cli:main",
         "toast_so_sim = sotodlib.toast.scripts.so_sim:cli",
         "toast_so_map = sotodlib.toast.scripts.so_map:cli",
         "toast_so_sat_transfer = sotodlib.toast.scripts.so_sat_transfer:cli",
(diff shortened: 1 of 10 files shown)


guanyilun (approved):
looks good to me!

PR #824, opened 2024-04-30
Adds the ability to overwrite existing fields.

msilvafe:
Currently if you try to wrap a field into an axis manager which already exists it will fail, this adds an overwrite argument which will write over the existing data in that field if true.

### DIFF
--- a/sotodlib/core/axisman.py
+++ b/sotodlib/core/axisman.py
@@ -553,7 +553,6 @@ class AxisManager:
     # Add and remove data while maintaining internal consistency.
 
     def wrap(self, name, data, axis_map=None,
-             accept_new=True, accept_merge=True,
              overwrite=False):
         """Add data into the AxisManager.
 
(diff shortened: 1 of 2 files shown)


sotodlib/core/axisman.py
```diff
@@ -553,7 +553,8 @@ def concatenate(items, axis=0, other_fields='exact'):
     # Add and remove data while maintaining internal consistency.
 
     def wrap(self, name, data, axis_map=None,
-             accept_new=True, accept_merge=True):
+             accept_new=True, accept_merge=True,
```
mhasself, line 556: Hmm, accept_new and accept_merge sure don't do anything.  You can remove them.

sotodlib/core/axisman.py
```diff
@@ -572,7 +573,12 @@ def wrap(self, name, data, axis_map=None,
             giving the name of an axis already described in the
             present object, and ax is an AxisInterface object.
 
+        overwrite (bool): If True then will write over existing data
+            in field ``name`` if present.
```
mhasself, line 577: Fix indentation to match other args.

mhasself (approved):
Thanks, this is long overdue.

Couple comments -- then please rebase, squash, and force push before merging.

mhasself (changes requested):
(Changing to "request changes")

Also -- please add a quick test for this.  There's lots of existing test code for axismanager ... should only be a couple of new lines.

mhasself (approved):
Thanks!

PR #830, opened 2024-05-01
fix: add in colon in jumpfixing that disapeared

skhrg:
I guess this disappeared at some point. Thanks to @ktcrowley for noticing.


ktcrowley (approved):
thx for the quick fix!

mhasself (approved):

PR #831, opened 2024-05-01
Major bug fix, function was not wrapping to aman

chervias:
should be `not(not(wrap))` instead of `not(wrap)`. The function was not wrapping to the aman


msilvafe:
Just get rid of the if not(not(wrap)) and change if else to

if wrap == True
…
if isinstance(wrap, str):
…

chervias:
ok, now it is working.

msilvafe (approved):

PR #820, opened 2024-04-27
Documentation for the wire grid pipeline

hnakata-JP:
Probably I am a beginner user of sphinx documentation, so I want to get any kind and helpful reviews from the developer experts. Thanks first.

PR #834, opened 2024-05-06
fixed a wrapping bug with wiregrid.py

hnakata-JP:
I fixed a wrapping issue about the function of correct_wg_angle. Please check it and merge this into master.

### DIFF
--- a/docs/calibration.rst
+++ b/docs/calibration.rst
@@ -1,2 +1,108 @@
+.. py:module:: sotodlib.site_pipeline.calibration
+
+===========
+calibration
+===========
+
+General description about the calibration pipeline scripts here.
+
+
+Command line interface
+======================
+
+Usage
+-----
+
+To get calibrated angles for detectors, ``'gamma'``, users have to load an axismanger,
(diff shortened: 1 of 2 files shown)


hnakata-JP:
This is a temporal pull request. If PR #820 will soon be merged into the master branch, this is not required any longer.

hnakata-JP:
Thanks for your review, Matthew. I inserted an unwrap method before the averaging. Also, I adjusted the name of the functions to PR #820. If there are any other critical concerns, I will merge this in a day or two.

mhasself:
> I inserted an unwrap method before the averaging

That's a decent start but please check:
- what does unwrap do, if there are nans?
- is the final result guaranteed to be in the range you want (previously it was guaranteed to lie in [0, pi], but now it is not).

hnakata-JP:
Thanks for the comments. Basically, we need to provide the calibration angles within [0, pi]. I adjusted that.
And np.unwrap is safer than just using mod np.pi. Then, I combined the unwrap function and corrected with returning angles in the expected range.
Further, the documentation is not the scope of this pull request, but I added a temporal documentation though.

PR #837, opened 2024-05-09
Additional Bookbinding error handling

kmharrington:
A small but important fix for finding azimuth / hk data (and additionally error raising for if it can't be found). 

Also add more named errors as a setup for future catching of named errors


guanyilun (approved):
I don't see any problems with this!

jlashner (approved):

Issue #798, opened 2024-04-09
Slow csr_array padding in preprocess


skhrg:
When running preprocess on LAT data I get the warning `SparseEfficiencyWarning: Changing the sparsity structure of a csr_matrix is expensive. lil_matrix is more efficient.`

This is coming from how the padded jump height arrays are being constructed. 
What happens right now is that a `csr_array` of all 0 and the correct shape gets made:
https://github.com/simonsobs/sotodlib/blob/83e572b37342a5f52befca90dc97cbdd1bfa5152/sotodlib/preprocess/core.py#L242
and then it gets filled with indexing like a normal array:
https://github.com/simonsobs/sotodlib/blob/83e572b37342a5f52befca90dc97cbdd1bfa5152/sotodlib/preprocess/core.py#L263

This is super inefficient for `csr_array`s (hence the warning), when I was profiling jumps I found that doing these sorts of suboptimal sparse operations could take more time than jump finding itself sometimes.

I recommend that we add a special case for `csr_array`s to `_expand` to avoid this.
Doing something along the lines of `out[k] = csr_array(v[nidx], row_idx, col_idx)` where `row_idx` and `col_idx` are derived from `oidx` should give a speedup.

PR #839, opened 2024-05-10
Faster sparse matrix expansion for preprocess

mhasself:
Random access (especially random writes) of sparse arrays is inefficient, so construct it from base data as efficiently as possible.

I also expanded the unit tests to make sure treatment of sparse arrays matches treatment of a full array.

On my test machine, with biggish arrays and 5% density, this is over 10 times faster and takes much less RAM.

### DIFF
--- a/sotodlib/preprocess/pcore.py
+++ b/sotodlib/preprocess/pcore.py
@@ -190,6 +190,29 @@ def _zeros_cls( item ):
     else:
         raise ValueError(f"Cannot find zero type for {type(item)}")
 
+def _reform_csr_array(arr, oidx, nidx, shape):
+    # Support function for _expand, to efficiently embed a csr_array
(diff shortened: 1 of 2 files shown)


msilvafe (approved):
I tested and got a ~7x speed up (but I'm sure this is a strong function of the details--the size of the array you're working with, how many entries you're expanding, etc.). [Documented line profiling here](https://simonsobs.atlassian.net/wiki/spaces/PRO/pages/413663320/TOD+Processing+preprocess+tod+profiling#Update-to-csr-array-expand).

Also thanks for adding tests and the nice description of the coo/csr array logic in the fn comments. I'm happy to merge it.

skhrg:
Just noting here that this closes #798

kmharrington:
can we merge this? would like to use it

kmharrington (approved):
Sorry if you were waiting on me to approve it.

Issue #838, opened 2024-05-10
preprocess submodule namespace collision


mhasself:
`preprocess` submodule directory contains:
```
__init__.py
core.py
preprocess_plot.py
processes.py
```

And that `__init__.py` contains `from processes import *`.  And processes.py imports sotodlib.core as core.  This makes it confusing to access the stuff in sotodlib/preprocess/core.py, because module namespace `sotodlib.preprocess.core` maps back to `sotodlib.core`.

Possible solutions:
- don't `import *`; or if you do then use a restrictive `__all__` declaration in processes.py to only pull in those names you really need (and core is not one of them).
- rename the file preprocess/core.py to something that's not already a name imported into preprocess module.

Here's an example of confusing behavior:
```
>>> from sotodlib.preprocess.core import _expand
>>> _expand
<function _expand at 0x743b7ec3ff70>

>>> from sotodlib.preprocess import core
>>> core._expand
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'sotodlib.core' has no attribute '_expand'
```

msilvafe:
I'll work on fixing this.

msilvafe:
Addressed in PR #841

PR #841, opened 2024-05-10
Preproc relpath

msilvafe:
This implements a fix to force the h5 filepath writing to always be relative to the sqlite filepath. We did not have this implemented properly before and the h5 datasets were just being writing to a file at the location where the prefect worker was running the script and filling up the docker volume. We ended up losing all of those h5 datasets when we restarted that docker so I am now catching up the satp1 and satp3 databases in my home user on this branch and then we will move things back over to the main metadata directory and turn back on prefect. Hopefully by that time we can have this addition in main.

This also addresses the `preprocess.core` package naming conflict with `sotodlib.core` brought up in issue #838.

It also fixes an issue introduced in pr #816 or #812 when we brought in tags to the querying.

I have started the satp1 and satp3 database builds running the preprocess_tod site pipeline script off of this branch and haven't had any failures so I think that's a good test of these updates.


sotodlib/site_pipeline/preprocess_tod.py
```diff
@@ -286,10 +288,11 @@ def main(
         if tot_query=="":
             tot_query="1"
 
-    for i, tag in enumerate(tags):
-        tags[i] = tags[i].lower()
-        if '=' not in tags[i]:
-            tags[i] += '=1'
+    if not(tags is None):
+        for i, tag in enumerate(tags):
```
guanyilun, line 292: tag is unused

guanyilun (approved):
The changes look good to me!

mhasself (approved):
Thanks!

PR #842, opened 2024-05-13
Improvements to toast context loading plus small fixes

tskisner:
- Add support for both collective and independent reading of data from the context.

- Add support for specifying the full detector selection dictionary.

- Add ability to apply preprocessing if a preprocessing config is specified.

- Avoid loading any data when distributing observations among process groups.

- In mlmapmaker wrapper, disable MPI in pixell if it is disabled in toast.  This prevents mpi4py imports on NERSC login nodes.

- Small fix in specifying workflow configs when running interactively.

### DIFF
--- a/sotodlib/toast/ops/load_context.py
+++ b/sotodlib/toast/ops/load_context.py
@@ -100,7 +100,9 @@ class LoadContext(Operator):
 
     bands = List(list(), help="Only load this list of band values")
 
-    dets_select = Dict(dict(), help="The full detector selection dictionary to use")
+    dets_select = Dict(
+        dict(), help="The `dets` selection dictionary to pass to get_obs()"
+    )
 
     ax_times = Unicode(
         "timestamps",
@@ -230,12 +232,17 @@ class LoadContext(Operator):
 
     def _open_context(self):
         if self.context is None:
+            # The user did not specify a context- create a temporary
+            # one from the file
             return Context(self.context_file)
         else:
+            # Just return the user-specified context.
             return self.context
 
     def _close_context(self, ctx):
         if self.context is None:
+            # We previously created a context from a file.  Close
+            # / delete that now.
             del ctx
 
     @function_timer
--- a/sotodlib/toast/ops/load_context.py
+++ b/sotodlib/toast/ops/load_context.py
@@ -4,6 +4,7 @@
 import os
 import re
 import datetime
+import yaml
 
 import numpy as np
 import traitlets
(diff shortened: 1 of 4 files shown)


sotodlib/toast/ops/load_context.py
```diff
@@ -87,6 +100,8 @@ class LoadContext(Operator):
 
     bands = List(list(), help="Only load this list of band values")
 
+    dets_select = Dict(dict(), help="The full detector selection dictionary to use")
```
keskitalo, line 103: This seems a little cryptic to me. It is not a list of detectors but a dictionary. Is each key a name of a detector set?

sotodlib/toast/ops/load_context.py
```diff
@@ -207,6 +228,16 @@ class LoadContext(Operator):
...
         super().__init__(**kwargs)
 
+    def _open_context(self):
+        if self.context is None:
+            return Context(self.context_file)
+        else:
+            return self.context
+
+    def _close_context(self, ctx):
+        if self.context is None:
```
keskitalo, line 238: Is this sufficient or should you set `self.context` to `None`?

sotodlib/toast/ops/load_context.py
```diff
@@ -367,7 +464,7 @@ def _exec(self, data, detectors=None, **kwargs):
 
             focalplane = toast.instrument.Focalplane(
                 detector_data=det_props,
-                sample_rate=rate,
+                sample_rate=1.0 * u.Hz,
```
keskitalo, line 467: Is this a leftover debugging setting?

keskitalo (approved):
Looks good. Please have a look at the small questions.

sotodlib/toast/ops/load_context.py
tskisner: This dictionary is passed [directly to the Context method](https://github.com/simonsobs/sotodlib/blob/8c47e388f65841e4bc7b5af16b320100324970f3/sotodlib/core/context.py#L223).  I'll add a note in the trait help.

sotodlib/toast/ops/load_context.py
tskisner: There are 2 traits at play here, `context` and `context_file`.  The user specifies one but not both.  If `context` is set to a pre-existing context instance, then "open" just returns that, else it instantiates a context from `context_file` and returns.  Similarly, if `context` was given, we don't try to close it.  But if we created it then we delete it.

sotodlib/toast/ops/load_context.py
tskisner: No, this is an optimization.  When we create the observation, we do not yet have timestamps (only the number of samples).  So we cannot compute the sample rate yet without reading through the entire dataset (which is what was happening before in the code).  Instead, we set the focalplane to a fake sample rate and later (after the timestamps are loaded) we go back and fix the sample rate.

sotodlib/toast/ops/load_context.py
tskisner: Ok, trait help is reworded.

PR #833, opened 2024-05-03
Act sim

keskitalo:
This is companion PR to https://github.com/hpc4cmb/toast/pull/754

PR #843, opened 2024-05-13
Add sample-diff white noise estimates to workflow tools

tskisner:
There are two instances of this operator:  one for flagging outliers prior to any processing, and one for quick noise estimation prior to mapmaking.


tskisner:
@keskitalo, let's reconcile this work with the changes you have in #833.

keskitalo (approved):
Awesome! Glad to see this operator in the SO workflows.

PR #844, opened 2024-05-13
Add a new Splits operator that runs the mapmaker over each split

tskisner:
- Implement 3 trivial splits to start ("all", "left_going", "right_going").

- An interval list and per-detector cuts are toggled to implement each split.


keskitalo (approved):
An elegant solution.

PR #846, opened 2024-05-15
Add an ACT-style readout filter to TOAST workflows

keskitalo:


tskisner (approved):
The code looks fine, but if this is ACT-specific, perhaps we could put this in `proc_act.py`?

keskitalo:
Let me double-check with Matthew. I think there might be something similar in SO contexts.

PR #848, opened 2024-05-15
Add analytic bandpasses to real data

keskitalo:


tskisner (approved):
Thanks, this is timely as we make simulations from real observations

PR #849, opened 2024-05-15
Add wrappers for TOAST deglitch and jumpcorrect

keskitalo:


tskisner (approved):
Thanks, useful to be able to include these toast operators in our workflows.

PR #850, opened 2024-05-15
Neaby source db

tterasaki:
This makes a database which contains flags to indicate whether the obs_id hit radio point sources, including non-planet sources such as RCW38 and galactic center.

An example of config file looks like below:
```
context_file: '/so/metadata/satp3/contexts/use_this.yaml'
query_text: 'type == "obs"'
min_ctime: null #1700000000
max_ctime: null
#update_delay: 1
query_tags: null
output_dir: '/so/home/tterasaki/code_dev/make_nearby_sources_db/nearby_sources'
distance: 2.

source_list:
 - 'mercury'
 - 'venus'
 - 'mars'
 - 'jupiter'
 - 'saturn'
 - 'uranus'
 - 'neptune'
 - ['taua', 83.6272579, 22.02159891]
 - ['rcw38', 134.7805107, -47.50911231]
 - ['iras16183-4958', 245.5154, -50.09168292]
 - ['iras19078+0901', 287.5575891, 9.107188994]
 - ['rcw122', 260.0339538, -38.95673421]
 - ['cenA', 201.3625336, -43.00797508]
 - ['3c279', 194.0409868, -5.79174024]
 - ['3c273', 187.2775626, 2.053532671]
 - ['G025.36-00.14', 279.5264042, -6.793169326]
 - ['QSO_J2253+1608', 343.4952422, 16.14301323]
 - ['galactic_center', -93.5833, -29.0078]
```

PR #856, opened 2024-05-16
New query style for obsdb which enables to use external database

tterasaki:
new query style for obsdb which enables to use external database.

### DIFF
--- a/sotodlib/core/metadata/obsdb.py
+++ b/sotodlib/core/metadata/obsdb.py
@@ -17,6 +17,101 @@ TABLE_DEFS = {
     ],
 }
 
+def _generate_query_components_from_tags(tags=None):
+    """
+    Generate query components from given tags.
+
+    This function creates SQL query components based on the provided tags.
+    It generates join clauses and extra fields to be included in the final query.
+
+    Args:
+        tags (list of str): Tags to include in the output; if they
+            are listed here then they can also be used in the query
+            string.  Filtering on tag value can be done here by
+            appending '=0' or '=1' to a tag name.
+
+    Returns:
+        tuple: A tuple containing:
+            - extra_fields (str): Comma-separated string of extra fields for the SELECT clause.
+            - joins (str): String of join clauses to be added to the query.
+    """
+    joins = ''
+    extra_fields = []
+    if tags is not None and len(tags):
+        for tagi, t in enumerate(tags):
+            if '=' in t:
+                t, val = t.split('=')
+            else:
+                val = None
+            if val is None:
+                join_type = 'left join'
+                extra_fields.append(f'ifnull(tt{tagi}.obs_id,"") != "" as {t}')
+            elif val == '0':
+                join_type = 'left join'
+                extra_fields.append(f'ifnull(tt{tagi}.obs_id,"") != "" as {t}')
+                query_text += f' and {t}==0'
+            else:
+                join_type = 'join'
+                extra_fields.append(f'1 as {t}')
+            joins += (f' {join_type} (select distinct obs_id from tags where tag="{t}") as tt{tagi} on '
+                      f'obs.obs_id = tt{tagi}.obs_id')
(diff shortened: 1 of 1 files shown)


tterasaki:
I think it is ready to go.

tterasaki:
Thank you, Matthew.
If you want to test, you can use:
`computeXX: /so/metadata/satp1/manifests/pwv_class/satp1_pwv_class_240516m/pwv_class.sqlite`

sotodlib/core/metadata/obsdb.py
```diff
@@ -253,69 +349,128 @@ def get(self, obs_id=None, tags=None, add_prefix=''):
...
+                for i, subdb_info in enumerate(subdbs_info_list):
+                    assert isinstance(subdb_info, dict)
+                    if 'filepath' not in subdb_info.keys():
+                        raise InputError(f'subdb_info does not have "filepath" in keys')
+                    filepath = subdb_info['filepath']
+                    alias = f'subdb{i}'
+                    aliases.append(alias)
+                    attach = f"ATTACH DATABASE '{filepath}' AS '{alias}'"
+                    cursor = cursor.execute(attach)            
+                    _extra_fields_sub, _join_sub, _query_sub = _generate_query_components_from_subdb(filepath=filepath,
```
mhasself, line 446: Please reduce line lengths here.

You could pass `**subdb_info` instead of all these .get(...)?

sotodlib/core/metadata/obsdb.py
```diff
@@ -253,69 +349,128 @@ def get(self, obs_id=None, tags=None, add_prefix=''):
...
+                            subdbs_info_list=[subdb_info])
+
+            This queries observations with a start_time greater than 1700000000, 
+            a tag of planet, a median pwv smaller than 2.0 [mm], and an rms of pwv
+            smaller than 0.1 [mm].
+
+            If you do not know the parameters in the sub-database, you can view the params 
+            as long as it is a ManifestDb like below::
+
+                from sotolib.core import metadata
```
mhasself, line 417: `sotodlib.core`

mhasself (changes requested):
Sorry it took me so long to get to this.  This is a really useful capability to have!

My requests for now are:
- deal with too-long lines (marked inline)
- squash and rebase

In the longer term I think we should update the interface so that:
- Extend ObsDb so that 'attach' can occur on the base ObsDb object, allowing access to merged fields in any request (including `.get`), without passing subdb_info to each call
- Extent Context so it can attach obsdb extensions automatically.
- Include more description of this feature in the docs.

But we don't need to deal with that stuff now.

davidvng:
In combination with #850, I added handling for querying an external ManifestDB in which a field, in this case `coverage` has a string value that is a list comma-separated strings in SQLite format. I added the translation of the query `'<substring> in coverage'` to become `'coverage LIKE "%<substring>%"'` to be understood by SQL as "any string that contains this substring"
I added a test to the test script and also tested on a real ManifestDB I created from #850 
In that database there is the following entry:
```
[{'_id': 1,
  'filename': 'preprocess_obs_sourcetags_archive_v2.h5',
  'obs:obs_id': 'obs_1721381147_satp1_1111111',
  'dataset': 'obs_1721381147_satp1_1111111',
  'coverage': 'saturn:ws0,saturn:ws4,saturn:ws5,neptune:ws4,neptune:ws5,QSO_J2253+1608:ws5'}]
```
I can successfully retrieve that entry with the following query:
```
r0 = obsdb.query(
                    subdbs_info_list = [{'filepath': extdb_path,
                                         'query_list': ['saturn:ws0 in coverage'],
                                         'params_list': ['coverage']}]
                   )
```

mhasself:
Hi @davidvng  and @tterasaki  -- who is going to deal with my requested changes to the original PR?

@davidvng  On the "coverage" changes -- those seem fine but you need to clean up the testing function (remove prints, restore mpi check).

PR #857, opened 2024-05-16
pointing_model: sat_v1 begins

mhasself:
Implements encoder offsets, base_tilt, and focal plane center of boresight rotation.

Given our current pointing products (measured at boresight=0), a model that specifies "bs_xi0" and "bs_eta0" should be able to cancel the pointing shifts in SATP1 arising from boresight != 0.


elle-shaw:
Hi Matthew I can take some more time to go over this in the next few days, one small comment so far --- only about the docstring on apply_pointing_model line 68
"Pass boresight=False to not store the result in tod." should say "Pass wrap=False..."

elle-shaw (approved):

PR #858, opened 2024-05-16
Change recency criterion for update_obsdb

AERAdler:
Look at timestamp of books instead of date of last modification of files.


mhasself (approved):
Thanks!

PR #862, opened 2024-05-17
hwp_metadata delete dets axis

ykyohei:
Delete dets axis from hwp_angles metadata, to fix the issue that the all the detectors get eliminated when we load data with hwp_angles metadata.

Actually dets axis was always included in hwp metadata. This became a problem because we changed the timestamp loading from ctx.get_obs(obs_id, no_signal=True) to ctx.get_obs(obs_id, dets=[]) in https://github.com/simonsobs/sotodlib/pull/809

### DIFF
--- a/sotodlib/hwp/g3thwp.py
+++ b/sotodlib/hwp/g3thwp.py
@@ -961,7 +961,6 @@ class G3tHWP():
 
         """
 
-        # write solution, metadata loader requires a dets axis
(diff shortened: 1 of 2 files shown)


ykyohei:
Yes, I tested by obs_1715427466_satp1_1111111.
I locally remade the metadata for this observation and tested to load the data by modifying the hwp_metadata of use_this.yaml. I confirmed that this issue is resolved.

PR #863, opened 2024-05-17
Change order of steps in nominal workflows to be more consistent

tskisner:
Proposed changes to the order of steps in `toast_so_sim` and `toast_so_map`.


sotodlib/toast/scripts/so_sim.py
```diff
@@ -70,15 +71,21 @@ def simulate_data(job, otherargs, runargs, comm):
             toast.ops.Reset(detdata=[defaults.det_data])
         # Append a weather model
         wrk.append_weather_model(job, otherargs, runargs, data)
-
+        # Before running simulations based on real data, we need
+        # a starting noise estimate
+        wrk.diff_noise_estimation(job, otherargs, runargs, data)
```
keskitalo, line 76: This needs to happen before zeroing the loaded data.

keskitalo (approved):
Only one issue about the order of operations.

sotodlib/toast/scripts/so_sim.py
tskisner: Done, and the load-or-simulate logic is now extracted as we discussed offline.

tskisner:
I think this PR can be safely merged, and then the data loading performance improvements can go in a separate PR.

keskitalo (approved):
Approved again.

PR #865, opened 2024-05-18
tod_ops.filters: restore iir_filter params as dict

mhasself:
Also minor code simplifications and tests for both the dict interface and the nested parameter support.


keskitalo (approved):
Awesome! Thanks a lot for the rapid turnaround.

keskitalo:
@tterasaki is it OK to merge this PR? I cannot merge #846 without it.

PR #866, opened 2024-05-18
Add venn package dependency

davidvng:
Adds `venn` package dependency to extras `sotodlib[site_pipeline]`, used in `plot_det_bias_flags`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -66,6 +66,7 @@ setup_opts["install_requires"] = [
 setup_opts["extras_require"] = {
     "site_pipeline": [
         "influxdb",
+        "venn"
     ],
     "tests": [
         "socs",


msilvafe (approved):

tskisner:
If this is just for one plotting utility used by a few people, maybe just put a try / except around the import and tell the user to install it?  More broadly, do we have any policy for introducing new dependencies @mhasself ? This particular package may be fine / easy to install, but it is a slippery slope sometimes.  Again, maybe this package is critical, but if it is an optional dependency then maybe just raise an exception if not importable and tell the user to install it.

davidvng:
This dependency is needed for prefect to run the script

skhrg:
If we want optional dependencies why not just do that in the packaging so you do `pip install sotodlib[prefect]` or `pip install sotodlib[site-pipeline]` to include any optional ones we need for that stuff but not the stuff the average user is interacting with?

tskisner:
Right, this is not a dependency of prefect, and it looks like `venn` was last updated 6 years ago (is it still maintained?), it looks like we only need it in [preprocess_plot.py](https://github.com/simonsobs/sotodlib/blob/master/sotodlib/preprocess/preprocess_plot.py) and those functions are not imported by default in [`preprocess.__init__.py`](https://github.com/simonsobs/sotodlib/blob/master/sotodlib/preprocess/__init__.py).  So this package is only needed when running prepocessing plots, which only happens in a limited set of use cases.  I definitely like @skhrg 's suggestion to add an extras_require entry for this kind of optional dependency.

PR #869, opened 2024-05-23
New functions for fft_ops

17-sugiyama:
This PR suggests adding `calc_masked_psd` and `calc_binned_psd` to fft_ops.
`calc_masked_psd` masks hwpss peaks and other peaks in the given PSD. This enables applying 1/f curve fitting to data with hwpss.
`calc_binned_psd` puts given PSD into bins. This makes `fit_noise_model` work much faster.
Here is a demonstration  for new functions.
https://github.com/17-sugiyama/scratch/blob/main/20250114_fft_ops_PR.ipynb


kmharrington (commented):
I'm noticing this PR brings in masked numpy arrays, I think for the first time in sotodlib. I'm worried that will break our IO setup going into and out of hdf5 files. 

Is there a reason we can't use Ranges / RangesMatrices to pass around masks here? 

I would also like the `calc_masked_psd` to accept more general masks. The function would be much more useful if it was generalized beyond HWPSS lines. Ex: what if I also want to mask PTC lines?

17-sugiyama:
I was using MaskedArray for no reason, so I modified the functions not to use it. 
The change also made the calculation faster. Thanks for your comments!
It's not apparent but the function `calc_masked_psd` has an option to mask a single peak like PTC lines.
`fft_ops.calc_masked_psd(aman, peak=True, peak_freq=peak_freq, peak_width=peak_width)` will do that.
Currently `peak_freq` accepts single frequency, so we need to run the code several times to mask multiple peaks.

sotodlib/hwp/hwp.py
```diff
@@ -524,8 +524,10 @@ def demod_tod(aman, signal_name='signal', demod_mode=4,
         the demodulated signal imaginary component filtered with `lpf` and multiplied by 2.
 
     """
+    if hwp_angle is None:
+        hwp_angle = aman.hwp_angle
     # HWP speed in Hz
-    speed = (np.sum(np.abs(np.diff(np.unwrap(aman.hwp_angle)))) /
+    speed = (np.sum(np.abs(np.diff(np.unwrap(hwp_angle)))) /
             (aman.timestamps[-1] - aman.timestamps[0])) / (2 * np.pi)
```
tterasaki, line 532: Can you implement below and use here and elsewhere?

```Python
def get_hwp_freq(timestamps=None, hwp_angle=None):
    hwp_freq = (np.sum(np.abs(np.diff(np.unwrap(hwp_angle)))) /
            (timestamps[-1] - timestamps[0])) / (2 * np.pi)
    return hwp_freq
```

sotodlib/tod_ops/fft_ops.py
```diff
@@ -338,6 +339,142 @@ def neglnlike(params, x, y):
...
+        p_mask: MaskedArray of the input PSD. p_mask.data returns the PSD before masking,
+            while p_mask.mask returns the bool array of the mask.
+    """
+    if f is None or pxx is None:
+        f = aman.freqs
+        pxx = aman.Pxx
+    if hwpss:
+        pxx_masked = []
+        if hwp_freq is None:
+            hwp_freq = np.median(aman['hwp_solution']['raw_approx_hwp_freq_1'][1])
```
tterasaki, line 405: Let's use the code I suggested above.

sotodlib/tod_ops/fft_ops.py
```diff
@@ -457,3 +606,105 @@ def fit_noise_model(
     if merge_fit:
         aman.wrap(merge_name, noise_fit_stats)
     return noise_fit_stats
+
+def hwpss_mask(f, hwp_freq, f_max=100, width_for_1f=(-0.4, +0.6), width_for_Nf=(-0.2, +0.2)):
```
tterasaki, line 610: I think we can implement it more straight forward.
Like
```Python
def get_mask_for_hwpss(freq, hwp_freq, modes=[1,2,3,4,5,6], width=0.4):
    if isinstance(width, float):
         "apply (-width/2, width/2) for all nf"
    elif isinstance(width, [np.array, list, tuple]):
         width = np.array(width)
         if len(width.shape) == 1:
             "apply symmetric mask for each nf"
         elif len(width.shape) = 2:
             "apply asymmetric mask for each nf"
```

sotodlib/tod_ops/fft_ops.py
```diff
@@ -457,3 +606,105 @@ def fit_noise_model(
...
+            (hwp_freq*N - 0.2) < f < (hwp_freq*N + 0.2).
+
+    Returns
+    -------
+        mask: Boolean array to mask frequency and power of the given PSD. 
+            False in this array stands for the index of hwpss to mask.
+    """
+    mask_arrays = [((f < hwp_freq + width_for_1f[0])|(f > hwp_freq + width_for_1f[1]))]
+    for n in range(int(f_max//hwp_freq-1)):
+        mask_arrays.append(((f < hwp_freq*(n+2) + width_for_Nf[0])|(f > hwp_freq*(n+2) + width_for_Nf[1])))
```
tterasaki, line 640: Let's use `get_mask_for_single_peak` defined below.

sotodlib/tod_ops/fft_ops.py
```diff
@@ -457,3 +606,105 @@ def fit_noise_model(
...
+    mask = np.all(np.array(mask_arrays), axis=0)
+    return mask
+
+def peak_mask(f, peak_freq, peak_width=(-0.002, +0.002)):
```
tterasaki, line 644: I think `get_mask_for_single_peak` is more straightforward naming.

sotodlib/tod_ops/fft_ops.py
```diff
@@ -457,3 +606,105 @@ def fit_noise_model(
...
+
+    Returns
+    -------
+        mask: Boolean array to mask the given PSD. 
+            False in this array stands for the index of the single peak to mask.
+    """
+    mask = (f < peak_freq + peak_width[0])|(f > peak_freq + peak_width[1])
+    return mask
+
+def binning_psd(psd, unbinned_mode=10, base=2, drop_nan=False):
```
tterasaki, line 668: Why don't we get errorbars of each bin?

17-sugiyama:
Thank you so much for your comments!
Besides applying your comments, I modified several things.

- [calc_pds_mask](https://github.com/simonsobs/sotodlib/blob/3c96965d55357e753c4fcb3f9445217a46fca1fa/sotodlib/tod_ops/fft_ops.py#L385) has a new option, `mode`. If `mode==replace`, it will calculate new mask array. If `mode==add`, newly calculated mask will be stacked on the existing one.
- [fit_noise_model](https://github.com/simonsobs/sotodlib/blob/3c96965d55357e753c4fcb3f9445217a46fca1fa/sotodlib/tod_ops/fft_ops.py#L546) has a new option, `fixed_parameter`. You can declare a parameter to fix during the PSD fitting.
- [fit_binned_noise_model](https://github.com/simonsobs/sotodlib/blob/3c96965d55357e753c4fcb3f9445217a46fca1fa/sotodlib/tod_ops/fft_ops.py#L724) is a new function to make PSD fittings much faster. besides knee frequency and white noise (fit parameters) are unbiased, the calculation becomes more than 20 times faster.

The demonstration of new features locates here.
https://github.com/17-sugiyama/scratch/blob/main/20240705_fft_ops_PR.ipynb

17-sugiyama:
Extended neglnlike function to apply to binned PSD.
This made fit results of binned and unbinned PSDs almost identical.
https://github.com/simonsobs/sotodlib/blob/9575bf99651e2781c99e4ea5aaa8ca393bfafa39/sotodlib/tod_ops/fft_ops.py#L360-L365

sotodlib/tod_ops/fft_ops.py
```diff
@@ -323,41 +334,201 @@ def calc_wn(aman, pxx=None, freqs=None, low_f=5, high_f=10):
...
+    elif 'wn' in fixed_param.keys():
+        if len(params)==2:
+            wn = fixed_param['wn']
+            fknee, alpha = params[0], params[1]
+        else:
+            raise ValueError('The number of fit parameters are invalid.')
+            return
+    elif len(fixed_param)==0:
+        if len(params)==3:
+            fknee, wn, alpha = params[0], params[1], params[2]
```
tterasaki, line 359: Can you modify the order of parameters?
I think the order of [wn, fknee, alpha] is more natural and less confusing.

sotodlib/tod_ops/fft_ops.py
```diff
@@ -413,38 +600,96 @@ def fit_noise_model(
...
+    # binning
+    if binning == True:
+        f, pxx, bin_size = get_binned_psd(aman, f=f, pxx=pxx, unbinned_mode=unbinned_mode,
+                                          base=base, merge=False)
     fitout = np.zeros((aman.dets.count, 3))
     # This is equal to np.sqrt(np.diag(cov)) when doing curve_fit
     covout = np.zeros((aman.dets.count, 3, 3))
-    for i in range(aman.dets.count):
+    if isinstance(fknee_est, (int, float)):
+        fknee_est = np.zeros(aman.dets.count)+fknee_est
```
tterasaki, line 627: `np.full` is a bit smarter.

tterasaki (changes requested):
Ok, you are almost there to get approved with minor my comments.
Another thing I want to ask you to add is a unit test for the fitting function.

I have already written basic component for the unit test like below.
Can you add the unit test using those at https://github.com/simonsobs/sotodlib/tree/fft_ops_hwpss_mask/tests ?
https://github.com/simonsobs/sotodlib/blob/fft_ops_hwpss_mask/tests/test_subpolyf.py (or some other tests) would be a good reference. 

```
import numpy as np
from sotodlib import core
from numpy.fft import rfftfreq,irfft
from numpy.fft import fftfreq,rfft
from scipy import interpolate
import matplotlib.pyplot as plt
from sotodlib.tod_ops import fft_ops

def model_func(x, sigma, fk, alpha):
    return sigma**2 * (1 + (fk/x)**alpha)

fs = 200.
dets = core.LabelAxis('dets', [f'det{di:003}' for di in range(20)])
nsamps = 200*3600

aman = core.AxisManager(dets)
ndets = aman.dets.count

white_noise_amp_array_input = 50 + np.random.randn(ndets) #W/sqrt{Hz}
white_noise_power_array_input = white_noise_amp_array_input**2 #W^2/Hz
fknee_array_input = 1 + 0.1*np.random.randn(ndets)
alpha_array_input = 3 + 0.2*np.random.randn(ndets)

freqs = rfftfreq(nsamps, d=1/fs)
pxx_input = model_func(freqs, 
           sigma=white_noise_amp_array_input[:, np.newaxis], 
           fk=fknee_array_input[:, np.newaxis],
           alpha=alpha_array_input[:, np.newaxis])

pxx_input[:, 0] = 0
nusamps_input = core.OffsetAxis('nusamps_input', len(freqs))

aman.wrap('freqs_input', freqs, [(0, nusamps_input)])
aman.wrap('wnl_input', white_noise_amp_array_input, [(0, 'dets')])
aman.wrap('fk_input', fknee_array_input, [(0, 'dets')])
aman.wrap('alpha_input', alpha_array_input, [(0, 'dets')])
aman.wrap('Pxx_input', pxx_input, [(0, 'dets'), (1, 'nusamps_input')])

T = nsamps/fs
ft_amps = np.sqrt(pxx_input * T * fs**2 / 2)

ft_phases = np.random.uniform(0, 2*np.pi, size=ft_amps.shape)
ft_coefs = ft_amps * np.exp(1.0j*ft_phases)
realized_noise = irfft(ft_coefs)
timestamps = 1700000000 + np.arange(0, realized_noise.shape[1])/fs
aman.wrap('timestamps', timestamps, [(0, core.OffsetAxis('samps', len(timestamps)))])
aman.wrap('signal', realized_noise, [(0, 'dets'), (1, 'samps')])
```

17-sugiyama:
I've reflected Tomoki's latest comments and added the unittest for fft_ops.
It seems that the tests failed for something not related to my changes...

tterasaki (dismissed):

sotodlib/tod_ops/fft_ops.py
```diff
@@ -459,45 +649,115 @@ def fit_noise_model(
...
+        if fixed_param == "wn":
+            initial_params = np.array([fknee_est, alpha_est])
+            fixed = wn_est
+        if fixed_param == "alpha":
+            initial_params = np.array([wn_est, fknee_est])
+            fixed = alpha_est
+
+        for i in range(len(pxx)):
             p = pxx[i]
-            wnest = np.median(p[((f > fwhite[0]) & (f < fwhite[1]))])
```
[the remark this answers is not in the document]
17-sugiyama: We have removed polyfit to reduce the calculations. 
White noise values of each detectors can be estimated with `calc_wn`.
I have confirmed that the initial parameter input of white noise is enough to perform fits as good as the polyfit version.

```
wn_initial_val = fft_ops.calc_wn(aman)
noise_fit_stats = fft_ops.fit_noise_model(aman, wn_est=wn_initial_val)
```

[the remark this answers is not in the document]
mmccrackan: Thanks for the explanation!

mmccrackan:
I wanted to revive this PR as we get set up to run ISO v2 and as we look to solve performance issues with the current noise fitting implementation.

I tested the new functionality for both a single scan and for subscans and find that it seems pretty robust.  Namely, I think setting an upper frequency limit of around 25 Hz (there typically a drop off in the region of 50 Hz that will drag the white noise down if included) and enabling the masking with the default parameters seems to recover many of the fits that failed with the older versions of the functions.  I explored a few other fitting methods outside of `Nelder-Mead,` but they seemed not to be as robust to features in the PSD.

I think one thing that needs to be added here is an update to the [preprocess noise class](https://github.com/simonsobs/sotodlib/blob/2a61e53e96b8d137b0322c56815709f249d35b82/sotodlib/preprocess/processes.py#L473) that will calculate the white noise (or use it if it exists) for `wn_est` if 'wn' is fixed.  The calculated white noise value (though biased: #995) seems to be a good estimate and fixing the white noise to that value seems to help a lot.  Otherwise, if I'm reading this correctly, when using a fixed 'wn', we will only be able to set all detectors and obsids to a single value when running through `preprocess_tod` (since one config file is used for all).

msilvafe:
> I wanted to revive this PR as we get set up to run ISO v2 and as we look to solve performance issues with the current noise fitting implementation.
> 
> I tested the new functionality for both a single scan and for subscans and find that it seems pretty robust. Namely, I think setting an upper frequency limit of around 25 Hz (there typically a drop off in the region of 50 Hz that will drag the white noise down if included) and enabling the masking with the default parameters seems to recover many of the fits that failed with the older versions of the functions. I explored a few other fitting methods outside of `Nelder-Mead,` but they seemed not to be as robust to features in the PSD.
> 
> I think one thing that needs to be added here is an update to the [preprocess noise class](https://github.com/simonsobs/sotodlib/blob/2a61e53e96b8d137b0322c56815709f249d35b82/sotodlib/preprocess/processes.py#L473) that will calculate the white noise (or use it if it exists) for `wn_est` if 'wn' is fixed. The calculated white noise value (though biased: #995) seems to be a good estimate and fixing the white noise to that value seems to help a lot. Otherwise, if I'm reading this correctly, when using a fixed 'wn', we will only be able to set all detectors and obsids to a single value when running through `preprocess_tod` (since one config file is used for all).

Can you add the update to the preprocess method required to allow using the wn_est to this PR and add to the example config entries in the docstring the typical values that you think work well?

mmccrackan:
> Can you add the update to the preprocess method required to allow using the wn_est to this PR and add to the example config entries in the docstring the typical values that you think work well?

Okay updated the preprocess class.  When fitting, you can pass in `wn_est` which is either a noise axis manager (raises ValueError out if not found) with `white_noise` in it or a `float`.  These will be used for the initial guess of white noise and for the fixed value of white noise if `fixed_param` is passed with the value `'wn'`.  If `wn_est` is not passed and `'wn'` is passed for `fixed_param`, it will call `calc_wn` and use those as the fixed values of white noise in the fit.  I made it only do `calc_wn` when fitting when the white noise is fixed so that we aren't always re-running `calc_wn` if we don't pass estimates, but we can change this if desired.

I added separate limits for the white noise frequencies (`wn_f_low` and `wn_f_high`; defaults to 5 and 10 as per `calc_wn` defaults) in the config since we may call `calc_wn` when fitting and we're going to want to use a restricted frequency range for the white noise.

sotodlib/tod_ops/fft_ops.py
```diff
@@ -365,44 +376,202 @@ def calc_wn(aman, pxx=None, freqs=None, low_f=5, high_f=10):
...
+    if pxx is None:
+        pxx = aman.Pxx
+            
+    f_bin, bin_size = log_binning(f, unbinned_mode=unbinned_mode, base=base, return_bin_size=True)
+    pxx_bin = log_binning(pxx, unbinned_mode=unbinned_mode, base=base, return_bin_size=False)
+
+    if merge:
+        aman.merge(core.AxisManager(core.OffsetAxis("nusamps_bin", len(f_bin))))
+        if overwrite:
+            if "freqs_bin" in aman._fields:
```
iparask, line 533: The same as my previous comment. :)

msilvafe (approved):
Looks good, thanks for getting this across the finish line. Sorry it sat around for so long.

PR #873, opened 2024-05-29
Preprocess methods to match processing in SAT Mapmaker

msilvafe:
Working to get new preprocess methods so that all processing done by @chervias and @susannaaz in the SAT mapmaking effort can be performed in a preprocess pipeline. I'm following the steps in this function: https://github.com/simonsobs/sotodlib/blob/0e1ce8bc6609f78c938966fef5adfa2468de21e9/sotodlib/site_pipeline/make_filterbin_map.py#L325

### DIFF
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -442,6 +442,7 @@ class Calibrate(_Preprocess):
       - name: "calibrate"
         process:
           kind: "single_value"
+          divide: True # If true will divide instead of multiply.
           # phase_to_pA: 9e6/(2*np.pi)
           val: 1432394.4878270582
       - name: "calibrate"
@@ -459,11 +460,21 @@ class Calibrate(_Preprocess):
 
     def process(self, aman, proc_aman):
         if self.process_cfgs["kind"] == "single_value":
-            aman[self.signal] *= self.process_cfgs["val"]
+            if self.process_cfgs.get("divide", False):
+                aman[self.signal] /= self.process_cfgs["val"]
+            else:
+                aman[self.signal] *= self.process_cfgs["val"]
         elif self.process_cfgs["kind"] == "array":
             field = self.process_cfgs["cal_array"]
             _f = attrgetter(field)
-            aman[self.signal] = np.multiply(aman[self.signal].T, _f(aman)).T
+            if self.process_cfgs.get("proc_aman_cal", False):
+                cal_arr = _f(proc_aman)
+            else:
+                cal_arr = _f(aman)
+            if self.process_cfgs.get("divide", False):
+                aman[self.signal] = np.divide(aman[self.signal].T, cal_arr).T
+            else:
+                aman[self.signal] = np.multiply(aman[self.signal].T, cal_arr).T
         else:
             raise ValueError(f"Entry '{self.process_cfgs['kind']}'"
                               " not understood")
@@ -630,7 +641,13 @@ class Demodulate(_Preprocess):
     name = "demodulate"
 
     def process(self, aman, proc_aman):
-        hwp.demod_tod(aman, **self.process_cfgs)
+        hwp.demod_tod(aman, **self.process_cfgs["demod_cfgs"])
+        if self.process_cfgs.get("trim_samps"):
+            trim = self.process_cfgs["trim_samps"]
+            aman.restrict('samps', (aman.samps.offset + trim,
+                                    aman.samps.offset + aman.samps.count - trim))
+            proc_aman.restrict('samps', (aman.samps.offset + trim,
+                                         aman.samps.offset + aman.samps.count - trim))
 
 
 class EstimateAzSS(_Preprocess):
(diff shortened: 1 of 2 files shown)


msilvafe:
To test the ``preproc_or_load_mpi`` and ``cleanup_mandb_mpi`` functions I ran in a script that's a stripped down version of the mapmaker found[ here in the pwg-scripts repo](https://github.com/simonsobs/pwg-scripts/blob/preproc_profiling/pwg-fct/tests/sat_process_pipe/test_mpi.py). I ran on NERSC with a tod list that had just 4 obs-ids and just processed one-wafer and one-bandpass in order to run the test script quickly. I confirmed that I could run the script both w/ and w/o the database entries present to confirm the "either generate db or load from existing db" functionality. 

I tested all of the new preprocess methods on both SATp1 and SATp3 data in the notebook: `NERSC: /global/u1/m/msilvafe/SO_FLP/20240529_Test_New_Preprocess.ipynb`

chervias (commented):
Looks great in general. One thing I wonder is what happens when an MPI job is interrupted for some reason, and you never get to the stage where you run `cleanup_mandb_mpi `. In this case you would run this function on its own after to preserve your progress. Can you do that without having errors and outputs lists? Can you easily recreate them from the written partial files?

sotodlib/site_pipeline/preprocess_tod.py
```diff
@@ -120,9 +126,196 @@ def swap_archive(config, fpath):
         os.makedirs(dname)
     return tc
 
+def preproc_or_load_mpi(obs_id, configs, dets, logger=None, 
+                        context=None, overwrite=False):
```
mhasself, line 130: I think you should downplay the "MPI" in this function name, and docstring.  Ok to mention it, but it's not really an essential to this function.  Lots of callers, parallel or not, might want to preproc_or_load for one of the standard logical groupings of dets.

So, just "preproc_or_load"?  Or "preproc_or_load_group"?

sotodlib/preprocess/processes.py
```diff
@@ -978,20 +1008,42 @@ def calc_and_save(self, aman, proc_aman):
...
-        relcal[~m0] = med1/pca_signal.weights[~m0,0]
-
-        rc_aman = core.AxisManager(aman.dets, aman.samps,
-                                   core.LabelAxis(name='bandpass',
-                                                  vals=bands))
-        rc_aman.wrap('relcal', relcal, [(0,'dets')])
-        rc_aman.wrap('medians', np.asarray([med0, med1]),
-                     [(0, 'bandpass')])
-        rc_aman.wrap('pca_mode0', pca_signal.modes[0], [(0, 'samps')])
+        if len(bands) == 2:
```
mhasself, line 1011: I don't understand why there are separate blocks for 1, 2, or other number of bands.  This could all be handled in a single block, which assembles results for each band, and writes them into rc_aman.  Too much duplicated code!

sotodlib/hwp/hwp.py
```diff
@@ -454,42 +481,48 @@ def estimate_sigma_tod(signal, hwp_angle):
...
-        The HWPSS template to be subtracted from the signal. If `hwpss_template`
-        is None (default), the HWPSS template stored in the axis manager under
-        the key 'hwpss_extract' will be used.
+        The axis manager containing the signal and the HWPSS template.
+    signal_name : str, optional
+        The name of the field in the axis manager containing the signal to be processed.
+        Defaults to 'signal'.
+    hwpss_template_name : str, optional
+        The name of the field in the axis manager containing the HWPSS template.
+        Defaults to 'hwpss_model'.
```
mhasself, line 499: The change over to `..._name` is no doubt to give some calling entity an easier job of things.  But it's much less general than accepting an array, because now the hwpss_template definitely needs to be in the AxisManager. 

In other parts of the code, such arguments are handled like this:
- If it's None, use the default.
- If it's a string, look up that string in the axisman.
- If it's an array, use the array.

That supports both modes of calling.  The logic can be written up into a helper function (and has, all over the place, I think), if you have multiple args to handle this way (which you do).

So I am hesitant to pass this breaking interface change that also is a bit of a regression.  I'd rather add additional logic to handle when "hwpss_template" is passed as a string, without removing the old interface.

mhasself (changes requested):
Lots of interesting stuff here ... happy to discuss some of my comments in more detail.

sotodlib/tod_ops/filters.py
```diff
@@ -364,7 +365,8 @@ def timeconst_filter(target, freqs, tod, timeconst=None, invert=False):
     if timeconst is None:
         timeconst = 'timeconst'
     if isinstance(timeconst, str):
-        timeconst = tod[timeconst]
+        _f = attrgetter(timeconst)
+        timeconst = _f(tod)
```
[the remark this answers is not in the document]
msilvafe: > 1. Why is this needed?  if tod is an axismanager, `tod[timeconst]` is actually better than `tod.<timeconst>`

Because time constant is buried multiple layers deep inside the axis manager it's in `tod['det_cal']['tau_eff']` passing `det_cal.tau_eff` to `attrgetter` can get things 2 layers deep like this.

> 2. Normally one would just do `timeconst = getattr(tod, timeconst)` instead of these two new lines

Just because I haven't use that/don't know how it works and we had an example of `attrgetter` for this already so I copied it. I'll try it out with `getattr` instead.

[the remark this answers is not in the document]
mhasself: Ah, neat.  Indeed attrgetter can do buried things, and getattr can't.  So just a add a brief comment to explain the purpose of that construction here.

sotodlib/hwp/hwp.py
msilvafe: Ok I think I've addressed this now up to the fact that we want to fix the core issue in the tod_ops filters module.

sotodlib/preprocess/processes.py
msilvafe: I also addressed this.

sotodlib/tod_ops/filters.py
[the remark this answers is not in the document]
msilvafe: Done.

sotodlib/site_pipeline/preprocess_tod.py
msilvafe: Done.

msilvafe:
@mhasself I believe I've addressed all of your comments. @guanyilun and I also added in some changes that fixed issues with running through prefect without any memory leak errors and I fixed an issue with the binning axis names for the azss binning. I've tested it by running it through the full nominal sat cmb preproc pipeline and all is making sense/no new errors w/ recent commits.

mhasself (approved):
What a work!  Please squash on merge, unless you really need the full gory story.

PR #874, opened 2024-05-30
Adds psycopg2-binary requirement for hkdb

jlashner:
This PR adds the `psycopg2-binary` requirement, which is necessary for reading and updating postgres databases such as hkdb.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -60,7 +60,8 @@ setup_opts["install_requires"] = [
     'pixell',
     'scikit-image',
     'pyfftw',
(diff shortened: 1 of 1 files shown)


guanyilun (approved):
looks good to me! Just want to clarify that everyone running analysis will need this dependency, not just site_pipeline users?

jlashner:
Yes, this is required for anyone wanting to read from the hkdb postgres databases.

mhasself (approved):

PR #875, opened 2024-05-31
Planets minor fix

tterasaki:
Removing 
https://github.com/simonsobs/sotodlib/blob/a26283a27a8e94dfbba686239d397370659e9090/sotodlib/coords/planets.py#L306

And changed 'tau_a' to 'tauA' to keep consistency with 'cenA'


davidvng:
Can you change tau_a here as well?
https://github.com/simonsobs/sotodlib/blob/fb2c856ea022617dd99d995f8c1e0bee89ebb2ce/sotodlib/obs_ops/sources.py#L38

tterasaki:
Thanks David. I have changed the 'tau_a'.

And I modified source_flags in preprocessing a bit so that it can receive parameters other than defaults for my pointing business.
Default values are not changed.

davidvng (approved):
Looks good, thanks for the fixes!

tterasaki:
Thank you!

Issue #825, opened 2024-05-01
SAT specific TOAST workflow script


tskisner:
This is in progress in #752.  We need a SAT-specific workflow that is flexible and can be used in several ways.  Consolidating previous discussion notes from various conversations, the general flow is:

- Load or simulate observing
  - If given an observation list, load that data from a context, optionally applying preprocessing
  - If given an observing schedule, simulate that observing and simulate any configured timestream components
  - Future Note: eventually, "loading data" could involve loading already-demodulated data through a context
  - Future Note: if context metadata included information about splits, we could construct them from that data.  Initially will just use named splits based on data properties (e.g. left-going, right-going, inner, outer, etc)
- Noise model
  - If per-detector noise weights were included in context metadata, then use that
  - Else, run simple sample-diff white noise estimation
  - If dictionary of per-observation noise weights is given, load that and scale noise model by those values.  
- If given list of signal realization maps, loop over those, else just run with loaded data
  - If using signal realization, load and replace detector timestreams with that
  - If doing more than one split, copy detdata to temp name
  - If doing more than one split, copy per-detector cuts to temp name
  - Loop over splits (could be "all data")
    - Restore detdata from temp copy (if multiple splits)
    - Restore per-det cuts from temp copy (if multiple splits)
    - Apply split (which toggles per-detector cuts and sets active interval list)
    - Apply filtering
    - Make map and write products.  Either to directory hierarchy or location in HDF5 file.

Given this core workflow, we can use input configuration files to run in different modes:

1. Simple mapmaking tests with real data or realistic sim
  - Specify list of observations (real data) or sim schedule
  - No list of realizations or per obs weights
  - No list of splits (default is single split with "all data")
  - Output is single map for all data

2. Atomic maps for real or realistic simulation
  - Specify a single observation or sim schedule with one obs (i.e. run many small jobs) 
  - Specify no list of realizations or per obs noise weights
  - Specify list of splits to use
  - Outputs are atomic maps for single observation

3. Transfer function sims for some number of realizations and sub-bundle
  - Specify list of observations (could be full bundle or sub-bundle to be combined in post processing)
  - Specify list of realizations (batching these saves on expensive I/O of loading data)
  - Specify per-obs noise weights (as determined by desired weighting when combining atomic maps of real data)
  - Specify list of splits to use
  - Outputs are (sub-)bundle maps for each realization

tskisner:
[...]

PR #876, opened 2024-06-03
Add Split operator support to so_sim workflow

keskitalo:


tskisner:
I know this is a draft, but thanks for adding this to `toast_so_sim`.  Summarizing offline discussion last week:

- We should add this to `toast_so_map` as well
- We should change the name of `sotodlib.toast.workflows.sat` to `sotodlib.toast.workflows.splits`, since this is not SAT-specific
- Separate from this, I will rename the `toast_so_sat` workflow in #825 to `toast_so_transfer` and rip out all the atomic map simulation / generation features (which can be made with the normal workflows modified in this PR) and focus that workflow on only the transfer function sims.

tskisner (approved):
This looks good, but can we rename the `sat.py` file to `splits.py`?

PR #877, opened 2024-06-04
Detector maps2

keskitalo:
This PR replaces #690 that ended up in irreconcilable conflict with the main branch.

### DIFF
--- a/sotodlib/toast/workflows/proc_mapmaker.py
+++ b/sotodlib/toast/workflows/proc_mapmaker.py
@@ -6,6 +6,7 @@
 import numpy as np
 from astropy import units as u
 import toast
+from toast.mpi import flatten
 import toast.ops
 from toast.observation import default_values as defaults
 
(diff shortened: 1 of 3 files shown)


sotodlib/toast/workflows/proc_mapmaker.py
```diff
@@ -191,103 +193,126 @@ def mapmaker_run(job, otherargs, runargs, data, map_op):
...
+                for iobs, obs in enumerate(data.obs):
+                    log.info_rank(
+                        f"{group} : mapping observation {iobs + 1} "
+                        f"/ {len(data.obs)}.",
+                        comm=new_comm.comm_world,
+                    )
+                    # Data object that only covers one observation
+                    obs_data = data.select(obs_uid=obs.uid)
+                    # Replace comm_world with the group communicator
+                    obs_data._comm = new_comm
```
tskisner, line 246: I know this was in the original code, but I think that this only works "accidentally" and is fragile (there is a reason this is an internal variable).  The `Data` constructor assigns the internal `_comm` variable to the input `Comm` object.  However, each observation has a reference to this original `Comm` instance as well.  And every Observation shared object has references to the original sub-communicators.  If we want to support getting a subset of observations on each group where we pretend that the group communicator is the world communicator, then we should probably make that kind of change within the `Data` class where we can do that more carefully and in one place.  Again, this is no worse than the original code, so it is fine to merge, but we should implement something cleaner upstream.

tskisner (approved):
Looks fine, except for a pre-existing problem that we should fix upstream.

PR #882, opened 2024-06-06
Bookbinder: Hk address specifications

### LINKED ISSUE #880: Bookbinder should load data based on full agent address not Block name
The bookbinder currently loads data based on the Block names, which is not a reliable way to load data. This should be changed to a more reliable method. See details from @jlashner below:

> I believe I know why this is happening. The bookbinder currently loads in hk data based on the block name ([see here](https://github.com/simonsobs/sotodlib/blob/04cc1f0ac82f8489f91c25f76ac3107691cb8d1a/sotodlib/io/bookbinder.py#L92-L101)). This is wrong for a number of reasons (which I believe @mhasself has mentioned in the past but we never got around to fixing)
> 1. Block names are meant to be private to agents, and shouldn't be relied upon by downstream users
> 2. There is no checking of agent id.
>
> The fact that there is no further check of agent id means that its possible for it to read HK data from multiple agent instances of the same agent type. This doesn't matter _as much_ for ACU hk data, however there are two HWP Encoder agents both writing to the same feed name. When the bookbinder compiles the HWP freq, it is taking data from two separate agents `satp1.hwp-bbb-e1.feeds.HWPEncoder` and `satp1.hwp-bbb-e2.feeds.HWPEncoder` with independent timestamps. It is possible for these two feeds to occasionally have exactly the same timestamp, in which case the duplicate timestamps error will be thrown.
>
> Thankfully this means that there are not actually duplicate timestamps in the hk files. I've double checked the raw HK data and found this to be the case. For existing books, this only effects the avg hwp_freq stored in the M_index.yaml, and since the encoders read roughly the same value, it should not really effect existing books much.
>
> To fix this, we need to change the bookbinder so that it takes in the full addresses of the data streams it needs to save or monitor as input, instead of relying on block-names and field-names alone.

_Originally posted by @jlashner in https://github.com/simonsobs/daq-discussions/discussions/100#discussioncomment-9655510_

jlashner:
This PR changes the bookbinder so that when loading in relevant HK data, instead of relying on hard-coded block names to retrieve data, it uses full user-configured addresses of desired feeds, of the form  ```<instance-id>.<feed-name>.<field-name>```. This resolves #880.

Hk field addresses now must be specified in the imprinter configuration files, with a block such as:
```
hk_fields:
    az: acu.acu_udp_stream.Corrected_Azimuth
    el: acu.acu_udp_stream.Corrected_Elevation
    boresight: acu.acu_udp_stream.Corrected_Boresight
    az_mode:  acu.acu_status.Azimuth_mode
    # corotator_enc: acu.acu_status.Corotator_current_position # (For LAT)
    hwp_freq: hwp-bbb-e1.HWPEncoder.approx_hwp_freq
```
Mount data quality checking will occur if the mount fields are set. If not set, a warning is logged.

I've tested this on the book that was failing due to duplicate data, but since this is a fairly large change, we should test in other scenarios.


sotodlib/io/bookbinder.py
```diff
@@ -94,73 +95,156 @@ def next_scan(it):
...
+
+        # Check for duplicates
+        clean_times, idxs = np.unique(self.times, return_index=True)
+        if len(self.times) != len(clean_times):
+            if not drop_duplicates:
+                raise DuplicateAncillaryData(
+                    f"HK data from {self.addr} has" 
+                    " duplicate timestamps"
+                )
+            self.times = self.times[idxs]
```
mhasself, line 165: Don't you need to take subset of self.data here too?

sotodlib/io/bookbinder.py
```diff
@@ -316,35 +396,36 @@ def add_acu_summary_info(self, frame, t0, t1):
...
+            if np.any(m):
+                dt = np.diff(az.times[m]).mean()
+                az_vel = np.diff(az.data[m]) / dt
+                frame['azimuth_velocity_mean'] = np.mean(az_vel)
+                frame['azimuth_velocity_stdev'] = np.std(az_vel)
+
+        if el is not None:
+            m = (t0 <= el.times) & (el.times <= t1)
+            if np.any(m):
+                el_vel = np.diff(el.data[m]) / dt
```
mhasself, line 422: This relies on dt computed in the "az is not None" block.  Granted it would be difficult for only one of az and el to be None.  But that is not what the block structure suggests ... it implies they are independent.

sotodlib/io/bookbinder.py
```diff
@@ -316,35 +396,36 @@ def add_acu_summary_info(self, frame, t0, t1):
...
-
-            frame['azimuth_velocity_mean'] = np.mean(az_vel)
-            frame['azimuth_velocity_stdev'] = np.std(az_vel)
-            frame['elevation_velocity_mean'] = np.mean(el_vel)
-            frame['elevation_velocity_stdev'] = np.std(el_vel)
+                frame['azimuth_mode'] = 'Preset'  # Slewing
+
+        if az is not None:
+            m = (t0 <= az.times) & (az.times <= t1)
+            if np.any(m):
```
mhasself, line 413: The block below will crash if only one element of m is True.  Consider (`if m.sum() >= 2`).

sotodlib/io/bookbinder.py
```diff
@@ -171,7 +255,13 @@ def __init__(self, files, book_id, drop_duplicates=False, log=None):
             self.log = logging.getLogger('bookbinder')
         else:
             self.log = log
-    
+
+        if self.hkdata.az is None:
+            self.log.warning("No ACU data specified in hk_fields!")
```
kmharrington, line 260: This needs to be a bypass-able error. If I'm reading through this setup correctly then if the mount agent changes name or if the imprinter configs are configured incorrectly then we'll just quietly get books without mount data which could take awhile to catch.

sotodlib/io/bookbinder.py
```diff
@@ -171,7 +255,13 @@ def __init__(self, files, book_id, drop_duplicates=False, log=None):
...
+
+        if self.hkdata.hwp_freq is None:
+            self.log.info("No HWP Freq data is specified in hk_fields.")
```
kmharrington, line 263: while we're at it, we should have a bypass-able error for "no HWP data"

sotodlib/io/bookbinder.py
```diff
@@ -226,36 +307,35 @@ def bind(self, outdir, times, frame_idxs, file_idxs):
...
-                )
-            if np.max(np.diff( block.times[m])) > 10:
+        def validate_mount_field(hk_field: HkDataField, max_dt=None):
+            m = (times[0] <= hk_field.times) & (hk_field.times <= times[-1])
+            if not np.any(m):
                 raise NoMountData(
-                    f"Max ACU data spacing {np.max(np.diff( block.times[m]))}s" 
-                    f" is higher than 10s. Interpolation may be "
-                    "questionable."
+                    f"No mount data overlappiing with detector data: {hk_field.addr}"
```
kmharrington, line 314: typo

sotodlib/io/bookbinder.py
kmharrington: yes, fixed in a soon to update push

kmharrington:
Ok, I have made a decent number of updates based on reviews / more tests with books. Now with matching site-pipeline-configs PR https://github.com/simonsobs/site-pipeline-configs/pull/82 

I have tested books from satp1, satp3, and the LAT. Have confirmed all the ACU fields are still there. Requiring HWP data and requiring ACU data is bypassable with a flag and more warning messages make it into the logs now.

I tested the duplicates on some older books where there actually was duplicate data as well as the satp1 books that were falsely saying there were duplicates. I've also tested the satp3 books where the mount data dropped out, errors throw correctly and it can be bypassed. 

@jlashner @mhasself  take a look?

jlashner:
Hi @kmharrington, the changes look good to me! So consider this my approval in case you want to approve yourself and merge.

sotodlib/io/bookbinder.py
```diff
@@ -873,25 +1015,23 @@ def get_metadata(self, telescope=None, tube_config={}):
...
-            dt = np.diff(block.times[m]).mean()
-            az = block.data['Corrected_Azimuth']
-            az_speed = np.abs(np.diff(az[m]) / dt)
+        if az is not None:
+            m = (t0 < az.times) & (az.times <= t1)
+            dt = np.diff(az.times[m]).mean()
+            az_speed = np.abs(np.diff(az.data[m]) / dt)
             if m.any():
                 meta['az_speed_mean'] = float(np.mean(az_speed))
                 meta['az_speed_stdev'] = float(np.std(az_speed))
```
mhasself, line 1037: My only remaining concern is how these meta are computed.  Right now there are a few possibilities:
- az is None -> speed & std are None
- az is found, but m.sum() == 0 -> speed & std are None
- az is found, but m.sum() == 1 -> speed & std are nan
- az is found, but m.sum() >= 2 -> speed & std are real numbers

Those invalid cases are non-uniform; it's annoying to have to deal with None and nan downstream, without the distinction really telling me anything definitive.  So I'd really like to see "if m.sum() >= 2" instead of "if m.any()" for all of these blocks that need to do a diff and store stats.

In practice this m.sum() == 1 won't happen very often... but then when it does, it will break a bunch of code and be annoying to retrofit.

jlashner: Ah thanks for this comment, I didn't realize that this check wasn't added. I think I correct this in the latest commit. I also made it so that the per-frame summary fields are set to `np.nan` when we can't determine them, instead of just being left out.

kmharrington (approved):
Approved!

PR #883, opened 2024-06-06
G3thk trim

sanahabhimani:
Draft PR for using the g3tsmurf_hwp_config.yaml file to obtain agent instance IDs for populating the Level 2 g3thk database for data packaging.

This will work for lat, satp1, satp2, satp3 g3thk level 2 databases. Was pointed to a site pipeline config file that uses the site g3thk database, but haven't yet looked into what is necessary for level 2 site g3thk database. Hence, the draft PR.

### DIFF
--- a/sotodlib/io/g3thk_db.py
+++ b/sotodlib/io/g3thk_db.py
@@ -131,7 +131,7 @@ class HKFields(Base):
 
 
 class G3tHk:
-    def __init__(self, hkarchive_path, finalize=None, db_path=None, echo=False): #finalization
+    def __init__(self, hkarchive_path, iids, db_path=None, echo=False):
         """
         Class to manage a housekeeping data archive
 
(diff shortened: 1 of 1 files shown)


sanahabhimani:
marking it ready for review as we confirmed that nothing's really needed for level 2 site hk data for data packaging

sotodlib/io/g3thk_db.py
```diff
@@ -144,13 +147,15 @@ def __init__(self, hkarchive_path, db_path=None, echo=False):
...
+        self.iids = iids
         self.engine = db.create_engine(f"sqlite:///{db_path}", echo=echo)
         Session.configure(bind=self.engine)
         self.Session = sessionmaker(bind=self.engine)
         self.session = Session()
         Base.metadata.create_all(self.engine)
 
-    def load_fields(self, hk_path):
+
+    def load_fields(self, hk_path, iids):
```
kmharrington, line 158: why do we need to pass `iids` if those are already saved in self?

kmharrington:
These updates have now been merged through the `dpkg-revamp` branch.

PR #884, opened 2024-06-06
Delete futures as they're processed.

msilvafe:
Trying out a fix for preprocess_tod memory leak introduced with futures based on this S.E. post found by @guanyilun https://stackoverflow.com/questions/73287939/threadpoolexecutor-threads-futures-do-not-release-memory-when-completed-and-t

### DIFF
--- a/sotodlib/site_pipeline/preprocess_tod.py
+++ b/sotodlib/site_pipeline/preprocess_tod.py
@@ -427,6 +427,7 @@ def main(
                      overwrite=overwrite, run_parallel=True) for r in run_list]
         for future in as_completed(futures):
             err, src_file, db_datasets = future.result()
+            futures.remove(future)
             db = _get_preprocess_db(configs, group_by)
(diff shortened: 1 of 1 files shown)


mhasself:
@msilvafe @guanyilun  I think Yilun is making a new worker docker image ... you want this fix included?

msilvafe:
> @msilvafe @guanyilun I think Yilun is making a new worker docker image ... you want this fix included?

I think that stack_overflow post seems to imply that its the right thing to do so I would say yes but I wasn't actually able to measure the memory leak we had seen so I wasn't able to see if this fixed anything or not.

mhasself:
> I think that stack_overflow post seems to imply that its the right thing to do so I would say yes but I wasn't actually able to measure the memory leak we had seen so I wasn't able to see if this fixed anything or not.

Ya, I think this will help.  I'm going to approve this, and if indeed you've tested this (which I think you have) then please feel free to  merge (use the "rebase + merge" option!!).

mhasself (approved):

PR #885, opened 2024-06-10
Add randomized drift rate option to HWPSS

keskitalo:


Helbouha:
Drift option looks good to me! 

I added a second option to simulate a random drift following a 1/f spectrum with a fixed power exponent and random couplings of detectors to it.

Here's a plot showing the PSD of a single detector for different options (alpha is the power exponent for the 1/f):
image

The signal is behaving in the way I would expect it to, so it's clear to merge for me!

sotodlib/toast/ops/sim_hwpss.py
```diff
@@ -97,6 +134,44 @@ def _exec(self, data, detectors=None, **kwargs):
...
+                w = toast.rng.random(
+                    nsamp,
+                    sampler="gaussian",
+                    key=(key1, key2),
+                    counter=(counter1, counter2),
+                ).array()
+                freqs = np.fft.rfftfreq(nsamp, 1/fs)
+                drift_psd = np.zeros_like(freqs)
+                drift_psd[1:] = np.abs(1/freqs[1:])**self.hwpss_drift_alpha
+                drift = np.fft.irfft(np.fft.rfft(w) * np.sqrt(drift_psd))
```
keskitalo, line 172: This produces zero-mean drift which is intuitively correct. However, since we apply the drift as a multiplicative factor, it should have unit mean (or median).

sotodlib/toast/ops/sim_hwpss.py
```diff
@@ -97,6 +134,44 @@ def _exec(self, data, detectors=None, **kwargs):
...
+                nsamp = obs.shared[self.times].data.size
+                w = toast.rng.random(
+                    nsamp,
+                    sampler="gaussian",
+                    key=(key1, key2),
+                    counter=(counter1, counter2),
+                ).array()
+                freqs = np.fft.rfftfreq(nsamp, 1/fs)
+                drift_psd = np.zeros_like(freqs)
+                drift_psd[1:] = np.abs(1/freqs[1:])**self.hwpss_drift_alpha
```
keskitalo, line 171: The user has no way of controlling the amplitude of the drift. There should probably be a `random_drift_rms` trait.

sotodlib/toast/ops/sim_hwpss.py
```diff
@@ -190,9 +265,28 @@ def _exec(self, data, detectors=None, **kwargs):
...
+                        self.hwpss_drift_coupling_center
+                        + gaussian * self.hwpss_drift_coupling_width
+                    )
+                else:
+                    coupling = 1.0
+                
                 # Co-add with the cached signal
 
-                signal += det_scale * iquss
+                signal += det_scale * iquss * drift * coupling
```
keskitalo, line 289: If the coupling truly applies to the drifting portion of the IQUSS signal, then it cannot be applied like this. Rather, you would need
```
signal += det_scale * iquss * (1 + (drift - 1) * coupling)
```
We could of course redefine `drift` to have zero mean, in which case
```
signal += det_scale * iquss * (1 + drift * coupling)
```
would do.

PR #886, opened 2024-06-10
Do not set turnarounds as invalid by default

keskitalo:


sotodlib/toast/workflows/proc_mapmaker.py
```diff
@@ -59,10 +59,21 @@ def setup_mapmaker(operators, templates):
...
         )
     )
-    operators.append(toast.ops.BinMap(name="binner", pixel_dist="pix_dist"))
     operators.append(
         toast.ops.BinMap(
-            name="binner_final", enabled=False, pixel_dist="pix_dist_final"
+            name="binner",
+            pixel_dist="pix_dist",
+            # Discard invalid data and turnarounds by default
+            shared_flag_mask=3,
```
tskisner, line 67: The default shared flag mask for the BinMap operator is to [cut all "non-science" data](https://github.com/hpc4cmb/toast/blob/697eefd199d61c25c082535078875bd3973256bf/src/toast/ops/mapmaker_binning.py#L83).  The hard-coded value here is equivalent to including the `processing` bit (which I agree with) and excluding the `irregular` and `unstable_scanrate` bits (why?).  It looks like we should actually change the meaning of the non-science flag in toast to include processing bit (which it does for the detector mask).

If you need a temporary workaround here in sotodlib, could we do something like
```
shared_flag_mask=(defaults.shared_mask_nonscience | defaults.shared_mask_processing)
```
to make it more clear the names of the flag bits?

sotodlib/toast/workflows/proc_mapmaker.py
```diff
@@ -59,10 +59,21 @@ def setup_mapmaker(operators, templates):
...
+        )
+    )
+    operators.append(
+        toast.ops.BinMap(
+            name="binner_final",
+            enabled=False,
+            pixel_dist="pix_dist_final",
+            # Discard invalid data and turnarounds by default
+            shared_flag_mask=3,
```
tskisner, line 76: Similarly for this.

sotodlib/toast/workflows/sim_observe.py
```diff
@@ -97,6 +97,10 @@ def setup_simulate_observing(parser, operators):
             detset_key="pixel",
             session_split_key="wafer_slot",
             enabled=False,
+            # Only raise the second bit for turnarounds.
+            # Marking them as invalid breaks the
+            # atmospheric simulation
+            turnaround_mask=2,
```
tskisner, line 103: The current defaults in the `SimGround` operator [set the turnarounds to "unstable_scanrate"](https://github.com/hpc4cmb/toast/blob/697eefd199d61c25c082535078875bd3973256bf/src/toast/ops/sim_ground.py#L301) (equivalent to `4`, not invalid).  The defaults of the `SimAtmosphere` operator [is to use the invalid mask](https://github.com/hpc4cmb/toast/blob/697eefd199d61c25c082535078875bd3973256bf/src/toast/ops/sim_tod_atm.py#L81) (equivalent to `1`).  So I don't see any conflict in the default flagging...

Is it possible there is a stale config somewhere with hardcoded flag values that is interfering with the defaults?

tskisner (changes requested):
The main thing I see from this PR is that the upstream toast non-science shared mask should include the processing bit (surprised that it does not actually).  But the other changes seem to be unnecessary with current defaults.  Happy to chat offline tomorrow and figure out what is going on.

keskitalo:
This branch has been rendered obsolete by changes in TOAST defaults.

PR #887, opened 2024-06-11
Apply net_corr factors by default to TOAST focalplanes

keskitalo:
Found another stale branch. This one seems still relevant.


tskisner (approved):
Better late than never :-)

PR #890, opened 2024-06-13
Add inner/outer focalplane split

keskitalo:
Extend the splits operator to support detector cuts


tskisner (approved):
Thanks for adding these new splits!

PR #894, opened 2024-06-17
Cleanup the structure of Split classes

tskisner:
Simplify structure of splits and expand docstrings.  Still need to add a unit test.


keskitalo (approved):
Looks elegant. Thanks for adding the unit test!

PR #895, opened 2024-06-19
SAT filter-bin clean-up

susannaaz:
Clean-up of the filter-bin code, including:
1. Rebase with master
2. Some bug fixes, e.g. maps were produced from obs.signal, without doing hwp synchronous signal removal, and relative calibration was only applied to the signal. Now the relative-calibrated hwpss subtracted signal is used to make the maps, to which the glitches and jumps are removed.
3. Added "--preprocess_config" flag to args for when reading from database
4. Added the option to calculate the parameters for the high-pass filter as a flag to args; 

TO-DOs:
- [ ] Currently glitches are cut rather than corrected; this needs to be fixed, although right now some of what the preprocessing selects as glitches are actually some artifacts in the data, and trying to fix them as you would with glitches might potentially not work well; for now, it seems safer to take them away from the data, but this is only a temporary work-around
- [ ] When selecting the fixed parameters for the high-pass filter, we currently use fknee = 0.1 which is higher than what the data suggests, and alpha = 2 , which I think would remove too much signal.


sotodlib/site_pipeline/make_filterbin_map.py
```diff
@@ -0,0 +1,1113 @@
...
+    return True
+            
+def select_data(obs,
+                det_left_right, det_in_out, det_upper_lower):
+
+    # Restrict non optical detectors, which have nans in their focal plane
+    # coordinates and will crash the mapmaking operation.
+    obs.restrict('dets', obs.dets.vals[obs.det_info.wafer.type == 'OPTC'])
+
+    splits.det_splits_relative(obs, det_left_right=det_left_right,
```
chervias, line 246: I think we probably want to do a relative detector split right before we go into mapmaking to make sure our splits are 50/50. This will probably change in the future but in the mean time lets move this line to the end of the `calibrate_obs_otf` function.

sotodlib/site_pipeline/make_filterbin_map.py
```diff
@@ -0,0 +1,1113 @@
...
+
+    filt = filters.timeconst_filter(timeconst = obs.det_cal.tau_eff,
+                                        invert=True)
+    
+    obs[signal_name] = filters.fourier_filter(obs, filt, signal_name=signal_name)
+        
+        
+def demodulate_hwp(obs):
+    apodize.apodize_cosine(obs)
+    hwp.demod_tod(obs)
```
chervias, line 409: This function by default will demodulate `obs.signal` but all the processing has been done over `obs[f'{signal_name}']`. They might not be the same.

chervias (changes requested):
I have this comments for now, I can do more comments later.

sotodlib/site_pipeline/make_filterbin_map.py
susannaaz: Done

sotodlib/site_pipeline/make_filterbin_map.py
```diff
@@ -0,0 +1,1113 @@
+from argparse import ArgumentParser
+import numpy as np, sys, time, warnings, os, so3g, logging, yaml, sqlite3, itertools
+from sotodlib import coords, mapmaking
+from sotodlib.core import Context,  metadata as metadata_core, FlagManager
```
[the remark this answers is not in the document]
susannaaz: Done

sotodlib/site_pipeline/make_filterbin_map.py
```diff
@@ -0,0 +1,1113 @@
...
+def get_psd(obs):
+    #freq, Pxx_demodQ = fft_ops.calc_psd(obs, signal=obs.demodQ, nperseg=nperseg, merge=True)
+    #_, Pxx_demodU = fft_ops.calc_psd(obs, signal=obs.demodU, nperseg=nperseg, merge=True)
+    # TODO: To use above, need to implement (to avoid bugs):
+    # nperseg = 2**x, use power as it's faster in fft operations
+    # set maxsample = greater than 2**x, so that it's the same as welch 
+    dt = obs.timestamps - obs.timestamps[0]
+    freq, Pxx_demodQ = welch(obs.demodQ, fs=1/np.mean(np.diff(dt)), nperseg=10*60*1/np.mean(np.diff(dt)))
+    _, Pxx_demodU = welch(obs.demodQ, fs=1/np.mean(np.diff(dt)), nperseg=10*60*1/np.mean(np.diff(dt)))
+    obs.merge( core.AxisManager(core.OffsetAxis("nusamps", len(freq))))
```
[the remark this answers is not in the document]
susannaaz: Done

sotodlib/site_pipeline/make_filterbin_map.py
susannaaz: Done

sotodlib/site_pipeline/make_filterbin_map.py
```diff
@@ -0,0 +1,1144 @@
...
+    flags.get_trending_flags(obs,
+                             n_pieces=10,
+                             max_trend=2.5) #,max_samples=obs.samps.count)
+    tdets = has_any_cuts(obs.flags.trends)
+    obs.restrict('dets', obs.dets.vals[~tdets])
+    #print('dets after trending cuts: ', {obs.dets.count})
+
+    
+def get_jumps(obs, signal_name):
+    jflags, _, jfix = jumps.twopi_jumps(obs,
```
chervias, line 281: This function is failing for some cases, I'll send you the script with the correction.

chervias (commented):
Another batch of bugs I have found.

PR #901, opened 2024-07-08
Binning improvement

tterasaki:
Akito pointed out an possible issue in binning hwpss and azss, and suggested a solution that apply apodizing window in binning. You can see what happens without the window in 
* `computeXX: /so/home/tterasaki/code_dev/240708_binning_improvement/test_azss.ipynb`
* `computeXX: /so/home/tterasaki/code_dev/240708_binning_improvement/test_hwpss.ipynb`

### DIFF
--- a/sotodlib/tod_ops/apodize.py
+++ b/sotodlib/tod_ops/apodize.py
@@ -13,7 +13,7 @@ def get_apodize_window_for_ends(aman, apodize_samps=1600):
     """
     w = np.ones(aman.samps.count)
     cosedge = np.cos(np.linspace(0, np.pi/2, apodize_samps))
-    w[-apodize_samps:] = cosedgejujj
+    w[-apodize_samps:] = cosedge
     w[:apodize_samps] = np.flip(cosedge)
     return w
     
--- a/sotodlib/tod_ops/apodize.py
+++ b/sotodlib/tod_ops/apodize.py
@@ -19,7 +19,8 @@ def get_apodize_window_for_ends(aman, apodize_samps=1600):
     
 def get_apodize_window_from_flags(aman, flags, apodize_samps=200):
     """
-    Generate an apodization window based on flag values.
+    Generate an apodization window based on flag values. Apply cosine tapering every 
+    continuous portion of data between flagged region.
 
     Args:
         aman: An axismanager
--- a/sotodlib/tod_ops/apodize.py
+++ b/sotodlib/tod_ops/apodize.py
@@ -1,24 +1,86 @@
 import numpy as np
 
+def get_apodize_window_for_ends(aman, apodize_samps=1600):
+    """
+    Generate an apodization window using a cosine taper at the beginning and end.
+
+    Args:
+        aman: An axismanager
+        apodize_samps (int): Number of samples to apply the cosine taper to at each end.
+
+    Returns:
+        numpy.ndarray: An array representing the apodization window.
+    """
+    w = np.ones(aman.samps.count)
+    cosedge = np.cos(np.linspace(0, np.pi/2, apodize_samps))
+    w[-apodize_samps:] = cosedgejujj
+    w[:apodize_samps] = np.flip(cosedge)
+    return w
+    
+def get_apodize_window_from_flags(aman, flags, apodize_samps=200):
+    """
+    Generate an apodization window based on flag values.
+
+    Args:
+        aman: An axismanager
+        flags (str or RangesMatrix): Flags of mask in RangesMatrix. If provided by 
+            a string, 'aman.flags[flags]' is used for the flags.
+        apodize_samps (int): Number of samples to apply the cosine taper.
+
+    Returns:
+        numpy.ndarray: An array representing the apodization window.
+    """
+    if isinstance(flags, str):
+        flags = aman.flags[flags]
+    apodizer = ~flags.mask()
+    apodizer = apodizer.astype(float)
+    cosedge = np.cos(np.linspace(0, np.pi/2, apodize_samps))
+    
+    for di, det in enumerate(aman.dets.vals):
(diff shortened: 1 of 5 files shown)


sotodlib/tod_ops/apodize.py
```diff
@@ -1,24 +1,86 @@
...
+    """
+    w = np.ones(aman.samps.count)
+    cosedge = np.cos(np.linspace(0, np.pi/2, apodize_samps))
+    w[-apodize_samps:] = cosedge
+    w[:apodize_samps] = np.flip(cosedge)
+    return w
+    
+def get_apodize_window_from_flags(aman, flags, apodize_samps=200):
+    """
+    Generate an apodization window based on flag values.
```
msilvafe, line 22: I think you want to elaborate on this doctoring a bit more. You are essentially cosine tapering every continuous portion of data between flagged regions, make that explicit in the docstring

sotodlib/tod_ops/apodize.py
```diff
@@ -1,24 +1,86 @@
...
+                if _apo_idxes_right[1] > aman.samps.count:                    
+                    apodizer[di][_apo_idxes_right[0]:] = 0
+                else:
+                    apodizer[di][_apo_idxes_right[0]:_apo_idxes_right[1]] = np.flip(cosedge)
+            else:
+                apodizer[di][_apo_idxes_left[0]:_apo_idxes_left[1]] = cosedge
+                apodizer[di][_apo_idxes_right[0]:_apo_idxes_right[1]] = np.flip(cosedge)
+    return apodizer
+    
 def apodize_cosine(aman, signal='signal', apodize_samps=1600, in_place=True,
```
msilvafe, line 65: `signal` here should be called `signal_name` to distinguish it from the usage of `signal` being an array in most top_ops fns.

sotodlib/tod_ops/apodize.py
```diff
@@ -1,24 +1,86 @@
...
-                   apo_axis='apodized'):
+                   apo_axis='apodized', window=None):
     """
     Function to smoothly filter the timestream to 0's on the ends with a
-    cosine function.
+    cosine function. If window is
```
msilvafe, line 69: "If window is" - incomplete sentence.

sotodlib/tod_ops/azss.py
```diff
@@ -24,11 +26,21 @@ def bin_by_az(aman, signal=None, az=None, range=None, bins=100, flags=None):
     range: array-like, optional
         A list specifying the range of azimuth angles to consider for binning. Defaults to None.
         If None, [min(az), max(az)] will be used for binning.
-    bins: integer
-        The number of bins to divide the azimuth angle range into. Defaults to 360.
+    bins: int or sequence of scalars
+        If bins is an int, it defines the number of equal-width bins in the given range (100, by default).
+        If bins is a sequence, it defines the bin edges, including the rightmost edge, allowing for non-uniform bin widths.
+        If `bins` is a sequence, `bins` overwrite `range`.
```
msilvafe, line 32: In docs/docstrings you need double backwards commas: "``" instead of "`" to get the text to render as you want.

sotodlib/tod_ops/azss.py
```diff
@@ -187,7 +237,7 @@ def get_azss(aman, signal=None, az=None, range=None, bins=100, flags=None,
     if merge_model:
         aman.wrap(azss_model_name, model_sig_tod, [(0, 'dets'), (1, 'samps')])
     if subtract_in_place:
-        aman.signal = np.subtract(signal, model_sig_tod, dtype='float32')
+        aman[signal_name] = np.subtract(signal, model_sig_tod, dtype='float32')
     return azss_stats, model_sig_tod
 
 def subtract_azss(aman, signal=None, azss_template=None,
```
msilvafe, line 243: Can you add an `in_place` type of argument to subtract_azss here?

sotodlib/tod_ops/binning.py
```diff
@@ -38,27 +42,33 @@ def bin_signal(aman, bin_by, signal=None,
...
     
     # prepare binned signal array
     binned_signal = np.full([aman.dets.count, nbins], np.nan)
     binned_signal_squared_mean = np.full([aman.dets.count, nbins], np.nan)
     binned_signal_sigma = np.full([aman.dets.count, nbins], np.nan)
-
+        
     # bin tod
     if flags is None:
+        bin_counts, _ = np.histogram(bin_by, bins=bins, range=range, weights = weight_for_signal)
```
msilvafe, line 61: Can we return `counts` in addition to `binned_signal_sigma` and also get that wrapped into the binning dict in the azss + hwpss binning fn's?

msilvafe (changes requested):
This is a bit subtle of an effect thanks for tracking it down and implementing the fix. The notebooks were quite helpful to follow along. I left a few inline comments but once those are addressed and you fix the failing tests this should be ready for re-review+merge.

sotodlib/tod_ops/azss.py
tterasaki: I am confused. Do you want another argument than `subtract_in_place`?

msilvafe: No, in the `subtract_azss` function we don't have an in_place option. We need that to work for the preprocess method to work.

tterasaki: Ah, I got it.

tterasaki: Ok, I have added that option.
https://github.com/simonsobs/sotodlib/blob/4d8eddf206cfcce61d47a0b6b3fb0502715645d1/sotodlib/tod_ops/azss.py#L247-L285

Changed subtract_hwpss to the same style as well.
https://github.com/simonsobs/sotodlib/blob/4d8eddf206cfcce61d47a0b6b3fb0502715645d1/sotodlib/hwp/hwp.py#L484-L525

PR #905, opened 2024-07-15
SAT filterbin mapmaker

chervias:
SAT mapmaker worked on by me and @susannaaz. There is also a very short script called `make-atomic-db` that writes hdf info files into the atomic map db, so you don't have to re-run maps if the mapmaker run is interrupted for some reason.


sotodlib/site_pipeline/make_atomic_db.py
```diff
@@ -0,0 +1,91 @@
...
+        print('output does not exist, exiting')
+        return True
+    # Open a DB, if it doesn't exist it will be created
+    # Write into the atomic map database.
+    conn = sqlite3.connect('./'+args['atomic_db']) # open the conector, if the database exists then it will be opened, otherwise it will be created
+    cursor = conn.cursor()
+        
+    # Check if the table exists, if not create it
+    # the tags will be telescope, frequency channel, wafer, ctime, split_label, split_details, prefix_path, elevation, pwv
+    cursor.execute("""CREATE TABLE IF NOT EXISTS atomic (
```
iparask, line 49: It would be very helpful if there is a primary key here. Primary keys are indexed and make reads and select much faster.

sotodlib/site_pipeline/make_atomic_db.py
```diff
@@ -0,0 +1,91 @@
...
+                  float(f['ctime']), 
+                  f['split_label'].decode(), 
+                  f['split_detail'].decode(), 
+                  f['prefix_path'].decode(), 
+                  f['elevation'], 
+                  f['azimuth'], 
+                  f['RA_ref_start'], 
+                  f['RA_ref_stop'], 
+                  float(f['pwv']))    
+        cursor.execute("INSERT INTO atomic VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)", tuple_)
```
iparask, line 83: Instead of `execute` you can do `executemany` and pass a list of dicts as input.

sotodlib/site_pipeline/make_atomic_db.py
iparask: Also having a primary key will make sure you do not write twice the same entry

PR #907, opened 2024-07-19
docs: pin sphinx-argparse version to fix readthedocs

mhasself:


mhasself:
Well, looks like they fixed sphinx-argparse already...

PR #910, opened 2024-07-29
Pin numpy<2 in package reqs and docs

mhasself:
Fixes a few issues with CI recently:
- pin numpy<2, because so3g probably segfaults otherwise.
- fix some warnings in docs build that have been rising to error level (including asserts that were malformed)

### DIFF
--- a/docs/g3tsmurf.rst
+++ b/docs/g3tsmurf.rst
@@ -304,9 +304,6 @@ an obs_id and a G3tSmurf instance and return paths or file lists.
 Usage with Context
 ------------------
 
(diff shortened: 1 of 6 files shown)


kmharrington (approved):
oh dang, when did 2.0 come out? 

but actually this type promotion looks nice for us if we can get so3g to stop vomiting 

https://numpy.org/doc/2.0/numpy_2_0_migration_guide.html#changes-to-numpy-data-type-promotion

mhasself:
> oh dang, when did 2.0 come out?

For my purposes, looks like June 16, 2024. https://pypi.org/project/numpy/#history

PR #914, opened 2024-08-02
Detmatch cache time limit

jlashner:
This PR protects against the problem found by SATP3 where a detset was added to the failed-file-cache before any observations were added, even though it was a valid tune.

To protect against this, the script will now find the associated timestamp of the detset and check that it has been more than `cfg.time_before_cache_failure` sec before adding it to the failed cache.


mhasself (approved):
Thanks for the quick fix.

Issue #916, opened 2024-08-05
Remove unique `obs_id` based tags being published to InfluxDB


BrianJKoopman:
The [QDS Monitor](https://sotodlib.readthedocs.io/en/latest/site_pipeline.html#qds-monitor) is [currently publishing](https://github.com/simonsobs/sotodlib/blob/1e08ce151864b6b709716bf60a2bb6be0996871d/sotodlib/site_pipeline/update_g3tsmurf_db.py#L169-L176) unique "tags" based on the `obs.obs_id`: https://github.com/simonsobs/sotodlib/blob/1e08ce151864b6b709716bf60a2bb6be0996871d/sotodlib/site_pipeline/update_g3tsmurf_db.py#L139

The uniqueness poses a problem for InfluxDB, which [recommends avoiding too many series](https://docs.influxdata.com/influxdb/v1/concepts/schema_and_data_layout/#avoid-too-many-series). Tags are indexed and per the documentation, "when too many indexes are created, both writes and reads may start to slow down." This might be contributing to our recent InfluxDB data source errors, which seem to be timed out queries.

We have hit the maximum values per tag for the "observation" tag, as shown by these errors from data packaging:
```
ERROR: sotodlib.io.load_smurf: Monitor Update failed for oper_ufm_mv5_1722876082 with 400: {"error":"partial write: max-values-per-tag limit exceeded (100000/100000): measurement=\"obs_process_log\" tag=\"observation\" value=\"oper_ufm_mv12_1722723652\" dropped=775"}
20:23:25.580 | ERROR   | sotodlib.io.load_smurf - Monitor Update failed for oper_ufm_mv5_1722876082 with 400: {"error":"partial write: max-values-per-tag limit exceeded (100000/100000): measurement=\"obs_process_log\" tag=\"observation\" value=\"oper_ufm_mv12_1722723652\" dropped=775"}
ERROR: sotodlib.io.load_smurf: Monitor Update failed for obs_ufm_mv5_1722876241 with 400: {"error":"partial write: max-values-per-tag limit exceeded (100000/100000): measurement=\"obs_process_log\" tag=\"observation\" value=\"oper_ufm_mv12_1722723652\" dropped=776"}
[...]

PR #917, opened 2024-08-05
Remove obs-id from tags

kmharrington:
Fix https://github.com/simonsobs/sotodlib/issues/916

### DIFF
--- a/docs/site_pipeline.rst
+++ b/docs/site_pipeline.rst
@@ -997,7 +997,7 @@ I'm just writing in SAT1)::
             det_tag = dict(base_tags)
             det_tag['detector'] = det
             tag_list.append(det_tag)
-        log_tags = {'observation': obs_id, 'wafer': wafer}
+        log_tags = {'telescope': 'SAT1', 'wafer': wafer}
         monitor.record('white_noise_level', det_white_noise, timestamps, tag_list, 'detector_stats', log_tags=log_tags)
         monitor.write()
 
--- a/sotodlib/site_pipeline/update_g3tsmurf_db.py
+++ b/sotodlib/site_pipeline/update_g3tsmurf_db.py
@@ -136,7 +136,10 @@ def _obs_tags(obs, cfgs):
         "stream_id" : obs.stream_id
     }]
 
-    log_tags = {"observation": obs.obs_id, "stream_id": obs.stream_id}
+    log_tags = {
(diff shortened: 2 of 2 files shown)


BrianJKoopman (changes requested):
This removes the 'observation' tag, but `record_tuning()` and `record_timing()` are going to still check for the `obs_id` in these tags. They do that with `monitor.check()`:
https://github.com/simonsobs/sotodlib/blob/1e08ce151864b6b709716bf60a2bb6be0996871d/sotodlib/site_pipeline/update_g3tsmurf_db.py#L148-L156

There must be a higher order tracking of which observations have already been processed by the site pipeline, yeah? This `monitor.check()` was meant to be used to avoid recomputing stats that were being monitored, the idea being that the `check()` could be used before the computation.

If there's some external tracking of this then I'm not really worried about accidentally re-writing the InfluxDB constantly. In either case, these checks also need to be removed, since they'll no longer function properly without the `observation` tags.

kmharrington:
This is level 2 observation building so the tracking here isn't that simple because the code can't know in advance how many observations there should be or if the ones it's finding are supposed to be complete. It marks completed observations by setting `obs.stop` but that's it. 

After thinking about the setup more, I think I'm just going to pull the monitor off of this script and raise errors instead. I was already doing it with timing so adding the readout ids is easy. Bookbinding also fails when it encounters these situations so we weren't really using this data anyway. 

The monitor implementation in `update_book_plan` is definitely a much better implementation and we run alerts off of that one.

BrianJKoopman (changes requested):
> After thinking about the setup more, I think I'm just going to pull the monitor off of this script and raise errors instead. I was already doing it with timing so adding the readout ids is easy. Bookbinding also fails when it encounters these situations so we weren't really using this data anyway.

That's fine by me. There are some lingering monitor related options, one pointed out in line below. The other is the `--use-monitor` argument setup in argparse on [lines 129-130](https://github.com/simonsobs/sotodlib/blob/7177b42aa2c3ac39d782b9b37277fa1eb277cbb2/sotodlib/site_pipeline/update_g3tsmurf_db.py#L129-L130). That should be dropped too.

If this is being dropped, does that mean we can drop the `g3tsmurf` measurement when we clean things up in the InfluxDB at site too?

BrianJKoopman (approved):
Looks good! Thanks for the changes.

PR #920, opened 2024-08-07
Demod common mode

ykyohei:
Added functions that are useful for estimating common mode of demodulated timestreams.
We had local discussions that implementing these functions and making plots of the common mode of demodulated timestreams will be useful for monitoring data quality.
Example plots. https://site.simonsobs.org/~ykyohei/satp3_tod_spectrum/gallery.html

### DIFF
--- a/sotodlib/coords/demod.py
+++ b/sotodlib/coords/demod.py
@@ -182,20 +182,22 @@ def from_map(tod, signal_map, cuts=None, flip_gamma=True, wrap=False, modulated=
         return signal_sim
 
 
-def rotate_demodQU(tod, zero_gamma=True):
+def rotate_demodQU(tod, update_focal_plane=True):
     """
     Apply detectors' polarization angle calibration to the HWP demodulated Q and U timestreams to
-    place all detectors' Q and U timestreams in a common telescope frame.
+    place all detectors' Q and U timestreams in a common telescope frame. This updates tod.demodQ
+    and tod.demodU, in place.
 
     Args:
         tod : an axisManager object
-        zero_gamma (bool, optional): Whether to set zero to gamma. Make this true for polarization
-        mapmaking using make_map.
+        update_focal_plane (bool, optional): Whether to set focal_plane.gamma angles to zero,
+        consistent with new coordinate reference. Make this true for polarization mapmaking
+        using make_map.
 
     """
     demodC = ((tod.demodQ + 1j*tod.demodU).T * np.exp(-2j*tod.focal_plane.gamma)).T
     tod.demodQ = demodC.real
     tod.demodU = demodC.imag
     del demodC
-    if zero_gamma:
+    if update_focal_plane:
         tod.focal_plane.gamma *= 0
--- a/sotodlib/coords/demod.py
+++ b/sotodlib/coords/demod.py
@@ -184,7 +184,8 @@ def from_map(tod, signal_map, cuts=None, flip_gamma=True, wrap=False, modulated=
 
 def rotate_demodQU(tod, zero_gamma=True):
     """
-    Apply detectors' polarization angle calibration to the HWP demodulated Q and U timestreams.
+    Apply detectors' polarization angle calibration to the HWP demodulated Q and U timestreams to
+    place all detectors' Q and U timestreams in a common telescope frame.
 
     Args:
         tod : an axisManager object
--- a/sotodlib/coords/demod.py
+++ b/sotodlib/coords/demod.py
@@ -180,4 +180,21 @@ def from_map(tod, signal_map, cuts=None, flip_gamma=True, wrap=False, modulated=
         if wrap:
             tod.wrap('signal', signal_sim, [(0, 'dets'), (1, 'samps')])
         return signal_sim
-    
+
+
+def rotate_demodQU(tod, zero_gamma=True):
+    """
+    Apply detectors' polarization angle calibration to the HWP demodulated Q and U timestreams.
+
+    Args:
+        tod : an axisManager object
+        zero_gamma (bool, optional): Whether to set zero to gamma. Make this true for polarization
+        mapmaking using make_map.
+
+    """
+    demodC = ((tod.demodQ + 1j*tod.demodU).T * np.exp(-2j*tod.focal_plane.gamma)).T
+    tod.demodQ = demodC.real
+    tod.demodU = demodC.imag
+    del demodC
+    if zero_gamma:
+        tod.focal_plane.gamma = 0
--- a/sotodlib/coords/demod.py
+++ b/sotodlib/coords/demod.py
@@ -190,9 +190,9 @@ def rotate_demodQU(tod, update_focal_plane=True):
 
     Args:
         tod : an axisManager object
-        update_focal_plane (bool, optional): Whether to set focal_plane.gamma angles to zero,
-        consistent with new coordinate reference. Make this true for polarization mapmaking
-        using make_map.
+            update_focal_plane (bool, optional): Whether to set focal_plane.gamma angles to zero,
+            consistent with new coordinate reference. Make this true for polarization mapmaking
+            using make_map.
 
     """
     demodC = ((tod.demodQ + 1j*tod.demodU).T * np.exp(-2j*tod.focal_plane.gamma)).T
--- a/sotodlib/coords/demod.py
+++ b/sotodlib/coords/demod.py
@@ -197,4 +197,4 @@ def rotate_demodQU(tod, zero_gamma=True):
     tod.demodU = demodC.imag
     del demodC
     if zero_gamma:
-        tod.focal_plane.gamma = 0
+        tod.focal_plane.gamma *= 0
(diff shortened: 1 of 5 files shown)


sotodlib/tod_ops/pca.py
```diff
@@ -215,3 +215,42 @@ def get_trends(tod, remove=False, size=1, signal=None):
...
+    Returns
+    -------
+        common mode timestream
+
+    """
+    signal = tod[signal_name]
+    if method == 'median':
+        common_mode = np.median(signal, axis=0)
+    elif method == 'average':
+        common_mode = np.average(signal, axis=0, weights=weights)
```
msilvafe, line 251: Probably a warning or error should be raised if weights are passed but method is chosen to be median.

sotodlib/coords/demod.py
```diff
@@ -180,4 +180,21 @@ def from_map(tod, signal_map, cuts=None, flip_gamma=True, wrap=False, modulated=
         if wrap:
             tod.wrap('signal', signal_sim, [(0, 'dets'), (1, 'samps')])
         return signal_sim
-    
\ No newline at end of file
+
+
+def rotate_demodQU(tod, zero_gamma=True):
+    """
+    Apply detectors' polarization angle calibration to the HWP demodulated Q and U timestreams.
```
msilvafe, line 187: I think worth stating that this puts all detectors in a common Q/U relative to the instrument frame.

sotodlib/tod_ops/pca.py
```diff
@@ -215,3 +215,42 @@ def get_trends(tod, remove=False, size=1, signal=None):
...
+
+
+def get_common_mode(
+    tod,
+    signal_name='signal',
+    method='median',
+    wrap_name=None,
+    weights=None,
+):
+    """Returns common mode timestream between detectors.
```
msilvafe, line 227: Mention in the description that this uses method 'median' or 'average' across detectors as opposed to a principle component analysis to get the common mode as this diverges from the rest of the module.

Also we are trying to get rid of the `signal_name` implementation everywhere so that these functions could be used with any np array with the right shape even it's not wrapped in aman. Currently the only thing holding us up from that is some baked in dependence in the `fourier_filters` module however once we fix that we're planning to get rid of `signal_name` completely. Since you don't use `fourier_filters` anywhere here I think this implementation example will work replacing `signal_name` with `signal` and allowing it to None, str, or ndarray:

https://github.com/simonsobs/sotodlib/blob/5c26756127403607358e55fbb0b642ab3a0bb523/sotodlib/hwp/hwp.py#L536-L547

sotodlib/tod_ops/pca.py
msilvafe, line 251: Probably a warning or error should be raised if weights are passed but method is chosen to be median.

msilvafe (changes requested):
This looks good, a couple inline comments. I would also like to add a preprocess method for this if you haven't done this before there are lots of examples in `sotodlib.preprocess.processes.py` or DM me and I can help you out or commit it for you.

sotodlib/tod_ops/pca.py
ykyohei: Addressed this suggestion. Also found typo in this example and fixed it https://github.com/simonsobs/sotodlib/pull/920/commits/7010a6eea7e4813ad22d45b93168278a7bf6a51f

ykyohei:
Addressed suggestions and added something to preprocess. If anything else is needed for preprocess, could you commit it?

sotodlib/coords/demod.py
```diff
@@ -180,4 +180,22 @@ def from_map(tod, signal_map, cuts=None, flip_gamma=True, wrap=False, modulated=
         if wrap:
             tod.wrap('signal', signal_sim, [(0, 'dets'), (1, 'samps')])
         return signal_sim
-    
\ No newline at end of file
+
+
+def rotate_demodQU(tod, zero_gamma=True):
```
mhasself, line 185: Can we change "zero_gamma" to "update_focal_plane"?  Then in the description say "whether to set focal_plane.gamma angles to zero, consistent with new coordinate reference."

sotodlib/coords/demod.py
```diff
@@ -180,4 +180,22 @@ def from_map(tod, signal_map, cuts=None, flip_gamma=True, wrap=False, modulated=
...
+    """
+    Apply detectors' polarization angle calibration to the HWP demodulated Q and U timestreams to
+    place all detectors' Q and U timestreams in a common telescope frame.
```
mhasself, line 188: Add "This updates tod.demodQ and tod.demodU, in place."

sotodlib/preprocess/processes.py
```diff
@@ -1171,6 +1172,43 @@ def process(self, aman, proc_aman):
...
+            signal: 'signal'
+            method: 'median'
+            wrap_name: 'signal_common_mode'
+
+    .. autofunction:: sotodlib.tod_ops.pca.get_common_mode
+    """
+    name = "get_common_mode"
+
+    def process(self, aman, proc_aman):
+        tod_ops.fft_ops.calc_psd(aman, **self.process_cfgs)
```
mhasself, line 1209: This does not seem like the right function call...

mhasself (changes requested):
Thanks -- a couple requests inline.

sotodlib/preprocess/processes.py
msilvafe: Ya the preprocess part of this PR needs work + testing. Kyohei left it up to me to fix this part so I'm holding that back. Will get to it soon apologies for the holdup!

ykyohei:
@msilvafe Do you have any objection to deleting the pre-process part of this PR and merging it?
The pre-process part is the only reason we cannot merge this and we can address it in a separate PR.

msilvafe (approved):

sotodlib/coords/demod.py
```diff
@@ -180,4 +180,24 @@ def from_map(tod, signal_map, cuts=None, flip_gamma=True, wrap=False, modulated=
...
+    """
+    Apply detectors' polarization angle calibration to the HWP demodulated Q and U timestreams to
+    place all detectors' Q and U timestreams in a common telescope frame. This updates tod.demodQ
+    and tod.demodU, in place.
+
+    Args:
+        tod : an axisManager object
+        update_focal_plane (bool, optional): Whether to set focal_plane.gamma angles to zero,
+        consistent with new coordinate reference. Make this true for polarization mapmaking
+        using make_map.
```
mhasself, line 195: Indent second+third lines of the parameter description, for proper formatting by sphinx.

sotodlib/tod_ops/pca.py
```diff
@@ -300,3 +303,53 @@ def pca_cuts_and_cal(tod, pca_aman, xfac=2, yfac=1.5, calc_good_medianw=False):
...
+        if weights is not None:
+            logger.warn('weights will be ignored because median method is chosen')
+        common_mode = np.median(signal, axis=0)
+    elif method == 'average':
+        common_mode = np.average(signal, axis=0, weights=weights)
+    else:
+        raise ValueError("method flag must be median or average")
+    if wrap_name is not None:
+        tod.wrap(wrap_name, common_mode, [(0, 'samps')])
+    return common_mode
```
mhasself, line 355: Add new line at eof.

sotodlib/tod_ops/pca.py
```diff
@@ -300,3 +303,53 @@ def pca_cuts_and_cal(tod, pca_aman, xfac=2, yfac=1.5, calc_good_medianw=False):
...
+        signal = tod[signal]
+    elif isinstance(signal, np.ndarray):
+        if np.shape(signal) != (tod.dets.count, tod.samps.count):
+            raise ValueError("When passing signal as ndarray shape must match (n_dets x n_samps).")
+    else:
+        raise TypeError("signal must be str, or ndarray")
+
+    if method == 'median':
+        if weights is not None:
+            logger.warn('weights will be ignored because median method is chosen')
```
mhasself, line 347: `logger.warn` is deprecated, use `logger.warning` instead.

sotodlib/tod_ops/pca.py
```diff
@@ -300,3 +303,53 @@ def pca_cuts_and_cal(tod, pca_aman, xfac=2, yfac=1.5, calc_good_medianw=False):
...
+    Arguments
+    ---------
+        tod: axis manager
+        signal: str, optional
+            The name of the signal to estimate common mode or ndarray with shape of
+            (n_dets x n_samps). Defaults to 'signal'.
+        method: str
+            method of common mode estimation. 'median' or 'average'.
+        wrap_name: str or None.
+            If not None, wrap the common mode into tod with this name.
```
mhasself, line 328: The convention in this module is to use "wrap" for the name of this argument.  Please make consistent.

mhasself (changes requested):
Thanks for your patience... just some minor requests.

ykyohei:
Thanks for your suggestions. I think I corrected them.

ykyohei:
@mhasself Just pinging to check if we can merge this.

mhasself (approved):
Looks good!

PR #922, opened 2024-08-09
Trivial fix for case of simulated elevation weighted noise models

tskisner:
When using simulated observing and nominal noise models, setup pointing first.


keskitalo (approved):
I tested the change and it fixes the issue.

PR #923, opened 2024-08-09
Modify planet mapmaker

yoshinori-0778:
Added xi/eta offset in get_scan_q in planets.py in order to take into account off-centered UFMs/detectors.
I only added those offsets so that this version has backward compatibility.
Demonstrations/Discussion using Jupiter observation for satp3 that having dipole T2P leakage from p.6 in https://docs.google.com/presentation/d/1H1T2U94in_STM8gw5q5HR3xolJYrKs4FQfjjAyozsLw/edit?usp=sharing.

### DIFF
--- a/sotodlib/coords/planets.py
+++ b/sotodlib/coords/planets.py
@@ -92,12 +92,16 @@ class SlowSource:
         return self.ra + self.v_ra * dt, self.dec + self.v_dec * dt
 
 
-def get_scan_q(tod, planet, refq=None):
+def get_scan_q(tod, planet, boresight_offset=None, refq=None):
     """Identify the point (in time and azimuth space) at which the
     specified planet crosses the boresight elevation for the
     observation in tod.  The rotation taking boresight coordinates to
     celestial coordinates at that moment in time (and pointing) is
     computed, and its conjugate is returned.
+
+    Boresight_offset is for taking into account off-centered wafer 
+    that have position offset in the focal plane, which is either None
+    or a (xi, eta) tuple in radian.
 
     The returned rotation is useful because it represents a fixed way
     to rotate the celestial such that the target source ends up at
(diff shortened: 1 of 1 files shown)


mhasself (changes requested):
Sorry to be slow to get to this.  Thanks for the useful extension.

Instead of `xioff, etoff`, please make this a single argument called `boresight_offset`, with default None.  The code may assume it's either None, or a (xi, eta) tuple.

Then rebase and force push the branch.

mhasself:
> Thank you for your comment. Because I am not familiar with git so far, I would like to clarify what should I do by following your comment "Then rebase and force push the branch" .
> 
>     1. modify the script based on your comment("boresight_offset")
> 
>     2. add and commit to the branch "modify_planet_mapmaker"
> 
>     3. `git rebase master modify_planet_mapmaker`
> 
>     4. `git checkout modify_planet_mapmaker`
> 
>     5. `git push --force-with-lease origin modify_planet_mapmaker`
> 
> 
> These are my understanding. Is this correct?

I think what you post above is correct, except that step 4 would normally be unnecessary (i.e. you should still be on that branch, and you won't have to check it out again).

I would do this, which I believe is equivalent:
1. make your changes
2. add and commit to the branch "modify_planet_mapmaker"
3. `git fetch` (to update the local understanding of `origin/master`)
4. `git rebase origin/master`  (no additional args needed if you're still on the modify_planet_mapmaker branch
5. `git push --force` (no additional args needed, usually... but make sure you're on the `modify_planet_mapmaker` branch!)

yoshinori-0778:
Thank you for your comments!
I followed what you suggested.

yoshinori-0778:
Thank you for additional comments.
I reflected your comments, "PEP-8 format" and "squash".

mhasself (approved):
Great!

PR #924, opened 2024-08-12
Adds `force_pointing` kw to match_params

jlashner:
- New `force_pointing` match param that will make it so only detectors with pointing information are matched.
- Removes the `force_src_pointing` param (which would force a match if pointing is provided) since this was always buggy and never had good results
- Allows you to set resonator pointing from the `ResSet.from_aman` function.


mhasself (changes requested):
As discussed I think a more complete matching rule for incorporating _strict position info_ would need to consider the detector type:

- NC: do not match (these don't have design_freq anyway)
- DARK, SLOT: _may or may not_ have position meas.
- UNRT, SQID, BARE: _must not_ have position meas.
- OPTC: _must_ have position meas.

jlashner:
Alright, I added different pointing checks that depend on the det-type like you described above, and confirmed that it still works and that I still get good matches. Info on the new match solutions + plots are here: https://simonsobs.atlassian.net/wiki/x/RwC2JQ

I also changed the config option from `force_pointing` to `enforce_pointing_reqs`.

sotodlib/coords/det_match.py
```diff
@@ -557,6 +557,17 @@ def add_pointing(self, bands, chans, xis, etas):
...
 
+
+def get_det_type_mask(arr: np.ndarray, det_types: List[str]) -> np.ndarray:
+    """
+    Returns a boolean mask of all dets that have specific detector types.
+    """
+    return np.logical_or.reduce([
+        arr['det_type'] == dt
+        for dt in det_types
+    ])
```
mhasself, line 568: I think you are looking for `np.isin`!

sotodlib/coords/det_match.py
```diff
@@ -687,11 +699,28 @@ def _get_biadjacency_matrix(self):
         m = src_arr['is_north'][:, None] != dst_arr['is_north'][None, :]
         mat[m] = np.inf
 
-        if self.match_pars.force_pointing:
-            m = np.isnan(src_arr['xi']) | np.isnan(src_arr['eta'])
-            mat[m, :] = np.inf
-            m = np.isnan(dst_arr['xi']) | np.isnan(dst_arr['eta'])
-            mat[:, m] = np.inf
+        if self.match_pars.enforce_pointing_reqs:
```
mhasself, line 702: So in this block, 'DARK' and 'SLOT' are implicitly untouched.  I think you need to at least _comment on_ DARK/SLOT, if not check explicitly that everything you haven't directly modified is listed as DARK or SLOT.

mhasself (changes requested):
Also remove .vscode files!

jlashner:
Hi @mhasself, I think I addressed your feedback so this is ready for a re-review when you can!

mhasself (approved):
Sorry this fell off my radar.  Please squash on merge.  Thanks!

PR #925, opened 2024-08-13
Fixes in optics code

skhrg:


sotodlib/coords/optics.py
```diff
@@ -330,7 +330,7 @@ def LAT_optics(zemax_path):
     x = LAT["x"][gi].ravel()
     y = LAT["y"][gi].ravel()
     xi = LAT["elev"][gi].ravel()
-    eta = LAT["eta"][gi].ravel()
+    eta = LAT["xel"][gi].ravel()
```
mhasself, line 333: Are we making an approximation here when converting "elev, xel" into "xi, eta"?  If so ... does that matter?  Elevation and cross-elevation are similar to xi, eta, if you're near elevation 0... but when you get a few degrees away from that origin, the difference starts to matter.

skhrg: The zemax model defines the origin to be at the boresite (since it doesn't know anything about sky coordinates and is modeled at a fixed configuration). So I think in this case el/xel is the same as xi/eta. So the only elevation dependent effect here is the LOS rotation. 

I think even if it was actual el/xel then it would still be OK since we could just compute the template at el=0 and rotate to get the correct orientation (which is basically what happens right now anyways).

mhasself: The point is that "el, xel" is not the same as "xi, eta", even stipulating that el=0.  The coordinate conversion to get (xi, eta) from (el, xel) is easy to do with so3g.proj.quat functions -- I guess I am proposing that we do that properly instead of ignoring the difference.

mhasself: > I suppose el/xel is actually a lonlat pair? 

That is the concern.  Is it?  If so, you'll want to do something like (depending on the sign convention for xel):
```
xi, eta, _ = quat.decompose_xieta(quat.euler(1, -np.pi/2) * quat.rotation_lonlat(-xel, el))
```
That takes, for example,
```
>>> el, xel = .01, .02
>>> xi, eta, _ = quat.decompose_xieta(quat.euler(1, -np.pi/2) * quat.rotation_lonlat(-xel, el))
>>> xi, eta
(0.019997666768331163, 0.009999833334166508)
```

Though it's also confusing to me that you associate (xi, eta) with (el, xel) ... rather than (xel, el).  That will need to be sorted.  Draw me a picture and we'll work it out...

sotodlib/coords/optics.py
```diff
@@ -519,32 +510,36 @@ def SAT_focal_plane(aman, x=None, y=None, pol=None, roll=0, mapping_data=None):
...
-        quat.euler(1, np.deg2rad(90)) * quat.rotation_lonlat(pol_minus_lon, pol_lat)
-    )
+    pol_r = (pol_x**2 + pol_y**2)**.5
+    pol_phi = np.arctan2(pol_y, pol_x)
+    pol_theta_iso = -fp_to_sky(pol_r)
+    pol_phi_iso = np.pi/2 - pol_phi + np.pi
+    _xi, _eta, _ = quat.decompose_xieta(quat.rotation_iso(pol_theta_iso, pol_phi_iso))
     pol_xi = _xi * np.cos(np.deg2rad(roll)) - _eta * np.sin(np.deg2rad(roll))
     pol_eta = _eta * np.cos(np.deg2rad(roll)) + _xi * np.sin(np.deg2rad(roll))
     gamma = get_gamma(pol_xi, pol_eta)
```
mhasself, line 545: I see you are insisting on computing some kind of projected angle here, which I have my concerns about.  Ok.

But if we're going to do it this way, I at least need it to not be done in a flat sky approximation.  The present approach uses delta xi and delta eta to measure an angle.  But the local xi and eta axes are not orthogonal on the sky.  The spherical way would be to take your mini-vectors, and rotate them back to near z^ before computing the angle.

To do that, I believe you can just modify the get_gamma function:
```
def get_gamma(pol_xi, pol_eta):
    # docstring etc

    xi = pol_xi.reshape((-1, 2))
    eta = pol_eta.reshape((-1, 2))

    q0 = quat.rotation_xieta(xi[:,0], eta[:,0])
    q1 = quat.rotation_xieta(xi[:,1], eta[:,1])
    dq = ~q0 * q1
    d_xi, d_eta, _ = quat.decompose_xieta(dq)

    gamma = np.arctan2(d_xi, d_eta) % (2 * np.pi)
    return gamma
```

mhasself (approved):
Ok -- I like what's in here for the SAT now but I still wonder about the LAT ... let's follow that up separately I guess.  Thanks!

skhrg:
Ok I am reasonably confident in what is going on with these LAT updates (+ a unit test now exists).

@mhasself can you take another look?

mhasself (approved):
Sorry for the delay!

Squash on merge please.

PR #926, opened 2024-08-15
Add a helper script for testing whether output directories are complete.

tskisner:
Moving some logic out of slurm scripts into a simple python helper script.

### DIFF
--- a/sotodlib/toast/scripts/so_batch_control.py
+++ b/sotodlib/toast/scripts/so_batch_control.py
@@ -1,0 +1,225 @@
+#!/usr/bin/env python3
+
+# Copyright (c) 2024-2024 Simons Observatory.
+# Full license can be found in the top level "LICENSE" file.
+
+"""
+Helper script for batching observations.
+
+The results are printed to stdout to enable use of this
+script inside a slurm script.
+
+"""
+
+import argparse
+import datetime
+import os
+import re
+import sys
+
+import numpy as np
+
+from astropy import units as u
+
+# Import sotodlib.toast first, since that sets default object names
+# to use in toast.
+import sotodlib.toast as sotoast
+
+import toast
+import toast.ops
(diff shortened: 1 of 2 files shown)


sotodlib/toast/scripts/so_batch_control.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+def set_obs_state(out_root, obs, state):
+    """Set the current processing state of an observation.
+    """
+    obs_dir = os.path.join(out_root, obs)
+    if not os.path.exists(obs_dir):
+        os.makedirs(obs_dir)
+    state_file = os.path.join(obs_dir, f"state")
+    temp_state = os.path.join(obs_dir, f"state_temp")
```
keskitalo, line 85: Should the code check if `state_temp` already exists?

sotodlib/toast/scripts/so_batch_control.py
```diff
@@ -0,0 +1,225 @@
...
+        None
+
+    """
+    selected = list()
+    for obs in all_obs:
+        state = get_obs_state(out_root, obs)
+        if state is not None:
+            if state == "done":
+                continue
+            if state == "running" and not ignore_running:
```
keskitalo, line 109: Could we ignore "running" jobs where the file is, say, over 24 hours old?

keskitalo (approved):
This should be very useful.

sotodlib/toast/scripts/so_batch_control.py
tskisner: Although this temp file should only exist for a fraction of a second, I added a check and raise an exception if it exists.

sotodlib/toast/scripts/so_batch_control.py
tskisner: I added a timeout for this situation

PR #929, opened 2024-08-19
Fix toast splits

keskitalo:
Small fixes emerging from the ISO simulations.


tskisner (approved):
Looks good, thanks for the fix!

PR #932, opened 2024-08-21
fix(qa.metrics.PreprocessQA): Reliably get db file from context.

tristpinsm:
A change in the preprocess context file was preventing the QA script fro starting.

@guanyilun How do I proceed to deploy this change to site?

### DIFF
--- a/sotodlib/qa/metrics.py
+++ b/sotodlib/qa/metrics.py
@@ -151,7 +151,7 @@ class PreprocessQA(QAMetric):
 
     def _get_available_obs(self):
         # find preprocess manifest file
-        man_file = [p["db"] for p in self.context["metadata"] if p.get("label", "") == "preprocess"]
+        man_file = [p["db"] for p in self.context["metadata"] if re.match(".*preprocess", p.get("db", ""))]
         if len(man_file) == 0:
             raise Exception(f"No preprocess metadata block in context {self.context.filename}.")
 


sotodlib/qa/metrics.py
```diff
@@ -151,7 +151,7 @@ def _process(self, meta):
 
     def _get_available_obs(self):
         # find preprocess manifest file
-        man_file = [p["db"] for p in self.context["metadata"] if p.get("label", "") == "preprocess"]
+        man_file = [p["db"] for p in self.context["metadata"] if re.match(".*preprocess", p.get("db", ""))]
```
msilvafe, line 154: I'm not sure what preprocess context we were using or what was changed but perhaps we can just add the preprocess metadata entries to the context dictionary if it doesn't exist based on the content of the preprocess config file like is done in `preprocess_tod._get_preprocess_context()` https://github.com/simonsobs/sotodlib/blob/master/sotodlib/site_pipeline/preprocess_tod.py#L59

msilvafe (commented):
I added a suggestion to handle this differently.

sotodlib/qa/metrics.py
tristpinsm: that is effectively what this is doing (I think), and the issue was that it was looking for the `label` key which used to be part of the config. I initially changed it to look for `unpack`, but then figured this way may be more robust to future changes (no guarantee this is true of course).

msilvafe (approved):
I understand this fix, I think matching this way makes sense.

PR #933, opened 2024-08-22
Fixing infinite loop if there are nans in the data for get_ptp_flags

chervias:
In `get_ptp_flags`, if there are nans in the data the `while True` loop will continue to infinity.

### DIFF
--- a/sotodlib/tod_ops/flags.py
+++ b/sotodlib/tod_ops/flags.py
@@ -577,6 +577,8 @@ def get_ptp_flags(aman, signal_name='signal', kurtosis_threshold=5,
     ratio = ptps_full/np.median(ptps_full)
     outlier_mask = (ratio<outlier_range[0]) | (outlier_range[1]<ratio)
     det_mask[outlier_mask] = False
+    if np.any(np.logical_not(np.isfinite(aman[signal_name][det_mask]))):
+        raise ValueError(f"There is a nan in {signal_name} in aman {aman.obs_info['obs_id']} !!!")
     while True:
         if len(aman.dets.vals[det_mask]) > 0:
             ptps = np.ptp(aman[signal_name][det_mask], axis=1)


sotodlib/tod_ops/flags.py
```diff
@@ -577,6 +577,8 @@ def get_ptp_flags(aman, signal_name='signal', kurtosis_threshold=5,
     ratio = ptps_full/np.median(ptps_full)
     outlier_mask = (ratio<outlier_range[0]) | (outlier_range[1]<ratio)
     det_mask[outlier_mask] = False
+    if np.any(np.logical_not(np.isfinite(aman[signal_name][det_mask]))):
+        raise ValueError(f"There is a nan in {signal_name} in aman {aman.obs_info['obs_id']} !!!")
```
msilvafe, line 581: Do you want it written this way so that the mapmaker can catch this error and exit? Since there's no `np.nanptp()` function I would have probably had `np.ptp()` run on a masked version of the array that excludes any samples that are nan in any of the detectors. 

But also this is a wider problem, I believe that many of our functions can't handle `nans` and it was just by chance that this function was the first in the order of our pipeline that couldn't handle a nan and crashed. It makes more sense to catch this at the time of the function that can produce the nans (i.e. the hwp angle estimation or hwpss estimation).

chervias: I'm not sure what we should do. I know for sure we need to fix the infinite loop now, because I would rather for it to crash than to send a job in nersc and burn those hours stuck in an infinite loop without me knowing

chervias: I see this as a last case scenario failsafe, we cannot have nans in the signal otherwise it might end up in an infinite loop, so we should throw an error if that happens

msilvafe: Ok, fair enough, eventually I want to kill the ptp flags anyway and I see the need to prevent an infinite loop so I'll approve this solution.

msilvafe (approved):
Squash + merge away.

PR #934, opened 2024-08-23
fix apodizer

ykyohei:
Current apodizer fails when there are many glitches near the start or end of the timestreams. This is because current apodizer only checkes first or last edge of flag.mask(). Fixed this error by simply checking all the edges. Also, we should loop over left edges and right edges independently because the number of left edges and right edges are not always same.
```
could not broadcast input array from shape (200,) into shape (90,)
  File "/pscratch/sd/y/ykyohei/so_site/scripts/cmb_process.py", line 127, in hwpss_subtraction
    hwp.get_hwpss(aman, flags=flags, merge_stats=False)
  File "/global/homes/y/ykyohei/scratch/sotodlib/sotodlib/hwp/hwp.py", line 122, in get_hwpss
    hwp_angle_bin_centers, bin_counts, binned_hwpss, hwpss_sigma_bin = get_binned_hwpss(
  File "/global/homes/y/ykyohei/scratch/sotodlib/sotodlib/hwp/hwp.py", line 252, in get_binned_hwpss
    weight_for_signal = weight_for_signal[np.newaxis, :] * apodize.get_apodize_window_from_flags(aman,
  File "/global/homes/y/ykyohei/scratch/sotodlib/sotodlib/tod_ops/apodize.py", line 100, in get_apodize_window_from_flags
    apodizer[di][_apo_idxes_right[0]:_apo_idxes_right[1]] = np.flip(cosedge)
ValueError: could not broadcast input array from shape (200,) into shape (90,)
```

Also improved apodize window for when neighboring glitches are closer than apodize_samps.
Figure 11 (1)

### DIFF
--- a/sotodlib/tod_ops/apodize.py
+++ b/sotodlib/tod_ops/apodize.py
@@ -1,5 +1,6 @@
 import numpy as np
 
+
 def get_apodize_window_for_ends(aman, apodize_samps=1600):
     """
     Generate an apodization window using a cosine taper at the beginning and end.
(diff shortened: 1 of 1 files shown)


tterasaki (approved):
Thank you Kyohei. This is really nice catch!

PR #935, opened 2024-08-26
Faster binning

tterasaki:
My binning was slow just because it loop np.histogram for all detectors.
Modified the binning function to the style with np.bincount, which result in ~2 times faster hwpss and azss calculation.


ykyohei (commented):
Looks okay to me.
I'm not sure about the calculation of `bin_indices` but I'm fine with merging this as long as you've confirmed that the result is the same as before.

tterasaki:
Thanks Kyohei. Yes, the new one is consistent with the previous one up to computational error level.

ykyohei (approved):

PR #936, opened 2024-08-26
Modified counter 1 over f filter to accept array-like input

tterasaki:
I want to apply counter 1 over f filter with each detector's filter value.


mhasself:
Sounds good -- can you take this opportunity to improve the docstring, and to add this filter to the tests (in both modes)?

Also I believe this will work for both scalar and array arguments:
```
fk = np.array(fk)[..., None]
n = np.array(n)[..., None]
return 1 / (1 + (fk/freqs)**n)
```
but perhaps that's not so transparent.

msilvafe:
> Sounds good -- can you take this opportunity to improve the docstring, and to add this filter to the tests (in both modes)?
> 
> Also I believe this will work for both scalar and array arguments:
> 
> ```
> fk = np.array(fk)[..., None]
> n = np.array(n)[..., None]
> return 1 / (1 + (fk/freqs)**n)
> ```
> 
> but perhaps that's not so transparent.

Also if you want to split out the array input vs scalar input you can just check `np.isscalar(fk)` and then if its not an int or float you'll get a type error which is ok.

mhasself:
@tterasaki Please let me know if you will be (a) fixing the docs and (b) adding to tests.  If not, I will make those changes.

PR #939, opened 2024-08-29
fix(site_pipeline.record_qa): Catch errors on loading metadata.

tristpinsm:
If crucial metadata like the det_match is missing, catch the exception and proceed with following obs_id's.


tristpinsm:
it would be nice to deploy this sometime soon

guanyilun (approved):

PR #940, opened 2024-08-29
Site pipeline script to run plotting on existing preprocess runs

### LINKED ISSUE #974: plot_flag_stats error with jumps
Found an error within `plot_flag_stats` after jumps during some preprocess runs . It seemed to happen specifically for mv18 in this case, but not fully convinced that's completely correlated. Will need to add handling for this.
```
394.749: ERROR: obs_1710033028_satp1_1111111 ['ufm_mv18_1709863558_tune']
<class 'ValueError'>: zero-size array to reduction operation minimum which has no identity
  File "/global/u1/d/davidvng/sotodlib/sotodlib/site_pipeline/update_preprocess_plots.py", line 113, in plot_preprocess_tod
    pipe.run(aman, aman.preprocess, update_plot=True)
  File "/global/homes/d/davidvng/.local/soconda/20240903_0.1.4/lib/python3.11/site-packages/sotodlib/preprocess/pcore.py", line 462, in run
    process.plot(aman, proc_aman, filename=os.path.join(self.plot_dir, '{ctime}/{obsid}', f'{step+1}_{{name}}.png'))
  File "/global/homes/d/davidvng/.local/soconda/20240903_0.1.4/lib/python3.11/site-packages/sotodlib/preprocess/processes.py", line 341, in plot
    plot_flag_stats(aman, proc_aman[name], flag_type='jumps', filename=filename.replace('{name}', f'{ufm}_jumps_stats'))
  File "/global/homes/d/davidvng/.local/soconda/20240903_0.1.4/lib/python3.11/site-packages/sotodlib/preprocess/preprocess_plot.py", line 479, in plot_flag_stats
    binmin = int(np.floor(np.min(glitchlog)))
                          ^^^^^^^^^^^^^^^^^
  File "/global/common/software/sobs/perlmutter/conda_envs/soconda_20240903_0.1.4/lib/python3.11/site-packages/numpy/core/fromnumeric.py", line 2953, in min
    return _wrapreduction(a, np.minimum, 'min', axis, None, out,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/global/common/software/sobs/perlmutter/conda_envs/soconda_20240903_0.1.4/lib/python3.11/site-packages/numpy/core/fromnumeric.py", line 88, in _wrapreduction
    return ufunc.reduce(obj, axis, dtype, out, **passkwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

davidvng:
Preprocess plotting is run on existing preprocess observations, given the same preprocess config pointing to the ManifestDB. Overall structure works, tested on ~30 observations.
An issue specific to plotting flag stats (`plot_flag_stats`) was found, described in #974. This PR includes handling that resolves #974.


sotodlib/preprocess/pcore.py
```diff
@@ -429,6 +434,7 @@ def run(self, aman, proc_aman=None, select=True, sim=False):
             proc_aman = core.AxisManager( aman.dets, aman.samps)
             full = core.AxisManager( aman.dets, aman.samps)
             run_calc = True
+            run_plot = False
```
msilvafe, line 437: This seems to force you to have to make plots only after the database is built. I think we want to retain the option to make the plots at the time the database is built i.e. when `run_calc = True` or make them from the prebuilt database as you've added here.

sotodlib/site_pipeline/update_preprocess_plots.py
```diff
@@ -0,0 +1,256 @@
...
+            group_by[i] = gb.split(':',1)[1]
+
+        if (gb == 'detset') and (len(group_by) == 1):
+            groups = context.obsfiledb.get_detsets(obs_id)
+            return group_by, [[g] for g in groups]
+        
+    det_info = context.get_det_info(obs_id)
+    rs = det_info.subset(keys=group_by).distinct()
+    groups = [[b for a,b in r.items()] for r in rs]
+    return group_by, groups
```
msilvafe, line 59: We're having a lot of repeated code between here, `preprocess_tod` and `preprocess_obs` we need to find places where code is repeated, make a common preprocess utility module to import those functions from, and then modify the existing scripts to use those utilities. If you don't do it in this PR please create an issue for it.

sotodlib/site_pipeline/update_preprocess_plots.py
```diff
@@ -0,0 +1,256 @@
...
+def get_parser(parser=None):
+    if parser is None:
+        parser = argparse.ArgumentParser()
+    parser.add_argument('configs', help="Preprocessing Configuration File")
+    parser.add_argument(
+        '--query', 
+        help="Query to pass to the observation list. Use \\'string\\' to "
+             "pass in strings within the query.",  
+        type=str
+    )
```
msilvafe, line 130: Seems like if this is purely for updating from an existing database, rather than passing a query you just want to pass the database and get back the list of obs-ids and groups from there. This would change this set of arguments a lot.

sotodlib/site_pipeline/update_preprocess_plots.py
```diff
@@ -0,0 +1,256 @@
...
+    parser.add_argument(
+        '--verbosity',
+        help="increase output verbosity. 0:Error, 1:Warning, 2:Info(default), 3:Debug",
+        default=2,
+        type=int
+    )
+    return parser
+
+def main(
```
msilvafe, line 172: Need docstring w/ intended usage example here or at top of the script.

sotodlib/site_pipeline/update_preprocess_plots.py
```diff
@@ -0,0 +1,256 @@
...
+            obs_list.extend(context.obsdb.query(tot_query, tags=[tag]))
+    else:
+        obs_list = context.obsdb.query(tot_query, tags=tags)
+    if len(obs_list)==0:
+        logger.warning(f"No observations returned from query: {query}")
+    run_list = []
+
+    if not os.path.exists(configs['archive']['index']):
+        #don't run if database doesn't exist
+        logger.warning(f"No database found at {configs['archive']['index']}.")
```
msilvafe, line 224: If you want this to break I think you should raise an exception instead of a warning here.

sotodlib/preprocess/pcore.py
```diff
@@ -437,6 +443,10 @@ def run(self, aman, proc_aman=None, select=True, sim=False):
                 proc_aman.restrict('dets', det_list)
             full = proc_aman.copy()
             run_calc = False
+            if update_plot:
```
msilvafe, line 446: I'm not sure `update_plot` is needed I think `run_plot` can be taken as `True` or `False` either in the `if proc_aman is None` or in the `else` block.

msilvafe (changes requested):

sotodlib/preprocess/pcore.py
davidvng: I've kept the plotting in the original `run_calc = True`:
https://github.com/simonsobs/sotodlib/blob/5a813f4d19e24b1c32c6d3c84ed7fcaaa2e2dfa8/sotodlib/preprocess/pcore.py#L457-L459

sotodlib/site_pipeline/update_preprocess_plots.py
davidvng: I can imagine cases where a custom-formatted ManifestDB would be less general to parse, so I figured using the more standard query to get the obs and groups from the same preprocess config that would have been used initially, containing the path to the ManifestDB, would be the most straightforward method to avoid any possible differences.

sotodlib/site_pipeline/update_preprocess_plots.py
davidvng: I've added these functions to `util.py` and `update_preprocess_plots` imports them. I will leave the change for `preprocess_tod` and `preprocess_obs` for another PR.

sotodlib/site_pipeline/update_preprocess_plots.py
davidvng: Latest commits add this capability by setting `use_db=True` (aka `--use-db` via command line). The obs list and groups are obtained directly from the database instead of needing a query.

msilvafe (approved):
Thanks for the updates! A couple comments then ok to merge. Please open issue or new PR to fix other preprocess scripts to use the utils as you added here, that's great.

PR #942, opened 2024-09-04
Detector Calibration Refactor

jlashner:
This PR adds the `update_det_cal.py` site pipeline script. This contains the det-cal update logic that was was previously in `update_smurf_caldbs`, along with the ability to run the TES parameter corrections documented [here](https://simonsobs.atlassian.net/wiki/spaces/~55705801070cc775254537ab15f2dcf6702b78/pages/596410380/2024-08-20+SATp1+Detector+Calibration+Corrections).

Due to the time consuming nature of the detcal correction and its fitting, some care had to be taken in the parallelization process. I've found that there are different optimal ways to parallelize this depending on whether you're running at NERSC or on site computing. IO to NERSC sqlite files is slow, so it is beneficial to parallelize the the obs-info loading in addition to the correction computation. The NERSC method chains together two separate multiprocessing pools, one for sqlite access and the other for param correction, and allows you to set the number of processes in each pool individually.

On site computing, sqlite IO is faster, so we don't need to parallelize the sqlite access. Instead we create a pool of however many processes we want, and run through observations serially, using the pool to parallelize the TES param corrections by running each detector correction in a separate process.

The method run is settable via the cfg object / file.

We can still run without the calibration correction by setting `cfg.apply_cal_correction = False`.

The TES param correction code itself is in sodetlib [here](https://github.com/simonsobs/sodetlib/blob/master/sodetlib/tes_param_correction.py).


mhasself (changes requested):
As requested in conversation:
- add to docs
- add to `__init__.py`
- deal with the fact that sodetlib is now a site-pipeline dependency
- consider including the "naive" versions of the fracRn / responsivity in the dataset, as well
- fix `main()` and arg parsing to be site_pipeline / prefect compliant.

Thanks!

jlashner:
@mhasself do you think you can point out the __init__.py file I need to change? I don't see where in the site pipeline init or the sotodlib init I should be adding something.

mhasself:
> @mhasself do you think you can point out the **init**.py file I need to change? I don't see where in the site pipeline init or the sotodlib init I should be adding something.

Ah, I was misremembering.  It's actually in `so-workflow` ([here](https://github.com/simonsobs/so-workflow/blob/main/sowork/flows/__init__.py)). So, not your job right now.

For compatibility with Prefect and the so-site-pipeline CLI, see https://sotodlib.readthedocs.io/en/latest/site_pipeline.html#module-sotodlib.site_pipeline.cli.  Adding stuff to `site_pipeline/cli.py` should break the CI, as desired...

sotodlib/site_pipeline/update_det_cal.py
```diff
@@ -0,0 +1,838 @@
...
+            else:
+                cal.phase_to_pW = pA_per_phi0 / (2 * np.pi) / cal.s_i * cal.polarity
+
+        res.result_set = np.array([astuple(c) for c in cals], dtype=CalInfo.dtype())
+        res.success = True
+    except Exception as e:
+        res.traceback = traceback.format_exc()
+        res.fail_msg = res.traceback
+        if cfg.raise_exceptions:
+            raise
```
mhasself, line 616: `raise e` would be better here, right, because the user would get some kind of informative traceback?

sotodlib/site_pipeline/update_det_cal.py
```diff
@@ -0,0 +1,838 @@
...
+        bgmap data taken immediately prior to IV. This will be -1 if the
+        detector is unassigned.
+    polarity: int
+        Polarity of the detector response for a positive change in bias current
+        while the detector is superconducting. This is needed to correct for
+        detectors that have reversed response.
+    r_n: float
+        Normal resistance of the TES [Ohms] calculated from IV curve data.
+    p_sat: float
+        "saturation power" of the TES [J] calculated from IV curve data.
```
mhasself, line 229: Aren't Psat / Pbias usually in W or pW?

mhasself (approved):
A few minor complaints -- once addressed feel free to squash + merge.

sotodlib/site_pipeline/update_det_cal.py
jlashner: Yep! Thanks for catching, the stored values should be watts.

sotodlib/site_pipeline/update_det_cal.py
jlashner: `raise` by itself re-raises the current exception without modifying the stack-trace. [Looking into it there's a slight difference between the two in how the trace is set](https://stackoverflow.com/questions/36153805/difference-between-raise-and-raise-e), but it looks like it doesn't really matter for python3.

sotodlib/site_pipeline/update_det_cal.py
```diff
@@ -0,0 +1,838 @@
...
+    log_level: str
+        Logging level for the logger.
+    """
+
+    def __init__(
+        self,
+        root_dir: str,
+        context_path: str,
+        *,
+        data_root: Optional[str] = None,
```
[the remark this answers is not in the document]
jlashner: I did run this through an auto-formatter (ruff), and it prefers spaces around arguments when type hints are present. Looking at the [pep-8 page](https://peps.python.org/pep-0008/#other-recommendations), it looks like they actually say this is correct and spaces around the equal sign is preferred when using type annotations. Is this ok with you?

sotodlib/site_pipeline/update_det_cal.py
mhasself: Ah, I hadn't appreciated the subtleties.  I think `raise e` is still slightly better.  It's helpful to know both where the error came from and that it was caught and re-raised.

sotodlib/site_pipeline/update_det_cal.py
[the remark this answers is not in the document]
mhasself: So it does!  Ok with me.

PR #944, opened 2024-09-06
Fix for flags and downsampling.

chervias:
This fix allows for the downsampling to be used in the ML mapmaker / depth-1 mapmaker. @iparask you left a comment about this some months ago. The partial solution we have for now is that obs.flags will contain all processing flags, including glitch_flags, etc. In the future we want to make the flags location and names to be user-defined in a config file or something, but this other solution is what we use in the mean time in other sections of the mapmaker. @tanaybhandarkar has also worked on this with me.

### DIFF
--- a/sotodlib/mapmaking/utilities.py
+++ b/sotodlib/mapmaking/utilities.py
@@ -510,15 +510,17 @@ def downsample_obs(obs, down):
     res.wrap("signal", resample.resample_fft_simple(obs.signal, onsamp), [(0,"dets"),(1,"samps")])
 
     # The cuts
-    # TODO: The TOD will include a Flagmanager with all the flags. Update this part
-    # accordingly.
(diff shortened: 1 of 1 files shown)


tanaybhandarkar (approved):

iparask:
@chervias do you mind pointing me to the comment I left? I'm working on a similar fix for the ML mapmaker that takes the flags path as an input.

mhasself (approved):

iparask (approved):
I run an ACT workflow and a LAT workflow. Everything seemed fine

chervias:
> @chervias do you mind pointing me to the comment I left? I'm working on a similar fix for the ML mapmaker that takes the flags path as an input.

Here https://github.com/simonsobs/sotodlib/commit/58681cfdb8cd1f1524adc82a9d01f37ec048053e

PR #945, opened 2024-09-06
Tag HWP metrics with telescope

tristpinsm:
@guanyilun sorry for the back-to-back PRs, but I noticed an oversight in the metrics that are being added by the currently running prefect flow. Once this fix is merged I can make a script to add the missing tags to the HWP metrics entered so far.


guanyilun (approved):

tristpinsm:
@guanyilun OK, I've updated the existing entries in the DB with the correct tags and we should be good to resume this flow.

PR #946, opened 2024-09-06
Include xtalk from other bandpasses and polarizations

tpsatt:
Add contributions from detectors of other bandpasses and polarizations to mumux crosstalk model.


keskitalo (approved):
A very useful generalization.

PR #947, opened 2024-09-09
Gp/fix/act flags

iparask:
This PR makes the glitch path configurable for glitch flags. It assumes that the flags are under a nested path and traverses it based on a separator. This can later be expanded to accept operations over paths.


mhasself:
I think we need to rethink this -- the additions actually break the abstraction (note the SignalCut inherits from Signal); MLMapmaker caches several Signal objects in `self.signals`, and loops over them.

So I suspect the right thing to do is have the `glitch_flags` setting enter through the SignalCut _constructor_, and be saved as an instance variable.

iparask:
I moved the path information in the `init` of the signal classes.

chervias:
I think going through the Signal objects is fine. We could also go through the init of the MLMapmaker object and pass it there as a "global" parameter that applies to all sub-signal objects included.

iparask:
> I think going through the Signal objects is fine. We could also go through the init of the MLMapmaker object and pass it there as a "global" parameter that applies to all sub-signal objects included.

Included. Also the ACT tod that was causing errors now works.

chervias (approved):
I tested the latest version and works fine.

sotodlib/mapmaking/utilities.py
```diff
@@ -439,8 +441,48 @@ def rangemat_sum(rangemat):
...
+    This function allows to pull data from an AxisManager based on a path.
+    Parameters:
+        - aman: An Axis Manager object
+        - path: a string with a recursive path to extract data. The path is separated via a sep.
+                For example 'flags.glitch_flags'
+        - sep: separator. Defaults to `.`
+    """
+
+    rpath = rpath.split(sep=sep)
+    flags = aman.copy()
```
mhasself, line 457: unneccessary copy

sotodlib/mapmaking/utilities.py
```diff
@@ -439,8 +441,48 @@ def rangemat_sum(rangemat):
...
     return res
 
-def find_usable_detectors(obs, maxcut=0.1):
-    ncut  = rangemat_sum(obs.flags.glitch_flags)
+def flags_in_path(
+    aman: core.AxisManager, rpath: str, sep: str = "."
+) -> bool:
+    """
+    This function allows to pull data from an AxisManager based on a path.
+    Parameters:
```
mhasself, line 449: add a blank line before "parameters:"

sotodlib/mapmaking/utilities.py
```diff
@@ -439,8 +441,48 @@ def rangemat_sum(rangemat):
         res[i] = np.sum(ra[:,1]-ra[:,0])
     return res
 
-def find_usable_detectors(obs, maxcut=0.1):
-    ncut  = rangemat_sum(obs.flags.glitch_flags)
+def flags_in_path(
+    aman: core.AxisManager, rpath: str, sep: str = "."
+) -> bool:
+    """
+    This function allows to pull data from an AxisManager based on a path.
```
mhasself, line 448: Inaccurate docstring.

sotodlib/mapmaking/utilities.py
```diff
@@ -439,8 +441,48 @@ def rangemat_sum(rangemat):
...
+        - path: a string with a recursive path to extract data. The path is separated via a sep.
+                For example 'flags.glitch_flags'
+        - sep: separator. Defaults to `.`
+    """
+
+    rpath = rpath.split(sep=sep)
+    flags = aman.copy()
+    while rpath and flags is not None:
+        path = rpath.pop()
+        flags = flags[path]
```
mhasself, line 460: This function does not do its job, if, e.g. "some.long.flags" is not in aman.  You need a:
```
if path not in flags._fields:
    return False
```
or similar.  Or do you want to look up the rpath, and return False if it's there but it's None?  (The docstring isn't clear about what exactly this function is supposed to do.)

sotodlib/mapmaking/utilities.py
```diff
@@ -439,8 +441,48 @@ def rangemat_sum(rangemat):
...
+    """
+    This function allows to pull data from an AxisManager based on a path.
+    Parameters:
+        - aman: An Axis Manager object
+        - path: a string with a recursive path to extract data. The path is separated via a sep.
+                For example 'flags.glitch_flags'
+        - sep: separator. Defaults to `.`
+    """
+
+    flags = aman.copy()
```
mhasself, line 477: Unnecessary copy.

sotodlib/mapmaking/ml_mapmaker.py
```diff
@@ -146,7 +155,7 @@ def transeval(self, id, obs, other, x, tod=None):
 
 class Signal:
     """This class represents a thing we want to solve for, e.g. the sky, ground, cut samples, etc."""
-    def __init__(self, name, ofmt, output, ext):
+    def __init__(self, name, ofmt, output, ext, glitch_flags: str = "flags.glitch_flags"):
```
mhasself, line 158: Does this need to be here?  Due to type checking issues?  Nothing is done with it...

sotodlib/mapmaking/ml_mapmaker.py
```diff
@@ -161,7 +170,8 @@ def __init__(self, name, ofmt, output, ext):
         self.ext    = ext
         self.dof    = None
         self.ready  = False
-    def add_obs(self, id, obs, nmat, Nd): pass
+        self.glitch_flags = glitch_flags
+    def add_obs(self, id, obs, nmat, Nd, glitch_flags:Optional[str]): pass
```
mhasself, line 174: Same question.

Can it be replaced with a generic **kwargs, to imply "yes sometimes a subclass will accept and process other random args."

sotodlib/mapmaking/ml_mapmaker.py
```diff
@@ -176,12 +186,12 @@ class SignalMap(Signal):
...
     def __init__(self, shape, wcs, comm, comps="TQU", name="sky", ofmt="{name}", output=True,
             ext="fits", dtype=np.float32, sys=None, recenter=None, tile_shape=(500,500), tiled=False,
-            interpol=None):
+            interpol=None, glitch_flags: str = "flags.glitch_flags"):
         """Signal describing a sky map in the coordinate system given by "sys", which defaults
         to equatorial coordinates. If tiled==True, then this will be a distributed map with
         the given tile_shape, otherwise it will be a plain enmap. interpol controls the
         pointing matrix interpolation mode. See so3g's Projectionist docstring for details."""
-        Signal.__init__(self, name, ofmt, output, ext)
+        Signal.__init__(self, name, ofmt, output, ext, glitch_flags)
```
mhasself, line 194: Nothing is done with glitch_flags (even in the Signal constructor).  You should cache it in self.glitch_flags.

My inclination is for that to be done, independently, in the SignalMap and SignalCut classes, rather than through the base Signal class.  I think one can imagine signals that don't care about glitch_flags / tod stuff at all (?).

sotodlib/mapmaking/ml_mapmaker.py
```diff
@@ -361,9 +373,9 @@ def transeval(self, id, obs, other, map, tod):
 
 class SignalCut(Signal):
     def __init__(self, comm, name="cut", ofmt="{name}_{rank:02}", dtype=np.float32,
-            output=False, cut_type=None):
+            output=False, cut_type=None, glitch_flags:str ="flags.glitch_flags"):
```
mhasself, line 376: I would prefer these default to None, as they do in SignalMap.  Then convert None -> flags.glitch_flags, to handle the default.

When you have cascading kwargs, e.g.:
```
  SignalMap(..., glitch_flags=None)
     -> Signal(glitch_flags=glitch_flags)
          ->  self.glitch_flags = glitch_flags
```
it tends to defeat the use of default value declarations for optional parameters.  My practice has been to prefer this format for setting default args:
```
    class Signal:
        def __init__(self, ..., glitch_flags=None):
            if glitch_flags is None:
                glitch_flags = 'flags.glitch_flags'
```
Then subclasses, or whatever, can pass in their default value of None and it propagates cleanly all the way down to whatever base function is comfortable with setting the default value.

However... this isn't necessary relevant if we take glitch_flags out of the base class Signal.  In that case you won't have cascading.  I still think it's a good practice, in code that _might_ become cascade, to use "None" to represent "the default value specified by some lower level".  (Though this makes it harder to use "None" to represent "disable this feature" -- I would normally use False to signal that in cases where it is needed.)

mhasself (changes requested):

sotodlib/mapmaking/ml_mapmaker.py
iparask: Yes, I can

sotodlib/mapmaking/ml_mapmaker.py
iparask: Generally, I prefer having the actual default value over a check and set in the code. It makes it very easy know the default value, and it avoids any question as to what is happening. I agree the base class needs to have a default as `None`. Using `kargs` will resolve that and make it much cleaner for sure.

sotodlib/mapmaking/utilities.py
iparask: Let me think about it and fix this

sotodlib/mapmaking/ml_mapmaker.py
mhasself: oh, sorry ... not sure how I didn't see that...  I was really looking for it...

iparask:
@mhasself let's move forward with this PR. ML mapmaker's multipass is not yet used from `sotodlib` mapmaking script and I still have not figured out what is happening with I try to run multiple passes with ACT data. I have not forgotten the bug and I will open an issue soon to keep track of that.

tests/test_mapmaking_utils.py
```diff
@@ -0,0 +1,63 @@
...
+from typing import cast
+
+from sotodlib.core import AxisManager
+from sotodlib.mapmaking.utilities import parameter_in_path, get_flags_from_path
+
+class TestParameterInPath(unittest.TestCase):
+    
+    def setUp(self):
+        # Mock objects for testing
+        obj1 = MagicMock()
```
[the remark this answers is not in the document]
iparask: It is an real Axismanager. Lines 20 to 25 make sure the first high level object is an AxisManager.

Oh... Lines 12-18 are not needed. Sorry I forgot to remove them

sotodlib/core/context.py
```diff
@@ -5,12 +5,15 @@
 import logging
 import numpy as np
 
+from typing import Union, Dict, Tuple, List
```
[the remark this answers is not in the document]
iparask: I will. I put them when I was trying to understand what was happening.

iparask:
I will also fix the conflict with master and sync again.

sotodlib/core/g3_core.py
```diff
@@ -6,7 +6,7 @@
 
 """
 
-from spt3g import core
+from so3g.spt3g import core
```
iparask, line 9: This was an import bug.

sotodlib/mapmaking/utilities.py
```diff
@@ -103,14 +105,14 @@ def inject_map(obs, map, recenter=None, interpol=None):
...
+    pmat  = coords.pmat.P.for_tod(obs, comps=comps, geom=(map.shape, map.wcs), rot=rot, threads="domdir", interpol=interpol)
     # And perform the actual injection
-    pmat.from_map(map.extract(shape, wcs), dest=obs.signal)
+    pmat.from_map(map.extract(map.shape, map.wcs), dest=obs.signal)
 
 def safe_invert_div(div, lim=1e-2, lim0=np.finfo(np.float32).tiny**0.5):
     try:
         # try setting up a context manager that limits the number of threads
-        from threadpoolctl import threadpool_limitse
+        from threadpoolctl import threadpool_limits
```
iparask, line 115: Typo fixed

tests/test_mapmaking_utils.py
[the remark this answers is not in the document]
iparask: This is for testing with an arbitrary object that has a dict representation and it is accessed from `test_parameter_exists_in_dict`. 

`test_parameter_exists_in_fields` tests with an AxisManager object:
```python
        flags = MagicMock(spec=AxisManager)
        flags._fields = {"glitch_flags": "some_value"}
        aman = MagicMock(spec=AxisManager)
        aman._fields = {"flags": flags,
                        "other_key": "another_value"}
```

I am not sure it makes sense to create a whole AxisManager object. I can give it a try if you think it is necessary

mhasself (commented):
Couple last questions but we're almost done with this...

sotodlib/mapmaking/utilities.py
```diff
@@ -440,8 +443,68 @@ def rangemat_sum(rangemat):
         res[i] = np.sum(ra[:,1]-ra[:,0])
     return res
 
-def find_usable_detectors(obs, maxcut=0.1):
-    ncut  = rangemat_sum(obs.flags.glitch_flags)
+
+def parameter_in_path(obj: Any, rpath: str, sep: str = ".") -> bool:
```
mhasself, line 447: Given that `"multi.component.key" in some_axis_manager_instance` now works, do you need this function?

iparask: I'm really sorry Matthew, I forgot to push the last sync that removes all these.

sotodlib/mapmaking/ml_mapmaker.py
```diff
@@ -382,14 +402,28 @@ def transeval(self, id, obs, other, map, tod):
             rot=rot, threads="domdir", weather=unarr(obs.weather), site=unarr(obs.site),
             interpol=self.interpol)
         # Build the RHS for this observation
-        pmap.from_map(dest=tod, signal_map=map, comps=self.comps)
+        # These lines are not activated during the first pass of mapmaking.
+        # if self.tiled:
```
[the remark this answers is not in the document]
iparask: Done

mhasself (approved):
Great!  Squash on merge!

PR #948, opened 2024-09-10
Add hwp model for satp2

17-sugiyama:
The hwp angle model for satp2 has been added.
The definition of the rotation angle and angle offset is based on this page.
https://simonsobs.atlassian.net/wiki/spaces/PRO/pages/185434251/CHWP+main+page#Rotation-direction-definition

### DIFF
--- a/sotodlib/hwp/hwp_angle_model.py
+++ b/sotodlib/hwp/hwp_angle_model.py
@@ -17,6 +17,13 @@ def default_model(telescope):
         model.wrap('mechanical_offset_1', np.deg2rad(-1.66 - 90 + 49.1))
         model.wrap('mechanical_offset_2', np.deg2rad(-1.66 + 90 + 49.1))
 
+    elif telescope == 'satp2':
+        sign = core.AxisManager()
+        sign.wrap('pid', 1)
+        sign.wrap('offcenter', -1)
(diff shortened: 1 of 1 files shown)


ykyohei:
Thanks for adding it. This is fine, but please add hwp_angle_model metadata too. Which is more official place to add hwp angle model.

Please make a new version of script that includes satp2 mode.
https://github.com/simonsobs/pipeline-metadata-scripts/tree/main/hwp_angle_model
Also please update confluence page.
https://simonsobs.atlassian.net/wiki/spaces/PRO/pages/610631740/hwp_angle_model+Metadata+Version+Log

17-sugiyama:
I would like to make a new PR in pipeline-metadata-scripts repository later.
https://github.com/simonsobs/pipeline-metadata-scripts/tree/hwp_angle_model_satp2/hwp_angle_model

I will only add satp2 parameters, but will not change the version of the angle model. Which is suitable for us;
- Add new file and set 'model_version' as 'v2',
- Add new file and set 'model_version' as 'v3',
- Edit existing file to add satp2 parameters?

ykyohei (approved):
We keep model_version as 'v2'. We update this when we update model, not only the model parameter.

The policies of metadata tracking is described here.
https://simonsobs.atlassian.net/wiki/x/PYBrEQ
Please make new file make_hwp_angle_model_240807r2.py

PR #949, opened 2024-09-10
small refactoring to avoid multiple ifs

iparask:
@ykyohei this PR refactors #948 and introduces a dictionary per telescope with its parameters and removes the ifs. We can move the parameters to a more general place if it is needed without significant code change.

The dictionary can also be expanded with different versions of HWP parameters with minimal code addition.


ykyohei:
Hi Giannis, we have hwp_angle_model metadata as the official place to save versioned the model parameters is https://github.com/simonsobs/pipeline-metadata-scripts/tree/main/hwp_angle_model.
The default model is only for backward compatibility with before we made hwp_angle_model metadata, and I plan to delete it later. We also have an older version of hwp_angle_model metadata.

If there is no strong opinion on your refactoring, I would like to keep the original one.

PR #953, opened 2024-09-12
add get_horizon_P

ykyohei:
Added a function to generate a projection matrix for an arbitrary source in horizon coordinates.
This will be especially useful for the drone mapmaking.

For example, one can make a drone map measured by SAT as follows.
```
P = coords.planets.get_horizon_P(tod, az, el, geom=geom, comps='TQU', hwp=True)
map = coords.demod.make_map(aman, P=P)
```
@axf295 already demonstrated drone mapmaking using an equivalent code.


mhasself (approved):
Very useful, thanks!

PR #954, opened 2024-09-13
CLI support to patch ObsDb and ObsFileDb from an upstream copy

mhasself:
Facilitate "in place" updates of ObsDb and ObsFileDb.  Basically if you have an upstream version that only contains new things, the code can cleanly add those new things into the downstream target.

Easiest usage is through the so-metadata cli. 

For ObsDb:
```
$ so-metadata obsdb-diff \
   satp1-may14-snapshot/satp1_obsdb_240911m_local.sqlite \
   satp1-may29-snapshot/satp1_obsdb_240911m_local.sqlite
```
which says:
```
Comparing to satp1-may29-snapshot/satp1_obsdb_240911m_local.sqlite ...
 ... upstream is different, but the target db can be patched to match.
```
and if you add `--patch`, it will modify the left one (may14) in place.

For ObsFileDb:
```
$ so-metadata obsfiledb satp1-may14-snapshot/satp1_obsfiledb.sqlite \
  diff satp1-may29-snapshot/satp1_obsfiledb.sqlite
```
will similarly print:
```
Comparing to satp1-may29-snapshot/satp1_obsfiledb.sqlite ...
 ... upstream is different, but the target db can be patched to match.
```
and you can cause the update by appending `--patch`.

### DIFF
--- a/sotodlib/core/metadata/cli.py
+++ b/sotodlib/core/metadata/cli.py
@@ -53,6 +53,13 @@ def get_parser():
     sp.add_argument('obs_id', nargs='?')
 
     # obsfiledb
+    sp = sps.add_parser('obsdb-diff', description=
+                        'Compare obsdb to an upstream copy and maybe patch it to match.')
+    sp.add_argument('db_file', help='Path to sqlite3 obsdb.')
+    sp.add_argument('upstream_db')
+    sp.add_argument('--patch', action='store_true')
+
+    # obsfiledb
     sp = sps.add_parser('obsfiledb', description=
                         'Inspect an ObsFileDb.  This can be used to list all files, '
                         'and to perform batch updates of filenames.')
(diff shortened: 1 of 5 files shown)


mhasself:
@tskisner Please comment especially on the interface, w.r.t. integration into metadata pull-down system.

This doesn't do ManifestDbs, but most hiccups in processing seem to come from obsdb/obsfiledb, as these are cached in the Context object.

tskisner (approved):
The logic for computing the DB differences looks fine, although I did not test this on actual databases.  The CLI interface is simple and sufficient for what we need.

I wanted to note that is change by itself does not resolve the issue with the `obsfiledb` containing absolute paths to the files on disk.  This means that metadata syncing will still need to "localize" those paths for a given system.  However, this change **will** hopefully prevent breaking open DB connections when the metadata is updated.

To summarize, after this is merged, the metadata sync will need to be changed in the following ways:

- The sync job will need to load the SO software stack to make these `sotodlib` tools available (currently it only uses bash, rsync, and system python3).
- After syncing from the upstream (e.g. site), the "localize" step will run and produce temporary DBs with the filesystem paths modified.  This is actually already done in the script.  Just the final `os.rename()` needs to be removed to stop overwriting the final localized DB.
- This new DB patching step will be run on each temp DB and patch the final localized DB in place (using sqlite connections).  Since we are just getting a brief write lock to the DBs (rather than overwriting them), it should not interrupt running jobs.

iparask:
Would the minimal `soconda` be enough for updating the DBs? At Princeton, this is run on a VM that does not have access to Tiger's `soconda` .

@diveshjain-phy FYI. This will also help with SO:UK.

tskisner:
> Would the minimal `soconda` be enough for updating the DBs? At Princeton, this is run on a VM that does not have access to Tiger's `soconda` .
> 
> @diveshjain-phy FYI. This will also help with SO:UK.

Yes, a `minimal` config would be fine for this.

Issue #956, opened 2024-09-16
update_det_match enters infinite loop when detset fails but is not old enough to be added to the failed file cache


jlashner:
Currently the way the update_det_match main script works, it will enter an infinite loop if a detset fails to compute but is too new to add to the failed cache. For instance:
```
2024-09-16 15:17:35,057 INFO update_det_match : Number of detsets remaining: 6
2024-09-16 15:17:35,484 INFO update_det_match : ufm_mv27_1726152561_tune is too recent to add to failed cache
2024-09-16 15:17:35,485 ERROR update_det_match : Cannot find obsid for detset ufm_mv27_1726152561_tune
2024-09-16 15:17:35,537 INFO update_det_match : Number of detsets remaining: 6
2024-09-16 15:17:35,993 INFO update_det_match : ufm_mv27_1726152561_tune is too recent to add to failed cache
2024-09-16 15:17:35,993 ERROR update_det_match : Cannot find obsid for detset ufm_mv27_1726152561_tune
2024-09-16 15:17:36,037 INFO update_det_match : Number of detsets remaining: 6
2024-09-16 15:17:36,498 INFO update_det_match : ufm_mv27_1726152561_tune is too recent to add to failed cache
2024-09-16 15:17:36,498 ERROR update_det_match : Cannot find obsid for detset ufm_mv27_1726152561_tune
2024-09-16 15:17:36,550 INFO update_det_match : Number of detsets remaining: 6
```
This is easy to fix, just need to compute all detsets to run on at the beginning of the script, instead of searching for the next detset to run dynamically.

PR #957, opened 2024-09-16
Fix infinite loop in det-match script

jlashner:
Currently SATP3 detmatches are being held up because the script loops infinitely if a match fails but is too new to add to the file cache. This addresses that bug, and resolves https://github.com/simonsobs/sotodlib/issues/956.

This has been tested by rebuilding the satp3 detmatch db in my home dir.

### DIFF
--- a/sotodlib/site_pipeline/update_det_match.py
+++ b/sotodlib/site_pipeline/update_det_match.py
@@ -248,11 +248,16 @@ def run_match_aman(runner: Runner, aman, detset, wafer_slot=None):
                      apply_dst_pointing=runner.cfg.apply_solution_pointing)
     return match
 
-def run_match(runner: Runner, detset: str):
+def run_match(runner: Runner, detset: str) -> bool:
     """
     Creates match files for specified detset, along with any other unmatched
     detsets in the loaded observation. If match fails for a known reason, this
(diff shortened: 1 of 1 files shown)


sotodlib/site_pipeline/update_det_match.py
```diff
@@ -429,13 +425,19 @@ def main(config_file: str, all: bool=False):
...
+            run_match(runner, detset)
             update_manifests_all(runner)
     else:
-        runner.run_next_match()
-        update_manifests_all(runner)
+        if len(remaining_detsets) == 0:
+            logger.info("No detsets to run.")
+        else:
+            run_match(runner, remaining_detsets[0])
+            update_manifests_all(runner)
```
mhasself, line 440: Would it make sense here to keep trying to run a remaining_detset until one actually succeeds?

In the current code, if it gets hung up on a failed one, but can't add it to failed cache, then it will not be able to run anything else in the mean time.

mhasself (approved):
Approving but consider my comment.

sotodlib/site_pipeline/update_det_match.py
jlashner: Sure, that makes sense to me.

PR #959, opened 2024-09-20
wip: postgres integration

iparask:
This is to continue the discussion about postgres as the metadata DB. I made some changes in ObsDB to connect to a postgres and run most of the ObsDB commands. I want to do some extra updates and then fix the unit test.

Having said that the size of the changes is not significant in terms of code that is changed. It is more to make sure the SQL code is correct.

Let me know what you think, because I will have to discuss it also with Research Computing at Princeton to have the ability to connect from the compute nodes.

### DIFF
--- a/sotodlib/core/metadata/common.py
+++ b/sotodlib/core/metadata/common.py
@@ -1,9 +1,96 @@
-import sqlite3
+import psycopg
 import gzip
 import os
 
+GET_TABLE_CREATE = """with table_info as (
+                      select
+                          c.column_name,
+                          c.data_type,
+                          c.character_maximum_length,
+                          c.is_nullable,
+                          tc.constraint_type,
+                          tc.constraint_name
+                      from
+                          information_schema.columns as c
+                      left join (
+                          select
+                              kcu.column_name,
+                              tc.constraint_type,
+                              tc.constraint_name,
+                              kcu.table_name
+                          from
+                              information_schema.table_constraints as tc
+                          join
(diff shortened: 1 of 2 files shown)


sotodlib/core/metadata/obsdb.py
```diff
@@ -1,19 +1,20 @@
...
+    "obs": [
+        "obs_id varchar(256) primary key",
+        "timestamp real",
     ],
-    'tags': [
-        "`obs_id` varchar(256)",
-        "`tag` varchar(256)",
-        "CONSTRAINT one_tag UNIQUE (`obs_id`, `tag`)",
+    "tags": [
+        "obs_id varchar(256)",
```
JBorrow, line 15: obs_id is not a foreign key?

sotodlib/core/metadata/common.py
```diff
@@ -1,9 +1,96 @@
-import sqlite3
+import psycopg
 import gzip
 import os
 
+GET_TABLE_CREATE = """with table_info as (
```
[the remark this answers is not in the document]
iparask: This is used to dump an existing DB to a file, similar to SQLite to file function. I do not know how the schema is created. @mhasself can you help here?

sotodlib/core/metadata/obsdb.py
iparask: It seems it is not. I think it should be and it will make search and joins even faster

PR #961, opened 2024-09-23
Fix unique constraint and indent logic error

### LINKED ISSUE #960: TimeCode table unique constraint is not enough
The time code tables need to have a unique constraint that also includes an agent name. Right now this constraint breaks the `utils.set_timecode_final` function. (could also re-write that function but the incorrectly set unique constraint is probably the more correct fix). 

https://github.com/simonsobs/sotodlib/blob/3ae43020bf851d383c8e7178f58ac6252b1aabce/sotodlib/io/g3tsmurf_db.py#L52

kmharrington:
Found while trying to close our some LAT time codes. Change the unique constraint to deal correctly with different servers. 

closes #960 

To update existing tables will need to run. (I'll take care of doing this on site)
```
CREATE TABLE time_codes_new (
        id INTEGER NOT NULL, 
        stream_id VARCHAR, 
        suprsync_type INTEGER, 
        timecode INTEGER, 
        agent VARCHAR, 
        PRIMARY KEY (id), 
        UNIQUE (stream_id, suprsync_type, timecode, agent)
);

INSERT INTO time_codes_new SELECT * from time_codes;
DROP TABLE time_codes;
ALTER TABLE time_codes_new RENAME to time_codes;
```

### DIFF
--- a/sotodlib/io/g3tsmurf_db.py
+++ b/sotodlib/io/g3tsmurf_db.py
@@ -49,7 +49,12 @@ class TimeCodes(Base):
     """
 
     __tablename__ = "time_codes"
-    __table_args__ = (db.UniqueConstraint("stream_id", "suprsync_type", "timecode"),)
+    __table_args__ = (db.UniqueConstraint(
+        "stream_id", 
+        "suprsync_type", 
+        "timecode",
+        "agent",
+    ),)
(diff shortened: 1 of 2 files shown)


guanyilun (approved):
Looks good to me!

PR #962, opened 2024-09-23
Add lmsi to preprocess_obs and preprocess_tod

davidvng:
Adds [lmsi](https://github.com/jborrow/lmsi) to `preprocess_obs` and `preprocess_tod` site pipeline scripts, to be used on prefect. To use, simply add the path to the lmsi config file as `lmsi_config` in the preprocess config. This will produce an HTML file to serve the plots as a simple gallery. @guanyilun has already run a catchup on the existing plot directories.


guanyilun (approved):
looks good to me!

PR #964, opened 2024-09-24
Merge healpix mapmaker

earosenberg:
Add healpix support to the sotodlib mapmaking module.
Includes some re-factoring of demod_mapmaker to minimize code duplication

### DIFF
--- a/sotodlib/coords/pmat.py
+++ b/sotodlib/coords/pmat.py
@@ -5,6 +5,7 @@ from pixell import enmap, tilemap
 
 from .helpers import _get_csl, _valid_arg, _not_both, _confirm_wcs
 from . import helpers
+from . import healpix_utils as hp_utils
 
 import logging
 logger = logging.getLogger(__name__)
@@ -73,8 +74,9 @@ class P:
       is passed as True then the gamma (polarization) angles will be
       reflected (gamma' = -gamma). [dets]
     - geom: The target map geometry. This is a pixell.enmap.Geometry
-      object, with attributes .shape and .wcs; or possibly (if tiled)
-      a pixell.tilemap.TileGeometry.
+      object, with attributes .shape and .wcs (for non-tiled rectpix);
+      a pixell.tilemap.TileGeometry (if tiled); or a Healpix geometry as
+      defined in coords.healpix_utils.
     - comps: String indicating the spin-components to include in maps.
       E.g., 'T', 'QU', 'TQU'.
     - rot: quat giving an additional fixed rotation to apply to get
@@ -133,13 +135,19 @@ class P:
         self.active_tiles = None
         self.det_weights = det_weights
         self.interpol = interpol
+        self.pix_scheme = _infer_pix_scheme(self.geom)
+
+        if self.pix_scheme == "healpix" and (self.geom.nside_tile is not None):
+            if isinstance(self.geom.nside_tile, int):
+                self.geom.nside_tile = min(self.geom.nside, self.geom.nside_tile)
+            self.geom.ntile = True # Not used for healpix but needed for tiled()
 
     @classmethod
     def for_tod(cls, tod, sight=None, geom=None, comps='T',
                 rot=None, cuts=None, threads=None, det_weights=None,
                 timestamps=None, focal_plane=None, boresight=None,
                 boresight_equ=None, wcs_kernel=None, weather='typical',
-                site='so', interpol=None, hwp=False):
+                site='so', interpol=None, hwp=False, qp_kwargs={}):
         """Set up a Projection Matrix for a TOD.  This will ultimately call
         the main P constructor, but some missing arguments will be
         extracted from tod and computed along the way.
@@ -153,6 +161,9 @@ class P:
         - the boresight= keyword argument
         - tod.get('boresight_equ')
         - tod.get('boresight')
+
+        qp_kwargs are passed through to qpoint to compute sight
+        if sight and boresight_equ args are None.
 
         If the map geometry geom is not specified, but the wcs_kernel
         is provided, then get_footprint will be called to determine
@@ -178,7 +189,7 @@ class P:
                 assert(boresight is not None)
                 sight = so3g.proj.CelestialSightLine.az_el(
                     timestamps, boresight.az, boresight.el, roll=boresight.roll,
-                    site=site, weather=weather)
+                    site=site, weather=weather, **qp_kwargs)
         else:
             sight = _get_csl(sight)
 
@@ -228,13 +239,17 @@ class P:
         """
         if super_shape is None:
             super_shape = (self._comp_count(comps), )
-        if self.tiled:
-            # Need to fully resolve tiling to get occupied tiles.
+        if self.pix_scheme == 'healpix':
             proj, _ = self._get_proj_threads()
-            return tilemap.from_tiles(proj.zeros(super_shape), self.geom)
-        else:
-            proj = self._get_proj()
-            return enmap.ndmap(proj.zeros(super_shape), wcs=self.geom.wcs)
+            return proj.zeros(super_shape)
+        elif self.pix_scheme == 'rectpix':
+            if self.tiled:
+                # Need to fully resolve tiling to get occupied tiles.
+                proj, _ = self._get_proj_threads()
+                return tilemap.from_tiles(proj.zeros(super_shape), self.geom)
+            else:
+                proj = self._get_proj()
+                return enmap.ndmap(proj.zeros(super_shape), wcs=self.geom.wcs)
 
     def to_map(self, tod=None, dest=None, comps=None, signal=None,
                det_weights=None, cuts=None, eigentol=None):
@@ -323,7 +338,13 @@ class P:
             weights_map = self.to_weights(
                 tod=tod, comps=comps, signal=signal, det_weights=det_weights, cuts=cuts)
 
-        if dest is None:
+        if self.pix_scheme == "healpix":
+            tile_list = self._get_hp_tile_list(weights_map)
+            if tile_list is not None:
+                weights_map = hp_utils.tiled_to_compressed(weights_map, -1)
+            if dest is None:
+                dest = np.zeros_like(weights_map)
+        elif (self.pix_scheme == "rectpix") and (dest is None):
             dest = self._enmapify(np.zeros_like(weights_map),
                                   wcs=_confirm_wcs(weights_map))
 
@@ -331,6 +352,8 @@ class P:
         dest[:] = helpers._invert_weights_map(
             weights_map, eigentol=eigentol, UPLO='U')
 
+        if self.pix_scheme == "healpix" and tile_list is not None:
+            dest = hp_utils.compressed_to_tiled(dest, tile_list, -1)
         return dest
 
     def remove_weights(self, signal_map=None, weights_map=None, inverse_weights_map=None,
@@ -361,15 +384,27 @@ class P:
         if signal_map is None:
             signal_map = self.to_map(**kwargs)
 
-        if dest is None:
-            wcs_to_use = _confirm_wcs(inverse_weights_map, signal_map)
-            dest = self._enmapify(np.empty(signal_map.shape, signal_map.dtype),
-                                  wcs=wcs_to_use)
-        else:
-            _confirm_wcs(inverse_weights_map, signal_map, dest)
+        if self.pix_scheme == "healpix":
+            tile_list = self._get_hp_tile_list(signal_map)
+            if tile_list is not None:
+                signal_map = hp_utils.tiled_to_compressed(signal_map, -1)
+                inverse_weights_map = hp_utils.tiled_to_compressed(inverse_weights_map, -1)
+            if dest is None:
+                dest = np.zeros_like(signal_map)
+        elif self.pix_scheme == "rectpix":
+            if dest is None:
+                wcs_to_use = _confirm_wcs(inverse_weights_map, signal_map)
+                dest = self._enmapify(np.empty(signal_map.shape, signal_map.dtype),
+                                      wcs=wcs_to_use)
+            else:
+                _confirm_wcs(inverse_weights_map, signal_map, dest)
 
         dest[:] = helpers._apply_inverse_weights_map(inverse_weights_map, signal_map)
+
+        if self.pix_scheme == "healpix" and tile_list is not None:
+            dest = hp_utils.compressed_to_tiled(dest, tile_list, -1)
         return dest
+
 
     def from_map(self, signal_map, dest=None, comps=None, wrap=None,
                  cuts=None, tod=None):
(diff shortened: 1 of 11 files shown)


chervias:
We changed the `tod_grouping` file to `obs_grouping`. Can you fix that please?

sotodlib/coords/pmat.py
```diff
@@ -429,10 +438,12 @@ def _comp_count(self, comps=None):
         return len(comps)
 
     def _get_proj(self):
-        if self.geom is None:
+        interpol_kw = _get_interpol_args(self.interpol)
+        if self.hp_geom is not None:
+            return so3g.proj.ProjectionistHealpix.for_healpix(self.hp_geom.nside, ordering=self.hp_geom.ordering, nside_tile=self.hp_geom.nside_tile, active_tiles=self.active_tiles, **interpol_kw)
+        elif self.geom is None:
             raise ValueError("Can't project without a geometry!")
         # Backwards compatibility for old so3g
```
mhasself, line 446: This comment is associated with the defn of interpol_kw so move it too.

sotodlib/coords/pmat.py
```diff
@@ -470,7 +480,8 @@ def _get_proj_threads(self, cuts=None):
             else:
                 tile_info = proj.get_active_tiles(self._get_asm())
             self.active_tiles = tile_info['active_tiles']
-            proj = self._get_proj()
+            self.hp_geom.nside_tile = proj.nside_tile # Update nside_tile if it was 'auto'
```
mhasself, line 483: This line is healpix specific, doesn't it break non-healpix projector?

sotodlib/coords/pmat.py
```diff
@@ -123,23 +123,26 @@ class P:
...
+        if geom is not None:
+            self.geom = wrap_geom(geom)
         self.comps = comps
         self.cuts = cuts
         self.threads = threads
         self.active_tiles = None
         self.det_weights = det_weights
         self.interpol = interpol
+        self.hp_geom = hp_geom
+        self.hp_tiled = self.hp_geom is not None and self.hp_geom.nside_tile is not None
```
mhasself, line 138: You can just reuse self.tiled, right?

It is a bug (which I have fixed in a brack somewhere) that this constructor doesn't set `self.tiled = None`.  So I suggest you do that, and than update it to self.hp_geom.nside_tile if conditions support that.

There are a lot of places where you need to do things differently depending on rectpix vs. healpix.  So consider inventing enumerator `self.pix_scheme`, with values none, rectpix, healpix, and choose branches based on that rather than depending on what things are None.

sotodlib/coords/pmat.py
```diff
@@ -123,23 +123,26 @@ class P:
...
 
     @classmethod
     def for_tod(cls, tod, sight=None, geom=None, comps='T',
                 rot=None, cuts=None, threads=None, det_weights=None,
                 timestamps=None, focal_plane=None, boresight=None,
                 boresight_equ=None, wcs_kernel=None, weather='typical',
-                site='so', interpol=None, hwp=False):
+                site='so', interpol=None, hwp=False, hp_geom=None, qp_kwargs={}):
```
mhasself, line 145: Update docstring.

mhasself (changes requested):
A few comments, but then I realized this is based on the SAT mapmaking branch, not master.

Do we want this all to go into master now, or do you want to do healpix -> SAT_mapmaking first, or healpix -> master first?

earosenberg:
> A few comments, but then I realized this is based on the SAT mapmaking branch, not master.
> 
> Do we want this all to go into master now, or do you want to do healpix -> SAT_mapmaking first, or healpix -> master first?

My thought was that this is similar in spirit to #891, so could go directly to master (and perhaps be merged from there into SAT-mapmaking). But I suppose if that is still the main branch for mapmaking the most important thing is to put it there, so we could go the other way. Happy to go with whatever you and @chervias think is easiest/best.

sotodlib/coords/pmat.py
earosenberg: Done

sotodlib/coords/pmat.py
earosenberg: It did; I had neglected to test tiled rectpix projectors. Fixed now

sotodlib/coords/pmat.py
earosenberg: Thanks for the suggestion, I have implemented it.
`self.tiled` is implemented here as a property toward the bottom of the file, so it can't be directly initialized to None. That should now work nicely with healpix as well though, so I have merged them (and geom / hp_geom)

mhasself:
@earosenberg  I think it will be fine to PR this towards master rather than a mapmaking branch.

Can you rebase + squash the commit history here, please? Currently it's deeply intertwined with other branches, but then includes reversion of some changes from those branches; so it would be better to present a simpler view of what's proposed here.

chervias:
I deleted the PR for the SAT_mapmaking branch, we are doing a refactor and we'll go through another route, so please merge into master.

sotodlib/coords/pmat.py
earosenberg: Done

sotodlib/coords/pmat.py
```diff
@@ -323,9 +340,16 @@ def to_inverse_weights(self, weights_map=None, tod=None, dest=None,
             weights_map = self.to_weights(
                 tod=tod, comps=comps, signal=signal, det_weights=det_weights, cuts=cuts)
 
+        if self.pix_scheme == "healpix":
+            tile_list = self._get_hp_tile_list(weights_map)
+            if None in weights_map: weights_map = hp_utils.untile_healpix_compressed(weights_map, -1)
```
mhasself, line 345: Two lines, please.  Or you can make this a one-liner with an idempotent form of the function, `weights_map = hp_utils.as_compressed(weights_map)`.

sotodlib/coords/pmat.py
```diff
@@ -501,9 +544,21 @@ def _get_asm(self):
         return so3g.proj.Assembly.attach(self.sight, so3g_fp)
 
     def _prepare_map(self, map):
-        if self.tiled: return map.tiles
+        if self.tiled and self.pix_scheme == "rectpix": return map.tiles
         else:          return map
```
mhasself, line 548: Make this 4 lines.  If you hadn't broken the alignment with the `else:` line I might have let this one slide. :)

The one liner is acceptable... `return (map.tiles if ... else map)`

sotodlib/coords/pmat.py
```diff
@@ -564,3 +619,19 @@ def _get_interpol_args(interpol):
         return {'interpol': interpol}
     assert interpol in [None, "nn", "nearest"], "Old so3g does not support interpolated mapmaking"
     return {}
+
+def check_pix_scheme(pix_scheme, geom=None):
```
mhasself, line 623: You might not need this function any more, or it might become private.  The docstring you have needs punctuation and refilling.  Or, remove it entirely; when a docstring for a short private function just restates in words what the code clearly manifests ... that's not super helpful and it can be left out.

sotodlib/mapmaking/demod_mapmaker.py
```diff
@@ -320,3 +289,138 @@ def write(self, prefix, tag, m):
             if self.comm is None or self.comm.rank == 0:
                 enmap.write_map(oname, m)
         return oname
+
+class DemodSignalMapHealpix(DemodSignal):
```
mhasself, line 293: I don't think it will be sustainable to have a separate class just to support Healpix projection.  Or more to the point: this class and DemodSignalMap contain too much complex, duplicated code.  We have to find a re-factoring that reduces such duplication.

sotodlib/mapmaking/demod_mapmaker.py
```diff
@@ -62,7 +67,7 @@ def __init__(self, signals=[], noise_model=None, dtype=np.float32, verbose=False
         self.ncomp        = len(comps)
         self.singlestream = singlestream
 
-    def add_obs(self, id, obs, noise_model=None, split_labels=None):
+    def add_obs(self, id, obs, noise_model=None, split_labels=None, det_weights='ivar', qp_kwargs={}):
```
mhasself, line 70: `det_weights argument` is not documented anywhere, but has rich semantics...

mhasself (changes requested):
Thanks for pushing this forward!  Lots of comments.  We might need to brainstorm the right way to enrich demod_mapmaker.py classes to support healpix without too much copy paste.

We don't have enough unit testing for Pmat... that's beyond your scope, I guess -- but if you want to take a stab at inventing that, it would be welcome.

sotodlib/mapmaking/demod_mapmaker.py
iparask: @mhasself these are very similar to the Signal class that are in the mapmaker.py. 

That can be the initial abstraction and move from there

mhasself: Ya, that's certainly the origin of DemodSignalmap... but stripped down for simple filter+bin use.  We should indeed be unforking these classes, not forking them more.

chervias: I agree with Matthew, I don't think we should have a new class.

earosenberg: Happy to refactor this. Perhaps a single class with two separate constructors `DemodSignalMap.for_rectpix` and `DemodSignalMap.for_healpix`, following the model of `pmat.P`? The base `__init__` would stay the same but with all args turned into kwargs and expanded with healpix kwargs, so existing code calling it will not break. 
Then I should be able to set a `pix_scheme` flag and navigate the branches straightforwardly as in `pmat`. 

If this sounds acceptable I'll proceed with the implementation, let me know of any other suggestions.

mhasself: That sounds good to me!

earosenberg:
OK addressed above comments, did the re-factor, and added some documentation

sotodlib/mapmaking/demod_mapmaker.py
```diff
@@ -185,29 +238,46 @@ def __init__(self, shape, wcs, comm, comps="TQU", name="sky", ofmt="{name}", out
...
-        if tiled:
-            geo = tilemap.geometry(shape, wcs, tile_shape=tile_shape)
-            self.rhs = tilemap.zeros(geo.copy(pre=(Nsplits,ncomp,)),      dtype=dtype)
-            self.div = tilemap.zeros(geo.copy(pre=(Nsplits,ncomp,ncomp)), dtype=dtype)
-            self.hits= tilemap.zeros(geo.copy(pre=(Nsplits,)),            dtype=dtype)
-        else:
-            self.rhs = enmap.zeros((Nsplits, ncomp)     +shape, wcs, dtype=dtype)
-            self.div = enmap.zeros((Nsplits,ncomp,ncomp)+shape, wcs, dtype=dtype)
-            self.hits= enmap.zeros((Nsplits,)+shape, wcs, dtype=dtype)
+        return cls(shape, wcs, comm, comps, name, ofmt, output, ext, dtype, sys, recenter, tile_shape, tiled, Nsplits, singlestream)
```
mhasself, line 241: Pass parameters by name, please!

sotodlib/mapmaking/demod_mapmaker.py
```diff
@@ -185,29 +238,46 @@ def __init__(self, shape, wcs, comm, comps="TQU", name="sky", ofmt="{name}", out
...
+            The data type to use for the time-ordered data.
+        Nsplits : int, optional
+            Number of splits that you will map simultaneously. By default is 1 when no
+            splits are requested.
+        singlestream : Bool, optional
+            If True, do not perform demodulated filter+bin mapmaking but rather regular
+            filter+bin mapmaking, i.e. map from obs.signal rather than from obs.dsT,
+            obs.demodQ, obs.demodU
+        """
+        return cls(None, None, None, comps, name, ofmt, output, ext, dtype, None, None, None, False, Nsplits, singlestream, nside, nside_tile)
```
mhasself, line 277: Pass params by name.

sotodlib/mapmaking/demod_mapmaker.py
```diff
@@ -310,13 +357,127 @@ def from_work(self, map):
...
-            if self.comm is None or self.comm.rank == 0:
-                enmap.write_map(oname, m)
+            if self.ext in ["fits", "fits.gz"]:
+                if hp is None:
+                    raise ImportError("Cannot save healpix map as fits; healpy could not be imported. Install healpy or save as npy or h5")
+                if m.ndim > 2:
+                    m = np.reshape(m, (np.product(m.shape[:-1]), m.shape[-1]), order='C') # Flatten wrapping axes; healpy.write_map can't handle >2d array
+                hp.write_map(oname, m.view(self.dtype), nest=(self.hp_geom.ordering=='NEST'), partial=write_partial_fits, overwrite=True)
+            elif self.ext == "npy":
+                np.save(oname, m.view(self.dtype))
```
mhasself, line 378: I think we'd be wise to write both the array and some supporting info the projection.  NEST is essential for understanding the data.  NSIDE is helpful.  Eventually we might throw in the celestial frame (galactic / equatorial).

sotodlib/mapmaking/demod_mapmaker.py
```diff
@@ -310,13 +357,127 @@ def from_work(self, map):
...
+            elif self.ext in ['h5', 'hdf5']:
+                if h5py is None:
+                    raise ValueError("Cannot save healpix map as hdf5; h5py could not be imported. Install h5py or save as npy or fits")
+                with h5py.File(oname, 'w') as f:
+                    dset = f.create_dataset("data", m.shape, dtype=self.dtype, data=m)
```
mhasself, line 383: Same here -- add attributes to this dataset with important projection info.

mhasself (changes requested):
Ok!  I think this is looking good.  A few comments about docs mostly.

Thanks for extending the existing tests -- that paves the way for more / better in the future.

Can you remind me what the use of `qp_kwargs` has been?  It is a bit suspicious to have that, and `det_weights`, added to the `DemodMapmaker.add_obs`.  You'd think that noise_model could somehow capture the effects of det_weights, and geom (or the class constructor, more generally) could handle whatever qp_kwargs is doing.  I have stopped looking into this in the spirit of getting this review in in finite time, but perhaps you and @chervias could consider that, from the perspective of eventually unforking the mapmaker & signal classes.

earosenberg:
> Can you remind me what the use of `qp_kwargs` has been? 

I am happy to leave or remove this as you and @chervias think. This was put in place to enable changing the settings of qpoint -- primarily so they could match TOAST settings. The point was to allow to_map match an input map that was scanned with TOAST at default settings. I think this is potentially useful, but maybe a sufficiently niche application that it's not worth complicating the code with.

>You'd think that noise_model could somehow capture the effects of det_weights

I think this is correct. The main function of `det_weights` now is to disable inverse variance weighting for testing, but this can easily be achieved with Nmat. I'll remove it, thanks for the suggestion.

sotodlib/mapmaking/demod_mapmaker.py
earosenberg: Done, added ordering and nside as attributes

sotodlib/mapmaking/demod_mapmaker.py
mhasself: Like this, basically:
```
np.save('test.npy', {'nside': 10, 'data': np.zeros(1000)}, allow_pickle=True)
```

earosenberg: Got it, thanks for the help!

earosenberg:
That should be the most recent comments done -- thanks for the continued feedback. For `qp_kwargs` I thought it seemed sensible to remove it from `demod_mapmaker` but leave it in `pmat`, but happy to change that if there are any objections.

By the way, I also have a unit test file for `healpix_utils`, mostly checking that the conversions between different healpix formats (tiled, compressed, full) are working right. Wasn't sure if you want this to be included, but will add it if so.

chervias:
Besides fixing the conflict, is there anything stopping us from merging? I need this to be merged first so that I can start moving towards merging #973.

earosenberg:
Not from my side -- was waiting for the go-ahead before doing a final rebase and resolving the conflict.

mhasself:
> Not from my side -- was waiting for the go-ahead before doing a final rebase and resolving the conflict.

This is good to go other than the conflict.  Please rebase, fix conflict, squash ... in some order ... and I'll approve.

earosenberg:
Rebase/conflict/squash all done. 
NB the new `test_pmat.py` is failing post-rebase on tiled rectpix maps. It seems that the new `enmapify` in `remove_weights` and `invert_weights` doesn't work for the tiled case. I can remove those lines from the test if it needs to pass before merging, but it seemed like an issue to address outside this PR.

earosenberg:
> Ok ... ya, just suppress that test (comment it out) and we can fix it separately.

Great, done

mhasself (approved):
Quick merge it before something else breaks

PR #967, opened 2024-09-26
update-det-cal: add ability to set multiprocessing start method

jlashner:
Allows you to switch between using multiprocessing "spawn" and "fork" start methods through the config file. Also changes the log-level for successful obs-ids to info.

This should help us debug some problems we're having right now with prefect.


guanyilun (approved):

guanyilun:
Thanks!

PR #971, opened 2024-09-27
Adding wobble correction to get_horizon_P

chervias:
The wobble correction will go in the pointing model, but in the meantime, we can have it here for making drone maps. Code by @ykyohei.

### DIFF
--- a/sotodlib/coords/planets.py
+++ b/sotodlib/coords/planets.py
@@ -178,14 +178,18 @@ def get_scan_P(tod, planet, boresight_offset=None, refq=None, res=None, size=Non
     return P, X
 
 
-def get_horizon_P(tod, az, el, **kw):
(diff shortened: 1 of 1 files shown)


ykyohei (commented):
Thanks for adding this.
I personally think that adding this correction only in this function is not a good idea.
I think that implementing an equivalent pointing correction as a support function in `sotodlib.coords.pointing_model.py` is better.

mhasself:
Sorry I missed this -- I agree with @ykyohei .

In the pointing matrix, `P.sight` holds the rotation from focal plane coordinates to celestial/map coordinates.  That's general for any `P`, so it is possible to construct a stand-alone function that modifies `P` to include wobble (`P.sight = P.sight * deflq`).  Putting such a function in pointing_model.py makes sense to me.

Thanks for trying to get this into sotodlib!

PR #972, opened 2024-09-27
fix: subtract off starting time during slope calculation

### LINKED ISSUE #968: get_trending_flags sometimes cuts all detectors due to division by zero in _slopes calculation
The function `get_trending_flags` in `tod_ops.flags` sometimes cuts all detectors due to a division by zero in the calculation of the slopes for each time chunk:

`_slopes = ((t * s).mean(axis=1) - t.mean() * s.mean(axis=1)) / (
            (t**2).mean() - (t.mean()) ** 2`

This leads to `inf` for `_slopes` and preprocessing will fail on the next step since no detectors remain.  An example observation where this occurs is:

obsid = 'obs_1718494724_satp3_1111111'
dets={'wafer_slot':'ws0', 'wafer.bandpass':'f150'}

Calculating some statistics of the `t` vector gives:

t.mean() = 1718494994.489052
(t^2).mean() =  2.953225046083927e+18
(t.mean())^2  = 2.953225046083927e+18
(t^2).mean() - (t.mean())^2 =  0.0 (this is the denominator)
np.min(t) = 1718494980.354052 
np.max(t)  = 1718495008.624052
t[-1] - t[0] = 28.269999980926514

The slope denominator should be zero if `t` is constant, but printing out the minimum and maximum of `t` shows that it is not.  The calculated denominator is still zero though.  Maybe this is a numerical precision issue?

skhrg:
Should fix #968

### DIFF
--- a/sotodlib/tod_ops/flags.py
+++ b/sotodlib/tod_ops/flags.py
@@ -457,6 +457,10 @@ def get_trending_flags(aman,
         timestamps = aman.timestamps
     assert len(timestamps) == signal.shape[1]
 
+    # This helps with floating point precision
+    # Not modifying inplace since we don't want to touch aman.timestamps
+    timestamps = timestamps - timestamps[0]
+
     slopes = np.zeros((len(signal), 0))
     cut = np.zeros((len(signal), 0), dtype=bool)
     samp_edges = [0]


mmccrackan (approved):
This should work!

mhasself:
This is fine but perhaps an even cleaner fix would be to zero out t.mean() inside the loop (t = t - t.mean()).

skhrg:
> (t = t - t.mean()).

I considered this but from poking around in a shell it seems like the loss of precision happens in the `mean` calculation. I haven't looked at the `numpy` source but assuming its just an accumulator summing over the array so that seems like a reasonable place to lose precision.

PR #973, opened 2024-09-28
Sat mapmaking refactor

chervias:
First step of the refactoring of the SAT mapmaking code, which is the general function to make a demodulated map (either atomic or depth-1). This will live in `mapmaking.demod_mapmaker`.


chervias:
Also, MPI is being removed from the functions in `demod_mapmaker`, because that is only useful for making depth-1 maps which we are not doing now, so for the time being its just annoying. Depth-1 mapmaking is still supported, just running in a single job.

sotodlib/mapmaking/demod_mapmaker.py
```diff
@@ -385,6 +379,206 @@ def write(self, prefix, tag, m):
...
         return oname
 
+def setup_demod_map(noise_model, shape=None, wcs=None, nside=None,
+                    comm=mpi.COMM_WORLD, comps='TQU', split_labels=None,
+                    singlestream=False, dtype_tod=np.float32,
+                    dtype_map=np.float32, recenter=None, verbose=0):
+    """
+        Setup the classes for demod mapmaking and return
+        a DemodMapmmaker object
+    """
```
mhasself, line 389: Please format your docstrings a bit more consistently with the rest of sotodlib. The extra indentation is not our practice, should be:
```
    """
    Setup the ...
```

Same for make_demod_map, below.

sotodlib/mapmaking/demod_mapmaker.py
```diff
@@ -385,6 +379,206 @@ def write(self, prefix, tag, m):
...
+        tag : str, optional
+            Prefix tag for the logger.
+        verbose : bool, optional
+        split_labels : list or None, optional
+            A list of strings with the splits requested. If None then no splits
+            were asked for, i.e. we will produce one map.
+        L : logger, optional
+            Logger for printing on the screen.
+        site : str, optional
+            Plataform name for the pointing matrix.
```
mhasself, line 510: Platform

sotodlib/mapmaking/demod_mapmaker.py
```diff
@@ -5,12 +5,14 @@
...
-import numpy as np
-from pixell import enmap, utils, tilemap, bunch
+__all__ = ['DemodMapmaker','DemodSignal','DemodSignalMap','make_demod_map']
+import numpy as np, os
+from pixell import enmap, utils as putils, tilemap, bunch, mpi
 import so3g.proj
 
+from .. import core
 from .. import coords
+from .. import site_pipeline
```
mhasself, line 15: I don't really like that library code imports things from `site_pipeline`.  That module is sort of special.  For now can you import that submodule from inside your driver function?

The load_or_preproc function, or whatever it's called, should probably move to library level... that's beyond the scope of this PR though.

mhasself (changes requested):
Minor comments -- thanks.

sotodlib/mapmaking/demod_mapmaker.py
chervias: Ok. I agree, probably this function should be in core next to get_obs. @msilvafe I will create an issue to remind us

PR #993, opened 2024-10-09
fix: switch to defaul factory for rx center

skhrg:


elle-shaw (approved):
Looks good to me, glad to hear its compatible with newer python now.

PR #996, opened 2024-10-10
hkdb: Fix bug where first G3 frame was being skipped

jlashner:
@mjrand found an issue with the HK loader where the first G3 frame was being skipped and not included in the result, leading to incomplete queries such as the one shown below:

image

This is caused by a small bug in the loader where the first time a field was found in the HKarchive it would be skipped. This PR fixes that issue.


mhasself (approved):
It might be faster if you cached misses as well; i.e. when agent/field/field_name does not match load_spec, you make a record of that for faster turn-around in the future.

PR #998, opened 2024-10-11
Hkdb relpath support

jlashner:
This PR adds support for relative paths in the hkdb, which is required for usage on data nodes such as nersc.

This changes the indexing function so that new HkFile paths are relative to `cfg.hk_root`, and changes the loading function such that it supports both absolute paths and relative paths in the database.

This will soon be followed by a migration of the hdkbs on site computing to change all HkFile paths to be relative instead of absolute.

This has been tested by building a new sqlite db and verifying that paths are relative and that loading works properly.


sotodlib/io/hkdb.py
```diff
@@ -220,7 +226,11 @@ def get_frames_from_file(file: HkFile, return_on_fail=True) -> List[HkFrame]:
         List of all HkFrames in the file
     """
     frames = []
-    reader = so3g.G3IndexedReader(file.path)
+    if os.path.isabs(file.path):
+        path = str(file.path)
+    else:
+        path = os.path.join(hkcfg.hk_root, str(file.path))
```
mhasself, line 232: I think you don't need this check.  Look:
```
>>> os.path.join('/some/prefix', '/an/abspath')
'/an/abspath'
```

sotodlib/io/hkdb.py
```diff
@@ -420,6 +430,13 @@ def load_hk(load_spec: Union[LoadSpec, dict], show_pb=False):
                 file_spec[frame.file.path] = []
             file_spec[frame.file.path].append(frame.byte_offset)
 
+    # Convert all paths to absolute paths based on cfg.hk_root
+    def create_abs_path(path):
+        if os.path.isabs(path):
+            return path
+        return os.path.join(load_spec.cfg.hk_root, path)
```
mhasself, line 437: same thing here

mhasself (approved):
Approving but see proposed simplifications.

jlashner:
Thanks Matthew, I never knew that about `os.path.join`!  Very useful.

I made those changes and also added some unit tests for hkdb.

PR #999, opened 2024-10-11
Added script that solves satv1 pointing model parameters.

elle-shaw:
Script that solves for pointing model parameters given an input config and saves manifest/metadata.


sotodlib/site_pipeline/solve_pointing_model.py
```diff
@@ -0,0 +1,278 @@
...
+    ufms = config.get("ufms")
+    rxs = fpc.Receiver.load_file(per_obs_fps)
+    if config.get("platform") == "satp1":
+        filelist = list(rxs.keys())
+        # the following are known to be bad fits:
+        filelist = [
+            item
+            for item in filelist
+            if "_1713" not in item and "1716423951" not in item
+        ]
```
mhasself, line 48: This stuff should come from the config file, not be hard-coded.

mhasself (changes requested):
Thanks!  Here's the round 1 review (there will be more rounds).

As a site_pipeline script, this should conform to site_pipeline standards (which I admit are not evenly enforced):
- Reformulate `main(...)`, add a `get_parser()` function, and use `util.main_launcher`.  See `site_pipeline/preprocess_tod.py` for a correct example.
- Register the function in `site_pipeline/cli.py` so it can be run from the command line using `so-site-pipeline` entry point.
- Add an entry in docs/site_pipeline.rst.  Make sure the text includes an example config file.  If command line args are non-trivial, it should include those two.  Both of those things can be "automagically" dropped in by sphinx, from the module code / docstrings, with the right constructions.  (See other examples in site_pipeline.rst)

One more big request for overall scope:  I know this fitter has been validated on this dataset, elsewhere.  But when it is used on future data sets it will be essential to visualize the results, to see the residuals, and explore param combos.  So... please include plotting functionality in here.

(Also a few minor comments inline.)

elle-shaw:
I had to do lots of the code editing on site computing while NERSC was down, so when I copied it back to NERSC to do the git push-ing, it makes it look like every line of code is new! Sorry

elle-shaw:
It looks like the test docs build is failing because the code needs lmfit module to run. I'm not sure how to fix this issue.

elle-shaw:
Ok, I'm finally going hands-off on this iteration.
I couldn't get the proper formatting to have some of the lmfit parameter fits (i.e. correlation, stderr...etc) saved in the axis manager --> .h5 output file. So, I tried to make everything as clear as possible in the log file. 

For the outlier rejection, the log files show what data points are rejected, (obs_Id, Roll, El, Weight, and fit residual). (Some outliers have zero weight, and so don't affect fits).
I also print out the weighted RMS after the first fit iteration, after excluding outliers but still first fit's parameters, and then after the second fit --- showing that the rms after the second iteration of fits goes down and it is not just the removal of the outliers.

tskisner:
Just a heads up that the tests on the most recent push will fail.  so3g and toast both had releases in the last couple hours and I'm about to merge a PR that updates the sotodlib interfaces in a couple places.  I'll post more info on dm-support shortly.

sotodlib/site_pipeline/solve_pointing_model.py
elle-shaw, line 1127: Had a question about why the if "ot_float_aman" statement is only on this second statement. We could have it make the fancy plots for both first and second fits. If that was the case, should the ot_float statement also apply to both minimizations?

elle-shaw (commented):
Quick changes look good to me, I just have the one question on the ot_float option handling on the analyze_PM_with_all_dets plotting section.

Issue #966, opened 2024-09-26
[Minor/Important] a part of the fourier filter method in tod_ops.filters can change all demod signal to nan


hnakata-JP:
I found this part in tod_ops.filters.fourier_filter erase or remove all the signals in the demodulation.
Especially, when I called hwp.demod_tod, the demodQ and demodU as wrapped signals came out as nan. I think this can cause critical trouble in map-making.

> logger.info('fourier_filter: IFFT.')
> t_2()

I am not sure of the details, but the problem can be solved temporarily when you comment this out.
**[To experts]** Please check this item, and treat it properly.

hnakata-JP:
As an example, you can refind this issue with this obs_id, obs_1725654705_satp1_1111111, but not with the previous one, obs_1725654492_satp1_1111111. This may come out just when you want to check some systematic effect.

hnakata-JP:
This issue can occur when the phasor or hwp_angle has a nan value in its array. One can avoid this by restricting TOD to the valid range.

ykyohei:
This happens in the observation that the hwp stopped in the middle of observations, so this does not become a problem for the cmb observations with no problem in hwp operation.
We can mitigate this by changing the timestamp interpolation method, and this is addressed in this draft PR https://github.com/simonsobs/sotodlib/pull/931

msilvafe:
> This happens in the observation that the hwp stopped in the middle of observations, so this does not become a problem for the cmb observations with no problem in hwp operation.
> 
> We can mitigate this by changing the timestamp interpolation method, and this is addressed in this draft PR https://github.com/simonsobs/sotodlib/pull/931

I see, thank you!

ykyohei:
I made separate PR that will fix this nan issue, because other hwp angle pipeline update that I plan in https://github.com/simonsobs/sotodlib/pull/931 will take some more time for merging it.

ykyohei:
This should be fixed by https://github.com/simonsobs/sotodlib/pull/1000

PR #1000, opened 2024-10-11
G3thwp nan fix

ykyohei:
There were hwp angle nan issues in the observations that the hwp stopped in the middle of observations. https://github.com/simonsobs/sotodlib/issues/966
This is because the old hwp angle extrapolation method returns nan when the data range of hwp angle is shorter than the range of observation. This PR improves the hwp data extrapolation methods in general and resolves the nan issue.
This PR also improves the logging of hwp solution pipeline.

This PR does not change the pipeline output except for edge cases, so I do not plan to update the version of hwp_angles metadata.

### DIFF
--- a/sotodlib/hwp/g3thwp.py
+++ b/sotodlib/hwp/g3thwp.py
@@ -586,7 +586,7 @@ class G3tHWP():
         period = (solved["fast_time_1"][offcenter_idx1+1] -
                   solved["fast_time_1"][offcenter_idx1])*self._num_edges
         # When data is extremely noisy, period gets zero and offcentering becomes nan
-        period[period==0] = np.median(period)
+        period[period == 0] = np.median(period)
         offset_angle = 2 * np.pi * offset_time / period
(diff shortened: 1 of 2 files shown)


17-sugiyama:
I guess the extrapolation runs e.g. when the hwp data is shorter than the observation data.
How do you think of adding the extrapolated indexes to the filled_flag?

17-sugiyama:
I'm sorry if I'm misunderstanding, but I guess the extrapolation of `filled_flag` fills with False where hwp data is extrapolated.
How do you think of filling `filled_flag` with True instead? Extrapolation is similar to fixing packet drops.
https://github.com/simonsobs/sotodlib/blob/cb51d0c53a99dedf7835d9c64509be9a3014bfc5/sotodlib/hwp/g3thwp.py#L1030C1-L1033C81

ykyohei:
@17-sugiyama That sounds good. Updated bool interpolation.

17-sugiyama (commented):
Looks very nice!

17-sugiyama (approved):

PR #1002, opened 2024-10-15
feat(site_pipeline.qa): Use JobDB to track processed observations.

tristpinsm:
I mocked up my understanding of how JobDB could be used to track processed observations in the `record_qa` metrics pipeline. I'm using it pretty minimally with the main objective to avoid re-attempting to process observations that failed previously. We could consider exercising some more functionality like retrying some number of times before quitting.

@mhasself if you have time to take a look and let me know if what I did makes sense, I'll proceed to testing this out next.

### DIFF
--- a/sotodlib/site_pipeline/record_qa.py
+++ b/sotodlib/site_pipeline/record_qa.py
@@ -5,6 +5,7 @@ from sotodlib import core
 import sotodlib.site_pipeline.util as sp_util
 from sotodlib.qa import metrics as qa_metrics
 from sotodlib.site_pipeline.monitor import Monitor
+from sotodlib.site_pipeline.jobdb import JobManager
 
 logger = sp_util.init_logger("qa_metrics")
 
@@ -18,6 +19,7 @@ def main(config):
 
     monitor = Monitor.from_configs(config["monitor"])
     context = core.Context(config["context_file"])
+    jdb = JobManager(sqlite_file=config["jobdb"])
 
     # get a list of metrics from the config
     metrics = []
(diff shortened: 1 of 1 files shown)


sotodlib/qa/metrics.py
```diff
@@ -105,13 +110,29 @@ def process_and_record(self, obs_id, meta=None):
...
+        return new_obs
+
+    def set_job_state(self, oid, state):
+        """Set the state of the job for this metric and obs_id in JobDB."""
+        job = self.jdb.get_jobs(jclass=self._jclass, tags={"obs_id": oid})
+        if len(job) != 1:
+            raise Exception(f"Didn't find exactly one job for {self._jclass}, obs_id={oid}. Found {len(job)}.")
+        job = job[0]
+        job.mark_visited()
+        job.jstate = state
```
mhasself, line 135: These lines should be inside a `with self.jdb.locked(job[0]) as job:` block.

mhasself:
> @mhasself if you have time to take a look and let me know if what I did makes sense, I'll proceed to testing this out next.

See my comment, but otherwise this is a fine first implementation, with goal of basic tracking of what has been run already.

I suspect you don't really need `jdb` intelligence in the `qa/metrics.py` module.  It could live entirely in the site_pipeline script side.

In a future where we might want to run multiple `qa` jobs in parallel, you'll want to replace your site_pipeline script loops over metric/obs_id with db queries that get and lock jobs while you process them.  Then multiple processes can check out different jobs.  But that can wait, and is not incompatible with what you've got so far.

tristpinsm:
I made changes based on Matthew's comments and tested this with the dev InfxluDB to check that it handles failures as expected.

Once it's ready to merge, I'll have to initialise the JobDB file with the already processed jobs, but it's not crucial that this happen before we deploy as it won't interfere with its operation and is just for completeness. The first time it runs, it should record all the failures in the JobDB and subsequent runs should skip those observations.

One question: What is the vision for storing the database? Will there be one centralised database file or will we create one for every pipeline? This will need to be specified in the config before this change is deployed.

mhasself (approved):
Thanks!

PR #1006, opened 2024-10-17
Remove the "Test docs build" action

mhasself:
Since we test docs build directly in readthedocs, this is not needed. And it's failing all of a sudden, so, it also needs to be fixed or else disappear.


tskisner (approved):
Looks good!

PR #1013, opened 2024-10-21
Fix caldb issue with bias lines >= 12

### LINKED ISSUE #1012: update_smurf_caldbs fails when bias_lines >= 12 voltages are not nan
Carlos noticed that the obs-id `obs_1727406001_satp3_1111111` does not have det_cal info, while the observations before and after (for instance `obs_1727409259_satp3_1111111`) do.

In prefect we see the error:
```Failed on obs_1727406001_satp3_1111111: 'ufm_mv35_b12'```
which occurs when trying to [check bias-line consistency in this loop](https://github.com/simonsobs/sotodlib/blob/fa9ecb78246a3cb730cc06dfc5794a011299e512/sotodlib/site_pipeline/update_smurf_caldbs.py#L314-L322).

I checked the bias step analysis files for the failing and working observations, and found that for the failing observation, Vbias is not nan for bias lines >= 12, while for the successful observations Vbias is nan for bias lines >= 12:
```
>>> bsa['Vbias']
array([  2.72476807,   2.76213257,   6.86217651,   6.88730469,
         2.99964844,   3.14976196,   6.94957886,   6.9996167 ,
         2.82462524,   2.82462524,   6.9996167 ,   6.94957886,
        22.84063599, -70.05560059,  42.25859619, -15.48594727])
```

I'm not sure exactly what is causing this in the data, but we need to change the existing update_smurf_caldbs script to skip bias lines >=12, rather than just checking for nans.

jlashner:
This resolves #1012 , where update_smurf_caldbs fails when the bias voltage on lines >= 12 are not nan in the bias step analysis.

### DIFF
--- a/sotodlib/site_pipeline/update_det_cal.py
+++ b/sotodlib/site_pipeline/update_det_cal.py
@@ -31,6 +31,7 @@ from sodetlib.operations.bias_steps import BiasStepAnalysis
 # stolen  from pysmurf, max bias volt / num_bits
 DEFAULT_RTM_BIT_TO_VOLT = 10 / 2**19
 DEFAULT_pA_per_phi0 = 9e6
+TES_BIAS_COUNT = 12  # per detset / primary file group
 
 logger = logging.getLogger("det_cal")
 if not logger.hasHandlers():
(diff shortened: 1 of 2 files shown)


mhasself (changes requested):
It's helpful to Future Us in cases like this to add a note explaining that the redundant check really is necessary. E.g.

`# Usually bias lines >= 12 show up as nan but we have seen cases where they are populated and need to be rejected based on count`

jlashner:
Thanks Matthew, I added a comment, and also noticed we would need to make the same change in the new `update_det_cal` script, so I added this check there as well.

mhasself (approved):
Thanks for quick fix!  Please remember to squash on merge.

PR #1021, opened 2024-10-28
Integrated changes from the pb2b angle solution pipeline to handle glitches better

bbixler500:
Here are some changes to the angle solution pipeline which should implement the pb2b glitch finding algorithm. I have tested only the `analyze` method and only on the dataset `obs_1713734012_satp1_00010000`. On that dataset the algorithm correctly cleans the glitches from the angle solution. More testing on other datasets is needed.

### DIFF
--- a/sotodlib/hwp/g3thwp.py
+++ b/sotodlib/hwp/g3thwp.py
@@ -1165,7 +1165,6 @@ class G3tHWP():
 
         return aman
 
-
     def _hwp_angle_calculator(
             self,
             counter,
@@ -1235,14 +1234,18 @@ class G3tHWP():
         # fix reference
         self._fix_ref()
 
-        # glitch removal
-        self._remove_glitch()
+        # correct references with glitches
+        self._correct_bad_refs()
 
         # reference filling
         if fast:
             self._fill_refs_fast()
         else:
             self._fill_refs()
+
+        # glitch removal
+        self._fix_datapoint_glitches()
+        self._fix_value_glitches()
 
         # assign IRIG synched timestamp
         self._time = scipy.interpolate.interp1d(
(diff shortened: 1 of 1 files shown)


sotodlib/hwp/g3thwp.py
```diff
@@ -1277,52 +1280,48 @@ def _hwp_angle_calculator(
...
+        ref_ind = np.where((self._encd_diff > ref_lo_cond) & (self._encd_diff < ref_hi_cond))[0]
+        # Potential glitched references
+        ref_ind_glitch = np.where((self._encd_diff < ref_lo_cond) & (self._encd_diff > ref_gl_cond))[0]
+
+        # Find and add glitched references to normal references
+        used_ind_pairs = []
+        for ind in ref_ind_glitch:
+            # Check that there is sufficient space before and after the glitched reference
+            before = min(ref_ind, key=lambda x:abs(x - ind) + 1e8*(1 + np.sign(x - ind)))
+            after = min(ref_ind, key=lambda x:abs(x - ind) + 1e8*(1 - np.sign(x - ind)))
```
ykyohei, line 1312: Thank you so much Bryce! I will test this on many observations before I merge.

By the way, I found that it is taking significantly long time for these two lines.
Replacing with something like following will improve the computation time a lot.

```
def testp(x, ind):
    return abs(x - ind) + 1e8*(1 - np.sign(x + ind))
def testm(x, ind):
    return abs(x - ind) + 1e8*(1 - np.sign(x - ind))

for ind in ref_ind_glitch:
    # Check that there is sufficient space before and after the glitched reference
    before = ref_ind[np.argmin(testp(ref_ind, ind))]
    after = ref_ind[np.argmin(testm(ref_ind, ind))]
```

ykyohei: These lines were making the typical processing time of 1hour observation data from ~30 seconds to 10 minutes.

I finally identified why it takes very long time on these lines for SO data.
It's because conditions of glitched reference is too strict and half of the all the data points were counted as the potential glitched references. I think this happens because SO encoder data is less uniform compared to pb2b data.

ykyohei (changes requested):
As commented, currently this hwp_angle solution pipeline takes significantly long time than the one in master branch. I need to improve it and test it to merge this.

sotodlib/hwp/g3thwp.py
ykyohei: I think this is fixed by https://github.com/simonsobs/sotodlib/pull/1021/commits/a12bf0422ae618f8102edc216f2674784ac1e3e0 I will work on testing this again.

sotodlib/hwp/g3thwp.py
```diff
@@ -1473,36 +1481,177 @@ def _fill_refs_fast(self):
...
+                # with a high or low clock difference
+                high, low = self._find_high_low_values(diff)
+                start_high = self._find_rot_start_type(diff)
+
+                # Generate a glitch mask for this rotation
+                result, mask = self._glitch_mask(diff, high, low, start_high)
+                # If the mask could not be generated, remove the rotation
+                if not result:
+                    mask = np.full(len(diff), False)
+                    dead_rots.append(i)
```
ykyohei, line 1540: I finished testing this on all the observations. This is working for most of the observations, but I found some edge cases where this PR performs worse than before. I listed up the obs_ids as follows. 
In these obs_ids, I think the glitch removal is failed, and the removal of failed rotation is leading to larger glitch. Could you take a look at these obs_ids? @bbixler500 

obs_1714436045_satp1_1111101
obs_1714439875_satp1_1111101
obs_1714443842_satp1_1111101
obs_1714451357_satp1_1111101
obs_1714462951_satp1_1111101
obs_1714470498_satp1_1111101
obs_1714474327_satp1_1111101
obs_1714948989_satp1_1111011
obs_1714979726_satp1_1111011
obs_1731357259_satp1_1111111

bbixler500:
Pushed updated so address errors seen in certain obs.

obs_1714436045_satp1_1111101
obs_1714439875_satp1_1111101
obs_1714443842_satp1_1111101
obs_1714451357_satp1_1111101
obs_1714462951_satp1_1111101
obs_1714470498_satp1_1111101
obs_1714474327_satp1_1111101 -> Error was caused by one datapoint per rotation having a strange duty cycle. Resulted in the algorithm sometimes identifying it as a glitch. (Fixed)
angle_pattern

obs_1714948989_satp1_1111011 -> "This is because of the incorrect packet drop filling due to datapoint glitches." (Fixed)
obs_1714979726_satp1_1111011 -> "Failed in data point glitch part. This is because of mis identification of ref edge." (Fixed)

Issue #1025, opened 2024-11-08
preproc_or_load_group in preprocess submodule


chervias:
This function currently lives in the site_pipeline section, but probably is better to be in the library side, for example should be in core next to get_obs

msilvafe:
Ya, thanks for creating an issue. @mmccrackan, and I have been speaking about the number of functions in site_pipeline that are being called outside of just the main script. I agree we need to move things out, we can discuss where the best place to do that is. We had moved some of the functions into `site_pipeline/util.py` however do you think that it's better that this goes into `preprocess/common.py` or `preprocess/util.py`. @mhasself since you raised this in @chervias's PR review perhaps you have an opinion?

mhasself:
Yes, as a general rule, don't import things from site_pipeline into other submodules.  And don't import other submodules into core.

So this functionality should live in preprocess/ submodule.  I think `common` doesn't quite capture the audience for this function so `util` or `main` might be better.  But not a strong feeling on that.

chervias:
Hi, is this fixed then by the new PR? If so I'll close the issue

mmccrackan:
Yes, this is addressed in #1027.  Closing.

PR #1014, opened 2024-10-21
Preprocess: replace common preprocess util functions

davidvng:
Resolves #991.

PR #1027, opened 2024-11-12
Unify preprocessing functions

mmccrackan:
This branch moves around some of the preprocessing functions into a common `preprocessing/preprocess_util.py` file.  The ones moved specifically are:

- `load_preprocess_det_select`
- `load_preprocess_tod`/`load_preprocess_obs` (merged into `load_and_preprocess`)
- `preproc_or_load_group` (to address #1025)
- `cleanup_mandb`

Some of these aren't shared among multiple files yet but they will be when #1026 is merged.  Note that the inputs of `load_and_preprocess` are slightly modified from `load_preprocess_tod`/`load_preprocess_obs`.  It now accepts `no_signal` which defaults to `None` and the `dets` and `meta` inputs not in `load_preprocess_obs` are now added.  The default config names were removed, though we could re-add them based on the value of `no_signal` perhaps.

Some additional files have also been moved into `site_pipeline/util.py`:

- `swap_archive`
- `get_preprocess_db` (edited to accept an optional logger as input, creates one if not given)

### DIFF
--- a/sotodlib/coords/helpers.py
+++ b/sotodlib/coords/helpers.py
@@ -321,35 +321,36 @@ def get_footprint(tod, wcs_kernel, dets=None, timestamps=None, boresight=None,
     fp1 = so3g.proj.FocalPlane.from_xieta(fake_dets, xieta1[0], xieta1[1])
 
     asm = so3g.proj.Assembly.attach(sight, fp1)
-    output = np.zeros((len(fake_dets), n_samp, 4))
+    planar = np.zeros((len(fake_dets), n_samp, 4))
     proj = so3g.proj.Projectionist.for_geom((1,1), wcs_kernel)
     if rot:
         # Works whether rot is a quat or a vector of them.
         asm.Q = rot * asm.Q
-    proj.get_planar(asm, output=output)
-
-    output2 = output*0
-    proj.get_coords(asm, output=output2)
+    proj.get_planar(asm, output=planar)
 
     # Get the pixel extrema in the form [{xmin,ymin},{xmax,ymax}]
     delts  = wcs_kernel.wcs.cdelt * DEG
-    planar = output[:,:,:2]
-    ranges = utils.minmax(planar/delts,(0,1))
-    # These are in units of pixel *offsets* from crval. crval
-    # might not correspond to a pixel center, though. So the
-    # thing that should be integer-valued to preserve pixel compatibility
-    # is crpix + ranges, not just ranges. Let's add crpix to transform this
-    # into offsets from the bottom-left pixel to make it easier to reason
-    # about integers
-    ranges += wcs_kernel.wcs.crpix
-    del output
+    ranges = utils.minmax(planar[:,:,:2]/delts,(0,1))
+    del planar
 
-    # Start a new WCS and set the lower left corner.
+    # These planar ranges are in units of pixels away from the
+    # reference point.  The reference point is not necessarily on a
+    # pixel center -- that depends on whether crpix is an integer.  So
+    # transform ranges by +(crpix - 1), making them relative to the
+    # bottom left pixel of wcs_kernel.  Note the -1 here accounts for
+    # FITS numbering of pixels starting at (1, 1).
+    ranges += (wcs_kernel.wcs.crpix - 1)
+
+    # Round ranges, which in pixel offset units, to nearest integer.
+    corners = utils.nint(ranges)
+
+    # Start a new WCS. Adjust crpix to put our footprint in the
+    # bottom-left pixel.
     w = wcs_kernel.deepcopy()
-    corners      = utils.nint(ranges)
     w.wcs.crpix -= corners[0]
-    shape        = tuple(corners[1]-corners[0]+1)[::-1]
+    shape = tuple(corners[1] - corners[0] + 1)[::-1]
     return (shape, w)
+
 
 def get_focal_plane_cover(tod=None, count=0, focal_plane=None,
                           xieta=None):
--- a/sotodlib/mapmaking/demod_mapmaker.py
+++ b/sotodlib/mapmaking/demod_mapmaker.py
@@ -523,6 +523,7 @@ def make_demod_map(context, obslist, noise_model, info,
         List of outputs from preprocess database. To be used in cleanup_mandb.
     """
     from .. import site_pipeline
+    from ..preprocess import preprocess_util
     context = core.Context(context)
     if L is None:
         L = site_pipeline.util.init_logger("Demod filterbin mapmaking")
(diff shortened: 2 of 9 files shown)


sotodlib/site_pipeline/preprocess_obs.py
```diff
@@ -12,35 +12,6 @@
 
 logger = sp_util.init_logger("preprocess")
 
-def _get_preprocess_context(configs, context=None):
```
davidvng, line 15: This removal is also done in #1014

sotodlib/site_pipeline/preprocess_tod.py
```diff
@@ -56,265 +57,6 @@ def dummy_preproc(obs_id, group_list, logger,
     if run_parallel:
         return error, dest_file, outputs
 
-def _get_preprocess_context(configs, context=None):
```
davidvng, line 59: This removal is also done in #1014

sotodlib/site_pipeline/preprocess_tod.py
```diff
@@ -56,265 +57,6 @@ def dummy_preproc(obs_id, group_list, logger,
...
-    if not found:
-        context["metadata"].append( 
-            {
-                "db" : configs["archive"]["index"],
-                "unpack" : "preprocess"
-            }
-        )
-    return configs, context
-
-def _get_groups(obs_id, configs, context):
```
davidvng, line 88: This removal is also done in #1014

davidvng (commented):
Thanks @mmccrackan, some comments on things that are also done in #1014 which has not been merged yet. How should we rectify this? @msilvafe

mmccrackan:
> Thanks @mmccrackan, some comments on things that are also done in #1014 which has not been merged yet. How should we rectify this? @msilvafe

It looks like the only major difference is that I didn't add the get_obslist function that you wrote.  I can add that here since it looks useful for the multilayer_preprocessing.  Since this pull request includes the preprocess_util function, I guess we can say that this branch supersedes #1014?


TODOs:
We are also going to change things so that stuff in preprocess_util currently imported from site_pipeline util are either copied or moved into preprocess_util.

Finally, we will attempt to merge preprocess_tod, preprocess_obs, and multilayer_preprocess_tod into a single script.

sotodlib/mapmaking/demod_mapmaker.py
```diff
@@ -523,6 +523,7 @@ def make_demod_map(context, obslist, noise_model, info,
         List of outputs from preprocess database. To be used in cleanup_mandb.
     """
     from .. import site_pipeline
```
chervias, line 525: This import here is only for the logger. @mhasself does not want us to import the site pipeline stuff as if it were a library. On the other hand, this logger is needed here because passing a logger from the script side to this function won't work with multiprocessing, as it won't print messages. Hence on this function, we have to create a new logger object to be able to print messages (this is how @msilvafe described it to me). Is there a workaround to this? Any suggestions? Maybe the logger stuff should also be moved to the library side.

mmccrackan: We've copied the logger stuff into preprocess/preprocess_util.py so that the preprocessing module also has its own logger available.  The mapmaking module could use that one or it could copy the logger functions into its own util file of some kind maybe.

chervias: ok, perfect. Can you change the logger to be created from preprocess_util then, please?

msilvafe:
> > Thanks @mmccrackan, some comments on things that are also done in #1014 which has not been merged yet. How should we rectify this? @msilvafe
> 
> It looks like the only major difference is that I didn't add the get_obslist function that you wrote. I can add that here since it looks useful for the multilayer_preprocessing. Since this pull request includes the preprocess_util function, I guess we can say that this branch supersedes #1014?
> 
> TODOs: We are also going to change things so that stuff in preprocess_util currently imported from site_pipeline util are either copied or moved into preprocess_util.
> 
> Finally, we will attempt to merge preprocess_tod, preprocess_obs, and multilayer_preprocess_tod into a single script.

Yes, I think this is a bit larger of an overhaul in preparation to merge preprocess_tod and preprocess_obs into one driving script, and we didn't really understand the overall repo philosophy about core -> main modules -> site pipeline only importing packages higher lower in the higherarchy and not higher so we need to pull everything out of site_pipeline.util so let's let this PR supersede #1014

sotodlib/coords/helpers.py
```diff
@@ -321,36 +321,37 @@ def get_footprint(tod, wcs_kernel, dets=None, timestamps=None, boresight=None,
     fp1 = so3g.proj.FocalPlane.from_xieta(fake_dets, xieta1[0], xieta1[1])
 
     asm = so3g.proj.Assembly.attach(sight, fp1)
-    output = np.zeros((len(fake_dets), n_samp, 4))
+    planar = np.zeros((len(fake_dets), n_samp, 4))
```
msilvafe, line 324: Why are there changes to `coords` in this PR?

sotodlib/preprocess/preprocess_util.py
```diff
@@ -0,0 +1,506 @@
...
+    """
+    def __init__(self, **kwargs):
+        self.root_dir = kwargs['root_dir']
+        self.pattern = kwargs['pattern']
+
+    def get_dest(self, **kw):
+        """Returns full path to destination directory.
+
+        """
+        return os.path.join(self.root_dir, self.pattern.format(**kw))
```
msilvafe, line 50: I think I'm a bit ambivalent about this as I think there were some hopes for these classes but they didn't really amount to much and just include the simplest return of h5group and file paths (so its probably fine to leave these here as is). But these are generally used by all manifestdb's not just preprocess so it seems like they could instead go in core.util or core.metadata.util (or were appropriately placed in site_pipeline.util since that's where the manifestdB generation scripts live). Do you care about this at all @mhasself or @kmharrington ?

sotodlib/preprocess/preprocess_util.py
```diff
@@ -0,0 +1,506 @@
...
+
+    logger.propagate = False
+    logger.setLevel(logging.DEBUG)
+
+    return logger
+
+
+def get_preprocess_context(configs, context=None):
+    """Load the provided config file and context file. To be used in
+    ``preprocess_*.py`` site pipeline scripts.
```
msilvafe, line 109: I've had a number of people asking about using this function lately so I think its worth adding to the docstring that the key difference between using this function and just `cfg = yaml.safe_load(open(configs, 'r'))`, `context = core.Context(cfg['context_file'])` is that if the provided `context.yaml` does not have a metadata entry for preprocess then it will add one based on the definition in the `preprocess.yaml` file.

msilvafe (approved):
Thanks for helping to restructure this. A couple comments inline to address but I'll approve.

sotodlib/coords/helpers.py
mmccrackan: This is from the fix for the failing pmat test that was pushed a few days ago.  I should have done a merge commit to make it more clear.

Issue #1011, opened 2024-10-19
Obs grouping: Get available wafers from obsdb


chervias:
Since the obsdb now includes the list of available wafers, we can read them from there instead of loading the metadata.

chervias:
Addressed by #1034

PR #1034, opened 2024-11-19
Load wafer list / freqs from obsdb

chervias:
Addressing #1011, we use the obsdb to get the wafer list. The freq list is hard coded (if mf we do f090, f150, etc.) We need to complete this for uhf and lf in the future.


msilvafe (approved):
Assuming you're tracking the eventual integration of LF this looks good.

Maybe add a placeholder in that if statement: i.e. 

```
elif row.tube_flavor == 'lf':
    raise ValueError('Band list for lf not implement.')
```

PR #1037, opened 2024-11-21
Faster book loading

mhasself:
Two improvements:

- Eliminate some unnecessary obsfiledb calls.
- Automatically g3 files to some faster disk before load.

The latter optimization is enabled if and only if `SOTODLIB_TOD_CACHE` envvar is set.  In which case, on `get_obs`, a tempdir is created under that CACHE dir, and files copied in, and loaded, and then the tempdir destroys itself.

The CACHE improvement, on NERSC, reduces load time by at least 50%.  In some cases (no_signal=True), the improvement is more like 80-90%.

This made no difference on site computing.

(An additional improvement which could be interesting would be to copy files in a thread, while reading the data in the foreground as each copy is completed.)

### DIFF
--- a/sotodlib/io/load_book.py
+++ b/sotodlib/io/load_book.py
@@ -16,6 +16,11 @@ The two access points in this submodule are:
     This can be used to load single G3 files (or a set of
     sample-contiguous G3 files) from a book, by filename.
 
+Reading from g3 files can be slow on some filesystems.  Use
+SOTODLIB_TOD_CACHE envvar to point to a directory where tempfiles can
+be safely made, for faster acccess, such as a scratch, local, or RAM
+disk.
+
 """
 
 import so3g
@@ -27,6 +32,8 @@ import itertools
 import logging
 import os
 import re
+import shutil
+import tempfile
 import yaml
 
 import sotodlib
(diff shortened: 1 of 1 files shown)


mhasself:
Concretely on NERSC, you can enable this feature like this:
```
mkdir $SCRATCH/tod_cache
export SOTODLIB_TOD_CACHE=$SCRATCH/tod_cache
```

tskisner:
Thanks for this work, I'll do some testing shortly.

chervias:
@mhasself I have used this a couple of times now in the mapmaker. I don't have exact numbers comparing before and after, but it seems to work great.

sotodlib/io/load_book.py
```diff
@@ -829,6 +850,10 @@ def _frames_iterator(files, prefix, samples, smurf_proc=None):
         filename = os.path.join(prefix, f)
         offset = i0
 
+        if use_temp_dir:
+            filename, orig_filename = os.path.join(use_temp_dir, 'framefile'), filename
+            shutil.copyfile(orig_filename, filename)
```
iparask, line 855: This can cause issues when two runs require the same observation and have the same tmp dir path. From `copyfile` docs: "If dst already exists, it will be replaced". Potentially this can happen while the file is open.

You can use pathlib to create the destination and check if it exists with something like:

```python
## Assuming use_temp_dir is an absolute path
filename = Path(use_temp_dir, 'framefile')
if not filename.exists():
    shutil.coyfile(...)
```

iparask (commented):
With a cache I would expect that the data are there for the next use, assuming that the next use is close in time. This cache dir works more like a tmp dir during runitme, while the `use_temp_dir` works like a cache from the user's side. 

I think the env name should reflect that, do you mind naming it with `_TMPDIR` instead of `_CACHE`?

Another issue I find is the recursive call just for the temp dir. How about something like this:
```python
    cache_dir = os.getenv('SOTODLIB_TOD_CACHE', None)
    if cache_dir is not None
        cache_dir = tempfile.TempDirectory(dir=cache_dir)
    use_temp_dir = use_temp_dir if use_temp_dir is not None else cache_dir

    ...

    if cache_dir is not None:
        cache_dir.cleanup()
```

tskisner (approved):
I agree with @iparask about the points concerning multiple processes writing to the same temp location.  One workaround at the job level would be to make a tempdir that is unique to each job.  We could also use the environment variable to specify a location in node-local ram disk.  In any case, I agree that these are nice features to have, and hopefully we can merge soon.

mhasself:
> With a cache I would expect that the data are there for the next use, assuming that the next use is close in time. This cache dir works more like a tmp dir during runitme, while the `use_temp_dir` works like a cache from the user's side.
> 
> I think the env name should reflect that, do you mind naming it with `_TMPDIR` instead of `_CACHE`?

Yes, agreed.  This is not a cache, it's a tmpdir.  I'll change the envvar.

> 
> Another issue I find is the recursive call just for the temp dir. How about something like this:

The single recursion is not expensive.  This was done to avoid needing to indent the whole function and wrap in a try/except.  "Users" aren't supposed to call this with `use_temp_dir` set.  The `_frames_iterator` is a private function here, and the only thing calling it with use_temp_dir!=None is ... itself.

Given that, and because the TempDir name is randomly generated, and not at all based on the obs_id, there's no real chance of two processes tripping each other up here.

So given that this _isn't a cache_, I don't think there's a functional problem with the implementation.  (But it might be nicer to avoid the recursion.)

iparask:
Based on the suggested signature and the docs, one can pass a string value to `use_temp_dir` and leave the env variable empty. Then the function will try to update the data in the same folder overwriting the existing.

If that is not the intended functionality, it is better if at least is not advertised in the docs, or make sure a call with a non-tempdir is not valid.

iparask (approved):

Issue #979, opened 2024-10-02
Preprocess: Generate detector level (within an observation) split flags


msilvafe:
The [DRP baseline design document](https://docs.google.com/document/d/1cBDbM8KD7DwPp6kfHI_dKUBOC2WwhxltWypEF6deiR0/edit?usp=sharing) outlines a set of null splits written by @chinoney. In order to make atomic maps with these null splits the mapmaker needs to take in a set of flags to choose which detectors (or samples in the case of left/right subscans) to bin into each map. These include:

**Splits based on focalplane/detector (detector-by-detector)**

1. High gain detectors vs. low gain detectors
2. High sensitivity detectors vs. low sensitivity detectors
3. High gain variation detectors vs. low gain variation detectors
4. High time constant detectors vs. low time constant detectors
6. Q-pixel/U-pixel
7. Sense-A/Sense-B for LF
8. “top”' (i.e. Ex^2) detectors vs. “bottom” (i.e. Ey^2) detectors - this is a jargon from PB. this means detectors of crossover or crossunder  
10. High leakage detectors vs. low leakage detectors
11. High 2f (4f) amplitude detectors vs. low amplitude detectors
12. High excess noise detectors at high frequency vs. low noise
14. The Left-half-side pixels vs. right-half-side pixels
15. The Top-half-side pixels vs. bottom-half-side pixels
16. The central pixels vs. the peripheral pixels

**Splits based on subscans**

1. Left vs. right-going sub-scans

We need a new low level Tod-ops module and associated preprocess method to generate these.

msilvafe:
Need to keep this open a bit longer because of the following:
1. We currently have no measure of gain drift. We need to add in a bias step before/after to `det_cal` or estimate this from variations in the HWP A2 but neither of these exist now.
2. I found A/B, Top/Bottom, and pol_angle but not Q/U. I chose to replace Q/U with a split on pol_angle.
4. I'm not sure I know the difference between "high sensitivity detectors vs low sensitivity detectors" and "high excess noise detectors at high frequency vs low noise" so right now I just have one high noise cut based on NEQ.
5. For the top/bottom and left/right focal plane split are those with respect to horizon coordinates or telescope coordinates (i.e. do I compensate for boresight rotation in those or not)? For now I've assumed fixed to telescope boresight coordinate system so that the subset of detectors does not change if boresight rotates.
6. We don't have subscan splitting now so I haven't implemented that but once we merge PR #1028 then this will be easy to add.

[...]

PR #1039, opened 2024-11-21
Iss979 map splits

msilvafe:
Addressing issue #979 still testing.


sotodlib/tod_ops/flags.py
```diff
@@ -672,3 +672,72 @@ def get_inv_var_flags(aman, signal_name='signal', nsigma=5,
...
+    split_aman.wrap('high_tau_avg', np.nanmean(aman.det_cal.tau_eff))
+    split_aman.wrap('det_A_flag', aman.det_info.wafer.pol <= split_cfg['det_A'],
+                    [(0, 'dets')])
+    split_aman.wrap('pol_angle_flag', aman.det_info.wafer.angle > split_cfg['pol_angle'], 
+                    [(0, 'dets')])
+    split_aman.wrap('det_top_flag', aman.det_info.wafer.crossover > split_cfg['det_top'], 
+                    [(0, 'dets')])
+    split_aman.wrap('high_leakage_flag', np.sqrt(proc_aman.t2p.lamQ**2 + proc_aman.t2p.lamU**2) > split_cfg['high_leakage'],
+                    [(0, 'dets')])
+    split_aman.wrap('high_leakage_avg', np.nanmean(np.sqrt(proc_aman.t2p.lamQ**2 + proc_aman.t2p.lamU**2)),
```
mmccrackan, line 734: Maybe this should either just be `leakage_average` if it is going to calculate it across all `lamQ` and `lamU` values or should we actually only calculate the average from the "high" values (or have two fields for high and low)?

sotodlib/tod_ops/flags.py
```diff
@@ -672,3 +672,72 @@ def get_inv_var_flags(aman, signal_name='signal', nsigma=5,
...
+                    [(0, 'dets')])
+    a2 = aman.det_cal.phase_to_pW*np.sqrt(proc_aman.hwpss_stats.coeffs[:,2]**2 + proc_aman.hwpss_stats.coeffs[:,3]**2)
+    split_aman.wrap('high_2f_flag', a2 > split_cfg['high_2f'], [(0, 'dets')])
+    split_aman.wrap('high_2f_avg', np.nanmean(a2), [(0, 'dets')])
```
mmccrackan, line 738: Same comment as `high_leakage_average` above for `high_2f_avg`.

mmccrackan (approved):
Looks great.  Just some optional comments/questions on naming and checking if some of the fields exist.

chervias:
Shouldn't the new `get_split_flags` function be in the `obs_ops.splits` module?

msilvafe:
> Shouldn't the new `get_split_flags` function be in the `obs_ops.splits` module?

Look ok now?

chervias (approved):
Look ok to me

chinoney:
Overall it looks nice to me.
I think the function generates flags for given criteria. In the function, only one side of the slits are wrapped into "am." I imagine "~" will be used to get the other side of the split.

I imagine we refer to names of various null splits everywhere in other codes. I think we need to unify/centralize them accordingly even it's not now.

Here do you turn on just "absolute" flags in this branch?

(The below are somehow general comments.) As I mentioned before, there're some types of the splits,
1. inter-observation splits
- observation splits based on detector summary statistics, e.g. high/low knee frequency observations=observations whose median of the knee frequency over the focalplane of the observation is higher/lower than the median from all the data
- the others, i.e. chronological, seasonal, environmental, and so on, e.g. 1st year vs. 2nd year, daytime vs. night time, close to Sun vs. far from Sun, etc,.
2. intra-observation splits
- absolute splits (=positional splits), i.e. left-side pixels vs. right-side pixels (on the focalplane),
- (another) absolute splits, e.g. high/low knee frequency detectors=detectors whose median of the knee frequency from all the observations is higher/lower than the median from all the data
- relative splits, e.g. relatively high/low knee frequency detectors=detectors whose knee frequency is higher/lower than the median from all the detectors of the observation
3. subscan splits
- left-going scans vs. right-going scans

Bundles of the inter-observations splits should be determined for each split because the atomic maps change for each split. Bundles of the intra-observation splits can be shared among the splits and common with bundles for the subscan splits and non-null test.

I think we should prioritize all of them except for the relative intra-obs splits because they are ambiguous compared to the others. For example, null tests of the high/low knee frequency observations/detectors are sensitive to corresponding observations/detectors. But the relative one is not.

msilvafe:
Ya I'm trying to do away completely with any relative splits in fact I prefer that we delete that function. The splits I've implemented here are only absolute, intra-observation splits. I've also provided some summary statistics (i.e. the mean fknee, mean gain, mean HWP A2, etc.) which can be used for inter-observation absolute splits however that will be handled in the atomic map dB + bundling further down the pipeline. Erik's recent PR for subscan processing that was just merged was what I was waiting on to add in the subscan split flags. I can add that in in a new PR. And yes the idea for all of these flags was one split is the flag and the other split is "~" of the flag.

PR #1040, opened 2024-11-22
fix missing preprocess logger

mmccrackan:
Fix the missing preprocess_util loggers bug introduced in #1027.


chervias (approved):
Looks good

PR #1041, opened 2024-11-22
Move PSD calculation to calc_and_save

mmccrackan:
This branch reworks the `calc_psd` handling in the preprocessing `PSD` pipeline stage.  Previously, the PSD calculation was carried out in the `process` function and wrapped into the axis manager.  When running the full pipeline, it was also wrapped into the processing axis manager, resulting in two copies.  The PSD would also be recomputed when loading from an existing database.

The `calc_psd` function has been moved into `calc_and_save` now so that it is only wrapped into the processing axis manager.  A slight change was made to the `Noise` stage as it expects the PSD to be present, so it now looks into the processing axis manager.  I didn't change the underlying behavior in `fit_noise_model` which will recompute the PSD and wrap it into the axis manager if the `PSD` is not an input.


msilvafe (approved):
Thank you for cleaning these all up!

PR #1043, opened 2024-11-25
Add lowpass filter to PCA Relcal

mmccrackan:
This branch adds an optional lowpass filter to the `PCARelCal `stage of the pipeline.  Previously, the lowpass was run as a separate filter stage before `PCARelCal` which results in a copy of timestreams being saved.  Now, the filtering occurs in the PCA `calc_and_save` function and is only temporarily stored.  The `lpf` configs, as well as the number of samples to trim, can be added to the `PCARelCal` config step.

### DIFF
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -1076,20 +1076,26 @@ class FourierFilter(_Preprocess):
 class PCARelCal(_Preprocess):
     """
     Estimate the relcal factor from the atmosphere using PCA.
-    
+
     Example configuration file entry::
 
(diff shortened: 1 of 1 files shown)


msilvafe (approved):
Looks good, happy to remove copies of data from the processing. Before merging can you just confirm that you ran this through a preproc pipeline w/ this branch and confirmed that the relcal outputs looked the same as when run through master?

mmccrackan:
Ran the same configuration through both the master branch and this branch, plotted all relcal fields against each other, and found all to be identical in dimensions and value.

PR #1044, opened 2024-11-27
SAT demod mapmaker updates

earosenberg:
Updates to SAT mapmaking scripts including:
- [x] ~~Add site_pipeline script to make atomic maps db~~ Write to atomic maps db from the mapmaker
- [x] Refactor `make_atomic_filterbin_map.py` to avoiding running `get_obs` twice
- [x] Make it possible to provide a wcs_kernel string (e.g. "car-5arcmin") rather than using the --area argument to provide a map filename


earosenberg:
@chervias
If we always make per-observation maps, should we use the obsid in the atomic map names instead of period start? This is sometimes confusing as the two often differ slightly. 
cf [these lines](https://github.com/simonsobs/sotodlib/blob/e09a212e2aac065e211de00a90a5bdb8bc34d59a/sotodlib/site_pipeline/make_atomic_filterbin_map.py#L383C4-L386)

chervias:
The actual important time is the ctime on the obsdb, which is when the obs actually started. In the atomic database we have both the obs_id and the path to the maps, so we can match obs and maps with the atomic database, right? I would say we can leave as it is now.

chervias:
I merged from the branch in #1037 to load obs faster. Please look at that PR for instructions, since an env variable has to be set up to use

chervias:
The PR is ready for review. Instead of using a separate script for appending to the atomic map DB, now everything is done in the same mapmaker as the maps are being written. Otherwise we have completed the other issues that is PR was meant to solve

chervias:
@mhasself could you please have a look? @JBorrow could you please have a look at the sqlalchemy bits? Thanks

sotodlib/mapmaking/utilities.py
```diff
@@ -533,3 +535,39 @@ def import_optional(module_name):
...
+
+class AtomicInfo(Base):
+    __tablename__ = "atomic"
+
+    obs_id: Mapped[str] = mapped_column(primary_key=True)
+    telescope: Mapped[str] = mapped_column(primary_key=True)
+    freq_channel: Mapped[str] = mapped_column(primary_key=True)
+    wafer: Mapped[str] = mapped_column(primary_key=True)
+    ctime: Mapped[int] = mapped_column(primary_key=True)
+    split_label: Mapped[str] = mapped_column(primary_key=True)
```
JBorrow, line 549: Is your aim here to have these be a composite primary key in the database?

sotodlib/site_pipeline/make_atomic_filterbin_map.py
```diff
@@ -419,49 +378,45 @@ def main(config_file: str) -> None:
...
-                                    split_label).encode(),
-                    elevation=obs_infos[obslist[0][3]].el_center,
-                    azimuth=obs_infos[obslist[0][3]].az_center,
-                    pwv=float(pwv_atomic)))
+                info = mapmaking.AtomicInfo(
+                    obslist[0][0],
+                    obs_infos[obslist[0][3]].telescope,
+                    band,
+                    detset,
+                    int(t))
```
JBorrow, line 407: I would recommend using kwargs here instead of positional

sotodlib/site_pipeline/make_atomic_filterbin_map.py
```diff
@@ -419,49 +378,45 @@ def main(config_file: str) -> None:
...
+                info.split_label = split_label
+                info.split_detail = ''
+                info.prefix_path = str(prefix + '_%s' % split_label)
+                info.elevation = obs_infos[obslist[0][3]].el_center
+                info.azimuth =obs_infos[obslist[0][3]].az_center
+                info.pwv = float(pwv_atomic)
+                info_list.append(info)
```
JBorrow, line 414: Why not just do this in the initializer?

sotodlib/mapmaking/utilities.py
JBorrow: Also why use ctime instead of a datetime object?

JBorrow (commented):
A few comments, happy to discuss further

sotodlib/mapmaking/utilities.py
chervias: yes, these are the things that combined define a unique atomic map. I thought a primary key was something that uniquely defines a row. Therefore, I made all these primary keys. ctime is just our tradition, we define observations by their ctime.

JBorrow: Yes, that's correct. You can certainly do this. One other option is to have a unique ID as the primary key, but if you are certain that _there will never be rows in the database with a duplicate value for these combined items_ then it is fine.

sotodlib/site_pipeline/make_atomic_filterbin_map.py
chervias: I guess I could do that in the init, but I reserved that for the primary keys that are mandatory, and then keep adding the optional things.

sotodlib/mapmaking/utilities.py
chervias: Ok. If I try to add another row to the database with the same set of these primary keys, it will error, right? That should never happen because these same keys are the ones used to check for existing maps at the beginning of the mapmaker. That way you don't re-map and existing map.

sotodlib/site_pipeline/make_atomic_filterbin_map.py
chervias: Could you elaborate a bit? I do not understand what you mean. Thanks

JBorrow: I would have written:
```
info = mapmaking.AtomicInfo(
    obs_id=obslist[0][0],
    telescope=obs_infos[obslist[0][3]].telescope,
    ...
)
```
This allows you to change the order of parameters, etc, later without impacting this code in potentially invisible ways.

sotodlib/mapmaking/utilities.py
chervias: It turns out this could happen. For example when you run a map with det_left_right splits and those are added to the database. Then let's say you run again but this time asking for the det_left_right split and the det_inner_outer split. Since you are missing the  det_inner_outer maps, you will run that entire obs again, and then you will make the det_left_right maps again, you will try to add to the database, there will be entries for those maps from the previous run, and it will complain. However we added a try catch statements to avoid the mapmaking from crashing.

JBorrow: Is that really preferable to having a non-composite primary key (like MapID or something that is just a self-increasing integer) and running a search for the map ahead of time?

sotodlib/mapmaking/demod_mapmaker.py
```diff
@@ -262,7 +264,7 @@ def for_healpix(cls, nside, nside_tile=None, comps="TQU", name="sky", ofmt="{nam
             The extension used for the files.
             May be 'fits', 'fits.gz', 'h5', 'h5py', or 'npy'.
         dtype : numpy.dtype
-            The data type to use for the time-ordered data.
+            The data type to use for the maps.
```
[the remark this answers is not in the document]
mhasself: (Whoops -- I am wrong about "missing some things" -- at least tiled and tile_shape.  I mis-read the source.  Drop that paragraph.)

iparask:
> In addition to inline comments -- I notice is that `mapmaking/__init__.py` does "import _" for all the submodules. I don't know how that slipped through, but please consider fixing as part of these changes. Likely no one is relying on the mapmaking._ population. In most cases it should simple be `from . import <submod>`. It's only acceptable to `import *` if you define an `__all__` list in the submodule, to limit namespace spamming. (That's a necessary, but not sufficient condition.)

As we said this is a much bigger work and I will refactor it as a separate development effort. That will include PR #1110.

chervias:
@mhasself comments addressed, except the treatment of the import in `__init__.py`. We agreed with @iparask that we would leave this for another PR. Also before merging to master I'll add a few more fields relevant for the atomic DB in the `AtomicInfo` class, which are mainly left unplemented in the mapmaker script (we fill them later on by hand).

mhasself (approved):
Great, thanks.

Please Squash on merge!

PR #1049, opened 2024-12-01
Make FlagManager.wrap_*() return the result, like AxisManager expects

amaurea:


chervias (approved):
I tested with the depth-1 mapmaker with downsampling and nothing seems broken. Thanks

Issue #1048, opened 2024-12-01
Union of flags into glitch_flags in preprocessing


chervias:
Right now, the wrapping of the flags for filter+bin demod mapmaking is [hard-coded in the map function](https://github.com/simonsobs/sotodlib/blob/307a65a1b4457def279e3a280c05b93f8f1b5d0a/sotodlib/mapmaking/demod_mapmaker.py#L524). This should be instead handled as a step in the preprocessing that makes the union of all desired flags into `obs.flags.glitch_flags`. I'm not completely sure, but I think this function does not exist yet.

chervias:
Addressed by PR #1050

PR #1050, opened 2024-12-02
Adding a union of flags function to preprocessing

chervias:
This PR intends to add a preprocessing function that creates a union of user-supplied flags into `aman.flags.glitch_flags`, which then the mapmaker cuts. This addresses #1048 
```
    - name: "union_flags"
      process:
        flag_labels: ['jumps_2pi.jump_flag', 'glitches.glitch_flags', 'turnaround_flags.turnarounds']
```

### DIFF
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -1388,25 +1388,6 @@ class UnionFlags(_Preprocess):
             aman['flags'].move(self.process_cfgs['total_flags_label'], None)
         aman['flags'].wrap(self.process_cfgs['total_flags_label'], proc_aman.total_flags)
 
-
-class PointingModel(_Preprocess):
-    """Apply pointing model to the TOD.
-
-    Saves results in proc_aman under the "pointing" field. 
-
-     Example config block::
-
-        - name : "pointing_model"
-          calc: True
-          
-    .. autofunction:: sotodlib.hwp.hwp_angle_model.apply_hwp_angle_model
-    """
-    name = "pointing_model"
-    
-    def calc_and_save(self, aman, proc_aman):
-        from sotodlib.coords import pointing_model
-        if self.calc_cfgs:
-            pointing_model.apply_pointing_model(aman)
             
 _Preprocess.register(SplitFlags)
 _Preprocess.register(SubtractT2P)
@@ -1438,4 +1419,3 @@ _Preprocess.register(DarkDets)
 _Preprocess.register(SourceFlags)
 _Preprocess.register(HWPAngleModel)
 _Preprocess.register(UnionFlags)
-_Preprocess.register(PointingModel)
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -1490,6 +1490,24 @@ class UnionFlags(_Preprocess):
             aman['flags'].move(self.process_cfgs['total_flags_label'], None)
         aman['flags'].wrap(self.process_cfgs['total_flags_label'], total_flags)
 
+class PointingModel(_Preprocess):
+    """Apply pointing model to the TOD.
+
+    Saves results in proc_aman under the "pointing" field. 
+
+     Example config block::
+
+        - name : "pointing_model"
+          process: True
+          
+    .. autofunction:: sotodlib.coords.pointing_model.apply_pointing_model
+    """
+    name = "pointing_model"
+
+    def process(self, aman, proc_aman):
+        from sotodlib.coords import pointing_model
+        if self.process_cfgs:
+            pointing_model.apply_pointing_model(aman)
             
 _Preprocess.register(SplitFlags)
 _Preprocess.register(SubtractT2P)
@@ -1522,3 +1540,4 @@ _Preprocess.register(SourceFlags)
 _Preprocess.register(HWPAngleModel)
 _Preprocess.register(GetStats)
 _Preprocess.register(UnionFlags)
+_Preprocess.register(PointingModel)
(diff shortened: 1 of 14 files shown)


chervias:
On the last line of the function, which wraps the union of flags into `aman`, I'm getting an error of mismatched shapes, which suggests `aman` and `proc_aman` might have different samps shape, which should not happen.

chervias:
#1051 fixed the bug, so now we are ready for a review

sotodlib/preprocess/processes.py
```diff
@@ -1355,6 +1357,28 @@ def save(self, proc_aman, split_flg_aman):
...
+    """Do the union of relevant flags for mapping
+    Typically you would include turnarounds, glitches, etc.
+
+    Saves results for aman under the "flags.glitch_flags" field.
+
+     Example config block::
+
+        - name : "union_flags"
+          process:
+            flag_labels: ['jumps_2pi.jump_flag', 'glitches.glitch_flags', 'turnaround_flags.turnarounds']
```
mmccrackan, line 1370: Possibly unrelated to this PR specifically, but I do see that `aman.flags` is already populated with `left_scan[samps]`, `right_scan[samps]`, and `turnarounds[samps]` (`merge=True` by default) but not by `glitch_flags` or `jump_flags` (`merge=False` is hardcoded).  These should probably standardized in another PR.  If all flags are merged into `aman.flags` by default, we could just search and merge everything in `aman.flags` instead of specifying them via the config file.

sotodlib/preprocess/processes.py
```diff
@@ -1355,6 +1357,28 @@ def save(self, proc_aman, split_flg_aman):
...
+
+    """
+    name = "union_flags"
+    
+    def process(self, aman, proc_aman):
+        total_flags = so3g.proj.RangesMatrix.zeros([proc_aman.dets.count, proc_aman.samps.count]) # get an empty flags with shape (Ndets,Nsamps)
+        for label in self.process_cfgs['flag_labels']:
+            _label = attrgetter(label)
+            total_flags += _label(proc_aman) # The + operator is the union operator in this case
+        aman['flags'].wrap('glitch_flags', total_flags)
```
mmccrackan, line 1380: Should the union of the jump, glitch, and turnarounds be called glitch_flags, or should they maybe be called something more general (maybe just `aman.flags['flags']` or something like that) since all of those are not really related to glitches?  Also, if the glitch flag axis manager is merged in `get_glitch_flags`, there can be a potential name conflict.

Also, it might be useful to have a check whether a flag axis manager exists and maybe a clean error message if it doesn't.

mmccrackan (approved):
Just a few minor comments, but otherwise I think everything makes sense and works.

sotodlib/preprocess/processes.py
chervias: So by default, an obs has an empty `FlagManager` in `aman.flags`. When you do the turnarounds function the `left_scan`, `right_scan`, and `turnarounds` flags are populated. I think we still should keep some amount of control such that in the preprocess config we can specify exactly what we want to go into the overall glitch flags.

sotodlib/preprocess/processes.py
chervias: This is the ACT legacy. In ACT the overall flags are called glitch_flags. We should do a more general name and we should discuss it, but for now, we have the ACT convention all over sotodlib.

sotodlib/preprocess/processes.py
```diff
@@ -14,6 +14,8 @@
 from .pcore import _Preprocess, _FracFlaggedMixIn
 from .. import flag_utils
 
+import so3g
```
msilvafe, line 17: Instead of a blanket import please import just the modules you need `from so3g.proj import RangesMatrix`

sotodlib/preprocess/processes.py
msilvafe: I agree that we want to keep the flexibility in which flags get combined to form the set of flags used in the mapmaking and not blinding search for all flags and combine them.

sotodlib/preprocess/processes.py
msilvafe: I agree with both of Michael's points, fine to push off the change away from the `glitch_flags` naming convention from ACT to another PR but please open an issue for it.

I also agree that there needs to be some logic here for handling `aman.flags` if it doesn't exist you need to create it, and if it exists but `glitch_flags` is already populated you need to first delete what's in that axis before writing to it.

sotodlib/preprocess/processes.py
```diff
@@ -1355,6 +1357,28 @@ def save(self, proc_aman, split_flg_aman):
...
+     Example config block::
+
+        - name : "union_flags"
+          process:
+            flag_labels: ['jumps_2pi.jump_flag', 'glitches.glitch_flags', 'turnaround_flags.turnarounds']
+
+    """
+    name = "union_flags"
+    
+    def process(self, aman, proc_aman):
```
msilvafe, line 1375: The union of the flags can be a `calc_and_save` i.e. creation of `total_flags` up to line 1379 then wrapping `total_flags` into `aman.flags.glitch_flags` (along with the added logic I mention in my comment below) can be a `process`

msilvafe (changes requested):
Couple requested changes to address then hit re-review. Thanks!

sotodlib/preprocess/processes.py
chervias: Ok I made a new PR

chervias:
Ok, bugs are fixed, now it is working

msilvafe (approved):
Thanks for patching this.

PR #1051, opened 2024-12-02
fix mismatching samples

mmccrackan:
This branch should address the comment made in #1050 where the number of samples between original axis manager and the processed axis manager are not the same.  The difference occurred because the original axis manager was trimmed during demodulation and the processed axis manager was then trimmed based on the already trimmed axis manager, resulting in even fewer samples.


chervias (approved):
Ok, tested and it works. Thanks

PR #1053, opened 2024-12-04
Toast mlmapmaker updates

keskitalo:


sotodlib/toast/ops/mlmapmaker.py
```diff
@@ -163,12 +185,15 @@ class MLMapmaker(Operator):
     weather = Unicode("vacuum", help="Weather to assume when making maps")
     site    = Unicode("so",     help="Site to use when making maps")
 
-    maxiter = Int(500, help="Maximum number of CG iterations")
+    maxiter = List(
+        [500],
+        help="Maximum number of CG iterations. Use commas to control each pass separately.",
```
tskisner, line 190: Maybe change the help string to explicitly say, "List of maximum number of CG iterations for each pass".

tests/test_mapmaker_pointing.py
```diff
@@ -26,7 +26,10 @@
 if toast_available:
     import healpy as hp
 
-from ._helpers import calibration_schedule, close_data_and_comm, simulation_test_data
+try:
+    from . import _helpers as helpers
+except ImportError:
+    import _helpers as helpers
```
tskisner, line 32: Why would the relative import fail?  We want to make sure we are always importing the `_helpers` module located in the `tests` directory, not in other locations.

tskisner (approved):
Just a couple of comments for consideration.

Thanks for refactoring the obs wrapping, which will be useful for the next steps of wrapping the sotodlib binner.

sotodlib/toast/ops/mlmapmaker.py
keskitalo: Good point. Changed.

tests/test_mapmaker_pointing.py
keskitalo: With this modification, I can execute this one unit test individually with ```python tests/test_mapmaker_pointing.py```.

tskisner: Instead of just doing
```
python -m unittest tests/test_mapmaker_pointing.py
```
?

keskitalo: yes

keskitalo: reverted

PR #1055, opened 2024-12-04
Changing the format of the input parameters of the main function in the mapmaker for prefect

chervias:
This updates the way the parameters are fed into the main function. This is for prefect compatibility.


sotodlib/site_pipeline/make_atomic_filterbin_map.py
```diff
@@ -226,7 +216,41 @@ def future_write_to_log(e, errlog):
...
+    dec_ref: Optional[float] = -40.0,
+    center_at: Optional[str] = None,
+    max_dets: Optional[int] = None,
+    fixed_time: Optional[int] = None,
+    min_dur: Optional[int] = None,
+    verbose: Optional[int] = 0,
+    quiet: Optional[int] = 0,
+    window: Optional[float] = None,
+    dtype_tod: Optional[str] = 'float32',
+    dtype_map: Optional[str] = 'float64'):
```
iparask, line 252: Carlos just an FYI, `Optional` is usually used when the parameter can be `None`. If it is not passed and it gets a non-None value then the optional is not needed.

PR #1056, opened 2024-12-04
Changing the format of input params for the main function for prefect compatibility

chervias:
This just changes the way parameters are fed to the main function so that it is compatible with prefect.


msilvafe (commented):
There are many arguments here. I would suggest with this many arguments that you put all but the most frequently changed into a yaml config file. And then the main function/argparse is just accepting the config file path and a few other arguments.

chervias:
I don't know how to do that, do you have an example I can look at?

chervias:
Ok, I changed the format to be the same as update_det_cal, now we receive a single parameter which is a yaml config.

msilvafe (approved):
I suggest you put an example config file entry showing all possible arguments in `docs/site_pipeline.rst` but otherwise I like this better than the very long set of main arguments. Thanks.

chervias:
Thanks. There is an example config file in docs, not with all params but the essential ones.

PR #1058, opened 2024-12-04
coords: get_focal_plane_cover masking

mhasself:
- ignores nan positions
- processes optional "det_weights" for user-specified considerations


chervias (approved):
Looks good

PR #1059, opened 2024-12-04
wip: removing python 38

iparask:


iparask:
This PR removes python 3.8 as a requirement for sotodlib. It increments the requirement in setup.py, removes it from the test workflow.

It also removes the deprecated `setuptools.command.test` and updates the test to use `python -m unittest discover`.

BrianJKoopman:
I'm not really a maintainer of this repo (though I am glad to see this) so I'm going to back out of the requested review. If there are any specifics that I might be able to help with feel free to ping me.

mhasself (approved):
Farewell python 3.8.  I will so miss not being able to use the | operator on dictionaries.

Wait for @tskisner approval on the TestCommand stuff -- I don't know about that...

.github/workflows/test.yml
```diff
@@ -62,13 +60,13 @@ jobs:
...
           export MPI_DISABLE=1
-          python3 setup.py test
+          python3 -m unittest discover
 
       - name: Run MPI Tests
         run: |
           export OMP_NUM_THREADS=1
           export OPENBLAS_NUM_THREADS=1
-          mpirun -np 2 python3 setup.py test
+          python3 -m unittest discover
```
tskisner, line 69: Can we restore the `mpirun` here?  Even 2 processes is enough to catch some problems.

tskisner (approved):
Looks good, after restoring the MPI tests.

PR #1060, opened 2024-12-09
Putting the mapmaker in the same split convention as the preprocessing

chervias:
This changes how the mapmaker handles the focal plane splits, so that they follow the same convention as preprocessing.
Also fixes a minor bug in preprocess.

### DIFF
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -1450,7 +1450,7 @@ class SplitFlags(_Preprocess):
     name = "split_flags"
 
     def calc_and_save(self, aman, proc_aman):
-        split_flg_aman = obs_ops.flags.get_split_flags(aman, proc_aman, split_cfg=self.calc_cfgs)
+        split_flg_aman = obs_ops.splits.get_split_flags(aman, proc_aman, split_cfg=self.calc_cfgs)
 
         self.save(proc_aman, split_flg_aman)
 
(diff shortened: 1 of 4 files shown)


sotodlib/preprocess/processes.py
```diff
@@ -1450,7 +1450,7 @@ class SplitFlags(_Preprocess):
     name = "split_flags"
 
     def calc_and_save(self, aman, proc_aman):
-        split_flg_aman = obs_ops.flags.get_split_flags(aman, proc_aman, split_cfg=self.calc_cfgs)
+        split_flg_aman = obs_ops.splits.get_split_flags(aman, proc_aman, split_cfg=self.calc_cfgs)
```
mmccrackan, line 1453: Though its not included in the changes here, there is a formatting issue for the docstring in `obs_ops.splits`. This might be a good PR to fix that on.  See https://github.com/simonsobs/sotodlib/blob/686e4ab01415deccfbe0e25fb45d331232c688ba/sotodlib/obs_ops/splits.py#L67

mmccrackan (approved):
Looks good.

sotodlib/preprocess/processes.py
chervias: ok, I'll add this

earosenberg (changes requested):
Looks okay to me; would be nice to address comment on splits.py

earosenberg (approved):
Looks great, thanks

PR #1063, opened 2024-12-09
load_book: don't over-optimize getting the file list

mhasself:
Although it's usually an error load dets=[], make-hwp-solution does that, using a basic context. This re-enables that support by optimizing obsfiledb.get_files.


guanyilun (approved):
Thanks for the fix! I confirm that it resolves the make-hwp-solution errors I see.

PR #1077, opened 2024-12-20
Add try except to catch det_match failures

mmccrackan:
Adds  some error checking around `preprocess_utils.get_groups` to catch failures related to getting det info.

### DIFF
--- a/sotodlib/site_pipeline/multilayer_preprocess_tod.py
+++ b/sotodlib/site_pipeline/multilayer_preprocess_tod.py
@@ -66,11 +66,31 @@ def multilayer_preprocess_tod(obs_id,
         configs_proc = yaml.safe_load(open(configs_proc, "r"))
     context_proc = core.Context(configs_proc["context_file"])
 
-    group_by_proc, groups_proc, _ = pp_util.get_groups(obs_id, configs_proc, context_proc)
+    group_by_init, groups_init, error_init = pp_util.get_groups(obs_id, configs_init, context_init)
+    group_by_proc, groups_proc, error_proc = pp_util.get_groups(obs_id, configs_proc, context_proc)
+
+    if error_init is not None:
+        if run_parallel:
+            return error_init[0], [None, None], [None, None]
+        else:
+            return
+
+    if error_proc is not None:
+        if run_parallel:
+            return error_proc[0], [None, None], [None, None]
+        else:
+            return
+
+    if len(groups_init) > 0 and len(groups_proc) > 0:
+        if (group_by_init != group_by_proc).any():
+            raise ValueError('init and proc groups do not match')
 
     all_groups_proc = groups_proc.copy()
     for g in all_groups_proc:
         if group_list is not None:
+            if g not in groups_init:
+                groups_proc.remove(g)
+                continue
             if g not in group_list:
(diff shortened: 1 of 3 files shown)


msilvafe:
Need to incorporate updates to file cleanup in the demod mapmaking functions.

sotodlib/site_pipeline/multilayer_preprocess_tod.py
```diff
@@ -281,76 +301,62 @@ def main(configs_init: str,
...
+                    found = True
+                    break
+
+            if found:
+                error = pp_util.save_group_and_cleanup(obs_id, configs_init, context_init,
+                                                       subdir='temp', remove=overwrite)
+                if error is not None:
+                    f = open(errlog, 'a')
+                    f.write(f'\n{time.time()}, cleanup error\n{error[0]}\n{error[2]}\n')
+                    f.close()
```
msilvafe, line 321: Can this whole block be a single function that takes in `policy_dir`, `obs_id`, `config`, `context`, `subdir`, `remove`, and `errlog` so we can just call that line twice here (and in other places where its needed i.e. preprocess_tod, demod mapmaker)

sotodlib/site_pipeline/multilayer_preprocess_tod.py
```diff
@@ -281,76 +301,62 @@ def main(configs_init: str,
...
-
-    nfile_proc = 0
-    folder_proc = os.path.dirname(configs_proc['archive']['policy']['filename'])
-    basename_proc = os.path.splitext(configs_proc['archive']['policy']['filename'])[0]
-    dest_file_proc = basename_proc + '_' + str(nfile_proc).zfill(3) + '.h5'
-    if not(os.path.exists(folder_proc)):
-        os.makedirs(folder_proc)
-    while os.path.exists(dest_file_proc) and os.path.getsize(dest_file_proc) > 10e9:
-        nfile_proc += 1
-        dest_file_proc = basename_proc + '_' + str(nfile_proc).zfill(3) + '.h5'
```
msilvafe, line 352: Why can we get rid of this?

sotodlib/site_pipeline/preprocess_tod.py
```diff
@@ -325,28 +332,41 @@ def main(
         logger.warning(f"No observations returned from query: {query}")
 
     # clean up lingering files from previous incomplete runs
-    for obs in obs_list:
-        obs_id = obs['obs_id']
-        pp_util.save_group_and_cleanup(obs_id, configs, context,
-                                       subdir='temp', remove=overwrite)
+    policy_dir = os.path.dirname(configs['archive']['policy']['filename']) + '/temp/'
```
msilvafe, line 335: join

msilvafe (approved):
Thanks for this update and the associated testing. Added a couple inline comments but also approved.

sotodlib/site_pipeline/multilayer_preprocess_tod.py
mmccrackan: This logic is inside `cleanup_mandb` so these lines are unused.  I will remove the corresponding lines from `preprocess_tod.py` since we don't need them there either.

PR #1079, opened 2024-12-20
Source flags update

mmccrackan:
Updates `SourceFlags` preprocess functions.  Adds a check to see if the input source is nearby and does the flagging as before.  Returns a zero `RangeMatrix` if no source is found.  Now, it accepts a list of sources and loops through them which are wrapped into `proc_aman.source_flags.source_name`.  Currently requires `FocalplaneNanFlags` to be run first so that any focal plane `NaNs` are removed.

### DIFF
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -1,6 +1,8 @@
 import numpy as np
 from operator import attrgetter
 
+from so3g.proj import Ranges, RangesMatrix
+
 import sotodlib.core as core
 import sotodlib.tod_ops as tod_ops
 import sotodlib.obs_ops as obs_ops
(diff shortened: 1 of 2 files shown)


sotodlib/preprocess/processes.py
```diff
@@ -1019,58 +1021,80 @@ class SourceFlags(_Preprocess):
...
+        source_list = []
+        for center_on in center_on_list:
+            # Get source from tags
+            if center_on == 'planet':
+                from sotodlib.coords.planets import SOURCE_LIST
+                matches = [x for x in aman.tags if x in SOURCE_LIST]
+                if len(matches) != 0:
+                    source = matches[0]
+                else:
+                    raise ValueError("No tags match source list")
```
msilvafe, line 1050: This doesn't give you a list, it'll always just pick out the first match. I think you want to delete this for loop and to have this check right after `calc_cfgs.get('center_on', 'planet')` then what's called `matches` here can be `source_list` and you don't need `center_on_list` at all you just need `source_list`. Also I don't think `center_on` can ever be planet (L1044) as its a string so when you try to iterate over it it'll fail (addressed via my previous comment) on L1027.

msilvafe (approved):
Couple inline comments to address but approved.

Issue #975, opened 2024-10-01
Jump Improvements


msilvafe:
We want to address 3 main issues with the current jump finding:

1. Speed it up
2. Reduce false positives
3. Reduce false negatives

This is to track progress on this. Current PR open addressing this is PR #807

kwolz:
Here is an example of the type of diagnostic plot I used to check the jump fixing algorithm. Left is the whole TOD, right is a zoom in to the jump-flagged region (yes, this requires the jump finder to have found a jump before). The function (tentatively documented, look for) is t[his one](https://github.com/simonsobs/pwg-scripts/blob/b1835a85ca19b39adbe785f28591be000cf65938/flp/sat-map-processing/kwolz/scripts/jump_validation.py#L170). ![Image](https://github.com/user-attachments/assets/4ca1a73d-3b88-49c3-8e69-f4ef5af9964d)

PR #1081, opened 2024-12-26
feat: move slow jumps and jumpfix to so3g

skhrg:
Relies on https://github.com/simonsobs/so3g/pull/196 
This is a direct rewrite for the most part, no algorithmic changes.

### DIFF
--- a/sotodlib/tod_ops/jumps.py
+++ b/sotodlib/tod_ops/jumps.py
@@ -4,14 +4,18 @@ import numpy as np
 import scipy.ndimage as simg
 import scipy.stats as ss
 from numpy.typing import NDArray
-from pixell.utils import block_expand, block_reduce, moveaxis
+from pixell.utils import moveaxis
 from scipy.sparse import csr_array
 from skimage.restoration import denoise_tv_chambolle
 from so3g import (
-    matched_jumps,
-    matched_jumps64,
+    block_minmax,
+    block_minmax64,
     find_quantized_jumps,
     find_quantized_jumps64,
+    matched_jumps,
+    matched_jumps64,
+    subtract_jump_heights,
+    subtract_jump_heights64,
 )
 from so3g.proj import Ranges, RangesMatrix
 from sotodlib.core import AxisManager
@@ -182,33 +186,39 @@ def jumpfix_subtract_heights(
         x_fixed: x with jumps removed.
                  If inplace is True this is just a reference to x.
     """
-
-    def _fix(jump_ranges, heights, x_fixed):
-        for j, jump_range in enumerate(jump_ranges):
-            for start, end in jump_range.ranges():
-                _heights = heights[j, start:end]
-                height = _heights[np.argmax(np.abs(_heights))]
-                x_fixed[j, int((start + end) / 2) :] -= height
-
+    orig_shape = x.shape
+    x = np.atleast_2d(x)
+    x = np.ascontiguousarray(x)
     x_fixed = x
     if not inplace:
         x_fixed = x.copy()
-    orig_shape = x.shape
-    x_fixed = np.atleast_2d(x_fixed)
+        x_fixed = np.ascontiguousarray(x_fixed)
     if isinstance(jumps, np.ndarray):
         jumps = RangesMatrix.from_mask(np.atleast_2d(jumps))
     elif isinstance(jumps, Ranges):
         jumps = RangesMatrix.from_mask(np.atleast_2d(jumps.mask()))
     if not isinstance(jumps, RangesMatrix):
         raise TypeError("jumps not RangesMatrix or convertable to RangesMatrix")
+    jumps = cast(RangesMatrix, jumps)
 
     if heights is None:
         heights = estimate_heights(x_fixed, jumps.mask(), **kwargs)
     elif isinstance(heights, csr_array):
         heights = heights.toarray()
     heights = cast(NDArray[np.floating], heights)
+    heights = heights.astype(x.dtype)
+    heights = np.ascontiguousarray(heights)
 
-    _fix(jumps.ranges, heights, x_fixed)
+    dtype = x.dtype.name
+    if len(x.shape) > 2:
+        raise ValueError("x may not have more than 2 dimensions")
+    if dtype == "float32":
+        fix = subtract_jump_heights
+    elif dtype == "float64":
+        fix = subtract_jump_heights64
+    else:
+        raise TypeError("x must be float32 or float64")
+    fix(x, x_fixed, heights, jumps)
 
     return x_fixed.reshape(orig_shape)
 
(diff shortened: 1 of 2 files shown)


kwolz (commented):
I've looked into the changes. I have only two small comments for my ease of understanding, otherwise it looks good to me.

sotodlib/tod_ops/jumps.py
```diff
@@ -182,33 +186,39 @@ def jumpfix_subtract_heights(
...
-    orig_shape = x.shape
-    x_fixed = np.atleast_2d(x_fixed)
+        x_fixed = np.ascontiguousarray(x_fixed)
     if isinstance(jumps, np.ndarray):
         jumps = RangesMatrix.from_mask(np.atleast_2d(jumps))
     elif isinstance(jumps, Ranges):
         jumps = RangesMatrix.from_mask(np.atleast_2d(jumps.mask()))
     if not isinstance(jumps, RangesMatrix):
         raise TypeError("jumps not RangesMatrix or convertable to RangesMatrix")
+    jumps = cast(RangesMatrix, jumps)
```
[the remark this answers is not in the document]
skhrg: Honestly that was just to make my typechecker shut up...

tests/test_tod_ops.py
```diff
@@ -310,6 +311,10 @@ def test_jumpfinder(self):
         heights = heights[heights.nonzero()].ravel()
         self.assertTrue(np.all(np.abs(np.array([10, -13, -8]) - np.round(heights)) < 3))
 
+        # Check fixing
+        ptp_fix = np.ptp(fixed.ravel()[~jumps.buffer(10).mask().ravel()])
+        self.assertTrue(ptp_fix < 1.1*ptp_orig)
```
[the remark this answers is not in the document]
skhrg: Yup, upped the buffer in the test to be conservative so that things never fail in CI. But also I set the PRNG key so I think that is unneeded.

mmccrackan (commented):
I think this looks fine other than the failing checks.

skhrg:
https://github.com/simonsobs/so3g/pull/196 is now merged so this should be good to go very soon

sotodlib/tod_ops/jumps.py
```diff
@@ -4,14 +4,18 @@
...
 import scipy.stats as ss
 from numpy.typing import NDArray
-from pixell.utils import block_expand, block_reduce, moveaxis
+from pixell.utils import moveaxis
 from scipy.sparse import csr_array
 from skimage.restoration import denoise_tv_chambolle
 from so3g import (
-    matched_jumps,
-    matched_jumps64,
+    block_minmax,
```
mhasself, line 11: One way to create some backwards-compatibility is to not import these by name.  Just import so3g.  Then the code will crash when you try to use those functions, rather than when you try to import them.

As written now, everyone's installation will break, if they try to use _anything_ from `sotodlib.tod_ops`, unless they've updated so3g.

skhrg:
Now that the new so3g is out can someone approve this?

sotodlib/tod_ops/jumps.py
```diff
@@ -3,16 +3,11 @@
 import numpy as np
 import scipy.ndimage as simg
 import scipy.stats as ss
+import so3g
 from numpy.typing import NDArray
-from pixell.utils import block_expand, block_reduce, moveaxis
+from pixell.utils import moveaxis
```
mhasself, line 8: Replace the one remaining moveaxis with np.moveaxis.

sotodlib/tod_ops/jumps.py
```diff
@@ -182,33 +177,39 @@ def jumpfix_subtract_heights(
...
-
+    orig_shape = x.shape
+    x = np.atleast_2d(x)
+    x = np.ascontiguousarray(x)
     x_fixed = x
     if not inplace:
         x_fixed = x.copy()
-    orig_shape = x.shape
-    x_fixed = np.atleast_2d(x_fixed)
+        x_fixed = np.ascontiguousarray(x_fixed)
```
mhasself, line 186: To reduce unnecessary reallocations, make use of `np.asarray`.  To do these two lines, for example, you can just:  `x_fixed = np.asarray(x, order='C', copy=True)`

mmccrackan (approved):

sotodlib/tod_ops/jumps.py
```diff
@@ -182,33 +177,39 @@ def jumpfix_subtract_heights(
...
         raise TypeError("jumps not RangesMatrix or convertable to RangesMatrix")
+    jumps = cast(RangesMatrix, jumps)
 
     if heights is None:
         heights = estimate_heights(x_fixed, jumps.mask(), **kwargs)
     elif isinstance(heights, csr_array):
         heights = heights.toarray()
     heights = cast(NDArray[np.floating], heights)
+    heights = heights.astype(x.dtype)
+    heights = np.ascontiguousarray(heights)
```
mhasself, line 201:
```
# This will only copy if needed for dtype/order.
heights = np.asarray(heights, dtype=x.dtype, order='C')
```

sotodlib/tod_ops/jumps.py
```diff
@@ -182,33 +177,39 @@ def jumpfix_subtract_heights(
...
-    def _fix(jump_ranges, heights, x_fixed):
-        for j, jump_range in enumerate(jump_ranges):
-            for start, end in jump_range.ranges():
-                _heights = heights[j, start:end]
-                height = _heights[np.argmax(np.abs(_heights))]
-                x_fixed[j, int((start + end) / 2) :] -= height
-
+    orig_shape = x.shape
+    x = np.atleast_2d(x)
+    x = np.ascontiguousarray(x)
```
mhasself, line 182: If user requested inplace, and x is not contiguous, then the change won't be in-place.  So you need to check for that.

For example
```
x = np.asarray(x, order='C', copy=(False if inplace else None))
```

mhasself (changes requested):
Minor comments ... looking forward to seeing this out in the wild.

sotodlib/tod_ops/jumps.py
skhrg: Do we really want to raise a value error if inplace is requested but the array is not contiguous?
I could instead raise a warning, make a copy, and then copy back into the original `x` later. But that may be babying the user too much.

skhrg: Actually assuming we want to raise the error in that case is the most correct block something like:

```
x_fixed = np.asarray(np.atleast_2d(x), order="C", copy=(False if inplace else None))
```

That way the cases are as follows:

* in place and already contiguous -> `x_fixed` just references `x`
* in place and not contiguous -> `ValueError`
* no in place and already contiguous -> `x_fixed` is a copy of `x`
* no in place and not contiguous -> `x_fixed` is a contiguous copy of `x`

Does that make sense?

mhasself: W.r.t. "no in place and already contiguous -> x_fixed is a copy of x" -- actually not true; `np.asarray(... copy=None)` will not copy if it doesn't need to.

Another way to approach this, to minimize copies and simplify logic:
- Apply all the conversions you need to get `x_fixed`, minimizing copies where possible.
- *Check* whether you ended up making a copy, or not -- `fixed_is_a_copy = np.may_share_memory(x, x_fixed)`
- Take different actions depending on `(fixed_is_a_copy, inplace)`.

I'm ok with your suggestion of making a copy and then updating x at the end.

The reason to ValueError would be that "inplace" is sometimes used not just to compress code, but to minimize copying of the data -- it's sort of implied that doing things inplace will use less RAM, less time.  But I think printing a warning is sufficient to help super-optimizers tweak things up.

skhrg:
Tests failing with ` TypeError: asarray() got an unexpected keyword argument 'copy'`, looks like that was added in numpy 2.0.0 (and I believe the PR for numpy 2.0 support in so3g is not yet ready).

I'll redo some logic to sidestep this... but hopefully in the future we can switch to it

skhrg:
...the problem with devving on both my laptop and site computing is sometimes I don't push all my fixes.
Should pass now.

mhasself (approved):

msilvafe:
@skhrg does this close out Issue #975 ?

PR #1082, opened 2025-01-02
Data Packaging: Updates for repairing failed books

kmharrington:
* Add autofixing for operations books with NoMountData errors (we generally don't care too strongly about the mount positions where operations take place, and/or it's find to miss the positions of some of these)
* Add option to skip the check that requires HK times to be monotonically increasing. We've seen this now around time periods where NoMountData was also getting thrown, we're going to want to be very careful about actually using this flag, (the time I saw this and I've checked that data, the mount was stationary the whole time, so I'm not worried about interpolations for that one)
* If librarian checking fails, try again (hoping that helps with caching errors at NERSC/Princeton)
* Allow over-riding ctimes for manual book making. Used once or twice when the ACU dropped out. 
* Check level 2 file size and throw errors if the files are too large. This makes it easier to flag and clear out smurf communication crashes
* Check for missing / dropped timesamples per stream_id. Throw error if more than 100 are dropped at once and report the number dropped (filled_samples) in the M_index.yaml so we can find them easier. This is all related to this discussion: https://github.com/simonsobs/daq-discussions/discussions/144

### DIFF
--- a/sotodlib/io/bookbinder.py
+++ b/sotodlib/io/bookbinder.py
@@ -15,7 +15,7 @@ import sys
 import shutil
 import yaml
 import datetime as dt
-from zipfile import ZipFile
+from zipfile import ZipFile, ZIP_DEFLATED
 import sotodlib
 from sotodlib.site_pipeline.util import init_logger
 from .datapkg_utils import walk_files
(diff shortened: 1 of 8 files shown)


kmharrington:
This has been running on-site for awhile and since I'm about to do some larger changes I'm just going to merge this setup.

PR #1033, opened 2024-11-17
coords.pmat: fix tiling .from_map behavior

mhasself:
Re-enable the test for rectpix tiled case.  This required a lot of custom checking for the rectpix+tiled case... and when combined with the healpix checking, this was hard to read.  So I've also abstracted that away using the notion of "flatten" and "unflatten" for maps.  This affects healpix and rectpix.

PR #1084, opened 2025-01-03
fix _apply_inverse_weights_map

ykyohei:
https://github.com/simonsobs/sotodlib/pull/1033 changed the return type of _apply_inverse_weights_map to numpy.ndarray with default arguments. This PR changes it back to pixell.enmap.ndmap.


mhasself (approved):
Thanks for the fix!

Issue #1071, opened 2024-12-17
Preprocess: Need to restrict noise fitting in low passed data


msilvafe:
Currently the demodulation lowpass filter is set by the hwp rotation speed if the filter parameters are not specifically provided. The noise fitting function should provide a check that frequencies above the lowpass filter are not included in the fits and raise a warning. @kmharrington raised this because she was analyzing stare observations where the HWP rotation speed was varying in each obs. The lpf was adjusting but the white noise/noise fit region was not fit accordingly so the results were biased leading to an erroneous interpretation of a large dependence of our noise levels on our choice of HWP rotation speed.

mmccrackan:
This has been added in #1085, though it does cause an error when loading without specifying wafers.  That is a separate issue though, so I think we can close this.

PR #1085, opened 2025-01-06
Noise fitting frequency limits from hwp freq

mmccrackan:
Branch to address #1071.  If the config file upper-frequency limits (either the white noise limit or the global upper limit of allowed frequencies) are greater than the hwp frequency (if hwp is present), they will be restricted to the hwp frequency and a warning is printed.  It will also fail if the lower frequency limit ends up being >= upper frequency limit because of this

Also a small change to the limits on the noise fitting for `fknee`.  Sometimes, zeros can occur and if `alpha` < 0, this gives `NaN `warnings.

### DIFF
--- a/sotodlib/tod_ops/fft_ops.py
+++ b/sotodlib/tod_ops/fft_ops.py
@@ -4,6 +4,7 @@ import sys
 import numdifftools as ndt
 import numpy as np
 import pyfftw
+import logging
 import so3g
 from scipy.optimize import minimize
 from scipy.signal import welch
@@ -11,6 +12,8 @@ from sotodlib import core
 
 from . import detrend_tod
 
+
+logger = logging.getLogger(__name__)
 
 def _get_num_threads():
     # Guess how many threads we should be using in FFT ops...
@@ -356,6 +359,18 @@ def calc_wn(aman, pxx=None, freqs=None, low_f=5, high_f=10):
     if pxx is None:
         pxx = aman.Pxx
 
+    # limit upper frequency cutoffs to hwp freq
+    if 'hwp_angle' in aman._fields:
+        hwp_freq = (np.sum(np.abs(np.diff(np.unwrap(aman.hwp_angle)))) /
+                     (aman.timestamps[-1] - aman.timestamps[0])) / (2 * np.pi)
+        hwp_freq_limit = 0.95 * hwp_freq
+        if high_f > hwp_freq_limit and hwp_freq > 0:
+            logger.warning(f"high frequency cutoff={high_f} > hwp_freq={hwp_freq_limit}. Limiting to hwp_freq.")
+            high_f = hwp_freq_limit
+            # make sure low_f < high_f
+            if low_f >= hwp_freq_limit:
+                raise ValueError(f"lower frequency cutoff={low_f} >= hwp_freq={hwp_freq_limit}")
+
     fmsk = np.all([freqs >= low_f, freqs <= high_f], axis=0)
     if pxx.ndim == 1:
         wn2 = np.median(pxx[fmsk])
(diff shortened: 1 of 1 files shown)


sotodlib/tod_ops/fft_ops.py
```diff
@@ -356,6 +359,18 @@ def calc_wn(aman, pxx=None, freqs=None, low_f=5, high_f=10):
...
+    # limit upper frequency cutoffs to hwp freq
+    if 'hwp_angle' in aman._fields:
+        hwp_freq = (np.sum(np.abs(np.diff(np.unwrap(aman.hwp_angle)))) /
+                     (aman.timestamps[-1] - aman.timestamps[0])) / (2 * np.pi)
+        hwp_freq_limit = 0.95 * hwp_freq
+        if high_f >= hwp_freq and hwp_freq > 0:
+            logger.warning(f"high frequency cutoff={high_f} > hwp_freq={hwp_freq}. Limiting to hwp_freq.")
+            high_f = hwp_freq_limit
+            # make sure low_f < high_f
+            if low_f >= hwp_freq_limit:
```
msilvafe, line 371: low_f > high_f is a check that should happen regardless of the check on hwp spin rate so I think this check/error raising should happen outside of the hwp_angle field check block.

sotodlib/tod_ops/fft_ops.py
```diff
@@ -459,12 +474,28 @@ def fit_noise_model(
...
+        if hwp_freq > 0:
+            hwp_freq_limit = 0.95 * hwp_freq
+            if f_max >= hwp_freq:
+                logger.warning(f"f_max={f_max} > hwp_freq={hwp_freq}. Limiting to hwp_freq.")
+                f_max = hwp_freq_limit
+            if fwhite[1] >= hwp_freq:
+                logger.warning(f"upper white noise freq={fwhite[1]} > hwp_freq={hwp_freq}. Limiting to hwp_freq.")
+                fwhite[1] = hwp_freq_limit
+                # make sure fwhite[0] < fwhite[1]
+                if fwhite[0] >= hwp_freq_limit:
```
msilvafe, line 491: Same comment as line 371

msilvafe (approved):
Looks good, one comment to address and just want to confirm you tested:

1. when high_f/fwhite[1] are > hwp spin rate you got the expected correction in the values 
2. the appropriate errors are raised when low f > high f
3. The divide by zero errors are eliminated by the addition of sys.float_info.min in the bounds

mmccrackan:
> Looks good, one comment to address and just want to confirm you tested:
> 
> 1. when high_f/fwhite[1] are > hwp spin rate you got the expected correction in the values
> 2. the appropriate errors are raised when low f > high f
> 3. The divide by zero errors are eliminated by the addition of sys.float_info.min in the bounds

Just realized this won't work for the raw signal PSD after testing on the ISO configs.  Will fix.

sotodlib/tod_ops/fft_ops.py
```diff
@@ -356,6 +359,18 @@ def calc_wn(aman, pxx=None, freqs=None, low_f=5, high_f=10):
     if pxx is None:
         pxx = aman.Pxx
 
+    # limit upper frequency cutoffs to hwp freq
+    if 'hwp_angle' in aman._fields:
```
iparask, line 363: `'hwp_angle' in aman` should return true if it exists in its fields. So there is no need to specifically check in `_fields`

sotodlib/tod_ops/fft_ops.py
```diff
@@ -459,12 +474,28 @@ def fit_noise_model(
             subscan=subscan,
             **psdargs,
         )
+
+    # limit upper frequency cutoffs to hwp freq
+    if 'hwp_angle' in aman._fields:
```
iparask, line 479: same

msilvafe:
> > Looks good, one comment to address and just want to confirm you tested:
> > 
> > 1. when high_f/fwhite[1] are > hwp spin rate you got the expected correction in the values
> > 2. the appropriate errors are raised when low f > high f
> > 3. The divide by zero errors are eliminated by the addition of sys.float_info.min in the bounds
> 
> Just realized this won't work for the raw signal PSD after testing on the ISO configs. Will fix.

We chatted about this but putting it here to not forget. I think the way we want to go forward with this is as follows:
1. When `proc_aman` is created in the `run` function add to the empty axismanager a field called `frequency_cutoffs`
2. Initially add a single field called `signal` which has a float matched to the cutoff of the IIR filter (should be 63 Hz but can compute from the iir filter params)
3. As additional signals are added and/or lowpass filtered (i.e. demodQ, demodU, dsT) their associated `frequency_cutoffs` should be added
4. In the noise calculation code check for the `frequency_cutoffs.<signal_name>` field for the `signal_name` you're computing the noise off of. If it exists check that the upper bound is < and if not set it equal to the cutoff if it doesn't exist do nothing.

mmccrackan:
Okay, rather delayed but I have added the new `frequency_cutoffs` axis manager into `proc_aman`.  Ended up being a little more complex than expected.  Flow is:

- Create `frequency_cutoffs` in `proc_aman` if it doesn't exist at the start.  Calculate cutoff for `signal` from IIR params, set to `np.nan` if IIR params don't exist.
- Set `signal` cutoff in `correct_iir_params` if it wasn't found before (is a `np.nan`).  Ensures things will fail if IIR params are not corrected when they need to be.
- Add `frequency_cutoffs` for `dsT`, `demodQ`, and `demodU` during demodulation.
- Add `frequency_cutoffs` for PSDs when they are calculated and are derived from the input signal `frequency_cutoff` (since the signals are not passed into noise fitting).
- Check lower and upper frequency limits for white noise and noise fitting using these cutoffs.  Limit the latter if it is too high and fail on the former if it is too high.

I checked the cutoffs for `signal` in the IIR calculation code and confirmed they are 62.9 Hz for both cases (in `pcore` and in `correct_iir`).  Ran through, saved db, and reloaded (should not be relevant currently on load anyway since noise is `calc_and_save` so this should be backwards compatible).

The other noise fitting limits changes in the original comment are no longer relevant.

sotodlib/preprocess/processes.py
```diff
@@ -15,10 +15,13 @@
                                    count_cuts, flag_cut_select,
                                    sparse_to_ranges_matrix)
 
+from sotodlib.preprocess import preprocess_util as pp_util
+
 from .pcore import _Preprocess, _FracFlaggedMixIn
 from .. import flag_utils
 from ..core import AxisManager
 
+logger = pp_util.init_logger("preprocess")
```
msilvafe, line 24: How does this logger interact with the loggers initialized in `site_pipeline` and `preprocess_util` functions?

mmccrackan: Went ahead and double checked this to make sure and confirmed that for both `preproc_or_load_group` and `preprocess_tod` when a `logger` is passed into the former and when running the latter, verbosity level is maintained even with adding the above `logger` into `processes`.

msilvafe:
This looks good to me now, thanks.

PR #1087, opened 2025-01-08
Add from_dict method to HkConfig

jlashner:
Small QOL update to properly instantiate HkConfig object from a dictionary.

### DIFF
--- a/sotodlib/io/hkdb.py
+++ b/sotodlib/io/hkdb.py
@@ -77,8 +77,13 @@ class HkConfig:
     def from_dict(cls, data: Dict[str, Any]) -> "HkConfig":
         """
         Generates an HkConfig object from a dictionary whose keys are fields of
-        the HkConfig dataclass. This will preprocess some specific fields, such
-        as ``db_url``, transforming them into the correct type.
+        the HkConfig dataclass.
+
+        If the ``db_url`` is specified, it can be set as a string, a
+        dictionary, or a sqlalchemy URL object. If it is of type dict, it
+        will be converted to a URL object by passing it through to the keyword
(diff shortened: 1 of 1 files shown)


sotodlib/io/hkdb.py
```diff
@@ -70,19 +70,22 @@ def __post_init__(self):
         if self.db_file is not None:
             self.db_url = f"sqlite:///{self.db_file}"
 
+    @classmethod
+    def from_dict(cls, data: Dict[str, Any]) -> "HkConfig":
```
mhasself, line 74: Please docstring for this... or perhaps better would be explain, in the class docstring, what this lets you get away with.

jlashner: Thanks Matthew, I added some additional info to the docstrings about when you might want to use this.

mhasself: I was hoping for something along the lines of
"db_url can be a string, or else a dict (which will be passed to db.URL.create()).  In either case, environment variables will be expanded."

jlashner: I see. Some of the awkwardness here comes from type-checking, and the fact that the type of the dataclass field "db_url" will always be of type `Union[str, db.URL]`, but it can be initialized a number of different ways.

These days I will generally avoid using dataclasses in these cases, and just use different type signatures for the `__init__` arguments and the class attributes. Would you rather I just change the implementation here so the init function accepts type Dict, and the behavior can be documented in the Args section?

mhasself: I don't find this particularly awkward, but maybe I just don't understand it. The comments you added are enough for me!

mhasself (approved):

Issue #1078, opened 2024-12-20
Detmatch `satp1_det_match_240220m` failing on 15 obs-ids


msilvafe:
We've found that 15 observations from satp1 fail to load due to missing detmatch metadata. The error that is returned when calling `context.get_det_info(obsid)` is pasted below.

<details>
  <summary>Click for traceback</summary>
  
```
ERROR: sotodlib.core.metadata.loader: An error occurred while processing a meta entry:

  spec:    {'db': '/global/cfs/cdirs/sobs/metadata/satp1/manifests/det_match/satp1_det_match_240220m/assignment_local.sqlite', 'label': 'assignment', 'det_info': True, 'on_missing': 'fail'}

  request: {'obs:obs_id': 'obs_1717265085_satp1_1111111'}


---------------------------------------------------------------------------
IncompleteMetadataError                   Traceback (most recent call last)
Cell In[13], line 1
----> 1 ctx.get_det_info(obsid)

File ~/.local/soconda/20240903_0.1.4/lib/python3.11/site-packages/sotodlib/core/context.py:495, in Context.get_det_info(self, obs_id, dets, samples, filename, detsets, meta, free_tags, on_missing)
    490 """Pass all arguments to :func:`get_meta(det_info_scan=True)`, and
    491 then return only the det_info, as a ResultSet.
    492 
    493 """
    494 if meta is None:
--> 495     meta = self.get_meta(obs_id=obs_id, dets=dets, samples=samples,
    496                          filename=filename, detsets=detsets, free_tags=free_tags,
    497                          on_missing=on_missing, det_info_scan=True)
    498 # Convert
    499 def _unpack(aman):

File ~/.local/soconda/20240903_0.1.4/lib/python3.11/site-packages/sotodlib/core/context.py:464, in Context.get_meta(self, obs_id, dets, samples, filename, detsets, meta, free_tags, check, ignore_missing, on_missing, det_info_scan)
    461     request['dets:readout_id'] = list(dets)
    463 metadata_list = self._get_warn_missing('metadata', [])
--> 464 meta = self.loader.load(metadata_list, request, det_info=det_info, check=check,
    465                         free_tags=free_tags, free_tag_fields=free_tag_fields,
    466                         det_info_scan=det_info_scan, ignore_missing=ignore_missing,
    467                         on_missing=on_missing)
    468 if check:
    469     return meta

[...]

PR #1088, opened 2025-01-08
Clarify error message when there is a detector count mismatch

jlashner:
Properly formats error message that occurs when there is a detector count mismatch between manifestdb and obsfiledb and adds some clarifying detail.

This error message quoted in several open issues and discussions:
- https://github.com/simonsobs/site-pipeline-configs/issues/60
- https://github.com/simonsobs/sotodlib/issues/1078

### DIFF
--- a/sotodlib/core/metadata/loader.py
+++ b/sotodlib/core/metadata/loader.py
@@ -638,8 +638,7 @@ def merge_det_info(det_info, new_info, multi=True, on_missing='trim'):
 
     if len(det_info) != len(i1) and on_missing != 'trim':
         raise IncompleteMetadataError(
-            f"Mismatching detector count: "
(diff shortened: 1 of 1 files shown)


mhasself:
It isn't necessarily the case that such a problem happens when expanding ObsFileDb with the first additional det_info.  It could occur elsewhere in the det_info construction chain.  So I think more generic "det_info had n_dets={}; after merge has n_dets={}" would be appropriate.

These IncompleteMetadata errors are always caught and re-raised with additional context (so to speak) -- so the actual "metadata" entry that caused the conflict is included in the traceback.  (As in the #1078 traceback)

jlashner:
Sure, that broader error message makes sense to me! I modified it to match your wording.

mhasself (approved):

PR #1089, opened 2025-01-08
Allows you to set hkdb object to use in LoadSpec

jlashner:
Max was finding that repeated calls to `load_hk` was eventually causing the number of connections to max out, due to the time it takes for hkdb.Session to be garbage collected.

This change allows you to manually pass in an HkDb instance to `load_hk`, avoiding repeated calls to sessionmaker if you are calling the function in a loop.


msilvafe (approved):
Tested looping over calls to LoadSpec and it seemed to work well to reduce number of connections to the database.

PR #1090, opened 2025-01-15
Fix for Source Flags

mmccrackan:
This branch is to fix a problem that arose when loading data back in that was introduced from the Source Flags preprocessing module.  Since fields are added into the axis manager, all detectors are cut when non-source entries are added.  For now, I copied the source list generation and use the fields that match sources in the source_flags axis manager.  We could alternatively just exclude "valid," but might not be future proof.


msilvafe (approved):
Thanks for the quick fix.

PR #1091, opened 2025-01-15
Allow plain filenames in cleanup_mandb

mmccrackan:
Changes cleanup_mandb to accept filenames without paths from the config file.


earosenberg (approved):
This seems to have solved that error, thanks. Now I can successfully pre-processing and map-making once with a plain filename. If I try to re-run it still fails, and seems to need an absolute path to the preprocessing db's rather than a relative path or plain filename. But we can treat that in a different issue

mmccrackan:
> This seems to have solved that error, thanks. Now I can successfully pre-processing and map-making once with a plain filename. If I try to re-run it still fails, and seems to need an absolute path to the preprocessing db's rather than a relative path or plain filename. But we can treat that in a different issue

The failure with loading plain filenames appears to be caused by something deeper into sotodlib's loading infrastructure, so I will try to address that in another PR.

PR #1095, opened 2025-01-16
Minor fix for logging in multilayer_load_and_preprocess

mmccrackan:
Minor fix to logger when running `multilayer_load_and_preprocess`.  The logging level would previously get reset due to the recursive call of `check_cfg_match`.


msilvafe (approved):
Thanks for catching this.

PR #1097, opened 2025-01-17
update_obsdb: new known_bad_books list in config file 

AERAdler:
won't attempt to index any books in that list


mhasself (approved):
Wow, that was fast.  Looks good!

PR #1098, opened 2025-01-17
Adding MPI style pool

iparask:
This is the second iteration of this work. I tried to do with minimal change in the code this time.

As a consequence, I felt that I had to remove the `util.main_launcher` as it does not offer any additional functionality in the main. That allows me to setup the pool and the executor in the main.

I also pull the number of nproc (number of worker processes) in the arg parser and outside of the config. Another change is that the `main()` now takes the executor as an argument. Under SLURM SLURM takes care of the process launching.

The result is that the rank is checked as soon as possible and only the master process runs the main function (`main()`).

I tested it on a single node and two nodes on Tiger3 and it utilized the cores properly distributed across the nodes.

~Note: It still depends on the pixel change which @amaurea will complete shortly. This means that there may be some changes to line 27 when this is done.~


sotodlib/utils/procs_pool.py
```diff
@@ -0,0 +1,28 @@
...
+    from mpi4py.futures import MPICommExecutor, as_completed
+    from mpi4py import MPI
+except Exception as e:
+    from concurrent.futures import ProcessPoolExecutor, as_completed
+    from multiprocessing import set_start_method as mp_set_start_method
+
+
+def get_exec_env(
+    nprocs: int = None,
+) -> tuple[int, Union["MPICommExecutor", "ProcessPoolExecutor"]]:
```
mhasself, line 14: It would be nice to let the user force a result here.  One way would be to add args:
`allow_mpi=True, allow_process_pool=True`, and then only return those executors if they are available.

Alternately, you could have a priorities / preference list: `priority=["mpi", "process_pool"]`.  Then the code would first try to deliver mpi, and if it can't it would deliver ProcessPool.  `priority=["process_pool"]` would deliver MPI directly.

The second one is perhaps more beneficial if you can imagine other *PoolExecutors coming in handy one day.

For the docstring, I think you need to show a very basic example of compliant usage -- because writing code that works with either MPI and ProcessPool is at least slightly tricky, w.r.t. needing to put the main driver inside an "if rank == 0:" block.

chervias:
@mhasself if you are happy with these changes can you approve please so we merge into the SAT mapmaker branch

sotodlib/utils/procs_pool.py
```diff
@@ -0,0 +1,142 @@
...
+        The as_completed function that is used to get the results of the parallel execution based on the type of the executor
+
+    Raises
+    ------
+    ValueError
+        When a valid executor is not available based on the execution priority list.
+    """
+    executor_created = False
+    while not executor_created:
+        executor_mode = priority.pop(0)
```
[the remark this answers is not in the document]
iparask: Sorry @mhasself. I fixed it.

sotodlib/utils/procs_pool.py
```diff
@@ -0,0 +1,143 @@
...
+    Raises
+    ------
+    ValueError
+        When a valid executor is not available based on the execution priority list.
+    """
+    user_priority = list(priority)
+    executor_created = False
+    while not executor_created:
+        executor_mode = user_priority.pop(0)
+        match executor_mode:
```
mhasself, line 131: "match" is python>=3.10

We have not dropped 3.9 support yet.

mhasself (approved):

PR #1099, opened 2025-01-17
site_pipeline: check_book funcs for update_obsdb use

mhasself:
This resolves issue where update_obsdb would sys.exit when a book check failed.

So now update_obsdb doesn't call main, it calls two other functions.

@AERAdler  I have not tested the update_obsdb changes!  Please do so to confirm they behave as we want.  In particular this should raise an informative error, not exit, when it encounters a bad book.

The logging will change slightly, because the update_obsdb logger is passed into the check_book functions.

### DIFF
--- a/sotodlib/site_pipeline/check_book.py
+++ b/sotodlib/site_pipeline/check_book.py
@@ -104,16 +104,19 @@ def get_parser(parser=None):
     return parser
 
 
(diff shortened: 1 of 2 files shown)


AERAdler (approved):
It looks good, I tested it and it caught the stray book properly.

PR #1102, opened 2025-01-21
Adding subscan HWPSS subtraction function to sotodlib.hwp.hwp

samdayweiss:
This PR adds a function in hwp.py to complete HWPSS subtraction by subscan, accounting for amplitude and phase variation within each scan and not relying on mean subtraction post-demodulation. It has also been found that glitch detection after HWPSS subtraction is more effective for identifying glitches, and this function introduces an iterative glitch detection approach, which also reduces bias in the HWPSS fit by improving the glitch mask.

### DIFF
--- a/sotodlib/hwp/hwp.py
+++ b/sotodlib/hwp/hwp.py
@@ -2,6 +2,7 @@ import numpy as np
 from scipy.optimize import curve_fit
 from sotodlib import core, tod_ops
 from sotodlib.tod_ops import bin_signal, filters, apodize
+from so3g.proj import RangesMatrix
 import logging
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


sotodlib/hwp/hwp.py
```diff
@@ -649,3 +650,114 @@ def demod_tod(aman, signal=None, demod_mode=4,
...
+            sub_flag = RangesMatrix.from_mask(flags.mask()[:,i[0]:i[1],])
+        
+        # clear sub axis manager of hwp axes so can populate new model
+        if 'hwpss_stats' in sub_aman._assignments.keys():
+            sub_aman.move('hwpss_stats',None)
+
+        if 'hwpss_model' in sub_aman._assignments.keys():
+            sub_aman.move('hwpss_model',None)
+
+        get_hwpss(sub_aman, signal=None, hwp_angle=None, bin_signal=True, bins=90,
```
ykyohei, line 719: I think we should use **kwargs, we should not hard code these arguments,

sotodlib/hwp/hwp.py
```diff
@@ -649,3 +650,114 @@ def demod_tod(aman, signal=None, demod_mode=4,
...
+    logger.info('subtract hwpss by subscan')
+    #define hwpss_avoid flags for mask to avoid in hwpss estimation
+    if 'hwpss_avoid' in aman.flags._assignments.keys():
+        aman.flags.move('hwpss_avoid',None)
+
+    aman.flags.reduce(flags=['hwpss_glitches', 'turnarounds'], method='union', wrap=True,
+                      new_flag='hwpss_avoid', remove_reduced=False)   
+
+    #prepare arrays for subscan subtracted signal and model
+    num_subscans = aman.subscan_info.subscan_flags.shape[0]
```
ykyohei, line 699: we need a check if subscan_info exists in aman

sotodlib/hwp/hwp.py
```diff
@@ -649,3 +650,114 @@ def demod_tod(aman, signal=None, demod_mode=4,
...
+    else:
+        aman.wrap('hwpss_subscan_remove', gfilled, [(0, 'dets'), (1, 'samps')])
+
+    #remove remaining by-products of the subtraction
+    aman.flags.move('hwpss_avoid',None)
+    del subtracted_sig
+    del hwpssmodel
+    del num_subscans
+
+    return None
```
ykyohei, line 763: return

sotodlib/hwp/hwp.py
```diff
@@ -649,3 +650,114 @@ def demod_tod(aman, signal=None, demod_mode=4,
...
+
+    aman.flags.reduce(flags=['hwpss_glitches_1', 'hwpss_glitches_2'], method='union', wrap=True,
+                      new_flag='hwpss_glitches', remove_reduced=True)
+    
+    logger.info('subtract hwpss by subscan')
+    #define hwpss_avoid flags for mask to avoid in hwpss estimation
+    if 'hwpss_avoid' in aman.flags._assignments.keys():
+        aman.flags.move('hwpss_avoid',None)
+
+    aman.flags.reduce(flags=['hwpss_glitches', 'turnarounds'], method='union', wrap=True,
```
ykyohei, line 695: Why do we need to take union with turnaround?

ykyohei (changes requested):
General comments.
- This function includes iterative hwpss subtraction and subscan hwpss subtraction. These two should be implemented separately.
- logger.info seems not necessary, please make them inline comments
- Please run linter such as flake8 and make corrections
- We also need to correction in docstring. this can be done in the next iteration,

ykyohei (commented):
The hwpss_model in this method has an inherent discontinuity in the boundaries of the subscans. This leads to 0.1 aW level of ringing at the boundaries of subscans.  This is below single detector white noise level 25-40 aW rt(s) but it can matter when we coadd data. We need to smooth the model to avoid masking the turnarounds too much.

Figure 25 (4)
Figure 25 (5)

sotodlib/hwp/hwp.py
```diff
@@ -911,3 +911,121 @@ def get_tau_hwp(
...
+
+    At the boundary of subscan_1 and subscan_2 (with constant coefficients c_s1, c_s2, respectively):
+    c(t)_i = {c_s1 for s < s_pivot - n_buf
+              c_s2 + (c_s1 - c_s2)*wn(s) for  s_pivot - n_buf >= s >= s_pivot + n_buf
+              c_s2 for s > s_pivot + n_buf}
+
+    Here, wn(s) is the window function chosen to span 2*n_buf for a smooth transition between coefficients, 
+    and s_pivot is the single sample transition point between subscan_1 and subscan_2.
+    '''
+    if f'{hwpss_model_name}' in aman._assignments:
```
ykyohei: we can simplify as `if hwpss_model_name in aman:`

sotodlib/hwp/hwp.py
```diff
@@ -911,3 +911,121 @@ def get_tau_hwp(
...
+            continue
+        else:
+            transition_indices = np.append(transition_indices,transition_idx[1])
+
+    #patch hwpss discontinuity regions between subscans
+    coeffs_diff = -1*np.diff(coeffs,axis=2) #minus sign gives us a1-a2 where a1,a2 time ordered
+    apodize = scipy.signal.windows.get_window(window, 4*n_buf, fftbins=False)
+    apodize = apodize[2*n_buf:] #only get range from wn(x) = 1 to wn(x) = 0
+    apodize = np.tile(apodize,(coeffs.shape[0],coeffs.shape[1])).reshape(coeffs.shape[0],coeffs.shape[1],2*n_buf)
+    for k in range(len(transition_indices)):
```
ykyohei: `for k, tidx in enumerate(transition_indices):`

sotodlib/hwp/hwp.py
```diff
@@ -911,3 +911,121 @@ def get_tau_hwp(
...
+    coeffs = np.zeros(shape=(aman.dets.count,2*len(modes),len(subscan_info.subscan_flags))) #list of hwpss_subscan harmonic coeff
+    transition_indices = np.array([])
+
+    for j in range(num_subscans):
+        subscan_flag = subscan_info.subscan_flags[j]
+        i = subscan_flag.ranges()[0]
+        sub_aman = aman.restrict('samps', (i[0] + aman.samps.offset, i[1] + aman.samps.offset), in_place=False)
+        sub_aman.move(hwpss_model_name, None)
+        if flags is None:
+            sub_flag = RangesMatrix.from_mask(np.zeros(aman.samps.count, dtype=int)[i[0]:i[1]])
```
ykyohei: dtype int is fine but bool is more natural

ykyohei (commented):
My main comments is about position of transition indices, I think it should be centered with respect to turnarounds or subscans.

Also, can you comment/clarify how much the ringing issues are reduced by apodization?

sotodlib/hwp/hwp.py
```diff
@@ -911,3 +911,121 @@ def get_tau_hwp(
...
+            merge_stats=True,
+            merge_model=True,
+            hwpss_model_name=hwpss_model_name,
+        )
+        aman.get(hwpss_model_name)[:, i[0]:i[1]] = sub_aman.hwpss_model
+        coeffs[:,:,j]+=sub_aman.hwpss_stats.coeffs
+        del sub_aman
+
+        transition_msk = aman.flags.turnarounds.mask() & subscan_flag.mask()
+        transition_idx = np.where(np.diff(transition_msk.astype(int)) == -1)[0] #determining index of falling edge which denotes transition to the next subscan
```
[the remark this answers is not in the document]
samdayweiss: No, I'm pretty sure this puts the transition at the middle of the turnaround. I will check again to make sure, but I designed it to be that way.

sotodlib/hwp/hwp.py
```diff
@@ -911,3 +911,121 @@ def get_tau_hwp(
...
+            aman.move(hwpss_model_name,None)
+    
+    aman.wrap(hwpss_model_name, np.zeros_like(aman.signal,dtype='float32'),[(0, 'dets'), (1, 'samps')])
+
+    #re-calculate subscan flags with turnarounds included, for apodization purposes
+    subscan_info = tod_ops.flags.get_subscans(aman, flags=None, merge=False, include_turnarounds=True, overwrite=False)
+    num_subscans = len(subscan_info.subscan_flags)
+
+    #transition_flag = np.zeros_like(aman.hwpss_model)
+    coeffs = np.zeros(shape=(aman.dets.count,2*len(modes),len(subscan_info.subscan_flags))) #list of hwpss_subscan harmonic coeff
```
msilvafe: for 3rd index of shape: `aman.subscan_info.subscans.count` can also remove comment after line `#list of hwpss_subscan...` not necessary (and this isn't a list...)

sotodlib/hwp/hwp.py
```diff
@@ -911,3 +911,121 @@ def get_tau_hwp(
...
+
+    for j in range(num_subscans):
+        subscan_flag = subscan_info.subscan_flags[j]
+        i = subscan_flag.ranges()[0]
+        sub_aman = aman.restrict('samps', (i[0] + aman.samps.offset, i[1] + aman.samps.offset), in_place=False)
+        sub_aman.move(hwpss_model_name, None)
+        if flags is None:
+            sub_flag = RangesMatrix.from_mask(np.zeros(aman.samps.count, dtype=int)[i[0]:i[1]])
+        else:
+            sub_flag = RangesMatrix.from_mask(flags.mask()[:, i[0]:i[1]])
```
msilvafe: This is super inefficient to expand the RangesMatrix out to boolean array and then back into RangesMatrix. What if you just always passed the same full AxisManager but combined the RangesMatrix for the subscan samples with the existing flags i.e. pass to `get_hwpss` the arg: `flags = subscan_flag + flags` and set `merge_model` and `merge_stats` to False so that they're returned and then wrap those into aman as you do on 990-991

sotodlib/hwp/hwp.py
msilvafe: You should be able to do this as RangesMatrices as well. Add in Ranges and then you will get a range which is the samples you care about and don't have to look for diff = -1.

sotodlib/hwp/hwp.py
```diff
@@ -911,3 +911,121 @@ def get_tau_hwp(
...
+        patch = np.einsum('ijk -> ik',coeff_t*vects)
+
+        aman.hwpss_model[:,tidx-n_buf:tidx+n_buf] = patch
+
+        del hwp_chi
+        del a2
+        del diff
+        del coeff_t
+        del vects
+        del patch
```
msilvafe: I think this can be one line.

PR #1103, opened 2025-01-22
Fix error logging

mmccrackan:
Fixes an issue with error logging when preprocessing failed and returned a list of [obsid, error].  Also added some more debugging statements to `cleanup_mandb` to help track some failure cases.


msilvafe (approved):
Thanks for that fix, this will help with parsing the error logs.

Issue #1105, opened 2025-01-23
HkDb: Frames can be loaded in the wrong order


jlashner:
Yoshinori [posted in dm-daq-support](https://simonsobs.slack.com/archives/CJ8M8HY7P/p1737662321242459) that hkdb was loading some frames in the wrong order:

![Image](https://github.com/user-attachments/assets/027d12a8-07a9-4267-8e69-00d58e5d6e03)

After looking into it, it seems as though this is caused by a single frame that has an abnormally long duration: 
```
frame.__dict__
>> {'_sa_instance_state': <sqlalchemy.orm.state.InstanceState object at 0x7f914c0d6c80>,
 'agent': 'acu',
 'byte_offset': 577164031,
 'end_time': 1737422107.8764656,
 'feed': 'acu_error',
 'file': <sotodlib.io.hkdb.HkFile object at 0x7f90d2b4eec0>,
 'file_id': 47326,
 'id': 20629910,
 'start_time': 1737108927.5658438}
```

This long frame makes it so [this query](https://github.com/simonsobs/sotodlib/blob/9121f2780cd38433f0e002ac7ff4d64964c6906f/sotodlib/io/hkdb.py#L451-L459) can potentially insert files out of order, since the query is order by Frame start time.

This bug can be fixed by sorting the file paths  and making sure we're loading them in the proper order.

PR #1106, opened 2025-01-23
HkDb: sort paths before loading frames

jlashner:
Resolves https://github.com/simonsobs/sotodlib/issues/1105.


mhasself (approved):
This is definitely the quickest solution. Unfortunate that it imposes a sorting condition on the filenames in the database.

There are alternatives, but not simple ones...

PR #1107, opened 2025-01-26
Update versioneer for compatibility with python-3.12

tskisner:
Our versioneer files were at version 0.18.  This updates the bundled file to version 0.29, which has a fix for changes to the python `configparser` API in python 3.12.


mhasself (approved):
Thanks!  The `__future__` is `__now__`!!

PR #1112, opened 2025-02-06
AxisMan: faster .wrap(axisman) and .merge(...)

mhasself:
Introduces notion of Axis equality. This is to make .load work faster -- in that case you are merging in lots of AxisManagers that have already concordant axes; and even if they weren't concordant you definitely don't need to keep a copy of the unrestricted objects around while you merge.

On test case '/global/cfs/cdirs/sobs/sat-iso/preprocessing/satp1_20250108_init/preprocess_archive_932.h5', group 'obs_1730532853_satp1_1111111_wafer_slot_ws0_wafer.bandpass_f150', this reduced load time from 4.5s to 1.0s

### DIFF
--- a/sotodlib/core/axisman.py
+++ b/sotodlib/core/axisman.py
@@ -129,6 +129,9 @@ class IndexAxis(AxisInterface):
         assert stop <= self.count
         return IndexAxis(self.name, stop - start), sl
 
+    def __eq__(self, other):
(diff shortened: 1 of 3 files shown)


amaurea (approved):
I get a factor 3x speed improvement for get_meta with this.

PR #1113, opened 2025-02-06
Use pytest

mhasself:


mhasself:
Hi @guanyilun  -- could you please fix test_imprinter.py?  Just commit fix to this branch.

I don't know how long this has been broken, but our CI wasn't running this test because it was pytest-based instead of unittest-based.  But now it's going to get run :)

guanyilun:
I see it passing now, should be good!

sotodlib/core/resources.py
```diff
@@ -83,7 +86,8 @@ def get_local_file(filename: str, cache: bool = True) -> str:
         target_path = local_cache if cache else "/tmp/"
         os.makedirs(name=target_path, exist_ok=True)
         target_file = os.path.join(target_path, filename)
-        _, headers = urlretrieve(de_url, target_file)
+        if download:
+            _, headers = urlretrieve(de_url, target_file)
```
iparask, line 90: If download is False, this function will return a filepath, but the file will not exist. Is it just for testing purposes?

iparask: Instead of that in the test you can do:
```
from unittest import mock

@mock.patch('urllib.request.urlretrieve', return_value=["some_answer", "some_header"])
def test_get_local_file(self, mocked_urlretrieve):
...
```

and pytest will never download anything.

.github/workflows/test.yml
```diff
@@ -56,17 +56,19 @@ jobs:
           python3 -m pip install -vvv .[tests]
 
       - name: Run Serial Tests
+        working-directory: ./tests
         run: |
           export OMP_NUM_THREADS=2
           export OPENBLAS_NUM_THREADS=2
           export MPI_DISABLE=1
-          python3 -m unittest discover
+          python3 -m pytest
```
iparask, line 64: Can you add `-vv` after pytest? The increased verbosity makes immediately clear which test exactly failed.

https://docs.pytest.org/en/stable/how-to/output.html#verbosity

iparask (approved):
I'm in favor of pytest as it offers a lot of features to make testing easier and better. I'm perfectly fine with the PR.

We may want to convert all the tests to pytest and use fixtures to mock things whenever needed.

sotodlib/core/resources.py
mhasself: Ahh, that is powerful.  However I am going to keep the download arg because it's generally useful; you might want to basically assert that a bunch of files are available.  Also if someone changed the implementation (to not use urlretrieve), this mock would become ineffective, with no warning.

PR #1114, opened 2025-02-12
Source subsampling + multipass fix + ml mapmaker merge

amaurea:
* implements source subsampling
* fixes the multipass issues
* merges simple_ml_mapmaker.py with make_ml_map.py

I started writing this in the ml_demod branch, so this might not work until #1110 is merged

Edit: Oops, yes, this contains all of ml_demod. I guess that means it can both be merged as is (which would make #1110 unnecessary, or after it. Either would work.


iparask:
@amaurea this renders #1035 obselete?

amaurea:
> @amaurea does this render #1035 obsolete?

If this is merged, then #1035 doesn't need to be merged. But if #1035 is merged first, then this pull request is still valid and should be compatible.

sotodlib/mapmaking/ml_mapmaker.py
```diff
@@ -87,7 +87,7 @@ def add_obs(self, id, obs, deslope=True, noise_model=None, signal_estimate=None)
         tod = nmat.apply(tod)
         # Add the observation to each of our signals
         for signal in self.signals:
-            signal.add_obs(id, obs, nmat, tod, glitch_flags=self.glitch_flags_path)
+            signal.add_obs(id, obs, nmat, tod)
```
iparask, line 90: Can you include the `cut_fields=self.glitch_flags_path` in this call?
SO data have their `glitch_flags` under `flags`, while ACT has it under the observation.

All the `glitch_flags` inputs where for that reason.

sotodlib/site_pipeline/make_ml_map.py
```diff
@@ -1,101 +1,69 @@
...
+    parser.add_argument(      "--nocal",   action="store_true", help="Disable calibration. Useful for sims")
+    parser.add_argument(      "--nmat-dir",  type=str, default="{odir}/nmats")
+    parser.add_argument(      "--nmat-mode", type=str, default="build", help="How to build the noise matrix. 'build': Always build from tod. 'cache': Use if available in nmat-dir, otherwise build and save. 'load': Load from nmat-dir, error if missing. 'save': Build from tod and save.")
+    parser.add_argument("-d", "--downsample", type=str, default="1",  help="Downsample TOD by this factor. ,-sep")
+    parser.add_argument(      "--maxiter",    type=str, default="500",help="Max number of CG steps per pass. ,-sep")
+    parser.add_argument(      "--interpol",   type=str, default="nearest", help="Pmat interpol per pass. ,-sep")
+    parser.add_argument("-T", "--tiled"  ,   type=int, default=1, help="0: untiled maps. Nonzero: tiled maps")
+    parser.add_argument(      "--srcsamp",   type=str, default=None, help="path to mask file where True regions indicate where bright object mitigation should be applied. Mask is in equatorial coordinates. Not tiled, so should be low-res to not waste memory.")
```
iparask, line 34: I think we need an argument for the cut fields path in the observations.

iparask (commented):
A few more comments I have as I try to use this with the ACT data.

sotodlib/site_pipeline/make_ml_map.py
```diff
@@ -1,101 +1,69 @@
...
-        cfg.update({k: v for k, v in cfg_from_file.items() if v is not None})
-    else:
-        print("No config file provided, assuming default values")
+import numpy as np, sys, time, warnings, os, so3g
+from sotodlib.core import Context, AxisManager, IndexAxis
+from sotodlib.io import metadata   # PerDetectorHdf5 work-around
+from sotodlib import tod_ops, mapmaking, core
+from sotodlib.tod_ops import filters
+from pixell import enmap, utils, fft, bunch, wcsutils, mpi, bench
+from enlib import log
```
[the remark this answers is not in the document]
chervias: yes, do we have to use enlib? sotodolib has it own logging module, right?

chervias:
@amaurea also could you please restore the configuration file parameter where you just pass a --config parameter with a yaml file with all other parameters. You remove that in your PR. Thanks!

sotodlib/mapmaking/utils.py
```diff
@@ -1,8 +1,10 @@
-from typing import Any, Union
+from typing import Optional, Any, Union
+from sqlalchemy import create_engine, exc
+from sqlalchemy.orm import declarative_base, Mapped, mapped_column, sessionmaker
 import importlib
 import numpy as np
 import so3g
-from pixell import enmap, fft, resample, tilemap, utils, bunch
+from pixell import enmap, fft, resample, tilemap, bunch utils as putils
```
iparask: You're missing a comma here.

amaurea:
Ok, Carlos and I have now restored the Argparse/configuration integration. I didn't know that that the one in the sotodlib version of the program was meant to be called from `so-site-pipeline`, so I only tested it in isolation when I merged the two versions of the program.

We also fixed a logic error in the tile redistribution that had crept in in my version. The problem was that when it was supposed to redistribute data to the workspace tiles, it just kept them as ownership tiles instead. This worked when I ran py.test originally, since I did that with just 1 mpi task. I was pretty sure I had done some bigger mapmaking tests with this, but I must have been wrong.

I haven't tested any bigger runs with this version yet. I'll try to find time, but it would be good if someone else could test it too.

amaurea:
Ok, I think we've arrived at something we're happy with now. In addition to the source sub-sampling etc. the ml mapmaker can now process ACT data and the sims. There were two difficulties with the latter:

1. Getting a list of all subids (=obs_id:wafer_slot:band) available. This is currently done by parsing the wafer mask at the end of the obs_id, combined with querying obsdb for the tube flavor of each, but act_ids didn't have such a mask, and the sims had wafer slots that were inconsistently numbered. Instead of making the code more complicated we updated the sim and act databases to be more consistent with the SO databases for real data. It might still be a good idea to replace all this with the wafer_slots_list instead though. That wasn't present in the ACT db, either though, but it is now.
2. Reading in data for a subid. ACT did not have any concept of wafer slots, so this broke. This is added in the updated ACT database.

Here's the script that made the changes to the ACT db:

```
import argparse, os
parser = argparse.ArgumentParser()
parser.add_argument("context")
parser.add_argument("odir")
args = parser.parse_args()
import numpy as np, yaml, h5py, shutil
from pixell import utils, sqlite, bunch

def update_sqlite_obsid(s, table, col="obs_id"):
	s.execute("update ? set ? = ? || '_1'", parameters=(table, col, col))
def update_hdf_obsid(ifile, ofile, dset, col):
	din   = bunch.read(ifile)
	dtype = din[dset].dtype
	dtype = [(dtype.names[0], 'S27')]+[(a,dtype.fields[a][0]) for a in dtype.names[1:]]
	dout  = bunch.Bunch()
	dout[dset] = np.zeros(len(din[dset]), dtype)
	dout[dset][col] = np.char.encode(np.char.add(np.char.decode(din[dset][col]), "_1"))
	for (name, dt) in dtype[1:]:
		dout[dset][name] = din[dset][name]
	bunch.write(ofile, dout)
def update_sqfile_file(s, fun):
	rows = s.execute("select id, name from files")
	for id, name in rows:
		name = fun(name)
		s.execute("update files set name = ? where id = ?", parameters=(name, id))

with open(args.context, "r") as ifile:
	context = yaml.safe_load(ifile)
# 0: Add a context tag for the original directory. This is needed for
# the entries we didn't modify
context["tags"]["old_root"] = os.path.dirname(os.path.abspath(args.context))
# Fix obs_colon_tags. band is wafer.bandpass for SO. Is it enough to rename the column in detdb.base for this?
context["obs_colon_tags"] = ["band", "wafer_slot"]

cdir = os.path.dirname(args.context)
pres = {key:os.path.join(cdir, val) for key,val in context["tags"].items()}

def findmeta(name):
	for i, entry in enumerate(context["metadata"]):
		if entry["name"] == name:
			return i
def getmeta(name):
	for entry in context["metadata"]:
		if entry["name"] == name:
			return os.path.join(cdir, entry["db"].format(**pres))
def replmeta(name, old, new):
	i = findmeta(name)
	context["metadata"][i]["db"] = context["metadata"][i]["db"].replace(old, new)

utils.mkdir(args.odir)

# Tasks:
# 1. Replace obs_id with obs_id_mask. These files must be updated:
#    1. obsdb
# 3. Add a tube_flavor, wafer_slots_lst to obs_id
# 4. Consider adding start_time, stop_time, n_samples, telescope, telescope_flavor, tube_slot, type="obs", subtype=obs_type, wafer_count, detector_flavor, duration, az_center, az_throw, el_center, el_throw, roll_center, roll_throw, hwp_freq_mean, hwp_freq_stdev.
obsdb_ifile = context["obsdb"].format(**pres)
obsdb_ofile = args.odir + "/obsdb.sqlite"
shutil.copy(obsdb_ifile, obsdb_ofile)
with sqlite.open(obsdb_ofile) as s:
	s.execute("update obs set obs_id = obs_id || '_1'")
	# Fix broken tube_flavor. Probably only the first of these is necessary
	s.execute("update obs set tube_flavor = replace(tube_flavor, 'pa4', 'hf')")
	s.execute("update obs set tube_flavor = replace(tube_flavor, 'pa5', 'mf')")
	s.execute("update obs set tube_flavor = replace(tube_flavor, 'pa6', 'mf')")
	# Add wafer slots list, which is trivial
	s.execute("alter table obs add column wafer_slots_list text default 'ws0'")
	# Add the other constant extra fields
	s.execute("alter table obs add column telescope text default 'act'")
	s.execute("alter table obs add column telescope_flavor text default 'act'")
	s.execute("alter table obs add column type text default 'obs'")
	s.execute("alter table obs add column detector_flavor default 'tes'")
	s.execute("alter table obs add column roll_center real default 0.0")
	s.execute("alter table obs add column roll_throw  real default 0.0")
	s.execute("alter table obs add column hwp_freq_mean   real default 0.0")
	s.execute("alter table obs add column hwp_freq_stdev  real default 0.0")
	# We want subtype instead of obs_type
	s.execute("alter table obs rename column obs_type to subtype");
	# The other fields will need data from the ACT database. Let's handle this later
	s.execute("commit")
context["obsdb"] = "{metadata_root}/obsdb.sqlite"
#    2. pointofs h5 file. Hardcoded assumption that it's relative, in ctx dir
isfile = getmeta("boresight_offset")
osfile = os.path.join(args.odir, os.path.basename(isfile))
shutil.copy(isfile, osfile)
with sqlite.open(osfile) as s:
	ihfile = list(s.execute("select name from files"))[0][0]
	ihfile = os.path.join(os.path.dirname(isfile), ihfile)
	ohfile = os.path.join(args.odir, os.path.basename(ihfile))
	update_sqfile_file(s, lambda name: os.path.relpath(ohfile, args.odir))
	update_hdf_obsid(ihfile, ohfile, "pointofs", "obs:obs_id")
	s.execute("commit")
#    3. abscal sqlite
#    4. abscal h5
isfile = getmeta("abscal&cal")
osfile = os.path.join(args.odir, "abscal.sqlite")
shutil.copy(isfile, osfile)
with sqlite.open(osfile) as s:
	s.execute("update map set [obs:obs_id] = [obs:obs_id] || '_1'")
	ihfile = list(s.execute("select name from files"))[0][0]
	ihfile = os.path.join(os.path.dirname(isfile), ihfile)
	ohfile = os.path.join(args.odir, os.path.basename(ihfile))
	update_sqfile_file(s, lambda name: os.path.relpath(ohfile, args.odir))
	update_hdf_obsid(ihfile, ohfile, "abscal", "tod_id")
	s.execute("commit")
replmeta("abscal&cal", "/metadata_", "/")
#    5. cal sqlite (don't need to change raw files it points to)
#    6. cuts sqlite
#    7. pcuts sqlite
#    8. timeconst sqlite
for iname, oname in [("relcal&cal", "cal"), ("glitch_flags&flags", "cuts"), ("source_flags&flags", "pcuts"), ("timeconst&", "timeconst")]:
	isfile = getmeta(iname)
	osfile = os.path.join(args.odir, oname + ".sqlite")
	shutil.copy(isfile, osfile)
	with sqlite.open(osfile) as s:
		s.execute("update map set [obs:obs_id] = [obs:obs_id] || '_1'")
		update_sqfile_file(s, lambda name: os.path.join(os.path.dirname(isfile), name))
		s.execute("commit")
	replmeta(iname, "/metadata_", "/")
# Update obsfiledb
isfile = context["obsfiledb"].format(**pres)
osfile = os.path.join(args.odir, os.path.basename(isfile))
shutil.copy(isfile, osfile)
with sqlite.open(osfile) as s:
	s.execute("update files set obs_id = obs_id || '_1'")
	s.execute("commit")
# 2. Add a wafer_slot to detdb base
isfile = context["detdb"].format(**pres)
osfile = os.path.join(args.odir, os.path.basename(isfile))
shutil.copy(isfile, osfile)
with sqlite.open(osfile) as s:
	s.execute("update base set wafer = array_name")
	s.execute("alter table base add column wafer_slot text default 'ws0'")
	# might need to set optics_tube to something sensible here
	s.execute("commit")
# focalplane is the only entry that didn't need any changes
i = findmeta("focal_plane")
replmeta("focal_plane", "{metadata_root}", "{old_root}")

# Finally write the updated context. Phew!
with open(args.odir + "/context.yaml", "w") as ofile:
	yaml.dump(context, ofile)
```

This could still use some improvements:

1. It's not self-contained, in that it still refers to the original ACT db for things that didn't change. This means that you can't just copy it to a new cluster. This would be some work to fix, but nothing too bad.
2. We still need to add things like az_center, az_throw etc, which will be needed for proper tile-friendly task distribution.

chervias (approved):

PR #1115, opened 2025-02-13
add pca filters to preprocess

ykyohei:
I would like to have these for source mapmaking using preprocess module. I tested these.
Also fixed some typos in PTPFlags and InvVarFlags.


msilvafe (approved):
Looks good, and thanks for fixing the misnamed variables in ptp flags and inv var flags!

PR #1116, opened 2025-02-14
Rework detector selection for preprocess loading

mmccrackan:
This branch changes the way detectors are restricted when loading from either a single layer or multilayer preprocessing run.  Detectors are now restricted to the number remaining at the end of the initial preprocess run.

### DIFF
--- a/sotodlib/preprocess/preprocess_util.py
+++ b/sotodlib/preprocess/preprocess_util.py
@@ -295,8 +295,9 @@ def load_preprocess_det_select(obs_id, configs, context=None,
     pipe = Pipeline(configs["process_pipe"], logger=logger)
 
     meta = context.get_meta(obs_id, dets=dets, meta=meta)
-    logger.info(f"Cutting on the last process: {pipe[-1].name}")
-    pipe[-1].select(meta)
+    logger.info("Restricting detectors on all processes")
+    for process in pipe[:]:
+        process.select(meta)
     return meta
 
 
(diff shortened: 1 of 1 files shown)


sotodlib/preprocess/preprocess_util.py
```diff
@@ -295,8 +295,9 @@ def load_preprocess_det_select(obs_id, configs, context=None,
     pipe = Pipeline(configs["process_pipe"], logger=logger)
 
     meta = context.get_meta(obs_id, dets=dets, meta=meta)
-    logger.info(f"Cutting on the last process: {pipe[-1].name}")
-    pipe[-1].select(meta)
+    logger.info("Restricting detectors on all processes")
+    for process in pipe[:]:
+        process.select(meta)
```
msilvafe, line 300: I think since you fixed this here, you don't need to run the select step `load_and_preprocess` so on line 347 you should be able to run `pipe.run` with `select=False` and if other functions call `load_preprocess_det_select` before `pipe.run` then the `select=False` should be used there as well. But we should test that it doesn't break other things. 

It would be awesome if `load_and_preprocess_det_select` could collect all of the restriction flags and union them and do one call the restrict as there is quite an overhead to each call to restrict (hence my suggestion to not call it again in pipe.run).

msilvafe (changes requested):
One inline comment to address/discuss before merging.

sotodlib/preprocess/preprocess_util.py
msilvafe: Sounds like a good plan.

sotodlib/preprocess/processes.py
```diff
@@ -141,8 +144,11 @@ def select(self, meta, proc_aman=None):
         else:
             raise ValueError(f"Entry '{self.select_cfgs['kind']}' not"
                                 "understood. Expect 'any' or 'all'")
-        meta.restrict("dets", meta.dets.vals[keep])
-        return meta
+        if in_place:
+            meta.restrict("dets", meta.dets.vals[keep])
+            return meta
+        else:
+            return keep
```
msilvafe, line 151: since this 5 line block is copied in every select step should we make a method of the Pipeline class called `restrict` or `cut` that takes in `meta`, `in_place`, and `keep` and then make these 5 lines: `return self.cut(meta, keep, in_place)`?

sotodlib/preprocess/processes.py
```diff
@@ -1168,12 +1185,20 @@ def select(self, meta, proc_aman=None):
...
-        return meta
+                if in_place:
+                    meta.restrict("dets", meta.dets.vals[keep])
+                    source_flags.restrict("dets", source_flags.dets.vals[keep])
+                else:
+                   keep_all &= keep
+        if in_place:
+            return meta
+        else:
+            return keep_all
```
msilvafe, line 1201: Why is this one different than the rest? This breaks my generalization :(

msilvafe (approved):

mmccrackan:
After some offline discussion we decided to just merge this without moving to a single function since there are differences in the handling between processing functions and other open PRs are adjusting them.

Added one more change to the noisy subscan flags which didn't have the `in_place` option added.  I removed following check in that function:

`if hasattr(proc_aman.noisy_dets_flags, "valid_dets"):`

Since I think it should be guaranteed to exist and we don't do this check elsewhere.  This makes sure it always returns either `meta` or `keep`.

PR #1117, opened 2025-02-14
simplify hwpss_subscan

ykyohei:
I tried to simplify this function to be modular #1102.
Now this function only estimates `hwpss_model` and we use `subtract_hwpss` for subtraction.
I removed iteration part because I think the iterative subtraction should be implemented by calling same function several times.
There is a difference in the treatment of turnaround, so please check.

### DIFF
--- a/sotodlib/hwp/hwp.py
+++ b/sotodlib/hwp/hwp.py
@@ -559,7 +559,7 @@ def subtract_hwpss(aman, signal='signal', hwpss_template_name='hwpss_model',
             aman.wrap(subtract_name, np.subtract(signal,
                       aman[hwpss_template_name], dtype='float32'),
                       [(0, 'dets'), (1, 'samps')])
-    
+
     if remove_template:
         aman.move(hwpss_template_name, None)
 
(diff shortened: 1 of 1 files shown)


sotodlib/hwp/hwp.py
```diff
@@ -652,112 +652,72 @@ def demod_tod(aman, signal=None, demod_mode=4,
...
+        hwpss_model_name=hwpss_model_name,
+    )
+
+    for subscan_flag in aman.subscan_info.subscan_flags:
+        i = subscan_flag.ranges()[0]
+        sub_aman = aman.restrict('samps', (i[0] + aman.samps.offset, i[1] + aman.samps.offset), in_place=False)
+        sub_aman.move(hwpss_model_name, None)
         if flags is None:
-            sub_flag = sub_aman.flags.hwpss_avoid
+            sub_flag = RangesMatrix.from_mask(np.zeros(aman.samps.count, dtype=int)[i[0]:i[1]])
```
samdayweiss, line 696: Do we want this to be zeros or ones? if no flags are applied, we'd like all samples to be selected in the subscan.

sotodlib/hwp/hwp.py
samdayweiss, line 1: @ykyohei so you'd like this function not to also complete the subtraction? It will just wrap the model calculated per subscan, and the user can use the subtract_hwpss function separately

ykyohei: Yes

sotodlib/hwp/hwp.py
ykyohei: I think this is correct, because one (True) will be masked.

sotodlib/hwp/hwp.py
```diff
@@ -652,112 +652,72 @@ def demod_tod(aman, signal=None, demod_mode=4,
...
+        apodize_flags_samps=apodize_flags,
+        merge_stats=False,
+        merge_model=True,
+        hwpss_model_name=hwpss_model_name,
+    )
+
+    for subscan_flag in aman.subscan_info.subscan_flags:
+        i = subscan_flag.ranges()[0]
+        sub_aman = aman.restrict('samps', (i[0] + aman.samps.offset, i[1] + aman.samps.offset), in_place=False)
+        sub_aman.move(hwpss_model_name, None)
```
samdayweiss, line 694: I think this will fail if the model name does not already exist in the parent axis manager?

ykyohei:
This has been tested to run with no errors, so I'm going to merge it. But we discuss more in original PR.

PR #1120, opened 2025-02-23
add scan_flags to azss modules

ykyohei:
I propose this way of estimation and subtraction of azss in left_going scans and right_going scans separately.
Also fixed the azss extrapolation at the boundaries of subscans. This resolves the glitches near turnarounds induced by azss subtraction.
I tested updated individual functions and a proprocess module.

### DIFF
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -828,7 +828,12 @@ class EstimateAzSS(_Preprocess):
           subtract: True
 
     If we estimate and subtract azss in left going scans only,
-    make union of gltich_flags and scan_flags first then
+    make union of gltich_flags and scan_flags first
+
+      - name : "union_flags"
+        process:
+          flag_labels: ['glitches.glitch_flags', 'turnaround_flags.right_scan']
+          total_flags_label: 'glitch_flags_left'
 
       - name: "estimate_azss"
         calc:
@@ -836,7 +841,7 @@ class EstimateAzSS(_Preprocess):
           azss_stats_name: 'azss_statsQ_left'
           range: [-1.57079, 7.85398]
           bins: 1080
-          flags: 'union of glitch_flags and right_scan (~left_scan)'
+          flags: 'glitch_flags_left'
           scan_flags: 'left_scan'
           merge_stats: False
           merge_model: False
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -873,6 +873,7 @@ class EstimateAzSS(_Preprocess):
                     scan_flags=self.calc_cfgs.get('scan_flags'),
                     method=self.calc_cfgs.get('method'),
                     max_mode=self.calc_cfgs.get('max_mode'),
+                    range=self.calc_cfgs.get('range'),
                     in_place=True
                 )
         else:
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -827,7 +827,8 @@ class EstimateAzSS(_Preprocess):
         process:
           subtract: True
 
-    If we estimate and subtract azss in left going scans only
+    If we estimate and subtract azss in left going scans only,
+    make union of gltich_flags and scan_flags first then
 
       - name: "estimate_azss"
         calc:
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -820,8 +820,24 @@ class EstimateAzSS(_Preprocess):
           azss_stats_name: 'azss_statsQ'
           range: [-1.57079, 7.85398]
           bins: 1080
+          flags: 'glitch_flags'
           merge_stats: False
           merge_model: False
+        save: True
+
+    If we estimate and subtract azss in left going scans only
+
+      - name: "estimate_azss"
+        calc:
+          signal: 'demodQ'
+          azss_stats_name: 'azss_statsQ_left'
+          range: [-1.57079, 7.85398]
+          bins: 1080
+          flags: 'union of glitch_flags and right_scan (~left_scan)'
+          scan_flags: 'left_scan'
+          merge_stats: False
+          merge_model: False
+          subtract_in_place: True
         save: True
 
     .. autofunction:: sotodlib.tod_ops.azss.get_azss
@@ -831,7 +847,7 @@ class EstimateAzSS(_Preprocess):
     def calc_and_save(self, aman, proc_aman):
         calc_aman, _ = tod_ops.azss.get_azss(aman, **self.calc_cfgs)
         self.save(proc_aman, calc_aman)
-    
+
     def save(self, proc_aman, azss_stats):
         if self.save_cfgs is None:
             return
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -823,6 +823,7 @@ class EstimateAzSS(_Preprocess):
           flags: 'glitch_flags'
           merge_stats: False
           merge_model: False
+          subtract_in_place: True
         save: True
         process:
           subtract: True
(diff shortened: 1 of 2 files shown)


mmccrackan (changes requested):
Just some initial comments while I'm testing this.  I think the main one is with regard to the `process` function.

sotodlib/preprocess/processes.py
```diff
@@ -820,18 +820,34 @@ class EstimateAzSS(_Preprocess):
...
 
+    If we estimate and subtract azss in left going scans only
+
+      - name: "estimate_azss"
+        calc:
+          signal: 'demodQ'
+          azss_stats_name: 'azss_statsQ_left'
+          range: [-1.57079, 7.85398]
+          bins: 1080
+          flags: 'union of glitch_flags and right_scan (~left_scan)'
```
mmccrackan, line 836: This is going to require the union of these flags to be calculated by `union_flags` right?  I don't think that is currently set up to do this so it might also need to be updated in this PR.

sotodlib/preprocess/processes.py
```diff
@@ -820,18 +820,34 @@ class EstimateAzSS(_Preprocess):
...
+          merge_stats: False
+          merge_model: False
+          subtract_in_place: True
+        save: True
+
     .. autofunction:: sotodlib.tod_ops.azss.get_azss
     """
     name = "estimate_azss"
 
     def calc_and_save(self, aman, proc_aman):
```
mmccrackan, line 847: I think we will also need a `process` function to calculate and subtract the AZSS since `calc_and_save` is not called when loading from an existing database.

sotodlib/tod_ops/azss.py
```diff
@@ -238,42 +251,44 @@ def get_azss(aman, signal='signal', az=None, range=None, bins=100, flags=None,
...
         if type(max_mode) is not int:
             raise ValueError('max_mode is not provided as integer')
         azss_stats, model_sig_tod = fit_azss(az=az, azss_stats=azss_stats, max_mode=max_mode, fit_range=range)
-        
+
     if method == 'interpolate':
-        f_template = interp1d(bin_centers, binned_signal, fill_value='extrapolate')
+        # mask az bins that has no data and extrapolate
+        mask = ~np.any(np.isnan(binned_signal), axis=0)
+        f_template = interp1d(bin_centers[mask], binned_signal[:, mask], fill_value='extrapolate')
```
mmccrackan, line 280: Just a heads up that we have a parallelized interpolation function in so3g now that should be faster.  It can be called with something like:

```
import so3g
model_sig_tod = np.zeros_like(aman.signal, dtype='float64')
_bin_centers = bin_centers[mask].copy()
_binned_signal = binned_signal[:,mask].copy()
so3g.interp1d_linear(_bin_centers, _binned_signal, aman.boresight.az, model_sig_tod)
```

Inputs need to be contiguous and have the same dtype.

mmccrackan (changes requested):

sotodlib/preprocess/processes.py
ykyohei: We have UnionFlags so my intention was to just call it beforehand.

sotodlib/tod_ops/azss.py
ykyohei: Thanks for the heads up. Let me try to minimize the changes this time. It's great to have this when we find this is the bottle neck.

sotodlib/preprocess/processes.py
ykyohei: I think this is fixed. I hope people like this simpler method.

sotodlib/tod_ops/azss.py
mmccrackan: Cool, we can do this in another PR as it is taking >20 or more seconds per TOD for the AZSS calculation when running on the NERSC login node.  We also have a detrend function that will help with this.

sotodlib/tod_ops/azss.py
```diff
@@ -238,45 +250,58 @@ def get_azss(aman, signal='signal', az=None, range=None, bins=100, flags=None,
...
+            raise ValueError('max_mode is not provided as integer')
+        azss_stats, model = fit_azss(az=az, azss_stats=azss_stats,
+                                     max_mode=max_mode,
+                                     fit_range=range)
+    if method == 'interpolate':
+        # mask az bins that has no data and extrapolate
+        mask = ~np.any(np.isnan(azss_stats.binned_signal), axis=0)
+        f_template = interp1d(azss_stats.binned_az[mask], azss_stats.binned_signal[:, mask], fill_value='extrapolate')
+        model = f_template(az)
+    return model
```
mmccrackan, line 300: In addition to the mask here, I think we actually do need to check the models for NaNs and set them to zero as was done here: https://github.com/simonsobs/sotodlib/commit/c5dcb5e4d78bef13e612a128a32145c18bf8a3fe. It was added not to remove the NaNs from the bin issue but due to some left or right scans being completely flagged, resulting in the model producing NaNs.

sotodlib/tod_ops/azss.py
```diff
@@ -238,45 +250,58 @@ def get_azss(aman, signal='signal', az=None, range=None, bins=100, flags=None,
...
-        model_sig_tod = f_template(aman.boresight.az)
+    model_sig_tod = get_azss_model(aman, azss_stats, az, method, max_mode, range)
 
     if merge_stats:
         aman.wrap(azss_stats_name, azss_stats)
     if merge_model:
         aman.wrap(azss_model_name, model_sig_tod, [(0, 'dets'), (1, 'samps')])
     if subtract_in_place:
-        aman[signal_name] = np.subtract(signal, model_sig_tod, dtype='float32')
+        aman[signal_name][:, scan_flags] = np.subtract(signal, model_sig_tod, dtype='float32')[:, scan_flags]
```
mmccrackan, line 278: Can `float32` be replaced with `dtype` here as was done on line 366?

sotodlib/preprocess/processes.py
```diff
@@ -820,24 +820,71 @@ class EstimateAzSS(_Preprocess):
...
+            else:
+                tod_ops.azss.subtract_azss(
+                    aman,
+                    proc_aman.get(self.calc_cfgs.get('azss_stats_name')),
+                    signal=self.calc_cfgs.get('signal'),
+                    scan_flags=self.calc_cfgs.get('scan_flags'),
+                    method=self.calc_cfgs.get('method'),
+                    max_mode=self.calc_cfgs.get('max_mode'),
+                    range=self.calc_cfgs.get('range'),
+                    in_place=True
```
mmccrackan, line 883: These will fail when loading from a config that doesn't specify all of these parameters.  For example, if you don't specify `method` in the config, it will pass in `None` and therefore crash as the model is not calculated.  I think you'll need to replace optional params with `**self.calc_cfgs`.

mmccrackan (changes requested):
A few more comments on things I found while testing this.

sotodlib/preprocess/processes.py
mmccrackan: `calc_and_save` worked and this looks pretty consistent with the `iso-sat-review` branch which is great.  See the comment below on the `process` function input handling.

sotodlib/preprocess/processes.py
ykyohei: I didn't do **self.calc_cfgs because it could lead to unexpected keyword arguments for subtract_azss.
I thought this is fine, for example if you don't specify method, it will crash both estimate_azss and subtract_azss.

sotodlib/tod_ops/azss.py
ykyohei: I think current method is slightly robust than that, but I didn't testes for all the observations, so I added zero set with warning.

mmccrackan (changes requested):
Looking good.  Found a minor typo.  Also, we should probably rename the preprocess function from `EstimateAzSS` to just `AzSS` since it also subtracts it.

sotodlib/preprocess/processes.py
```diff
@@ -820,24 +820,72 @@ class EstimateAzSS(_Preprocess):
...
+        if self.calc_cfgs.get('azss_stats_name') in proc_aman and self.process_cfgs["subtract"]:
+            if sim:
+                tod_ops.azss.get_azss(aman, **self.calc_cfgs)
+            else:
+                tod_ops.azss.subtract_azss(
+                    aman,
+                    proc_aman.get(self.calc_cfgs.get('azss_stats_name')),
+                    signal=self.calc_cfgs.get('signal', 'signal'),
+                    scan_flags=self.calc_cfgs.get('scan_flags'),
+                    method=self.calc_cfgs.get('method', 'interpolation'),
```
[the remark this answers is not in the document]
ykyohei: Sorry! Thanks for catching.

mmccrackan (approved):
Okay, I tested both running from scratch and loading back from an existing database and I get consistent results in both cases for interpolation and fitting.

ykyohei:
Thanks so much for checking it!

PR #1121, opened 2025-02-24
fix azss extrapolation

ykyohei:
Cherry pick from https://github.com/simonsobs/sotodlib/pull/1120
When azss bin range is wider than scanned az, the azss will be nan. We need to mask nan bins to correctly extrapolate azss.
This resolves the glitches near the boundaries of subscan induced by azss subtraction.


mmccrackan (approved):
Looks good.

PR #1125, opened 2025-02-26
azss template subtraction

ykyohei:
A different approach to azss subtraction, with a milder transfer function.

link to a presentation showing motivation and status https://simonsobs.atlassian.net/wiki/spaces/PRO/pages/806846478/SAT+LAT+Analysis+Section+1+telecons#06-March-2025-(A-slot---10am-ET)


msilvafe:
Hi @ykyohei I think I missed your presentation about this (can you link it) but reading the code it seems like you're creating a common mode AzSS template by averaging the AzSS across all detectors in a process group (obs-wafer-band) then using that template to project out through least-squares the coefficient coupling to each detector. This ends up being a non-linear filter as the template shape is changing per obs right? Or are you envisioning a template that gets built up over the full season such that you have one template per wafer-band-<left/right> and then we use that template over the whole season? If the latter how is that template provided/loaded? And how do you quantify the time-variability of such a template?

ykyohei:
Hi Max, yes, it will be challenging but I was envisioning the possibility of using same template over certain period of time, if it works sufficiently. In that case we will make new azss metadata with range-match and this module will fit the amplitude coefficient (or something) for each detector for each observation on the fly. if our relcal works sufficiently we might not need this "fitting".
Independent studies for time variability of azss will determine the time-range of azss template metadata.

msilvafe:
> Hi Max, yes, it will be challenging but I was envisioning the possibility of using same template over certain period of time, if it works sufficiently. In that case we will make new azss metadata with range-match and this module will fit the amplitude coefficient (or something) for each detector for each observation on the fly. if our relcal works sufficiently we might not need this "fitting". Independent studies for time variability of azss will determine the time-range of azss template metadata.

Hi Kyohei, that sounds good. But for the code as written here you're computing the template either using PCA or with a weighted average over detectors for every obs-wafer-band and then subtracting it from the data. I don't think we want to implement it this way right? I think this needs to be able to take in a template from the metadata as you suggest perhaps initially we just have 1 template that spans all time that's basically the template you show in your presentation linked above and then this code just computes the per-det coefficients with least squares and subtracts that scaled version of the template.

ykyohei:
I don't think so, I think we should start from less assumptions, and I thought this is a good start point. 
Using one template for every obs-wafer-band is very aggressive. I'm happy to add metadata or the interface with it when we have more confidence.

msilvafe:
> I don't think so, I think we should start from less assumptions, and I thought this is a good start point. Using one template for every obs-wafer-band is very aggressive. I'm happy to add metadata or the interface with it when we have more confidence.

But this is non-linear and what we are doing now (binned template w/ interpolation or legendre polynomials) is not. And the full season template would also not be non-linear. So what do you think is the benefit of this intermediate approach relative to what we have now?

ykyohei:
Better transfer function is the motivation. I think this approach is effective for SAT because of its larger field of view.

msilvafe (approved):
Ok @ykyohei and I spoke about the approach here. My one qualm with this is that averaging over _detectors_ as is done here makes a non-linear filter as the template is unique per observation and thus incompatible with the observation matrix however the longer term strategy will be to build up a template averaging over _observations_ not over _detectors_ which is compatible with obsmat.

The benefit of this method for the time being is that we can average over the cmb signal at different sky rotation angles per az bin reducing cmb mode loss and improving the TF without having to include data from a long time period which requires deeper understanding of the time variability of the Az/Ground-template. In the longer run we will want to build up an Az/Ground-template and understand all of the time variability to figure out how many templates we need over a season but this allows us another tool to use now as a step toward that goal. 

I did not run this so I would just like you to confirm the following before squash+merge:
1. This was run through a preproc pipeline without failure exercising all of the new optional arguments you've implemented.
2. This doesn't break our existing pipelines. I.e. we can still run the older versions of AzSS estimation/subtraction without errors (including what we've been running in the sat-iso-review branch)

PR #1127, opened 2025-02-27
obsfiledb: v3, with index on obs_id and detset

### LINKED ISSUE #1123: obsfiledb needs "indexes" defined.
obsfiledb is way faster if you define obs_id as an index.  Like this:

`CREATE INDEX IF NOT EXISTS idx_column_name ON {table}(obs_id);`

mhasself:
so-metadata script (fix-db) can be used to add this index, in situ.

Thanks to @JBorrow

Resolves #1123


AERAdler (approved):
This looks fine. Approved.

PR #1128, opened 2025-02-27
resolve azss conflict

ykyohei:
resolved conflict of azss module and merged master.
After this is merged. the difference of #1065 from master will be subscan_flag only


msilvafe:
I think we probably want resolve merge conflicts with respect to the sat-iso-review branch right and then at the end of iso we handle all of the merge conflicts from sat-iso-review -> master. But welcome to hear @chervias, @susannaaz, and @earosenberg's thoughts.

ykyohei:
Close this because we resolve in [iso-sat-review](https://github.com/simonsobs/sotodlib/pull/1147) instead

PR #1129, opened 2025-02-28
Workflow to increase patch version monthly

iparask:
Hello, this workflow increases the patch number from master monthly and makes a release. The purpose of this workflow is to provide some level of reproducibility in our socondas and not installing from master, but from a specific tag.

It gets launched at midnight of the first of every month. The action it uses requires a token with content write permissions as a secret (https://github.com/softprops/action-gh-release?tab=readme-ov-file#permissions)

I can also trigger it manually to verify if it works and then we can disable it. Let me know what you think!


mhasself:
Oh, thanks!  This looks good to me.  I'm going to release 0.6.0 on master and then we can start patch-bumping from there.

tskisner:
What do you guys think about versioning like `25.3.1` (i.e. date) since these are just made up versions anyway?  At least then we could easily tell how old the version is...

iparask:
I will update this after #1130 as it introduces `pyproject.toml`.

tskisner (approved):
This is great!  If there is some additional option to the `action-gh-release` action that would insert a list of the merged PRs since last release that would be awesome.

mhasself:
> What do you guys think about versioning like `25.3.1` (i.e. date) since these are just made up versions anyway? At least then we could easily tell how old the version is...

Just confirming for the record that we decided *not* to do date-based version codes.  Carry on with patch number bumping.

iparask:
There is nothing more to add. I changed it to use `gh` it is much simpler and allows for automations. This release https://github.com/simonsobs/so_mapmaking_campaign_manager/releases was created with the `gh` command in the workflow.

PR #1130, opened 2025-02-28
Update build system to work with python-3.12 and numpy-2

tskisner:
- Remove numpy<2 constraint

- Add pyproject.toml and move most package metadata out of setup.py

- Switch to non-vendored mode for versioneer


mhasself (approved):
Hooray!

PR #1131, opened 2025-02-28
Fix axisman.concat to merge string fields properly

mhasself:
Because https://github.com/numpy/numpy/issues/16377

### DIFF
--- a/sotodlib/core/axisman.py
+++ b/sotodlib/core/axisman.py
@@ -8,6 +8,8 @@ try:
 except ImportError:
     from scipy.sparse import csr_matrix as csr_array
 
+import so3g
+
 from .util import get_coindices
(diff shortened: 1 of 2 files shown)


msilvafe (approved):
I've tested this and it fixed the data loading error with the preproc metadata I was having. FYI @tristpinsm and @ahincks when this is merged record_QA should start working again.

tristpinsm:
> I've tested this and it fixed the data loading error with the preproc metadata I was having. FYI @tristpinsm and @ahincks when this is merged record_QA should start working again.

I confirmed that with this fix the preprocess metrics can be recorded for an `obs_id` that was failing before (for a single `obs_id` using the dev InfluxDB).

PR #1132, opened 2025-03-04
Fixes to drone simulation range

gabrielecoppi:
The PR finally solves the issue for drone scan size. Now the argument is required


tskisner (approved):
Although I did not test this, it looks fine to me.  Thanks for the formatting cleanup!

PR #1133, opened 2025-03-05
Implementing units for fits maps

chervias:
This adds the writing of units into the fits keyword BUNIT for maps produced by the ML and atomic-filterbin mapmakers. It requires the latest branch of pixell.


mhasself:
> It requires the latest branch of pixell.

Seems fine but we this to be passing tests.  Those pixell features need to be released on PyPI.

JBorrow (approved):

PR #1134, opened 2025-03-05
Action pull all history

iparask:
We want the whole history for this action. The checkout job pulls everything and prints also the old and new version.


tskisner (approved):
Thanks for the fix!

PR #1138, opened 2025-03-07
check-book/load-book: handle case where obsfiledb detset is just a bounding set

### LINKED ISSUE #1136: Obsfiledb doesn't support re-tuning on subset of bands
Max [started a discussion in slack](https://simonsobs.slack.com/archives/C034C6WQEHH/p1741217696076939) where for some wafers, the number of detectors loaded from an observation didn't match the number we expected: 
```
[IN]:
for i, obs in enumerate(obslist[-7:]):
    if np.isin(obs['stream_ids_list'], ['ufm_mv48r1','ufm_mv50r2']):
        aman = ctx.get_obs(obs)
        print(aman.obs_info.stream_ids_list, aman.dets.count)

[OUT]:
ufm_mv50r2 219
ufm_mv48r1 817
```
where we expect ~1600 channels per wafer.

This is similar to an [issue I posted](https://github.com/simonsobs/site-pipeline-configs/issues/60), however we believe this instance was caused by a partial retune. Max manually re-optimized wafers, and in the process created new tunes for a subset of bands while leaving the existing tunes for the remaining bands. I believe this created a new detset for just the re-tuned bands, and so only the re-tuned bands are being loaded by the observation.

I realize that supporting metadata involving retuning subsets of bands is potentially difficult, however, since this behavior is fully supported from the control-side, it is very easy for this to happen and for us to not notice it is a problem. I think when max loaded the data, no errors were thrown, and he only noticed that the yield was lower than expected. I think if the official stance is that this behavior is unsupported, we need to see alerts ASAP that books are in a degraded state.

mhasself:
Two fixes needed:
- when updating obsfiledb, allow a detset to expand.
- when loading books, accept that obs might not have all the dets that listed in obsfiledb detset.

Tested on:
```
oper_1741050783_satp1_0010000
oper_1741051126_satp1_0010000
oper_1741115973_satp1_0100000
oper_1741116451_satp1_0100000
```

Resolves #1136 .  Once this is merged I can update site obsfiledb with new dets, and also add the obsfiledb index from #1127 .

### DIFF
--- a/sotodlib/core/metadata/obsfiledb.py
+++ b/sotodlib/core/metadata/obsfiledb.py
@@ -186,11 +186,12 @@ class ObsFileDb:
         self.conn.commit()
 
     def add_detset(self, detset_name, detector_names, commit=True):
-        """Add a detset to the detsets table.
+        """Add a detset to the detsets table (by adding detectors with
+        specific names to it).
 
         Arguments:
-          detset_name (str): The (unique) name of this detset.
-          detector_names (list of str): The detectors belonging to
+          detset_name (str): The name of the detset.
+          detector_names (list of str): New detectors belonging to
             this detset.
 
         """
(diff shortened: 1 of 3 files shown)


AERAdler (approved):
Having read your exchanges with Katie in the issue page I believe this can be approved.

Issue #969, opened 2024-09-27
`iir_params` keys `a` and `b` are not populated for some observations


mmccrackan:
I searched and didn't see any issues for this error specifically, but for some observations, the `iir_params` dictionary populated by `_load_book_detset` in `load_book.py` contains `None` for the IIR filter parameters `a` and `b` such that the `fourier_filter` step fails with:

`"/global/homes/m/mmccrack/so_home/repos/20240819_catchup/sotodlib/sotodlib/tod_ops/filters.py", line 522, in iir_filter
    B, A = np.polyval(b[::-1], z), np.polyval(a[::-1], z)`
`TypeError: 'NoneType' object is not subscriptable`

The other parameters in the `iir_params` dictionary (`enabled` and `fscale`) are populated.  There are checks in the `iir_filter` function in `filters.py` to see `a` and `b` were function inputs and it defaults to those in the `tod` axis manager if they are not, though there are no checks to see if they are `None` in the axis manager dictionary itself.

An example observation where this occurs is:

obsid = 'obs_1717755016_satp3_1111111'
dets={'wafer_slot':'ws4', 'wafer.bandpass':'f150'}

chervias:
Yes, we have known about these for a while, they crashed the mapmaker many many times and we had to put specific try catch statements to avoid that. I'm not sure why this happens.

msilvafe:
This issue goes all the way down our software stack to the streamer (https://github.com/simonsobs/smurf-streamer/issues/40) so I will assign @jlashner

mhasself:
This is a data integrity issue, and might indicate deeper problems; thanks for creating an issue.

Questions I'd like to have answers too ASAP:
- How often does this happen?
- Do we detect any cases where the readout filter indeed seems to be misconfigured (not just reporting incorrectly)?

Whether or not this gets fixed upstream, we will probably want a mitigation for existing data (unless it's extremely rare).  I don't really like the approach of "oh just put in default filter params if they're not there"... that defeats all the care we've taken to record the state of the system.  I would rather do this with an external (obsdb/obs_info) tag that confirms, for certain obs, that it's safe to jam in filter parameters.

jlashner:
I don't have a full quantitative response, but I did write a script to scan satp1 data to check for this problem, and though I didn't check the full archive, I couldn't find any examples of it.

Is it true this only happens for satp3?

My current theory is that this happens whenever Satp3 uses `jackhammer hammer --soft` instead of doing a full hammer.
[...]

PR #1139, opened 2025-03-07
correct missing iir_params

ykyohei:
This should be a temporary solution for this issue but it's helpful to have..

### DIFF
--- a/sotodlib/obs_ops/utils.py
+++ b/sotodlib/obs_ops/utils.py
@@ -1,0 +1,42 @@
+from sotodlib import core
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+def iir_params_correction(aman):
+    """
+    Add iir_params if missing
+
+    See `sotodlib.tod_ops.filters.iir_filter` for more detail
+
+    Parameters
+    ----------
+    aman: AxisManager of observation
+
+    Returns
+    -------
+    List of field names that have no iir_params
+
+    """
+    a = [1., -3.74145562, 5.25726624, -3.28776591,
+         0.77203984, 0., 0., 0.,
+         0., 0., 0., 0.,
+         0., 0., 0., 0.]
+    b = [5.28396689e-06, 2.11358676e-05, 3.17038014e-05, 2.11358676e-05,
+         5.28396689e-06, 0.00000000e+00, 0.00000000e+00, 0.00000000e+00,
+         0.00000000e+00, 0.00000000e+00, 0.00000000e+00, 0.00000000e+00,
+         0.00000000e+00, 0.00000000e+00, 0.00000000e+00, 0.00000000e+00]
+
+    iir_missing = []
+    for _field, _sub_iir_params in aman.iir_params._fields.items():
+        if isinstance(_sub_iir_params, core.AxisManager):
+            if 'a' in _sub_iir_params._fields:
+                if _sub_iir_params['a'] is None:
+                    iir_missing.append(_field)
+                    logger.warning(f'iir_params are missing on {_field}. '
+                                   'Fill default params.')
+                    _sub_iir_params['a'] = a
+                    _sub_iir_params['b'] = b
+
+    return iir_missing
--- a/sotodlib/obs_ops/utils.py
+++ b/sotodlib/obs_ops/utils.py
@@ -4,7 +4,7 @@ import logging
 logger = logging.getLogger(__name__)
 
 
-def iir_params_correction(aman):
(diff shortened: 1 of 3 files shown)


sotodlib/obs_ops/utils.py
```diff
@@ -0,0 +1,42 @@
+from sotodlib import core
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+def iir_params_correction(aman):
```
mhasself, line 7: Use a verb in function name: "correct_iir_params".

sotodlib/preprocess/processes.py
```diff
@@ -1760,6 +1760,23 @@ def process(self, aman, proc_aman):
         if self.process_cfgs:
             pointing_model.apply_pointing_model(aman)
 
+class IIRParamsCorrection(_Preprocess):
+    """Correct missing iir_params
+
+    Example config block::
+
+        - name : "iir_params_correction"
+          process: True
```
mhasself, line 1769: Can you add a "time_range" option, e.g.
```
time_range: ["2023-01-01", "2025-01-01"]
```

As discussed (#969), I don't like just painting over missing metadata, assuming that any failures in the future will be the same cause as failures of the past.  If we believe that all SATP3 filter issues in the past have been just reporting errors, but the filter was configured correctly, then we can _explicitly_ say "it's safe to patch these data by just writing in default values".

mhasself (commented):
A couple of comments in line, but ...

This was discussed a bunch in #969.  Have all the concerns there been resolved?  The conclusion there was that these dets should be _cut_, not patched.

ykyohei:
Thank you for the comments.
Yes. I was aware of the discussion. I thought there was a suggestion of having this kind of patch preprocess, so I did it.
I was doing this patch for my analyses, and I have not found obvious issues so far.

sotodlib/preprocess/processes.py
```diff
@@ -1760,6 +1760,34 @@ def process(self, aman, proc_aman):
...
+    """
+    name = "correct_iir_params"
+    def __init__(self, step_cfgs):
+        self.time_range = step_cfgs.get('time_range')
+        super().__init__(step_cfgs)
+
+    def process(self, aman, proc_aman):
+        import datetime
+        from sotodlib.obs_ops import correct_iir_params
+        if self.time_range is not None:
```
mhasself, line 1783: It is odd to do nothing, if that parameter is missing.  It should either be:
- an error to omit time_range argument, or
- time_range argument is optional.

mhasself (commented):
There are hard-coded params here, but that can be generalized in the future (hopefully we won't need to...).

I have only that one last inline comment ... adding @msilvafe as the real reviewer.

sotodlib/preprocess/processes.py
```diff
@@ -1760,6 +1760,37 @@ def process(self, aman, proc_aman):
...
+    name = "correct_iir_params"
+    def __init__(self, step_cfgs):
+        self.time_range = step_cfgs.get('time_range')
+        super().__init__(step_cfgs)
+
+    def process(self, aman, proc_aman):
+        import datetime
+        from sotodlib.obs_ops import correct_iir_params
+        if self.time_range is None:
+            correct_iir_params(aman)
```
msilvafe, line 1785: I didn't interpret @mhasself's comment to be if time range is optional you by default fill the values. I think we want something like this:
1. Decide on a time range where we believe it's safe to fill the values. Hard code that time-range in per telescope in the obs_ops.util function
2. Check if the observation is in this time range.
3. If it's not in the time range and the iir params are missing raise an error, if it is in that time range and the iir params are missing fill them, if the iir params aren't missing don't do anything

I have a slack message from Jack that says he told @samdayweiss to stop soft-hammering on Oct 29th 2024 so I would say we make anything before Oct 29th the cutoff and if we have any missing iir params after that date we need to know about them and understand why as its a new issue.

msilvafe (changes requested):

sotodlib/preprocess/processes.py
ykyohei: Thank you. I determined the time range based on the screening of iir_params https://github.com/simonsobs/smurf-streamer/issues/40
Tested time range check is working.

msilvafe (approved):
I like this level of strictness, thank you for updating it. When we have more epochs or iir params issues we'll need to implement something fancier as discussed in #969 but this should address the concerns in that issue about blanket filling missing values!

Issue #985, opened 2024-10-03
Preprocess: Add preprocess method that checks on stability of HWPSS


msilvafe:
Calculate the variation in the HWPSS over an observation. In particular we are most interested in A2 but could generalize this to any harmonic. This would essentially be demodulating that harmonic and computing statistics on that timestream (variance, RMS, peak-peak, etc.) and saving those stats into the preprocess manifest db.

PR #1144, opened 2025-03-12
Preprocess method for HWPSS A2 stats

jsyleung:
Addresses #985

Use `hwp.demod_tod` to demodulate the 2nd HWP harmonic and compute statistics on the demodulated timestream.

The current implementation runs into a name conflict if `hwp.demod_tod` has previously been called, since the field names “dsT”, “demodQ”, and “demodU” are hardcoded. Short of changing that function, one possible solution might be to make a copy of the AxisManager with those fields removed, but making a copy of the data on RAM is not ideal.

Suggestions welcome.

### DIFF
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -779,6 +779,50 @@ class SubtractHWPSS(_Preprocess):
                 subtract_name = self.process_cfgs["subtract_name"]
                 )
 
+class A2Stats(_Preprocess):
+    """
+    Calculate statistical metrics for A2, the 2f-demodulated Q and U signals.
+
+    Example config block::
+
+      - name: "a2_stats"
+        calc:
+          stat_names: ["mean", "median", "var", "ptp"]
+          subscan: True
+        save: True
+
+    """
+    name = "a2_stats"
+
+    def calc_and_save(self, aman, proc_aman):
+        # Get A2 signal using the demod_tod function; this function generates intermediary
+        # dsT, demodQ, demodU fields in the AxisManager that will be deleted
+        hwp.demod_tod(aman, demod_mode=2)
+
+        # Compute A2 stats
+        stat_names = self.calc_cfgs.get("stat_names", ["mean", "median", "var", "ptp"])
+        split_subscans = self.calc_cfgs.get("subscan", False)
+        am_name = "a2_stats"
+        # Q
+        tod_ops.flags.get_stats(aman, aman.demodQ, stat_names=stat_names, split_subscans=split_subscans, name=am_name, merge=True)
+        for sn in stat_names: aman[am_name].move(sn, f"{sn}Q")
+        # U
+        aman[am_name].merge(tod_ops.flags.get_stats(aman, aman.demodU, stat_names=stat_names, split_subscans=split_subscans))
+        for sn in stat_names: aman[am_name].move(sn, f"{sn}U")
+
+        # Delete the intermediary data fields from the AxisManager
+        aman.move('dsT', None)
+        aman.move('demodQ', None)
+        aman.move('demodU', None)
+
+        self.save(proc_aman, aman[am_name])
+
+    def save(self, proc_aman, a2_stats):
+        if self.save_cfgs is None:
+            return
+        if self.save_cfgs:
+            proc_aman.wrap("a2_stats", a2_stats)
+
 class Apodize(_Preprocess):
     """Apodize the edges of a signal. All process configs go to `apodize_cosine`
 
@@ -1780,6 +1824,7 @@ _Preprocess.register(Noise)
 _Preprocess.register(Calibrate)
 _Preprocess.register(EstimateHWPSS)
 _Preprocess.register(SubtractHWPSS)
+_Preprocess.register(A2Stats)
 _Preprocess.register(Apodize)
 _Preprocess.register(Demodulate)
 _Preprocess.register(AzSS)
--- a/sotodlib/tod_ops/flags.py
+++ b/sotodlib/tod_ops/flags.py
@@ -901,7 +901,7 @@ def get_stats(aman, signal, stat_names, split_subscans=False, mask=None, name="s
         Name of axis manager to add to aman if merge is True.
     """
     stat_names = np.atleast_1d(stat_names)
-    fn_dict = {'mean': np.mean, 'median': np.median, 'ptp': np.ptp, 'std': np.std,
+    fn_dict = {'mean': np.mean, 'median': np.median, 'ptp': np.ptp, 'std': np.std, 'var': np.var,
                      'kurtosis': stats.kurtosis, 'skew': stats.skew}
 
     if isinstance(signal, str):


sotodlib/preprocess/processes.py
```diff
@@ -779,6 +779,50 @@ def process(self, aman, proc_aman):
...
+        # Get A2 signal using the demod_tod function; this function generates intermediary
+        # dsT, demodQ, demodU fields in the AxisManager that will be deleted
+        hwp.demod_tod(aman, demod_mode=2)
+
+        # Compute A2 stats
+        stat_names = self.calc_cfgs.get("stat_names", ["mean", "median", "var", "ptp"])
+        split_subscans = self.calc_cfgs.get("subscan", False)
+        am_name = "a2_stats"
+        # Q
+        tod_ops.flags.get_stats(aman, aman.demodQ, stat_names=stat_names, split_subscans=split_subscans, name=am_name, merge=True)
```
msilvafe, line 807: Let's not `merge=True` see some of the other processes for an example but create a new axis_manager and wrap all saved data products into that axis_manager then send that to `save`

sotodlib/preprocess/processes.py
```diff
@@ -779,6 +779,50 @@ def process(self, aman, proc_aman):
...
+          subscan: True
+        save: True
+
+    """
+    name = "a2_stats"
+
+    def calc_and_save(self, aman, proc_aman):
+        # Get A2 signal using the demod_tod function; this function generates intermediary
+        # dsT, demodQ, demodU fields in the AxisManager that will be deleted
+        hwp.demod_tod(aman, demod_mode=2)
```
msilvafe, line 800: I think modify hwp.hwp.demod_tod to have a `wrap` argument that defaults to True lines 643-657 would then be wrapped in a `if wrap:` block and the `else` would just return the 3 arrays. Then here you could just return the 3 arrays and not worry about wrapping them and deleting them from `aman` later.

msilvafe (changes requested):
Thanks for adding this! A couple inline comments!

jsyleung:
See if this is what you’re looking for.

Is a unit test necessary here?

sotodlib/hwp/hwp.py
```diff
@@ -640,18 +643,29 @@ def demod_tod(aman, signal=None, demod_mode=4,
...
+            aman, lpf, signal_name='demodU', detrend=None) * 2.
+    else:
+        demod_aman = core.AxisManager(aman.dets, aman.samps)
+        demod_aman.wrap("timestamps", aman.timestamps, axis_map=[(0, 'samps')])
+        # Filter the demodulated signal
+        demod_aman.wrap("demodQ", demod.real, axis_map=[(0, 'dets'), (1, 'samps')])
+        demod_aman["demodQ"] = tod_ops.fourier_filter(demod_aman, lpf, signal_name="demodQ", detrend=None) * 2.
+        demod_aman.wrap("demodU", demod.imag, axis_map=[(0, 'dets'), (1, 'samps')])
+        demod_aman["demodU"] = tod_ops.fourier_filter(demod_aman, lpf, signal_name="demodU", detrend=None) * 2.
+        return aman[signal_name], demod_aman['demodQ'], demod_aman['demodU']
```
msilvafe, line 671: How about something like below the prevents some duplicated lines and treats T equivalently in both cases?
```
demod_aman = core.AxisManager(aman.dets, aman.samps)
demod_aman.wrap("timestamps", aman.timestamps, axis_map=[(0, 'samps')])
demod_aman.wrap("dsT", demod.real, axis_map=[(0, 'dets'), (1, 'samps')])
demod_aman["dsT"] = tod_ops.fourier_filter(aman, lpf, signal_name='dsT', detrend=None)
demod_aman["demodQ"] = tod_ops.fourier_filter(demod_aman, lpf, signal_name="demodQ", detrend=None) * 2.
demod_aman.wrap("demodQ", demod.real, axis_map=[(0, 'dets'), (1, 'samps')])
demod_aman["demodQ"] = tod_ops.fourier_filter(demod_aman, lpf, signal_name="demodQ", detrend=None) * 2.
demod_aman.wrap("demodU", demod.imag, axis_map=[(0, 'dets'), (1, 'samps')])
demod_aman["demodU"] = tod_ops.fourier_filter(demod_aman, lpf, signal_name="demodU", detrend=None) * 2.
if wrap:
    for fld in ['dsT', 'demodQ', 'demodU']:
        aman.wrap(fld, demod_aman[fld], axis_map=[(0, 'dets'), (1, 'samps')])
    del demod_aman
else:
    return demod_aman['dsT'], demod_aman['demodQ'], demod_aman['demodU']
 ```

File: sotodlib/preprocess/processes.py [hunk 4]
```diff
@@ -779,6 +779,47 @@ def process(self, aman, proc_aman):
...
+        # Q
+        a2stats_aman = tod_ops.flags.get_stats(aman, demodQ, stat_names=stat_names, split_subscans=split_subscans)
+        for sn in stat_names:
+            a2stats_aman.move(sn, f"{sn}Q")
+        # U
+        a2stats_aman.merge(tod_ops.flags.get_stats(aman, demodU, stat_names=stat_names, split_subscans=split_subscans))
+        for sn in stat_names:
+            a2stats_aman.move(sn, f"{sn}U")
+
+        aman.wrap("a2_stats", a2stats_aman)
```
msilvafe, line 814: We don't want to wrap this into the axis manager, just send it to `save` which will save it in `proc_aman` and will be loaded back into the axis manager when loaded from `load_and_preprocess`. I recognize that we're a bit inconsistent about this across the module now, there are _some_ preprocess methods that save to `aman` and `proc_aman` but we should standardize to not modify `aman` in `calc_and_save` if possible.

msilvafe (approved):
Looks good, just a couple suggestions. Sorry for the delay.

sotodlib/preprocess/processes.py
jsyleung: Okay; I’ve deleted line 814

sotodlib/hwp/hwp.py
jsyleung: I had been keeping old code intact, but now that I’m more comfortable working with axis managers, I’ve understand what you’re doing. Just to check:

- Line 3: `demod.real` → `aman[signal_name]`
- Line 4: `aman` → `demod_aman` (like following lines) since `'dsT'` got wrapped into the latter in the previous line
- Line 5: deleted; I believe this was a spurious addition (repeat of Line 7, but cannot work without Line 6)

sotodlib/preprocess/processes.py
```diff
@@ -779,6 +779,46 @@ def process(self, aman, proc_aman):
...
+    Uses tod_ops.flags.get_stats; refer there for list of available statistics.
+
+    Example config block::
+
+      - name: "a2_stats"
+        calc:
+          stat_names: ["mean", "median", "var", "ptp"]
+          subscan: True
+        save: True
+
```
msilvafe, line 794: Add references to tod_ops.flags.get_stats (See other preprocess methods for examples but should look like `.. autofunction:: sotodlib.tod_ops.flags.get_stats`) and `sotodlib.hwp.hwp.demod_tod` also usually we pass in `calc_cfgs` using the unpack `**` operation however here you're getting specific arguments only so please list in the docstring which specific arguments are allowed and what values/types they can take.

msilvafe (approved):
Looks good, one last in-line comment that's docs related. Thanks!

sotodlib/preprocess/processes.py
jsyleung: - Is sequential listing the only way to add two “autofunction” calls? i.e.,
```python
.. autofunction:: sotodlib.hwp.hwp.demod_tod
.. autofunction:: sotodlib.tod_ops.flags.get_stats
```

- To clarify, **are you asking me to change the** `get_stats()` **call to use** `**self.calc_cfgs`**?** e.g., earlier, in the `GetStats` class, https://github.com/simonsobs/sotodlib/blob/4c2a562b74a57c8b3784c8af4a1bee6690331537/sotodlib/preprocess/processes.py#L451 Unfortunately, `GetStats` also doesn’t list in its docstring all of the allowed argument values/types, so I’m not following what you’re looking for. The only two relevant `calc_cfgs` are `stat_names` and `subscan`, which are currently listed in the “Example config block” with example values (type is inferred). Since these aren’t arguments to the class, I wouldn’t be listing them under a (new) `Parameters` section. **Should I describe it in prose?**

- For reference, the “allowed” stats are hard-coded into `tod_ops.flags.get_stats()`: https://github.com/simonsobs/sotodlib/blob/4c2a562b74a57c8b3784c8af4a1bee6690331537/sotodlib/tod_ops/flags.py#L904-L905 but these aren’t listed anywhere, including the [docs](https://sotodlib.readthedocs.io/en/latest/tod_ops.html#sotodlib.tod_ops.flags.get_stats), nor is there a way “get” them.

PR #1146, opened 2025-03-17
Uhf labels

AERAdler:
f230/290 changed to f220/280 to match data labels.


kmharrington:
I'd like to check with @smsimon  and @keskitalo  that this won't be a breaking change for the sims. But we've confirmed these will be the names that come out of the real data

smsimon:
There are a lot of references to these elsewhere. The proposal we had within DM was to keep the labels in the sims the same and just have the sim and data labels be different.

AERAdler:
There are bad UHF labels in pwg-scripts as well... https://github.com/search?q=repo%3Asimonsobs%2Fpwg-scripts%20f290&type=code

PR #1147, opened 2025-03-18
ISO SAT review

chervias:
branch containing some of the updates required to run the SAT ISO analysis in https://github.com/simonsobs/pwg-scripts/tree/master/iso-sat-review


sotodlib/tod_ops/flags.py
```diff
@@ -720,7 +719,7 @@ def get_inv_var_flags(aman, signal_name='signal', nsigma=5,
 
     return mskinvar
 
-def get_subscans(aman, merge=True, include_turnarounds=False, overwrite=True):
+def get_subscans(aman, flags="flags", merge=True, include_turnarounds=False, overwrite=True):
```
ykyohei, line 722: flags here is unused. Also no docstring.

PR #1150, opened 2025-03-19
numpy 2 compatibility fixes

iparask:
These are numpy 2 compatibility issues:

https://numpy.org/doc/stable/reference/generated/numpy.in1d.html

The `unicode_` is deprecated and removed. Numpy says to use `str_` instead.


mhasself (approved):
Thanks.  Seems like this will work back to at least numpy 1.20, which should be enough for everybody.

PR #1151, opened 2025-03-19
Gp/act

iparask:
These are the proposed changes for continuing supporting ACT data with sotodlib


iparask:
This runs on Tiger3. In addition to the code, it required changes in the context file of ACT and the following DBs: `obsdb`, `obsdb_reduced`, and `detdb`.

In the context file, I added `array_name` and `tube_flavor` as `obs_colon_tags`. A column `tube_flavor` was added in the DBs with values based on the corresponding frequencies.

chervias (approved):
Looks good to me

PR #1153, opened 2025-03-20
Fix typos for LAT det match

mmccrackan:
Fixing some issues I found while trying to run LAT pointing fits through the detector matching code.


mhasself (approved):
Looks good to me.

Careful about mixing "reformatting" (even helpful whitespace cleanup) with substantial code changes.  I think both things are minor here, so don't worry about it.  But generally we will want those split into separate PRs.

Issue #1154, opened 2025-03-20
IndexError in jumps.std_est


ykyohei:
I see this error in new twopi_jumps, this happens when samps.count is multiple of win_size = 20
sotodlib version '0.6.2+11.ga3d5dd22'

```
from sotodlib import tod_ops, core
ctx = core.Context('/scratch/gpfs/SIMONSOBS/so/tracked/metadata/satp3/contexts/use_this_local.yaml')
aman = ctx.get_obs(obs_id = 'obs_1723454945_satp3_1111111', dets = {'wafer_slot': 'ws0', 'wafer.bandpass': 'f090'})
tod_ops.jumps.twopi_jumps(aman, signal=aman.signal, fix=False, overwrite=True)

---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Cell In[10], line 5
      2 ctx = core.Context('/scratch/gpfs/SIMONSOBS/so/tracked/metadata/satp3/contexts/use_this_local.yaml')
      3 aman = ctx.get_obs(obs_id = 'obs_1723454945_satp3_1111111', dets = {'wafer_slot': 'ws0', 'wafer.bandpass': 'f090'})
----> 5 tod_ops.jumps.twopi_jumps(aman, signal=aman.signal, fix=False, overwrite=True)

File [~/tools/sotodlib/sotodlib/tod_ops/jumps.py:447](https://mytiger.princeton.edu/node/tiger-g04c6n10/41547/lab/tree/home/kyoheiy/scratch/so_site/nb/~/tools/sotodlib/sotodlib/tod_ops/jumps.py#line=446), in twopi_jumps(aman, signal, win_size, nsigma, atol, max_tol, fix, inplace, merge, overwrite, name, ds, **filter_pars)
    445     raise TypeError("Signal is not an array")
    446 if atol is None:
--> 447     atol = nsigma * std_est(
    448         signal.astype(float), ds=win_size * ds, win_size=win_size
    449     )
    450     np.clip(atol, 1e-8, max_tol)
    452 _signal = _filter(signal, **filter_pars)

File [~/tools/sotodlib/sotodlib/tod_ops/jumps.py:60](https://mytiger.princeton.edu/node/tiger-g04c6n10/41547/lab/tree/home/kyoheiy/scratch/so_site/nb/~/tools/sotodlib/sotodlib/tod_ops/jumps.py#line=59), in std_est(x, ds, win_size, axis, method)
[...]

PR #1155, opened 2025-03-20
fix jump.std_est

ykyohei:
Fix slicing of std_est https://github.com/simonsobs/sotodlib/issues/1154


skhrg (changes requested):
Nice catch!

I think a slightly cleaner fix would be:

```
x = x[..., : x.shape[1] - (x.shape[-1] % win_size)]
```

That way it solves the `stop == 0` case but we can avoid adding the extra logic block.

skhrg (approved):
Looks good to me

PR #1156, opened 2025-03-24
Refactor preprocess_tod to use per-group parallelization

mmccrackan:
Work in progress.  Refactors main preprocess functions to parallelize over each individual detector grouping instead of just over observation id's so as to line up with mapmaking.

To-do:
- Refactor `multilayer_preprocess_tod`
- Merge both `preprocess_tod` and `multilayer_preprocess_tod's` underlying functions with preproc_or_load_group.


mmccrackan:
Just realized that making the error db an obsdb instead of a manifestdb would make more sense here as it isn't pointing to any files, though since it is keyed by obs and wafer we might need #1159 first.  Will see about updating it.

msilvafe:
Let's close this PR and split up this into two PRs one for the improved error raising and one that implements the per-group parallelization.

PR #1159, opened 2025-03-25
Obs wafer band in obsdb

msilvafe:
This is an update to obsdb to allow storing information keyed by obs-wafer-band without breaking the current obs-id primary key behavior and allowing us to retrieve common entries for an obs-id between 2 databases that could be keyed by obs-wafer-band or obsid.

### DIFF
--- a/sotodlib/core/metadata/obsdb.py
+++ b/sotodlib/core/metadata/obsdb.py
@@ -1,5 +1,6 @@
 import sqlite3
 import os
+import numpy as np
 
 from .resultset import ResultSet
 from . import common
@@ -17,6 +18,20 @@ TABLE_DEFS = {
     ],
 }
 
+TABLE_DEFS_OWB = {
+    'obs': [
+        "`obs_id` varchar(256)",
+        "`wafer_slot` varchar(256)",
+        "`bandpass` varchar(256)",
+        "`timestamp` float",
+        "PRIMARY KEY (`obs_id`, `wafer_slot`, `bandpass`)"
+    ],
+    'tags': [
+        "`obs_id` varchar(256)",
+        "`tag` varchar(256)",
+        "CONSTRAINT one_tag UNIQUE (`obs_id`, `tag`)",
+    ],
+}
 
 class ObsDb(object):
     """Observation database.
@@ -40,7 +55,7 @@ class ObsDb(object):
         "`obs_id` varchar(256)",
     ]
 
-    def __init__(self, map_file=None, init_db=True):
+    def __init__(self, map_file=None, init_db=True, obs_wafer_band=False):
         """Instantiate an ObsDb.
 
         Args:
@@ -65,6 +80,8 @@ class ObsDb(object):
             if map_file is None:
                 map_file = ':memory:'
             self.conn = sqlite3.connect(map_file)
+        if obs_wafer_band:
+            TABLE_DEFS = TABLE_DEFS_OWB
 
         self.conn.row_factory = sqlite3.Row  # access columns by name
         if init_db:
@@ -81,10 +98,38 @@ class ObsDb(object):
                     changes = True
             if changes:
                 self.conn.commit()
+        self.primary_keys = self.get_primary_keys()
+        if ~np.isin(len(self.primary_keys), [1,3]):
+            raise ValueError("Unsupported number of primary keys in the 'obs' table"
+                             " only 1 or 3 are allowed.")
 
+    def get_primary_keys(self):
+        """Retrieve the primary keys of the specified table.
+           This is used whether to index by obs_id or obs-wafer-band."""
+        query = "PRAGMA table_info('obs')"
+        c = self.conn.execute(query)
+        primary_keys = [row['name'] for row in c.fetchall() if row['pk'] > 0]
+        return primary_keys
+    
+    def warn_primary_keys(self, wafer_band):
+        """Warn the user if the primary keys are not specified 
+           and we're defaulting to using _all."""
+        if (None in wafer_band):
+            warn_str = 'WARNING: Primary key(s)'
+            for i, wb in enumerate(wafer_band):
+                if wb is None:
+                    wafer_band[i] = '_all'
+                    warn_str += f' wafer_band[{i}],'
+            warn_str += """
+            are not specified and ObsDb is indexed by obs_id, wafer_slot,
+            and bandpass. These keys will be set to _all.
+            """
+            print(warn_str)
+        return wafer_band
+    
     def __len__(self):
-        return self.conn.execute('select count(obs_id) from obs').fetchone()[0]
-
+        return self.conn.execute(f'SELECT COUNT({self.primary_keys[0]}) FROM obs').fetchone()[0]
+    
     def add_obs_columns(self, column_defs, ignore_duplicates=True, commit=True):
         """Add columns to the obs table.
 
@@ -146,11 +191,14 @@ class ObsDb(object):
             self.conn.commit()
         return self
 
-    def update_obs(self, obs_id, data={}, tags=[], commit=True):
+    def update_obs(self, obs_id, wafer_band=(None, None),
+                   data={}, tags=[], commit=True):
         """Update an entry in the obs table.
 
         Arguments:
           obs_id (str): The id of the obs to update.
+          wafer_band (tuple of str): The wafer_band[0] = wafer_slot and wafer_band[1] = bandpass
+            to update.  The default value is ('_all', '_all').
           data (dict): map from column_name to value.
           tags (list of str): tags to apply to this observation (if a
             tag name is prefxed with '!', then the tag name will be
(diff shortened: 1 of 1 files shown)


sotodlib/core/metadata/obsdb.py
```diff
@@ -65,6 +80,11 @@ def __init__(self, map_file=None, init_db=True):
             if map_file is None:
                 map_file = ':memory:'
             self.conn = sqlite3.connect(map_file)
+        
+        global TABLE_DEFS
+        global TABLE_DEFS_OWB
+        if obs_wafer_band:
+            TABLE_DEFS = TABLE_DEFS_OWB
```
mhasself, line 87: This will modify the module level global var, affecting any future instantiations of ObsDb.  So instead you'll want to store the table_defs as an instance variable, and have all the internal functions refer to that. (`self._table_defs = TABLE_DEFS` / `TABLE_DEFS_OWB`).

sotodlib/core/metadata/obsdb.py
```diff
@@ -160,22 +211,33 @@ def update_obs(self, obs_id, data={}, tags=[], commit=True):
             self.
 
         """
+        if (len(self.primary_keys) == 3):
+            wafer_band = self.warn_primary_keys(wafer_band)
+            obs_id = {'obs_id': obs_id, 'wafer_slot': wafer_band[0],
+                      'bandpass': wafer_band[1]}
+        else:
+            obs_id = {'obs_id': obs_id}
```
mhasself, line 219: Instead of re-using `obs_id`, call this aggregate var something like `obs_key`.

That's a terminology we could adopt globally.

sotodlib/core/metadata/obsdb.py
```diff
@@ -314,11 +392,41 @@ def query(self, query_text='1', tags=None, sort=['obs_id'], add_prefix=''):
...
         q = 'select obs.* %s from obs %s where %s %s' % (extra_fields, joins, query_text, sort_text)
-        c = self.conn.execute(q)
+        c = self.conn.execute(q, query_values)
         results = ResultSet.from_cursor(c)
         if add_prefix is not None:
             results.keys = [add_prefix + k for k in results.keys]
         return results
+    
+    def query_linked_dbs(primary_db, secondary_db, query_condition,
+                         query_values=(), add_prefix=''):
```
mhasself, line 402: This function is odd to see in the present context because it doesn't pay much attention to the 3-parameter keys... I guess it's intended for primary=normal_obsdb, and secondary=an_owb_obsdb.  Then you get the usual per-obs stuff, and an enhanced wafer/band entry for each obs.  Ok.  Might be worth a comment that either or neither of the dbs can be an owb.

Also `primary_db` is in the "self" position, so call it self.  Or make this a staticmethod.

Why only primary and secondary?  You could accept an arb number of dbs from the user.  (Though the primary db query is not the slow part -- the follow up, looping over obs for (each) secondary db will be.)  That being said, a helper function that accepts "primary_results", and looks up all this stuff in secondary, could be helpful; that's what you chain together to get primary -> secondary, thirdary, etc.

Instead of return list of dict of resultset.... easier to just return list of list/tuple of resultset?

sotodlib/core/metadata/obsdb.py
```diff
@@ -81,10 +101,38 @@ def __init__(self, map_file=None, init_db=True):
...
+        return primary_keys
+    
+    def warn_primary_keys(self, wafer_band):
+        """Warn the user if the primary keys are not specified 
+           and we're defaulting to using _all."""
+        if (None in wafer_band):
+            warn_str = 'WARNING: Primary key(s)'
+            for i, wb in enumerate(wafer_band):
+                if wb is None:
+                    wafer_band[i] = '_all'
```
mhasself, line 124: This failed for me because I passed in a tuple.  Also it's not great to modify someone's inputs if you don't need to.  Instead, start the func with `wafer_band = list(wafer_band)`; then you'll definitely be able to change it and return it.

sotodlib/core/metadata/obsdb.py
```diff
@@ -81,10 +101,38 @@ def __init__(self, map_file=None, init_db=True):
...
+                    warn_str += f' wafer_band[{i}],'
+            warn_str += """
+            are not specified and ObsDb is indexed by obs_id, wafer_slot,
+            and bandpass. These keys will be set to _all.
+            """
+            print(warn_str)
```
mhasself, line 130: use warnings.warn or whatever.

sotodlib/core/metadata/obsdb.py
```diff
@@ -81,10 +101,38 @@ def __init__(self, map_file=None, init_db=True):
                     changes = True
             if changes:
                 self.conn.commit()
-
+        self.primary_keys = self.get_primary_keys()
+        if ~np.isin(len(self.primary_keys), [1,3]):
+            raise ValueError("Unsupported number of primary keys in the 'obs' table"
+                             " only 1 or 3 are allowed.")
+
+    def get_primary_keys(self):
```
mhasself, line 109: This conflicts with database terminology "primary key", which is the combination of the fields.  So a better name for this function is "get_primary_key_keys" or "get_primary_key_fields".  So how about "get_primary_fields".

Indeed the obsdb code is pretty rigorous in referring to the obsdb columns as "fields", rather than "keys".  So let's keep doing that.

mhasself (changes requested):
This is totally the right direction for obsdb, thanks for figuring it out.

Obviously this needs tests and stuff :) so there will be a few rounds of iteration.  But I'm on-board with the concept.

You've hard-coded the extension to OWB, which makes sense for SO.  But it's a small step to _fully_ generalize it for more arbitrary index combos.  So I might push for that in next iteration.  (The "get_primary_keys" function shows that you can figure out the obs_key for a given database.)

msilvafe:
> This is totally the right direction for obsdb, thanks for figuring it out.
> 
> Obviously this needs tests and stuff :) so there will be a few rounds of iteration. But I'm on-board with the concept.
> 
> You've hard-coded the extension to OWB, which makes sense for SO. But it's a small step to _fully_ generalize it for more arbitrary index combos. So I might push for that in next iteration. (The "get_primary_keys" function shows that you can figure out the obs_key for a given database.)

Ok @mhasself I think I've addressed all of your comments, please have another look when you can. Thanks.

sotodlib/core/metadata/obsdb.py
```diff
@@ -93,42 +84,48 @@ def __init__(self, map_file=None, init_db=True, obs_wafer_band=False):
...
-           This is used whether to index by obs_id or obs-wafer-band."""
+           This is used whether to index by obs_id or
+           obs_id plus additional fields defined by wafer_info."""
         query = "PRAGMA table_info('obs')"
         c = self.conn.execute(query)
         primary_keys = [row['name'] for row in c.fetchall() if row['pk'] > 0]
         return primary_keys
     
-    def warn_primary_keys(self, wafer_band):
+    def warn_primary_keys(self, wafer_info):
```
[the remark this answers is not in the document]
msilvafe: @mhasself what structure are you thinking that users could pass in a dict for this wafer_info?
Like:
```
{`wafer_slot':'ws0', 'wafer.bandpass':'f150'}
```
And I would return `['wafer_slot', 'bandpass']` or something like: `{wafer_info:['wafer_slot', 'bandpass']}`? I guess I'm not really sure what the useful aspect of the values in the key-value pairs of a dict would be in this case.

sotodlib/core/metadata/obsdb.py
```diff
@@ -211,35 +208,38 @@ def update_obs(self, obs_id, data={}, tags=[],
...
-            wafer_band = self.warn_primary_keys(wafer_band)
-            obs_id = {'obs_id': obs_id, 'wafer_slot': wafer_band[0],
-                      'bandpass': wafer_band[1]}
-        else:
-            obs_id = {'obs_id': obs_id}
+        obs_key = {'obs_id': obs_id}
+        if (len(self.primary_keys) > 1):
+            wafer_info = self.warn_primary_keys(wafer_info)
+            for i, k in enumerate(self.primary_keys[1:]):
+                obs_key[k] = wafer_info[i]
```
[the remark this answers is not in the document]
msilvafe: I see this is also a similar comment about generalizing to a dict structure as you added under the `warn_primary_keys` function. Can you please write out an example of what the `obs_id` as str + `wafer_info` as dict, or `obs_id` as dict + `wafer_info` as list of str  or `obs_id` as dict and `wafer_info` as dict, etc. would actually look like?

sotodlib/core/metadata/obsdb.py
```diff
@@ -92,39 +92,63 @@ def __init__(self, map_file=None, init_db=True, wafer_info=None):
...
+            if sorted(pkeys) != sorted(primary_keys): # sorted allows for different order
+                raise ValueError(f"Primary keys do not match: {primary_keys} != {pkeys}"+
+                                f" must use `wafer_info`=={primary_keys} or create a new dB with {pkeys}")
         return primary_keys
     
-    def warn_primary_keys(self, wafer_info):
+    def _convert_wafer_info(self, obs_id, wafer_info):
+        """Helper function to allow flexibility in way obs_id and wafer_info are passed in."""
+        if isinstance(wafer_info, dict):
+            wafer_info = wafer_info.values()
```
mhasself, line 115: I think you need to grab these by key -- they might not be in the order you expect.

sotodlib/core/metadata/obsdb.py
```diff
@@ -92,39 +92,63 @@ def __init__(self, map_file=None, init_db=True, wafer_info=None):
...
+        if isinstance(obs_id, tuple):
+            if len(obs_id) == len(self.primary_keys):
+                wafer_info = tuple([wi for wi in obs_id[1:]])
+                obs_id = obs_id[0]
+            else:
+                raise ValueError(f"obs_id tuple must be of length {len(self.primary_keys)}")
+        if isinstance(obs_id, dict):
+            if len(obs_id) == len(self.primary_keys):
+                wafer_info = tuple([wi for ki, wi in obs_id.items() if ki != 'obs_id'])
```
mhasself, line 124: Same here -- you're relying on the order of .items().

sotodlib/core/metadata/obsdb.py
```diff
@@ -92,39 +92,63 @@ def __init__(self, map_file=None, init_db=True, wafer_info=None):
                     changes = True
             if changes:
                 self.conn.commit()
-        self.primary_keys = self.get_primary_fields()
+        self.primary_keys = self.get_primary_fields(wafer_info)
 
-    def get_primary_fields(self):
+    def get_primary_fields(self, wafer_info=None):
```
mhasself, line 97: I guess this should be a private function.  Seems like you cache `self.primary_keys` on init, so that's what users should access.

sotodlib/core/metadata/obsdb.py
```diff
@@ -204,13 +228,29 @@ def update_obs(self, obs_id, data={}, tags=[],
             tag name is prefxed with '!', then the tag name will be
             un-applied, i.e. cleared from this observation.
 
-        Returns:
-            self.
+        Example of ways to pass updates to obsdb when there are multiple primary keys.
+        1) obs_id as str and wafer_info as tuple::
```
mhasself, line 232: Add a blank line above 1) so it renders properly ([docs build](https://github.com/simonsobs/sotodlib/pull/1159)).

tests/test_obsdb.py
```diff
@@ -145,13 +145,32 @@ def test_owb_primary_linked_query(self):
...
         self.assertNotEqual(result_other['data1'], 999)
+        # Method 2
+        db.update_obs('myobs0', wafer_info={'wafer_slot':'ws3', 'bandpass':'f150'},
+                      data={'data1': 998})
+        result_m2 = db.get('myobs0', wafer_info={'wafer_slot':'ws3', 'bandpass':'f150'})
+        self.assertEqual(result_m2['data1'], 998)
+        # Method 3
+        db.update_obs({'obs_id':'myobs0', 'wafer_slot':'ws5', 'bandpass':'f090'},
+                      data={'data1': 997})
+        result_m3 = db.get({'obs_id':'myobs0', 'wafer_slot':'ws5', 'bandpass':'f090'})
```
mhasself, line 167: For at least one of these dict tests, pass the keys in the non-canonical order.

mhasself (changes requested):
Minor comments -- sorry for slow on this.

Issue #1158, opened 2025-03-25
apply_pointing_model rotates azimuth by about 2pi


ykyohei:
I find that apply_pointing_model rotates azimuth by about 2pi, this is fine mathematically, but this would leads to problem in azimuth synchronous signal estimation/filtering because the azimuth goes to unexpected range. The amount of correction should be small.
 
```
aman = ctx.get_obs(obs_id = 'obs_1730092656_satp3_1011111',dets={'wafer.type': 'OPTC', 'wafer_slot': 'ws0', 'wafer.bandpass': 'f090', })
print(aman.boresight.az)
coords.pointing_model.apply_pointing_model(aman)
print(aman.boresight.az)
```

```
[3.63942698 3.63942719 3.63942706 ... 3.64619045 3.64614672 3.64610314]
[-2.66229314 -2.66229293 -2.66229307 ... -2.65552968 -2.65557341
 -2.65561699]
```

elle-shaw:
@mhasself,  Kyohei pointed out this issue to me about the pointing model corrections and I wonder if it is similar to something I have seen before when using quat.decompose_lonlat. For my purposes, I knew that the correct answer would never be a negative Az value, so I did a simple check like:

``
    if -1*negaz/DEG < 0:
        newaz = -1*negaz/DEG + 360
    else:
        newaz = -1*negaz/DEG

``
But I'll look into this more.

mhasself:
I'm interested in talking about this more... there are two options here:
1.  Adjust the "branch" of the output az to minimize the difference to the input az.
2.  Ask that people doing things like template binning use the `.ancil.az_enc` (or whatever it is called) -- i.e. the uncorrected, full range, encoder value.

I lean towards (2), and here's why:
- A minor problem with (1) is that in some cases the az perturbation won't be small.  For example, if you have base tilt, and are near zenith, then az can be basically anything.  (This is unlikely to affect our CMB obs, admittedly ... but I like it when I don't have to worry about such possible singularities.)
- An advantage of (2) is that the template results you derive are tied to _the data_, rather than to some kind of corrected data.  That means one is free to modify the pointing model, later, and the az-sync result can be re-used (it was based on az_enc, and thus can be applied based on that same az_enc). 

I'm open to (1), just raising a couple thoughts :)

ykyohei:
(2) is good point, thanks for pointing out
I will close this since https://github.com/simonsobs/sotodlib/pull/1162 is merged. Thank you for your quick fix.

PR #1162, opened 2025-03-28
Pointing model quat.deconstruct_lonlat issue

elle-shaw:
I am making this PR in response to issue #1158 raised by Kyohei. I have added the following to the end of the pointing_model.model_sat_v1 function to add or subtract 2 pi from the azimuth values if any of them are off by over 2pi from what is expected. However if not all of the azimuth values have the same issue, perhaps this should be changed to a np.all() or make changes to individual values instead of all the tod. 
```
    neg_az, el, roll = quat.decompose_lonlat(q_hs)
    
    if np.any((-neg_az - az_orig) > 2*np.pi):
        new_az = -neg_az - 2*np.pi
    elif np.any((-neg_az - az_orig) < -2*np.pi):
        new_az = -neg_az + 2*np.pi
    else:
        new_az = -neg_az
    return new_az, el, roll
```

### DIFF
--- a/sotodlib/coords/pointing_model.py
+++ b/sotodlib/coords/pointing_model.py
@@ -196,13 +196,11 @@ def model_sat_v1(params, az, el, roll):
             * q_fp_offset * ~q_fp_rot * quat.euler(2, roll) * q_fp_rot)
 
     neg_az, el, roll = quat.decompose_lonlat(q_hs)
-    
-    if np.any((-neg_az - az_orig) > 2*np.pi):
-        new_az = -neg_az - 2*np.pi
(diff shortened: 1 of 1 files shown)


mhasself (changes requested):
I'd like to suggest the following:
```
# Make corrected az as close as possible to input az.
change = ((az - az_orig) + np.pi) % (2 * np.pi) - np.pi
az = az_orig + change
```

This is a thing I have typed about a billion times...

We'll have bigger troubles coming when we get to the LAT, since, e.g. (az, el) = (180, 100) could easily pop out of decompose_lonlat as (az, el) = (0, 80)!

Issue #1163, opened 2025-03-28
hkdb interface changes


mhasself:
I intend to adjust the `hkdb.load_hk` interface, so heads up and let me know if you're relying on these behaviors that I want to change...below...

1. Currently a field with no data will be returned to the user as an array with shape (0,).

This is a bug because all other returned things are 2d arrays.  **It would make more sense to return an array with shape (2, 0).**

2. Currently each field is returned as a shape (2, n) array, where the first row is the timestamps and the second row is the field.

This is problematic because when a field is string-like, or integer, then stuff gets converted in weird and annoying ways.  For example:
```
 array([['1717199990.4303348', '1717200050.6193132', '1717200110.772649',
        ..., '1717286448.0141983', '1717286508.2101457',
        '1717286568.3621724'],
       ['outlet7', 'outlet7', 'outlet7', ..., 'outlet7', 'outlet7',
        'outlet7']], dtype='<U32')
```
Note how the timestamps are strings :(

So to be general we should just **always return a tuple of two shape (n, ) arrays, rather than a (2, n) array.**

mhasself:
Implemented in #1164

PR #1164, opened 2025-03-28
hkdb updates -- note some API changes

mhasself:
- Each result item is a tuple of arrays, not a (2, n) array.
- Empty datasets are returned as a tuple of 0-len arrays.
- New args to load_hk for fine-tuning fields and times.
- New function get_feed_list, for inspecting db.
- "subdirs" option to make scanning hk archive subsets easier.
- Update docs for new function; add more links; more autodoc.

cf #1163 

"round2" of changes, coming later, will include a change in db schema that makes it possible to retrieve a field list without tooooo much G3 file reading.

### DIFF
--- a/docs/hkdb.rst
+++ b/docs/hkdb.rst
@@ -1,7 +1,8 @@
+.. py:module:: sotodlib.io.hkdb
 
-==============================
-Loading L3 Houskeeping Data
-==============================
+============================
+Loading L3 Housekeeping Data
+============================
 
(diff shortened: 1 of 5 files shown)


kmharrington:
For the recorded, data packaging does not use these files

mhasself:
Last chance to object @msilvafe  :)

msilvafe:
> Last chance to object @msilvafe :)

I'm happy to have interface changes since it comes along with all of these useful QOL improvements. Thanks!

PR #1168, opened 2025-03-31
fix bug SlowSource

yoshinori-0778:
This PR is for fixing a bug for `SlowSource` Class.
Ra/Dec used in this class is in radian, but there is a line that treats in degree units.
This does not affect much, though.


mhasself (approved):
Thanks!

PR #1170, opened 2025-04-03
Update executor call.

msilvafe:
Updates the executor to use @iparask 's new utility method that choses between MPI or multiprocessing pool executors to work on tiger. I copied what was done in [make_atomic_filterbin_map](https://github.com/simonsobs/sotodlib/blob/master/sotodlib/site_pipeline/make_atomic_filterbin_map.py) but haven't tested yet.

### DIFF
--- a/sotodlib/site_pipeline/multilayer_preprocess_tod.py
+++ b/sotodlib/site_pipeline/multilayer_preprocess_tod.py
@@ -2,12 +2,12 @@ import os
 import yaml
 import time
 import logging
+from typing import Optional, Union, Callable
 import numpy as np
 import argparse
 import traceback
 from typing import Optional
-import multiprocessing
-from concurrent.futures import ProcessPoolExecutor, as_completed
+from sotodlib.utils.procs_pool import get_exec_env
 import h5py
 import copy
 from sotodlib.coords import demod as demod_mm
@@ -271,7 +271,9 @@ def get_parser(parser=None):
     )
     return parser
 
-def main(configs_init: str,
+def main(executor: Union["MPIPoolExecutor", "ProcessPoolExecutor"],
+         as_completed_callable: Callable,
+         configs_init: str,
          configs_proc: str,
          query: Optional[str] = None, 
          obs_id: Optional[str] = None, 
@@ -291,7 +293,6 @@ def main(configs_init: str,
 
     errlog = os.path.join(os.path.dirname(configs_proc['archive']['index']),
                           'errlog.txt')
-    multiprocessing.set_start_method('spawn')
 
     obs_list = sp_util.get_obslist(context_proc, query=query, obs_id=obs_id, min_ctime=min_ctime,
                                    max_ctime=max_ctime, update_delay=update_delay, tags=tags,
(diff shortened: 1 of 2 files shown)


sotodlib/site_pipeline/multilayer_preprocess_tod.py
```diff
@@ -332,41 +333,43 @@ def main(configs_init: str,
...
+                logger.info(f'Processing future dependent result db_dataset: {db_datasets_proc}')
+                for db_dataset in db_datasets_proc:
+                    pp_util.cleanup_mandb(err, db_dataset, configs_proc, logger, overwrite)
+            else:
+                pp_util.cleanup_mandb(err, db_datasets_proc, configs_proc, logger, overwrite)
 
 if __name__ == '__main__':
-    sp_util.main_launcher(main, get_parser)
+    args = get_parser().parse_args()
+    rank, executor, as_completed_callable = get_exec_env(args.nprocs)
```
mmccrackan, line 373: Should be `arg.nproc`.

sotodlib/site_pipeline/preprocess_tod.py
```diff
@@ -305,7 +305,8 @@ def get_parser(parser=None):
     return parser
 
 
-def main(
+def main(executor: Union["MPIPoolExecutor", "ProcessPoolExecutor"],
```
mmccrackan, line 308: I think these should be: `Union["MPICommExecutor", "ProcessPoolExecutor"]`

mmccrackan (changes requested):
Some initial comments regarding things found while testing.

mmccrackan:
After merging in `master` which included #1188, running with 1 node with or without `srun` now calls `ProcessPoolExecutor` and running with `srun` and 2 nodes calls `MPIPoolExecutor`, so I think it is working as expected.

msilvafe:
> After merging in `master` which included #1188, running with 1 node with or without `srun` now calls `ProcessPoolExecutor` and running with `srun` and 2 nodes calls `MPIPoolExecutor`, so I think it is working as expected.

Ok, I've marked this as ready for review.

mmccrackan (approved):
Approving.

I tested both `preprocess_tod` and `multilayer_preprocess_tod` in a `slurm` script with both `srun -n 1` and `srun -n 2`, with the former using `ProcessPoolExecutor` and the latter using `MPICommExecutor` as expected.  Confirmed process archives, DBs, and output logs were written.

mmccrackan:
> This PR has unfortunately broken prefect. Prefect automatically wraps around each argument of `main` function and builds a custom form in the web interface to set them. This PR adds an executor and a callable as arguments, both are not supported by prefect. I've currently rolled back the sotodlib version on site computing to keep it working. We should design a better interface.

I'll try to come up with a fix for this.  Thanks for the heads up.

iparask:
> After merging in `master` which included #1188, running with 1 node with or without `srun` now calls `ProcessPoolExecutor` and running with `srun` and 2 nodes calls `MPIPoolExecutor`, so I think it is working as expected.

@mmccrackan, are you running with 1 core, or 1 node with multiple cores?

iparask:
> This PR has unfortunately broken prefect. Prefect automatically wraps around each argument of `main` function and builds a custom form in the web interface to set them. This PR adds an executor and a callable as arguments, both are not supported by prefect. I've currently rolled back the sotodlib version on site computing to keep it working. We should design a better interface.

@guanyilun, can you point me to the perfect script about this? The reason the executor is a callable is that we work around the MPI4py's context requirement for the executor.

guanyilun:
> @guanyilun, can you point me to the perfect script about this? The reason the executor is a callable is that we work around the MPI4py's context requirement for the executor.

This line [here](https://github.com/simonsobs/so-workflow/blob/e05ddf46c3bd0fe1b171fef2802e2bfdfa174f93/sowork/flows/__init__.py#L64-L65) automatically wraps around the `main` function in each site-pipeline module and transforms it into a `prefect` compatible "flow". You can think of prefect args requirement as an argparser -- it doesn't like callable, or a class. 

One idea is to make the `main` function only accept arguments from argparser and then rename the current function to something like `_main` or similar. This would allow auto-wrapping in prefect to work. 

An alternative method is to write a customized wrapper for this module. I've done something like this for the `make_atomic_filterbin_map`.

PR #1172, opened 2025-04-04
update_obsdb: exclude NONE dets (fixed tones) from detsets

mhasself:
This was sort of done in io.check_book, but that code segment wasn't called from update_obsdb / site_pipeline.check_book.  Now io.check_book never releases the names of the fixed tones.  The "remove_fixed_tones" work-around setting is removed, since we just always want to remove them.


AERAdler (approved):
Looks good! Is there an explanation somewhere of why we get fixed tones in the first place?

PR #1176, opened 2025-04-08
Iso sat review

chervias:


sotodlib/preprocess/preprocess_util.py
```diff
@@ -378,6 +383,8 @@ def multilayer_load_and_preprocess(obs_id, configs_init, configs_proc,
         the (large) TOD blob.  Not all loaders may support this.
     logger: PythonLogger
         Optional. Logger object or None will generate a new one.
+    init_only: bool
+        Optional. Whether or not to run the dependent pipeline.
```
mhasself, line 387: Unclear.  "If True, do not run the dependent pipeline."

sotodlib/preprocess/processes.py
```diff
@@ -819,18 +825,18 @@ class AzSS(_Preprocess):
         calc:
           signal: 'demodQ'
           azss_stats_name: 'azss_statsQ'
-          range: [-1.57079, 7.85398]
+          azrange: [-1.57079, 7.85398]
```
mhasself, line 828: Why isn't this called "az_range"?  I suspect it's "too late now" but please be careful in the future to be consistent in things like word separation.

sotodlib/preprocess/preprocess_util.py
```diff
@@ -429,6 +438,116 @@ def multilayer_load_and_preprocess(obs_id, configs_init, configs_proc,
             raise ValueError('Dependency check between configs failed.')
 
 
+def multilayer_load_and_preprocess_sim(obs_id, configs_init, configs_proc,
```
mhasself, line 441: It looks like the only two functions from this submodule included in readthedocs are:
```
autofunction:: sotodlib.preprocess.preprocess_util.load_preprocess_det_select
autofunction:: sotodlib.preprocess.preprocess_util.load_and_preprocess
```
Consider adding more, or auto-docing the whole submodule to reduce maintenance burden.

mhasself (commented):
Minor comments.

What is your plan for merging, in terms of the commit history?  Can you squash + rebase this, or will you need to keep one of these commits, for the permanent record?

sotodlib/preprocess/preprocess_util.py
chervias: @msilvafe @mmccrackan

mmccrackan: Sure we can definitely improve things on this front.  Maybe we should do this on a separate PR?

mmccrackan: Good point.  Will add more docstrings/auto-docing here!

mhasself:
> @mhasself now I always squash and merge, what would be a reason to not do that?

That's fine.  There had been talk in the past about preserving some version of the code that was used for ISO stuff.  Squash + merge is definitely ok with me.

chervias:
Ok, we have the code preserved in the branch `iso-sat-review-v1`

mhasself (approved):
Sorry for the delay.

Squash on merge!

Thanks.

PR #1180, opened 2025-04-09
More generic downsampling

skhrg:
I find myself using this function a lot but then making a copy with slight changes.
Here is a pass at making it more generic, docs and tests to come.

Note that this is a cleaned up version of something I had laying around so I may have broken it in that process,
will leave it as a draft until I get the time to test properly but feedback on if this is an okay way of doing it is still useful.

### DIFF
--- a/sotodlib/mapmaking/utils.py
+++ b/sotodlib/mapmaking/utils.py
@@ -605,7 +605,16 @@ def downsample_cut(cut, down):
     factor down."""
     return so3g.proj.ranges.RangesMatrix([downsample_ranges(r,down) for r in cut.ranges])
 
-def downsample_obs(obs, down, skip_signal=False):
+def get_wrappable(axman, key):
+    val = getattr(axman, key)
+    if isinstance(val, core.AxisManager):
+        return key,val
+    else:
+        axes   = axman._assignments[key]
+        axdesc = [(k,v) for k,v in enumerate(axes) if v is not None]
+        return key, val, axdesc
+
+def downsample_obs(obs, down, skip_signal=False, fft_resample=["signal"]):
     """Downsample AxisManager obs by the integer factor down.
 
     This implementation is quite specific and probably needs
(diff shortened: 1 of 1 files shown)


msilvafe:
I had opened this one quite awhile ago that I tried to make generic but we never pushed it across the finish line. Can you take a look to see if it's suitable or we should close that one in favor of this one? https://github.com/simonsobs/sotodlib/pull/474

skhrg:
Oh I remember seeing that PR a while ago and totally forgot about it. Long term I think something combining the two would be useful. I agree with Sigurd's comments on the downsampling which is why I have used the original version of the function many times. But that PR seems to have more robust cuts handling as well as upsampling which is nice. 

I think at the very least combining things so we get the better cuts handling from there would be very nice in the short term.

But also interested in what @mhasself  thinks since he is the grand architect of our lives.

msilvafe:
> Oh I remember seeing that PR a while ago and totally forgot about it. Long term I think something combining the two would be useful. I agree with Sigurd's comments on the downsampling which is why I have used the original version of the function many times. But that PR seems to have more robust cuts handling as well as upsampling which is nice. 
> 
> 
> 
> I think at the very least combining things so we get the better cuts handling from there would be very nice in the short term.
> 
> 
> 
> But also interested in what @mhasself  thinks since he is the grand architect of our lives.

My suggestion is to commit your version of a PR that is the best of both worlds and we review that

skhrg:
I am of the opinion that its in our best interest to just do multiple smaller PRs. This one extends functionality that people (or at least I) currently use.
We can bring over thing from the other PR in multiple smaller ones as we find time/need.

I'll try to test this tonight so its ready for review.

msilvafe:
> I am of the opinion that its in our best interest to just do multiple smaller PRs. This one extends functionality that people (or at least I) currently use. We can bring over thing from the other PR in multiple smaller ones as we find time/need.
> 
> I'll try to test this tonight so its ready for review.

Ya I’m generally in agreement that code that people are using is most useful to get in over code that’s been sitting in review for 2 years. But I’ll review with an eye for consistency with future plans to generalize with the older PR please assign me for review when you convert from draft!

skhrg:
Where does the time go...
Finally got around to testing this.

sotodlib/mapmaking/utils.py
```diff
@@ -616,57 +625,42 @@ def downsample_obs(obs, down, skip_signal=False):
...
-        new_key = key.split(".")[-1]
-        res.flags.wrap(new_key, downsample_cut(obs[key], down),
-                       [(0,"dets"),(1,"samps")])
+        elif key == "signal" and skip_signal:
+            continue
+        elif isinstance(obs[key], (core.AxisManager, core.FlagManager)):
+            res.wrap(key, downsample_obs(obs[key], down, skip_signal, fft_resample))
+        elif isinstance(obs[key], so3g.proj.ranges.RangesMatrix):
+            res.wrap(key, downsample_cut(obs[key], down))
+        elif key in "fft_resample":
```
mmccrackan, line 644: I think this should be `fft_resample` and not `"fft_resample"` right?

mmccrackan (approved):
Okay, talked with Max and we will do the preprocessing implementation (adding in the flag downsampling and related features from Max's PR, adding preprocess functions) in some standalone PRs after this one so as not to hold this up any longer.

PR #1182, opened 2025-04-10
Changes to handling of errors in `record_qa`

tristpinsm:
Changes requested in today's site pipeline call.

I haven't tested these, but they are pretty minimal. If the implementation looks ok we could test it out.

### DIFF
--- a/sotodlib/site_pipeline/record_qa.py
+++ b/sotodlib/site_pipeline/record_qa.py
@@ -19,7 +19,14 @@ def main(config):
 
     monitor = Monitor.from_configs(config["monitor"])
     context = core.Context(config["context_file"])
-    jdb = JobManager(sqlite_file=config["jobdb"])
+
+    # get JobDB configuration
+    jdb_max_retry = 5  # mark a job as failed after this number of tries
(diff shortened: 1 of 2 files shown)


sotodlib/site_pipeline/record_qa.py
```diff
@@ -98,9 +107,14 @@ def main(config):
                         f"Processing metric {m._influx_meas}.{m._influx_field} failed.",
                         exc_info=True
                     )
+                    n_fail += 1
```
mhasself, line 110: I think the idea would be that this would only count as a `n_fail += 1` (raised to prefect error level) if `visit_count > max_retry`

This is the right track -- thanks.

tristpinsm: right, that make sense ! thanks

mhasself:
Sounds like the right approach -- but needs testing; consider converting to draft unless that's imminent.

tristpinsm:
I will try to test in the next few days -- need to identify a set of failing obs_id

tristpinsm:
OK, I've tested and made a few tweaks.

guanyilun (approved):
I don't see anything wrong with this!

Issue #1174, opened 2025-04-07
make-hwp-solutions does not work with numpy=2


mhasself:
Problem seems to be use of np.where.  Usually you can do `some_array[condition]` rather than `some_array[np.where(condition)]`.

ykyohei:
This should be resolved by #1183

PR #1183, opened 2025-04-11
fix numpy2 compatibility of hwp-make-solutiuons

ykyohei:
This makes hwp-make-solutiuons numpy 2 compatible https://github.com/simonsobs/sotodlib/issues/1174
Tested by numpy 2.2.4


iparask:
@ykyohei can you also test it with numpy 2.1.6 as it is the one that exists in soconda?

ykyohei:
2.1.6 is not pip installable so I instead confirmed that this also works at 2.0.0 and 2.1.3.

mhasself (approved):
Thanks!

PR #1165, opened 2025-03-28
hkdb updates - round2

mhasself:
This is a modification to the database schema, so existing database will not be compatible until they are upgraded.  Get in touch with me if you're worried about this (I will work on getting the site dbs ready...).

- Adds a hash on the field names, for each frame.
- This makes it possible to do "get_field_list" fairly efficiently.
- The new-style database can still be consumed by older code

This PR will remain a draft until #1163 is through and I've prepped the site hkdbs.

PR #1186, opened 2025-04-15
hkdb changes -- includes schema update

mhasself:
Re-opening here, in favor of #1165.

Summary of changes:
- Adds a hash on the field names, for each frame.
  - This makes it possible to do "get_field_list" fairly efficiently.
  - The new-style database can still be consumed by older code
- Adds purge_unindexed files function.
- Declares index on a frames.file_id -- postgres doesn't do this for you.

I've updated our active databases to have this new column, so they'll just start supporting this function naturally when local code is updated.

In general though, the new code will fail when trying to query or update a database created by the prior code.  If you have a giant database you'd like to modify for use with the new code... there's sql for that, get in touch with me.  You don't need to start over.

The new schema lets you do this:
```
hkdb.get_field_list(lspec, ['timing-m1000-office-1.*.*']
>> ['timing-m1000-office-1.m1000.mbgLtNgPtpPortState_1', 'timing-m1000-office-1.m1000.mbgLtNgPtpPortState_1_description', ...]
```
and it's not even slow ...

### DIFF
--- a/sotodlib/io/hkdb.py
+++ b/sotodlib/io/hkdb.py
@@ -11,6 +11,7 @@ import time
 import numpy as np
 import pathlib
 import fnmatch
+import hashlib
 
 import sqlalchemy as db
 from sqlalchemy.ext.declarative import declarative_base
@@ -146,6 +147,8 @@ class HkFrame(Base):
         instance-id of the OCS agent for the frame
     feed: str
         Name of the OCS feed for the frame
+    fields_hash: str
+        Hash to identify the combination of fields included in this frame.
     byte_offset: int
         Offset of the frame in the G3 file in bytes
     start_time: float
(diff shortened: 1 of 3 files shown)


sotodlib/io/hkdb.py
```diff
@@ -292,12 +299,16 @@ def get_frames_from_file(
...
         _, agent, _, feed = addr.split('.')
         start_time, stop_time = 1<<32, 0
+        fields = []
         for block in frame['blocks']:
             ts = np.array(block.times) / spt3g_core.G3Units.s
+            fields.extend(block.keys())
             start_time = min(start_time, ts[0])
             stop_time = max(stop_time, ts[-1])
+
+        fields_hash = hashlib.sha256(','.join(sorted(fields)).encode('ascii')).hexdigest()[:16]
```
msilvafe, line 309: What's the idea with the `fields_hash` addition to the db?

mhasself: "Hash to identify the combination of fields included in this frame."

To determine what fields are available, we either have to list them all in the database (major expansion) or else read them from the data.  Reading from the data is very slow, since the list of field names can change, so you need to read every frame in the time range.  But agents actually do tend to have the same fields, for long stretches of time.  So the hash allows us to identify a small representative set of frames, where there are potentially a new set of field names.  It keeps the DB lean while also being pretty efficient in minimizing data archive reads to get the full field list.

msilvafe: Ahh makes sense, thanks for the explanation!

PR #1187, opened 2025-04-16
added traits for azimuth and elevation starting point of the drone

gabrielecoppi:
It is a quick modification of the drone operator to allow starting azimuth and elevation

### DIFF
--- a/sotodlib/toast/ops/sim_source.py
+++ b/sotodlib/toast/ops/sim_source.py
@@ -514,6 +514,10 @@ class SimSource(Operator):
         help="Initial distance of the artificial source in meters",
     )
 
+    source_azimuth_start = Quantity(
+        u.Quantity(0, u.degree), help="Start of the scan along the azimuth axis"
+    )
+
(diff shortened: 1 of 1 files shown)


sotodlib/toast/ops/sim_source.py
```diff
@@ -514,6 +514,10 @@ class SimSource(Operator):
         help="Initial distance of the artificial source in meters",
     )
 
+    source_azimuth_start = Quantity(
+        u.Quantity(0, u.degree), help="Start of the scan along the azimuth axis"
+    )
```
tskisner, line 519: This is technique (setting to zero and comparing to zero) is fine.  You can also set the trait to `None` as the default value and pass the `allow_none=True` option.  Then later in the code do `if self.source_azimuth_start is None: ...`.  But comparing a floating point number to zero should always work- just not comparisons to other values.

tskisner (approved):
I did not test this, but it looks fine.

PR #1191, opened 2025-04-22
TOAST intervals are now open-ended

keskitalo:
Update all instances of TOAST `Interval` to use the new open-ended definition. The unit tests will fail until a new tag of TOAST is uploaded.


keskitalo:
@tskisner this PR needs a new TOAST tag before it can pass the unit tests.

tskisner:
> @tskisner this PR needs a new TOAST tag before it can pass the unit tests.

Yes, our policy is that sotodlib master must always work with the latest toast tag, so we cannot merge this at the moment.

PR #1192, opened 2025-04-23
Fix noisy_subscan flags config dictionary copying

mmccrackan:
Fix for `noisy_subscan` flags.  In the loop `signal` is not a copy but a reference to `self.calc_cfgs`, so this ends up being modified when `subscan_stats` is overwritten.


msilvafe (approved):
Thanks for catching and fixing this. Sorry for the delay in review.

PR #1194, opened 2025-04-25
Parallelize make_book for operations books

kmharrington:
This update is starting a parallelization effort for book binding  with just the operation books (there's lots of them and they're small). 

* Makes a new helper function in imprinter to separate the database reads and database writes. Bookbinding involved a whole but of reads from the imprinter and g3tsmurf databases, but at the end we only write book.status and book.message. sqlite / sqlalchemy can handle multiple active sessions if only one active session is writing. 
* `imprinter.bind_book` still acts the same as it always has
* I tested this on ~3000 operation books on the LAT and the speed-up factor was basically `n_processes`. I also verified that the output books had `timestamps`, `signal` and `Z_smurf` files that were identical between the parallelly and serially written books.
  * I think 3000 should have also been enough to catch write conflicts if I missed something
* small bugfix for cleanup level 2

### DIFF
--- a/sotodlib/site_pipeline/make_book.py
+++ b/sotodlib/site_pipeline/make_book.py
@@ -1,13 +1,42 @@
 import os
 import traceback
 import argparse
+
 import datetime as dt
 from typing import Optional
-from sotodlib.io.imprinter import Imprinter
+from sotodlib.io.imprinter import Imprinter, Books, FAILED
 import sotodlib.io.imprinter_utils as utils
 
+import multiprocessing
+from concurrent.futures import ProcessPoolExecutor, as_completed
 
-def main(config: str):
+
+def bind_books_parallel(platform, book_list, n_proc ):
+    imprint = Imprinter.for_platform(platform)
+    session = imprint.get_session()
+    bid_list = [book.bid for book in book_list]
+    failed_list = []
+    with ProcessPoolExecutor(n_proc) as exe:
+        futures = [
+            exe.submit(_bookbinding_helper, platform, bid) for bid in bid_list
+        ]
+        for future in as_completed(futures):
+            bid, status, message, _  = future.result()
+            imprint.logger.info(f"Just finished book {bid}")
+            book = session.query(Books).filter(Books.bid == bid).one()
+            book.status = status
+            book.message = message
+            session.commit()
+            if status == FAILED:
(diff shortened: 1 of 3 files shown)


sotodlib/site_pipeline/make_book.py
```diff
@@ -1,13 +1,42 @@
...
+                failed_list.append( book.bid)
+    return failed_list
+
+def _bookbinding_helper(platform, bid ):
+    imprint = Imprinter.for_platform(platform)
+    return imprint._run_book_binding(bid)
+
 
-def main(config: str):
+def main(config: str, parallel_oper:bool=False):
```
mhasself, line 39: Instead of parallel_oper, just take n_proc?  Then if it it's more than 1, trigger parallel behavior.

sotodlib/site_pipeline/make_book.py
```diff
@@ -67,11 +104,14 @@ def get_parser(parser=None):
...
         help="Path to imprinter configuration file"
     )
-    parser.add_argument('output_root', type=str, help="Root path of the books")
+    #parser.add_argument('output_root', type=str, help="Root path of the books")
+    parser.add_argument("--parallel-oper", action="store_true")
     return parser
 
 
 if __name__ == "__main__":
+    multiprocessing.set_start_method('spawn')
```
mhasself, line 113: So this doesn't need to be in main()?

mhasself (approved):
Cool!  There are some formatting things you might want to clear up (e.g. the #parser.add_arg that got removed).

Parallelization looks great though.

sotodlib/site_pipeline/make_book.py
kmharrington: You're right, I probably do need to put it in `main()` I haven't tested this through prefect yet but I don't think this section gets executed by prefect.

iparask:
Why not use this: https://github.com/simonsobs/sotodlib/blob/master/sotodlib/utils/procs_pool.py#L66 instead of implementing the executor again?

If you need a process pool no matter what, then just ask for that.

kmharrington:
I hadn't realized there was more parallelization functions in sotodlib. I'm still hesitant to switch to it though because we've been trying to keep the data packaging code fairly independent from the rest of sotodlib. Less likely changes made for later analysis affect on-site operations.

PR #1196, opened 2025-04-28
hwp_angle solution; Add special fix for irig desynchronisation problem

ykyohei:
Add special fix for irig desynchronization problem. This is tested to reproduce corrected hwp_angle solution.
We specify the period and constant time offset in site-pipeline-configs/satpX/hwp_config.yaml.
Companion PR: https://github.com/simonsobs/site-pipeline-configs/pull/125
This correction was validated by hwpss angle study and Tau A map.

### DIFF
--- a/sotodlib/hwp/g3thwp.py
+++ b/sotodlib/hwp/g3thwp.py
@@ -90,6 +90,10 @@ class G3tHWP():
         self._encoder_disk_radius = self.configs.get(
             'encoder_disk_radius', 346.25)
 
+        # The time period and amount of irig desynchronization
+        # [ start_time, stop_time, amount of time shift ]
+        self._irig_desync = self.configs.get('irig_desync', None)
+
         # Output path + filename
         self._output = self.configs.get('output', None)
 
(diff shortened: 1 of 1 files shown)


sotodlib/hwp/g3thwp.py
```diff
@@ -1385,6 +1391,20 @@ def _irig_quality_check(self):
             self._irig_time = np.array([])
             self._rising_edge = np.array([])
 
+    def _fix_irig_desync(self):
+        """ Fix IRIG desynchronization by adding constant time offset """
+        if self._irig_desync is None:
+            return
+
+        if len(self._irig_time) < 1:
+            return
```
mhasself, line 1400: In an "and", Python won't evaluate second condition if first one is False.  So it's safe to:
```
if self._irig_desync is None or len(self._irig_time) == 0:
  return
```

But you really don't need the second check, since the loop over an empty list will do it for you.

mhasself (approved):
Minor comment but not a big deal.

mhasself:
We should wait for tests to pass ... probably need this first: #1191

PR #1197, opened 2025-04-28
preprocess_obs: filter obs list for LAT formatted obs_ids

davidvng:
Adds bool argument to `preprocess_obs.py` for LAT observations. Since LAT obs_ids are split by optics tubes with timestamps within about 10 seconds of each other, we only need to run `preprocess_obs` on one of them since it loads the entire aman with no signal. Without this, `preprocess_obs` runs redundant processes.

### DIFF
--- a/sotodlib/site_pipeline/preprocess_obs.py
+++ b/sotodlib/site_pipeline/preprocess_obs.py
@@ -17,7 +17,8 @@ def preprocess_obs(
     obs_id, 
     configs,  
     overwrite=False, 
-    logger=None
+    logger=None,
+    obs_group=None
 ):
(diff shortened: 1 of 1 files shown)


msilvafe:
Ok but this approach I think begs the question how do you connect your single obsid which is for a single lat tube to all of the 7 obsids associated with this? I think you actually need to add some sort of database or link file that maps between your single obsid and the 7 associated ones.

davidvng:
@msilvafe I added handling for obs grouping by adding the list of grouped obsids directly to the db. I tested this on the existing LAT planet obs so far and everything seems to work.

msilvafe (approved):
We talked about this a bit offline. Thanks.

PR #1204, opened 2025-05-01
Start making det match db time aware

mmccrackan:
Work in progress.  Attempt to make `det_match` dbs time aware in the same way as the `focal_plane` dbs.  Adds a start and end time which are used for a `range_match` in the db and a group for the h5 files.


mmccrackan:
Will close in favor of opening a new one after I merge #1627.

PR #1205, opened 2025-05-01
Mss3 tweaks

keskitalo:


sotodlib/toast/ops/splits.py
```diff
@@ -544,34 +592,57 @@ def _exec(self, data, detectors=None, **kwargs):
...
+                    # to the view name
+                    FlagIntervals(
+                        shared_flags=self.shared_flags,
+                        shared_flag_bytes=1,
+                        view_mask=[
+                            (f"~{spl.split_intervals}", np.uint8(self.shared_flag_mask))
+                        ],
+                        reset=True,
+                    ).apply(data)
+                    map_binner.shared_flag_mask |= self.shared_flag_mask
```
tskisner: Inside the `Split` base class, when a split is created a interval list is created and the per-detector flags are saved.  When the split is removed, that interval list is purged and the per-detector flags are restored.

If you are applying the `FlagIntervals` operator, that will modify the shared flags and leave it in a different state at the end.  I think in order to do this cleanly, you would need to save the shared_flag values, overwrite them here for each split, and then restore them when each split is removed.

tskisner (changes requested):
Thanks these useful features, especially to the Splits operator!  My main concern is the manipulation of the shared_flags (and restoring them) during each split.

sotodlib/toast/ops/splits.py
keskitalo: Good point. Just modifying a user-specified flag bit is not enough. I added saving and restoring of the shared flags. I also expanded on the existing unit test to confirm that intervals and flags produce the same output maps.

sotodlib/toast/workflows/proc_mapmaker.py
```diff
@@ -174,6 +174,11 @@ def mapmaker_select_noise_and_binner(job, otherargs, runargs, data):
 
         if job_tmpls.baselines.enabled:
             job_tmpls.noise_model = noise_model
+    elif job_ops.filterbin.enabled:
+        job_ops.filterbin.binning = job_ops.binner_final
+        job_ops.filterbin.binning.full_pointing |= otherargs.full_pointing
+        job_ops.filterbin.det_data = job_ops.sim_noise.det_data
+        job_ops.filterbin.output_dir = otherargs.out_dir
```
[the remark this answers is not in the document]
keskitalo: That is already communicated in the binner.

tskisner (approved):

PR #1206, opened 2025-05-01
Add loader and det cal functionality to load in bandpass info using bg

tdalford:
Adds custom functionality to be able to load in bandpass information using the `det_cal` bias group information. The pull request does a combination of:

1. Modify `update_det_cal.py` to calculate and save bandpass information inside the ManifestDB from the bias group.
2. Modify the PCARelCal module in `processes.py` to allow a fieldname for the bandpass as input rather than using the hard-coded `wafer.bandpass`
3. Add a custom metadata loader which allows input of which fields to load into a ResultSet, ignoring the rest. The loader also allows a user to modify the field names.

With the custom loader we can load up the `bandpass` field from the new `det_cal` into `aman.det_info` using e.g. a field name `dets:det_cal.bandpass`. The rest of the det cal information can be then loaded into `aman.det.cal` with a separate context entry using the same database.

### DIFF
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -1324,6 +1324,7 @@ class PCARelCal(_Preprocess):
     def __init__(self, step_cfgs):
         self.signal = step_cfgs.get('signal', 'signal')
         self.run = step_cfgs.get('pca_run', 'run1')
+        self.bandpass = step_cfgs.get('bandpass', 'wafer.bandpass')
         self.run_name = f'{self.signal}_{self.run}'
 
         super().__init__(step_cfgs)
@@ -1346,7 +1347,7 @@ class PCARelCal(_Preprocess):
             if self.plot_cfgs:
                 self.plot_signal = filt_aman[self.signal]
 
-        bands = np.unique(aman.det_info.wafer.bandpass)
+        bands = np.unique(aman.det_info[self.bandpass])
         bands = bands[bands != 'NC']
         # align samps w/ proc_aman to include samps restriction when loading back from db.
         rc_aman = core.AxisManager(proc_aman.dets, proc_aman.samps)
@@ -1354,7 +1355,7 @@ class PCARelCal(_Preprocess):
         relcal = np.zeros(aman.dets.count)
         pca_weight0 = np.zeros(aman.dets.count)
         for band in bands:
-            m0 = aman.det_info.wafer.bandpass == band
+            m0 = aman.det_info[self.bandpass] == band
             rc_aman.wrap(f'{band}_idx', m0, [(0, 'dets')])
             band_aman = aman.restrict('dets', aman.dets.vals[m0], in_place=False)
 
@@ -1408,7 +1409,7 @@ class PCARelCal(_Preprocess):
             det = aman.dets.vals[0]
             ufm = det.split('_')[2]
 
-            bands = np.unique(aman.det_info.wafer.bandpass)
+            bands = np.unique(aman.det_info[self.bandpass])
             bands = bands[bands != 'NC']
             for band in bands:
                 pca_aman = aman.restrict('dets', aman.dets.vals[proc_aman[self.run_name][f'{band}_idx']], in_place=False)
@@ -1900,9 +1901,9 @@ _Preprocess.register(SourceFlags)
 _Preprocess.register(HWPAngleModel)
 _Preprocess.register(GetStats)
 _Preprocess.register(UnionFlags)
-_Preprocess.register(RotateQU) 
-_Preprocess.register(SubtractQUCommonMode) 
-_Preprocess.register(FocalplaneNanFlags) 
-_Preprocess.register(PointingModel)  
+_Preprocess.register(RotateQU)
+_Preprocess.register(SubtractQUCommonMode)
+_Preprocess.register(FocalplaneNanFlags)
+_Preprocess.register(PointingModel)
 _Preprocess.register(BadSubscanFlags)
 _Preprocess.register(CorrectIIRParams)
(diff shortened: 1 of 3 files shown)


sotodlib/site_pipeline/update_det_cal.py
```diff
@@ -617,6 +620,28 @@ def find_correction_results(band, chan, dset):
...
+                highband_str = 'f280'
+
+            elif tube_flavor == 'lf':
+                lowband_str = 'f030'
+                highband_str = 'f040'
+
+            if cal.bg in bgs_lowband:
+                cal.bandpass = lowband_str
+            elif cal.bg in bgs_highband:
+                cal.bandpass = highband_str
```
msilvafe, line 643: How about something more compact like this? 
```
bgs = {'lb':[0, 1, 4, 5, 8, 9], 'hb':[2, 3, 6, 7, 10, 11]}
tube_flavor = am.obs_info.tube_flavor
band_str = {'mf':{'lb':'f090', 'hb':'f150'},
            'uhf':{'lb':'f220', 'hb':'f280'},
            'lf':{'lb':'f030', 'hb':'f040'}}
if cal.bg in bgs['lb']:
    cal.bandpass = band_str[tube_flavor]['lb']
elif cal.bg in bgs['hb']:
    cal.bandpass = band_str[tube_flavor]['hb']
```

And maybe `bgs` and `band_str` can be defined once/globally at the top of the script.

sotodlib/site_pipeline/update_det_cal.py
```diff
@@ -621,26 +627,10 @@ def find_correction_results(band, chan, dset):
...
-                highband_str = 'f040'
-
-            if cal.bg in bgs_lowband:
-                cal.bandpass = lowband_str
-            elif cal.bg in bgs_highband:
-                cal.bandpass = highband_str
+            if cal.bg in bgs['lb']:
+                cal.bandpass = band_str[tube_flavor]['lb']
+            elif cal.bg in bgs['hb']:
+                cal.bandpass = band_str[tube_flavor]['hb']
```
msilvafe, line 633: If referencing globals I think this needs to be capitalized variable names here.

tdalford: Yeah sorry, missed this earlier

kmharrington:
I think I like this setup but will leave it to @mhasself to approve the additional loader option. 

To deploy this we'll need to update the existing det cal databases correct? I believe these are the instructions for setting up to deploy the update. https://simonsobs.atlassian.net/wiki/spaces/PRO/pages/306249767/Site+Pipeline+Metadata+Logs

mhasself:
Sorry for the delay.  I'll review this soon but one thing to do immediately is add tests for `ModifyFieldsResultSetHdfLoader`.  I think the current place where such things is exercised is `tests/test_context.py`.  Adding a basic test of the loader features in there would be appreciated.

tdalford:
> Is there any reason you can think of (and I'll have a go, too) to not include this functionality directly in ResultSetHdfLoader? Seems backwards compatible.

My reasons were a combination of:

1. The documentation mentioned that 'load_fields' was only used for the AxisManagerHDF loader
2. If I add the functionality to change the fieldnames using load_fields then I think I would have to modify the AxisManagerHDF load_fields behavior to do this also, which I could, but would be more changes.

But these aren't that good.

kmharrington:
@mhasself  Tommy has added tests, could we get a review here?

tests/test_context.py
```diff
@@ -311,6 +311,19 @@ def test_120_load_fields(self):
         self.assertCountEqual(tod.ondisk._fields.keys(), ['disk1', 'subaman'])
         self.assertCountEqual(tod.ondisk.subaman._fields.keys(), ['disk2'])
 
+    def test_130_load_fields(self):
```
mhasself, line 314: "load_fields_resultset" would distinguish this from test_120 more clearly.

mhasself (changes requested):
Focusing on the metadata changes -- I think it would be good to just add this functionality directly to the base class.  This is because the "load_fields" is already implemented on AxisManager loader, with similar semantics (which we can make more similar, in the future) -- and you've taken advantage of the fact that "load_fields", in metadata definitions, is propagated through to the loader.  Which is great.

Once that is done there are a couple of documentation fixes -- i.e. remove the "this only works on AxisManager" warning; and mention instead that the ResultSet loader also lets you rename fields.  The syntax should be included somehow in readthedocs -- I should be able to reconstruct it starting from [this autodocced class](https://sotodlib.readthedocs.io/en/latest/context.html#sotodlib.core.metadata.MetadataSpec).

Thanks for the tests.

sotodlib/preprocess/processes.py
```diff
@@ -1343,6 +1343,7 @@ class PCARelCal(_Preprocess):
     def __init__(self, step_cfgs):
         self.signal = step_cfgs.get('signal', 'signal')
         self.run = step_cfgs.get('pca_run', 'run1')
+        self.bandpass = step_cfgs.get('bandpass', 'wafer.bandpass')
```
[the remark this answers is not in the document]
tdalford: Are you saying to use 'self.bandpass_key' as the class variable name or to use 'bandpass_key' as the config parameter name? Or both?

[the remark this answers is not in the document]
tdalford: I'm going to guess you mean the config parameter name, that is much more visible

mhasself (approved):
Ok, I think this makes sufficient sense.  Sorry for the delay!

I have not looked at the det_cal changes in detail, but I understand that's being looked at carefully in situ.

Please just address my minor comments, then squash on merge.

PR #1212, opened 2025-05-09
Update per-band array handling for relcal

mmccrackan:
The PCA relcal function wrapped in arrays that included the bandpasses in their names, which breaks when trying to load in multiple bandpasses from an existing preprocessing database.  Those arrays were only used in the PCA plotting, so this branch changes things such that they are only made when plotting is enabled and are deleted once that band's plots have been made.  While it might be useful to have these values available, it is currently more or less impossible to store per-band information (unless they are per-detector indexed) in the preprocessing axis manager and be able to load it back due to restrictions from the axis manager loading/concatenation functions.  This requires changes at a deeper level in the axis manager classes.

Made a slight change to the plotting which was broken.  Since `aman` is not trimmed after the lowpass filter, I added a trim to the restricted copy of `aman` in the plotting function if lowpassing.

Tested by preprocessing two band independently and running `get_meta(obsid)` with both plotting disabled and enabled.

Confirmation that these parameters are not used elsewhere other than plotting:
https://github.com/search?q=repo%3Asimonsobs%2Fsotodlib+pca_mode0&type=code
https://github.com/search?q=repo%3Asimonsobs%2Fsotodlib+xbounds&type=code
https://github.com/search?q=repo%3Asimonsobs%2Fsotodlib+ybounds&type=code
https://github.com/search?q=repo%3Asimonsobs%2Fsotodlib+%7Bband%7D_idx&type=code

### DIFF
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -1362,7 +1362,8 @@ class PCARelCal(_Preprocess):
         pca_weight0 = np.zeros(aman.dets.count)
         for band in bands:
             m0 = aman.det_info.wafer.bandpass == band
-            rc_aman.wrap(f'{band}_idx', m0, [(0, 'dets')])
+            if self.plot_cfgs is not None:
+                rc_aman.wrap(f'{band}_idx', m0, [(0, 'dets')])
             band_aman = aman.restrict('dets', aman.dets.vals[m0], in_place=False)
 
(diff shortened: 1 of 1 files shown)


msilvafe (approved):
Looks good, thanks for the fix Michael. I think the issue is not in the AxisManager class but instead the HDF5 loader the combines manifestdb's into the AxisManager. Perhaps @mhasself could weigh in on how we may extend those in the future but I think it probably makes sense to move those conversations to a separate issue.

PR #1213, opened 2025-05-09
add intersect_flags

yoshinori-0778:
This PR adds new function of intersect flags. This function is useful for planet mapmaking because we want to do preprocess outside of planet mask. Intersect flags from inversed source_flags and something can be used for this situation.
I checked that intersect flags is as expected
Screenshot 2025-05-09 at 4 35 36 PM

### DIFF
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -1,6 +1,7 @@
 import numpy as np
 from operator import attrgetter
 import copy
+import warnings
 
 from so3g.proj import Ranges, RangesMatrix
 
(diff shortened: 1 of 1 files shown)


msilvafe:
I suggest instead we generalize "UnionFlags" to "CombineFlags" or something similar and it can take as an input argument `+, *, union, intersect` to perform either union or intersect (or maybe a list of +/* matching the length of the list of flags passed in to identify which should be unions and which intersected).

yoshinori-0778:
Thank you for your comment. I remove intersect_flags and add combine_flags as you suggested.
I allow list type as input. Is this what you expected?

PR #1216, opened 2025-05-12
LAT pointing model and ffp updates

skhrg:


sotodlib/coords/pointing_model.py
```diff
@@ -104,19 +110,86 @@ def apply_pointing_model(tod, pointing_model=None, ancil=None,
...
+        "az_offset": -6.431949463522369e-07,
+        "el_offset": 0.10784213916209494,
+        "cr_offset": 0.1068009204473848,
+        "el_xi_offset": 0.07524197484976551,
+        "el_eta_offset": 0.13704409641376075,
+        "rx_xi_offset": -0.03523515856319279,
+        "rx_eta_offset": -0.005858421274865713,
+        "mir_xi_offset": -0.17792288910750848,
+        "mir_eta_offset": 0.21440894087003562,
+    }
```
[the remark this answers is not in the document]
skhrg: Fixed

sotodlib/coords/pointing_model.py
```diff
@@ -115,8 +120,75 @@ def apply_pointing_model_sat(vers, params, tod, ancil):
...
+            continue
+        params[key] = np.deg2rad(val)
+    q_enc = quat.rotation_lonlat(
+        -1 * (az.copy() + params["az_offset"]), el.copy() + params["el_offset"]
+    )
+    q_mir = quat.rotation_xieta(params["mir_xi_offset"], params["mir_eta_offset"])
+    q_el_roll = quat.euler(2, el.copy() + params["el_offset"] - np.deg2rad(60))
+    q_tel = quat.rotation_xieta(params["el_xi_offset"], params["el_eta_offset"])
+    q_cr_roll = quat.euler(2, -1 * cr - params["cr_offset"])
+    q_rx = quat.rotation_xieta(params["rx_xi_offset"], params["rx_eta_offset"])
```
mhasself, line 182: So with regards to `q_mir, q_tel, q_rx`.  The coordinates that go into these are all offsets, in a sense.  But not in the same way as encoder offsets are... so I might drop that word in favor of some other descriptive token.

But more importantly we should decide whether we want to apply them as `rotation_xieta(xi, eta)` or `rotation_xieta(-xi, -eta)`.

Consider the combo `q_cr_roll * q_rx`.  If `q_rx = rotation_xieta(+xi, +eta)`, then you're saying "where (xi, eta) represents a shift applied to the focal plane, prior to rotating by the corotator angle."  If you use `(-xi, -eta)`, then you can say "(xi, eta) represents the position in the focal plane which that remains fixed under corotation."

It's equivalent, mathematically.  But the signs of the parameters change.  So it's really a choice of how you want to be able to describe this, in words, in code comments and supporting writeups, going forward.  I _think_ it's nice to be able to say that second thing -- "xi, eta is a position in the focal plane where _some thing happens_".  This is the rationale for the docstring re-wording above.  But there is probably a reasonable opposite view.

(In the current SAT parametrization I think these might be mixed ... fp_rot seems to the "position of" convention but fp_offset uses the "shift applied" convention.  Something to consider for v2...)

sotodlib/coords/pointing_model.py
```diff
@@ -115,8 +120,75 @@ def apply_pointing_model_sat(vers, params, tod, ancil):
...
+#
+def model_lat_v1(params, az, el, roll):
+    """Applies pointing model to (az, el, roll).
+
+    Args:
+      params: AxisManager (or dict) of pointing parameters.
+      az, el, roll: naive horizon coordinates, in radians, of the
+        boresight.
+
+    The implemented model parameters are:
```
mhasself, line 144: On naming of the parameters ... here are some things that I like about the SAT model naming:
- (xi, eta) coordinate pairs are of the form `something_{xi,eta}0`.  The 0 indicates it's a single number, rather than a vector.  The xi, eta last means you can easily just say `something`, to mean the pair / the whole rotation.
- Encoder offsets are called "enc_offset_{axis}".

I'm on board with "cr" for the corotator.

I left some comments in the model about how the various (xi, eta) pairs should be described, and whether or not that means you apply (xi, eta) or (-xi, -eta) in the model... it bears a bit of thought, in case we ever want to describe this carefully to someone.  The specific problem I have with the current param descriptions is that "the offset between A and B" is ambiguous to me ... it doesn't have a specific direction attached to it (whereas "the position of A relative to B" does).

mhasself (changes requested):
Basically I'd like to converge this a bit better with the SAT model and params ... and take an opportunity to decide how we want to think about those (xi, eta) "offsets".

sotodlib/coords/pointing_model.py
```diff
@@ -141,52 +145,82 @@ def model_lat_v1(params, az, el, roll):
...
-    q_rx = quat.rotation_xieta(params["rx_xi_offset"], params["rx_eta_offset"])
-    new_az, el, roll = (
-        quat.decompose_lonlat(q_enc * q_mir * q_el_roll * q_tel * q_cr_roll * q_rx)
-        * np.array([-1, 1, 1])[..., None]
+
+    # Elevation component of roll motion
+    q_el_roll = quat.euler(2, el - np.deg2rad(60))
+    
+    # Rotation that takes a vector in telescope's elevation-hub centered 
+    # coordinates to mirror-centered coordinates
```
mhasself, line 197: I think this should be "takes a vector in telescope's corotator-centered coordinates to el-hub-centered coordinates"

mhasself (changes requested):
Exciting times!  Just minor comments.

sotodlib/coords/pointing_model.py
```diff
@@ -141,52 +145,82 @@ def model_lat_v1(params, az, el, roll):
...
+    The implemented model parameters are all in Radians:
+    - enc_offset_{az, el, cr}: Encoder offsets in Radians.
+      Sign convention: True = Encoder + Offset
+    - cr_center_{xi,eta}0: The (xi,eta) coordinate in the LATR-centered 
+      focal plane that remains fixed under corotation. 
+    - el_axis_center_{xi,eta}0: The (xi,eta) coordinate in the CR-centered
+      focal plane that appears fixed when the elevation structure is rotated 
+      about its axis.
+    - mir_center_{xi,eta}0: The (xi,eta) coordinate in the El-structure-centered
+      focal plane about which any mirror misalignment rotates.
```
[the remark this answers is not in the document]
skhrg: Let me know if the latest update makes more sense. Will try to make diagrams for all these things soon, but image I attach you to M2 and rigidly tilt the M1+M2 system as a unit, you will see one point stay fixed.

In reality its more complicated than that, M1 and M2 tilt independently about their internal origins but that adds up to an effective offset of the system on sky which can be reproduced by a joint tilt about some effective origin.

sotodlib/coords/pointing_model.py
```diff
@@ -145,16 +145,17 @@ def model_lat_v1(params, az, el, roll):
...
       Sign convention: True = Encoder + Offset
     - cr_center_{xi,eta}0: The (xi,eta) coordinate in the LATR-centered 
       focal plane that remains fixed under corotation. 
     - el_axis_center_{xi,eta}0: The (xi,eta) coordinate in the CR-centered
       focal plane that appears fixed when the elevation structure is rotated 
       about its axis.
     - mir_center_{xi,eta}0: The (xi,eta) coordinate in the El-structure-centered
-      focal plane about which any mirror misalignment rotates.
+      focal plane that appears fixed when the mirrors are rotated about the ray from
+      sky that kits the center of both mirrors.
```
mhasself, line 158: hits

mhasself (approved):
Ok!  Squash on merge.

PR #1217, opened 2025-05-12
fix typo in get_horiz

davidvng:


msilvafe (approved):
Looks like this doesn't get used too often because it looks like this was broken ~2 months ago by Sigurd in PR #1114

PR #1218, opened 2025-05-12
Update glitches to allow unique names.

msilvafe:
Allow glitches to be saved into fields with names other than `glitch_flags`


mmccrackan (approved):
Looks good.

PR #1220, opened 2025-05-13
Add runtime error check in save_group_and_cleanup

mmccrackan:
Previously this function was set up to only find corrupted files that didn't completely write out, but an existing db entry raises a RuntimeError, so this branch adds a catch for those.  Probably will improve `cleanup_mandb` to identify existing db entries in a future update (maybe along with the features of #1156).

### DIFF
--- a/sotodlib/preprocess/preprocess_util.py
+++ b/sotodlib/preprocess/preprocess_util.py
@@ -733,7 +733,12 @@ def save_group_and_cleanup(obs_id, configs, context=None, subdir='temp',
                 if not remove and g not in db_groups:
                     cleanup_mandb(None, outputs_grp, configs, logger)
                 else:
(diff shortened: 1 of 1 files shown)


mmccrackan:
Changed this so that instead of just catching a RuntimeError it will open the db and check what groups are inside of it.  If they are already found, it will delete the temporary file.  This way we will still be alerted to other errors we may not have encountered yet.  Will upgrade cleanup_mandb to better handle overwritting in a separate PR.

msilvafe (approved):
Ok this makes sense. We hopefully won't run into this case often but since we did run into it recently when 2 instances of preprocess_tod were running over top of each other at site its good that we protect ourselves against this case. I would like us to log if we are removing temp files and why (i.e. because remove = True or g in db_groups)

PR #1221, opened 2025-05-13
feat: add more flexibility when inserting default iir params

skhrg:


msilvafe (approved):
I'm ok with these changes although I would prefer that in addition to adding these optional arguments you add the known time range for the lat to the time_range dictionary. I also think a simple extension of this is to allow the time_range dictionary to contain a list of tuples of start/end times and loop over that list to allow for multiple epochs. The latter you don't need to implement but I would like you to implement the former.

skhrg:
Yeah having some known time ranges would be good. I don't currently have them recorded in a convenient way. @mmccrackan you have a list of bad obs right? Do you have ranges you recommend?

mmccrackan:
> Yeah having some known time ranges would be good. I don't currently have them recorded in a convenient way. @mmccrackan you have a list of bad obs right? Do you have ranges you recommend?

I only know it started after 04/25 and no Mars observations were missing them before that.  It is not only Mars observations, but I also saw some rcw38 that are missing them as well.  I will run a script to get a complete list of the ones with missing values.

mmccrackan:
[20250513_iir_stats.txt](https://github.com/user-attachments/files/20196368/20250513_iir_stats.txt)

Here's the full list of `obs_ids`, wafers, and ufms for the query:

`'type="obs" and start_time > 1745452800'
`
With 1745452800 being April 24th.  It looks like the range is between 1745970733 and 1746624934.

msilvafe:
> [20250513_iir_stats.txt](https://github.com/user-attachments/files/20196368/20250513_iir_stats.txt)
> 
> Here's the full list of `obs_ids`, wafers, and ufms for the query.
> 
> `'type="obs" and start_time > 1745452800' ` With 1745452800 being April 24th. It looks like the range is between 1745970733 and 1746624934.

Confirm that the end date is around when I slacked @tpsatt about hard hammering to reset these.

skhrg:
Okay @msilvafe I think I got those changes in. Let me know if it looks okay to merge.

ykyohei (commented):
lat time_ranges has typo. Other points look good to me.

skhrg:
Good catch! Fixed

ykyohei (approved):

PR #1222, opened 2025-05-13
Fix AzSS subtract logic.

msilvafe:
In the current version of AzSS there's different logic for the `process` method of the `AzSS` preprocess class depending if this is the first time the pipeline is being run or a subsequent time (i.e. if `proc_aman` is being loaded back from the archive or being created for the first time in the `calc_and_save` method). Currently if `process_cfgs` is set to `subtract = True` it will only subtract if the `proc_aman` is loaded from the archive (i.e. the 2nd+ time the pipeline is run) since its checking for `self.calc_cfgs.get('azss_stats_name') in proc_aman` which is not added to `proc_aman` until the `calc_and_save` step which is run after the `process` step in the `Pipeline.run()` function. Here I've updated the logic so that the subtraction happens regardless of if its the first time or subsequent time. 

I suspect this impacts the ISO-v1 run since this is in the 2nd layer it was run on-the-fly with the mapmaker so unless the mapmaker was run twice for certain obsids reusing the existing preprocessing "proc" archive the subtraction step was not being called. Tagging @susannaaz, @ykyohei, and @chervias to see this.

### DIFF
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -889,21 +889,28 @@ class AzSS(_Preprocess):
             proc_aman.wrap(self.calc_cfgs["azss_stats_name"], azss_stats)
 
     def process(self, aman, proc_aman, sim=False):
-        if self.calc_cfgs.get('azss_stats_name') in proc_aman and self.process_cfgs["subtract"]:
-            if sim:
-                tod_ops.azss.get_azss(aman, **self.calc_cfgs)
+        if self.process_cfgs["subtract"]:
+            if self.calc_cfgs.get('azss_stats_name') in proc_aman:
+                if sim:
+                    tod_ops.azss.get_azss(aman, **self.calc_cfgs)
+                else:
+                    tod_ops.azss.subtract_azss(
+                        aman,
+                        proc_aman.get(self.calc_cfgs.get('azss_stats_name')),
+                        signal=self.calc_cfgs.get('signal', 'signal'),
+                        scan_flags=self.calc_cfgs.get('scan_flags'),
+                        method=self.calc_cfgs.get('method', 'interpolate'),
+                        max_mode=self.calc_cfgs.get('max_mode'),
+                        azrange=self.calc_cfgs.get('azrange'),
+                        in_place=True
+                    )
             else:
-                tod_ops.azss.subtract_azss(
-                    aman,
-                    proc_aman.get(self.calc_cfgs.get('azss_stats_name')),
-                    signal=self.calc_cfgs.get('signal', 'signal'),
-                    scan_flags=self.calc_cfgs.get('scan_flags'),
-                    method=self.calc_cfgs.get('method', 'interpolate'),
-                    max_mode=self.calc_cfgs.get('max_mode'),
-                    azrange=self.calc_cfgs.get('azrange'),
-                    in_place=True
-                )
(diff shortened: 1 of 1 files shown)


chervias:
So the mapmaking was not run twice I think, because if it ran once the maps were created and written into the atomic DB, then it would never run again since it would skip subsequent runs. 
On the other hand, we have maps from the before times when we didn't have azss subtraction, like this Q/U plot here, and the large scales look quite different from what we have in ISO. So I don't know what's going on.

satp1_20240903_f150_P_0
satp1_20240903_f150_P_1

sotodlib/preprocess/processes.py
```diff
@@ -889,21 +889,28 @@ def save(self, proc_aman, azss_stats):
...
-                    signal=self.calc_cfgs.get('signal', 'signal'),
-                    scan_flags=self.calc_cfgs.get('scan_flags'),
-                    method=self.calc_cfgs.get('method', 'interpolate'),
-                    max_mode=self.calc_cfgs.get('max_mode'),
-                    azrange=self.calc_cfgs.get('azrange'),
-                    in_place=True
-                )
+                cfgs = dict(copy.deepcopy(self.calc_cfgs))
+                cfgs.pop('subtract_in_place', None)
+                tod_ops.azss.get_azss(aman, subtract_in_place=True, **cfgs)
```
ykyohei, line 910: Why do we not do `tod_ops.azss.get_azss(aman, **cfgs)` here instead?

ykyohei (commented):
Looks mostly good to me but I have one question.

sotodlib/preprocess/processes.py
msilvafe: Because we in this segment of the logic we want to run subtract_in_place=True so if `subtract_in_place` is present in calc_cfgs I'm deleting it and then directly passing subtract_in_place=True to the get_azss fn

susannaaz:
I think this might have been introduced with the modifications on the azss operation between the ISO version [here](https://simonsobs.slack.com/archives/C08SA15JMA6/p1747168572140329) we worked on initially and the latest implementation by @ykyohei, and maybe we messed up when we merged the two into master. The maps we ran for ISO used the previous version, and I think I did a run before and after adding the azss (ISO version) and it was visibly removing some of the synchronous signal in the maps. Also in the previous version we had a loophole such that we had the calc_and_save step before preproc precisely for the reasons you mention, so at least ISO maps should be safe, but thanks for finding this Max

sotodlib/preprocess/processes.py
ykyohei: I'm confused.. We should have subtract_in_place=True is present in cal_clfgs if we want to subtract azss, and passing **self.calc_cfgs will do same thing.

Are you saying that subtract_in_place=False should be overwritten to True if process.subtract True because it's redundunt?

msilvafe: Yes, perhaps I should just have the check at the beginning if `process.subtract` and `calc_cfgs['subtract_in_place']` conflict and just raise a ValueError there?

ykyohei: Okay, I was not aware of genral rules of proprocess. 
Then I think it's better to change the subtract_in_place part of docstring and change example config to omit subtract_in_place from calc_config and specify subtraction by process config.

ykyohei (approved):
Looks good to me. Thanks for the changes.

PR #1228, opened 2025-05-16
Add a db check to save_group_and_cleanup

mmccrackan:
Adds a check for if the preprocessing manifest db exists in save_group_and_cleanup to make sure it works when no database exists.


msilvafe (approved):

chervias (approved):
Just tested on the mapmaker and it is working.

PR #1229, opened 2025-05-21
Common mode glitch

ykyohei:
Modified get_glitch_flags to allow 1d array input and added preprocess for getting common mode. Also added unit test.

The motivation is to find and remove the glitch which is common among detectors. We have physical motivation for this kind of glitches. We can also do this on hwp demodulated timestreams to find the glitch induced by hwp angle.

Example preprocess config for common mode glitch.
```
    - name: "get_common_mode"
      calc:
        signal: "signal"
        method: "median"
        wrap: "signal_commonmode"
      save: True

    - name: "glitches"
      skip_on_sim: False
      glitch_name: "glitches_commonmode"
      calc:
        signal_name: 'signal_commonmode'
        t_glitch: 0.007
        buffer: 100
        hp_fc: 6.0
        n_sig: 10
      save: True
```

### DIFF
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -1496,6 +1496,35 @@ class PCAFilter(_Preprocess):
         model = tod_ops.pca.get_pca_model(aman, signal=signal, n_modes=n_modes)
         _ = tod_ops.pca.add_model(aman, model, signal=signal, scale=-1)
 
+class GetCommonMode(_Preprocess):
+    """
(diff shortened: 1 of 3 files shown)


msilvafe (approved):
This looks good to me what testing have you done other than the unit tests?

ykyohei:
Thanks. Another test I did was to make sure the example preprocess config runs as expected.
I found minor typo, so let me fix it and merge this.

PR #1230, opened 2025-05-22
feat: add dataclass for storing rfft object, allow fourier filter to …

skhrg:
…take in a rfft object, cache superior/inferion ints,some minor error handling adjustments

### DIFF
--- a/sotodlib/tod_ops/fft_ops.py
+++ b/sotodlib/tod_ops/fft_ops.py
@@ -1,5 +1,9 @@
 """FFTs and related operations
 """
+from dataclasses import dataclass, field
+from functools import lru_cache, partial
+from typing_extensions import Callable
+from numpy.typing import NDArray
 import sys
 import numdifftools as ndt
 import numpy as np
@@ -59,24 +63,21 @@ def rfft(
 
         freqs: the frequencies it is value at (since resizing is an option)
     """
-    if len(aman._assignments[signal_name]) > 2:
-        raise ValueError("rfft only works for 1D or 2D data streams")
 
     axis = getattr(aman, axis_name)
 
     if len(aman._assignments[signal_name]) == 1:
         n_det = 1
-        main_idx = 0
         other_idx = None
-
     elif len(aman._assignments[signal_name]) == 2:
         checks = np.array(
             [x == axis_name for x in aman._assignments[signal_name]], dtype="bool"
         )
-        main_idx = np.where(checks)[0][0]
         other_idx = np.where(~checks)[0][0]
         other_axis = getattr(aman, aman._assignments[signal_name][other_idx])
         n_det = other_axis.count
+    else:
+        raise ValueError("rfft only works for 1D or 2D data streams")
 
     if detrend is None:
         signal = np.atleast_2d(getattr(aman, signal_name))
@@ -101,16 +102,16 @@ def rfft(
     else:
         raise ValueError('resize must be "zero_pad", "trim", or None')
 
-    a, b, t_fun = build_rfft_object(n_det, n, "FFTW_FORWARD")
+    rfft = RFFTObj.for_shape(n_det, n, "FFTW_FORWARD")
     if resize == "zero_pad":
-        a[:, : axis.count] = signal
-        a[:, axis.count :] = 0
+        rfft.a[:, : axis.count] = signal
+        rfft.a[:, axis.count :] = 0
     elif resize == "trim":
-        a[:] = signal[:, :n]
+        rfft.a[:] = signal[:, :n]
     else:
-        a[:] = signal[:]
+        rfft.a[:] = signal[:]
 
-    t_fun()
+    rfft.t_forward()
 
     if delta_t is None:
         if "timestamps" in aman:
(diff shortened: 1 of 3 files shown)


sotodlib/tod_ops/fft_ops.py
```diff
@@ -120,9 +121,60 @@ def rfft(
...
+    _t_backward: Optional[Callable]
+
+    def validate(self, shape):
+        if shape != (self.n_det, self.n):
+            raise ValueError("Data is wrong shape for the rfft object")
+        if self._t_forward is None and self._t_backward is None:
+            raise ValueError("Neither t_forward nor t_forward specified!")
+
+    @property
+    def t_forward(self):
```
mhasself, line 168: Alternately you could have a default implementation of the t_* that just raises an error (i.e. slightly more informative than attribute not found). Then no need for private members.

sotodlib/tod_ops/fft_ops.py
```diff
@@ -141,33 +193,35 @@ def build_rfft_object(n_det, n, direction="FFTW_FORWARD", **kwargs):
...
 
-    return a, b, t_fun
+    rfft_obj = RFFTObj(n_det, n, a, b, t_forward, t_backward)
+
+    return rfft_obj 
 
 
-def find_inferior_integer(target, primes=[2, 3, 5, 7, 11, 13]):
+@lru_cache
+def find_inferior_integer(target, primes=(2, 3, 5, 7, 11, 13)):
```
mhasself, line 224: Can you cache the outermost calls only?  It probably doesn't matter much, but this function is recursive and the value of caching all the intermediate calculations is much less important.

That would entail creating private _find_inferior_integer, which is not cache-decorated.

sotodlib/tod_ops/fft_ops.py
```diff
@@ -120,9 +121,60 @@ def rfft(
     freqs = np.fft.rfftfreq(n, delta_t)
 
     if other_idx is not None and other_idx != 0:
-        return b.transpose(), freqs
+        return rfft.b.transpose(), freqs
+
+    return rfft.b, freqs
+
+
+@dataclass
```
mhasself, line 129: No doubt you're doing this to get the handy constructor.  dataclass isn't supposed to be used for objects carrying custom methods ... but I've certainly done that.

Since this object is only constructed from the `build_rfft_object` anyway, that code could become the constructor.  There would be considerable charm in simply `rfft = RFFTObj(tod.signal.shape)` or `rfft = RFFTObj.for_tod(tod)`

skhrg: > dataclass isn't supposed to be used for objects carrying custom methods

Not sure if I agree with that given that the first example in the docs does that https://docs.python.org/3/library/dataclasses.html

I've always interpreted the point of dataclasses (beyond the handy constructor) to be that the instance is totally defined by its attributes rather than some other inputs that may not be as directly related (which is why I like them so much, the abstraction between the constructor inputs and the instance is one of those object oriented things I never liked much). But that is a stylistic/philosophical issue so I am willing the relent...

Would something like:

```
@dataclass
class RFFTObject:
...

@classmethod
def for_shape(cls, n_det, n, ...):
    ...

@classmethod
def for_tod(cls, tod, ...):
    return cls.for_shape(*tod.shape, ...)
```

sotodlib/tod_ops/fft_ops.py
skhrg: Sure

sotodlib/tod_ops/fft_ops.py
mhasself: Great points ... I don't know what I was reading that gave me that impression about dataclass...

Yes your proposal is acceptable.

sotodlib/tod_ops/fft_ops.py
```diff
@@ -120,54 +121,109 @@ def rfft(
...
+        t_backward = partial(_t_null, "backward") 
+        if direction == "FFTW_FORWARD":
+            t_forward = pyfftw.FFTW(a, b, direction=direction, **fftargs)
+        elif direction == "FFTW_BACKWARD":
+            t_backward = pyfftw.FFTW(b, a, direction=direction, **fftargs)
+        elif direction == "BOTH":
+            t_forward = pyfftw.FFTW(a, b, direction="FFTW_FORWARD", **fftargs)
+            t_backward = pyfftw.FFTW(b, a, direction="FFTW_BACKWARD", **fftargs)
+        else:
+            raise ValueError("direction must be FFTW_FORWARD or FFTW_BACKWARD")
```
mhasself, line 197: "or BOTH"

sotodlib/tod_ops/fft_ops.py
```diff
@@ -180,13 +236,14 @@ def find_inferior_integer(target, primes=[2, 3, 5, 7, 11, 13]):
     while n > 0:
         n -= 1
         base = p**n
-        best_friend = find_inferior_integer(target / base, primes[1:])
+        best_friend = getattr(find_inferior_integer, "__wrapped__", find_inferior_integer)(target / base, primes[1:])
```
mhasself, line 239: Why not just `find_inferior_integer.__wrapped__`?

mhasself (changes requested):
Two minor things.  Thanks!

sotodlib/tod_ops/fft_ops.py
skhrg: Paranoia that someone may want to disable the cache by doing

`find_inferior_integer = find_inferior_integer.__wrapped__`

in a script.

sotodlib/tod_ops/fft_ops.py
skhrg: Fixec

PR #1231, opened 2025-05-23
fix: convert iir pars to array

skhrg:
Realized that the original iir pars are an array. This mostly doesn't matter but there are some edge cases where it is nice to keep that consistent (if you want to wrap this into something for example).


ykyohei (approved):
Looks good to me.

PR #1232, opened 2025-05-23
Adding pointing "wobble" correction

susannaaz:


msilvafe:
I would think this would be a `process` instead of a `calc_and_save` and it would just modify the boresight or per-detector pointing similar to how apply_pointing_model works. Storing the pointing n_dets x n_samps for az, el, roll triples the data size I think. Also I think pointing stuff falls under @elle-shaw's domain so she should have an eye on this implementation even if its still a WIP at the moment.

mhasself (changes requested):
Great to see this important feature entering the code!  Some comments on this:

1.  Wobble would be much easier to implement as part of the pointing model, if the parameters were the same for all wafers and bands.  That's not the case, I guess.  So agreed that this will not be simply added into the "pointing_model", as we currently use it.
2. As Max said, we can't be storing multiple TOD-sized things in the preprocessing archive.  But also, there is a tremendous amount of redundant information in the data structures that are being used here.  Specifically the complexity of the parameters is [wafer, band] and the corrected sightlines is [wafer, band, time].  But the parameters are provided as [det], and the sightlines as [det, time].  So that's 2000 times more numbers than we need; and for the sightlines, this is a huge RAM inflation.
3.  Presumably the modified sightlines are used for pointing operations, at some point.  Like, when you're making a map of some (wafer, band).  Wouldn't that be a good moment to apply the wobble correction, just for that limited set of detectors?  I.e. that is the time when you're interested in asking "what single set of wobble params apply to these dets I'm working on?", then "what's the single modified sightline, given these wobble params?" and then "great, give a Pmat using this sightline".

So I think you should do this:
1. Store wobble parameters as a metadata result, and load it through context metadata.  It can be an AxisManager with structure like: `wobble_params.{wafer_slot}.{band}.[amp, phase]`.
2. Remove the preprocessing stuff.
3. Add a function to coords/helpers.py, or a similar place in coords, that extracts the correct wobble_params for a set of detectors and returns the sightline that you need for pmat construction.  (This can do the validation that you only have dets from a single wafer/band.)

susannaaz:
Thanks @msilvafe @mhasself @elle-shaw! Sorry for being slow on this, I've been travelling and only now managed to spend some time looking at this. This still needs quite a bit more testing, so for now, this is mostly a note to myself as I continue working on it between trips:

For now I've updated the branch following @mhasself comments by

1. removing what I added in `preprocess` and `pointing_model`
2. adding to coords/helpers (https://github.com/simonsobs/sotodlib/blob/5b2c3844db4fd5f91f5ca3808eda42a89f7621e1/sotodlib/coords/helpers.py#L706-L767)
3. modifying the [pmap in demod_mapmaker](https://github.com/simonsobs/sotodlib/blob/5b2c3844db4fd5f91f5ca3808eda42a89f7621e1/sotodlib/mapmaking/demod_mapmaker.py#L312-L325) to be able to ingest the modified sightline (right now this is set by default, this probably needs to be more flexible, maybe with a True/False argument somewhere in [here](https://github.com/simonsobs/sotodlib/blob/sa_dev/wobble_deflection/sotodlib/site_pipeline/make_atomic_filterbin_map.py) where we specify if the modified sightline is needed) (relevant to @chervias) 

One note: I attempted storing the metadata for amplitude/phase in a single database (to load it directly as `wobble_params.{wafer_slot}.{band}.[amp, phase]` which would be cleaner) but went back to using separate metadata databases because storing both amp and phase together for the same (wafer_slot, band) key causes a `Duplicate entries for combined unique key error` (but maybe there's a smarter way of making the metadata than what I've done) -- checking with @msilvafe ++ who have more experience in building metadata.

The metadata and updated context files are available in `/scratch/gpfs/SIMONSOBS/users/sa5705/wobble/20250603/`
if people want to play with this -- any inputs/comments/suggestions very welcome

susannaaz:
Stored wobble parameters (amp/phase) within a single metadata as a ResultSet with structure `wobble_params.{wafer_slot}.{band}.[amp, phase]` that can be loaded with context metadata.

Followed logic described in https://github.com/simonsobs/sotodlib/pull/1232#issuecomment-2937905057:
- removed what I added in preprocess and pointing_model
- added function `get_deflected_sightline` to coords/helpers.py which extracts the wobble_params for a set of detectors in a given sat/wafer/band and returns the sightline needed for pmat
- included in the [pmap in demod_mapmaker](https://github.com/simonsobs/sotodlib/blob/5b2c3844db4fd5f91f5ca3808eda42a89f7621e1/sotodlib/mapmaking/demod_mapmaker.py#L312-L325) option `pdeflect` to ingest the modified sightline 


TODO:
- [ ] right now `pdeflect` is set by default: this probably needs to be more flexible, maybe with a True/False argument somewhere in [here](https://github.com/simonsobs/sotodlib/blob/sa_dev/wobble_deflection/sotodlib/site_pipeline/make_atomic_filterbin_map.py))

sotodlib/coords/pointing_model.py
```diff
@@ -4,7 +4,7 @@
 import numpy as np
 from .. import core
 from .helpers import _valid_arg
-from so3g.proj import quat
+from so3g.proj import quat, CelestialSightLine
```
chervias, line 7: `CelestialSightLine` is not used anywhere in the file, why do you need to import it?

mhasself (changes requested):
Thanks -- this is looking pretty good so I'm mostly complaining about docstrings...

sotodlib/coords/helpers.py
```diff
@@ -702,3 +702,62 @@ def to_g3(self):
...
+
+    Parameters
+    ----------
+    aman : AxisManager
+        AxisManager for the observation, must include hwp_angle, timestamps, 
+        and boresight.az/el, as well as det_info with wafer and band info.
+
+    wobble_meta : AxisManager
+        Metadata tree containing both amp and phase fields under
+        wobble_meta.{wafer_slot}.{band}.{amp, phase}
```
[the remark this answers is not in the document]
chervias: yes this is the first approach that Susanna tried, but I don't think the metadata works like that, you just have amp and phase directly available at wobble_meta. I'll change it

tests/test_demod_mapmaker.py
```diff
@@ -45,7 +45,7 @@ def make_map(obs, nside=None, nside_tile=None, shape=None, wcs=None, comps='TQU'
 
     signals    = [signal_map]
     mapmaker   = DemodMapmaker(signals, noise_model=NmatWhite(), comps=comps)
-    mapmaker.add_obs('obs0', obs, use_psd=False)
+    mapmaker.add_obs('obs0', obs, use_psd=False, apply_wobble=False)
```
mhasself, line 48: You've modified this test in order to suppress wobble correction. 

Do you have a test that _does_ do wobble correction?  Seems like that would be good to have.

mhasself (changes requested):
Thanks for all the doc fixes!

Please add a test of the wobble correction.

PR #1233, opened 2025-05-27
Also trim on proc_aman in fft_trim

mmccrackan:
The `fft_trim` function only trimmed the axis on `aman` and not `proc_aman`, leading to a miss matched samples error later in the processing stages.


chervias (approved):
Works fine.

PR #1234, opened 2025-05-27
ML mapmaker I/O

chervias:
This implements the preprocessing module for loading obs into the ml mapmaker, which requires to run preprocessing previously. The idea is to run with LAT data and leave the absolute minimum required preprocessing in this code, and we will remove that eventually


sotodlib/site_pipeline/make_ml_map.py
```diff
@@ -94,13 +111,27 @@ class DataMissing(Exception): pass
         sys.exit(1)
     L.info("Found %d tods" % (len(sub_ids)))
 
+    # Define our task distribution
+    obsinfo = mapmaking.get_obsinfo_subids(sub_ids, context)
+    owner   = mapmaking.distribute_tods_ra(obsinfo, comm.size) # assumes site=so
```
iparask: Since there is a `site` parameter, why does this take it as input?

sotodlib/site_pipeline/make_ml_map.py
```diff
@@ -282,12 +353,13 @@ class DataMissing(Exception): pass
...
-            except (DataMissing,IndexError,ValueError) as e:
+            except (DataMissing,IndexError,ValueError,LoaderError,TypeError) as e:
                 L.debug("Skipped %s (%s)" % (sub_id, str(e)))
+                L.debug("Datacount: %s full" % (sub_id))
                 continue
 
-        nkept = comm.allreduce(nkept)
-        if nkept == 0:
+        nkept_all = np.array(comm.allgather(nkept))
+        if np.sum(nkept_all) == 0:
```
iparask: Why not `comm.Allreduce(nkept, nkept_all, op=MPI.SUM)` here?

sotodlib/mapmaking/utils.py
```diff
@@ -484,10 +502,16 @@ def get_pos(name, ctime, sys=None):
...
                 return np.array([0,0]) # would use geom here
             else:
-                obj = getattr(ephem, name)()
-                djd = ctime/86400 + 40587.0 + 2400000.5 - 2415020
-                obj.compute(djd)
-                return np.array([obj.a_ra, obj.a_dec])
+                try:
+                    planet = coords.planets.SlowSource.for_named_source(
+                        name, ctime)
+                    ra0, dec0 = planet.pos(tod.timestamps.mean())
```
iparask: `tod` is undefined. This will always fail

sotodlib/site_pipeline/make_ml_map.py
```diff
@@ -25,34 +29,46 @@ def get_parser(parser=None):
...
-    from sotodlib import tod_ops, mapmaking, core
-    from sotodlib.tod_ops import filters
+    import sys, time, warnings, os, so3g, yaml
+
+    from sotodlib import mapmaking, core
+    from sotodlib.coords import sidelobes
     from sotodlib.mapmaking import log
+    from sotodlib.preprocess import preprocess_util as pp_util
+    from sotodlib.site_pipeline.utils import depth1_utils as d1u
+    from sotodlib.site_pipeline.utils.config import _get_config
```
iparask: `yaml`, `log` and `fft` are not used. Can you remove them?

sotodlib/site_pipeline/make_ml_map.py
```diff
@@ -94,13 +111,27 @@ class DataMissing(Exception): pass
...
     if args.srcsamp:
         srcsamp_mask  = enmap.read_map(args.srcsamp)
 
+    # set up the preprocessing
+    try:
+        preproc = _get_config(args.preprocess_config)
+    except:
+        if comm.rank==0:
+            L.info(f"{args.preprocess_config} is not a valid config")
+        sys.exit(1)
```
iparask: Why not raise an exception here instead of a `sys.exit`?

sotodlib/site_pipeline/make_ml_map.py
```diff
@@ -144,29 +175,37 @@ class DataMissing(Exception): pass
...
+        sidelobe_cutters = {}
 
         nkept = 0
+        to_skip_all = comm.allreduce(to_skip)
+
         # TODO: Fix the task distribution. The current one doesn't care which mpi
         # task gets which tods, which sabotages the pixel-saving effects of tiled maps!
         # To be able to distribute the tods sensibly, we need a rough estimate of where
         # on the sky each tod is. We should be able to get this using the central
         # ctime, az and el for each tod.
```
iparask: Isn't this done with redistributing the tods

sotodlib/site_pipeline/make_ml_map.py
```diff
@@ -144,29 +175,37 @@ class DataMissing(Exception): pass
...
             signal_srcsamp = mapmaking.SignalSrcsamp(comm, srcsamp_mask, dtype=dtype_tod)
             # This *must* come after all the other signals due to how it zeros out the
             # affected samples in the tod (inherited from SignalCut). Might have been better
             # to factorize out that zeroing into its own thing that's easier to control.
             signals.append(signal_srcsamp)
         mapmaker   = mapmaking.MLMapmaker(signals, noise_model=noise_model, dtype=dtype_tod, verbose=verbose>0)
+        sidelobe_cutters = {}
 
         nkept = 0
+        to_skip_all = comm.allreduce(to_skip)
```
iparask: This is very counter-intuitive for me. Reducing a variable that is set later in the code.

Is the purpose for that we do not go though the same tods in the second pass?

sotodlib/site_pipeline/make_ml_map.py
```diff
@@ -179,8 +218,28 @@ class DataMissing(Exception): pass
...
+                    #obs = context.get_obs(sub_id, meta=meta)
+                    obs = pp_util.load_and_preprocess(obs_id, preproc, context=context, meta=meta)
+                if obs.dets.count < 50:
+                    L.debug("Skipped %s (Not enough detectors)" % (sub_id))
+                    L.debug("Datacount: %s full" % (sub_id))
+                    to_skip += [sub_id]
+                    continue
+                # Check nans
+                mask = np.logical_not(np.isfinite(obs.signal))
+                if mask.sum() > 0:
```
iparask: Why not something like:
```python
if not np.all(np.isfinite(obs.signal))
```

instead? 

The name is also a bit confusing. It is a check if there are any NaNs and if you want to keep it the way it is, rename the variable to something like `nan_exist`.

sotodlib/mapmaking/utils.py
```diff
@@ -322,6 +323,18 @@ def get_subids_file(fname, context=None):
...
+    """Given a list of subids, return a ResultSet with one entry
+    from the obsdb for each subid. Ideally one would have this
+    from the original query that defined the subids, but since
+    subids are often read in directly from a file, we can't rely
+    on that. This function is a bit inefficient, since it starts
+    from the full obsdb and then looks up the subids from it"""
+    obsinfo = context.obsdb.query("1")
+    obsids  = split_subids(subids)[0]
+    inds    = putils.find(obsinfo["obs_id"], obsids)
+    return obsinfo.subset(rows=inds)
```
iparask: Why not do something like:
```
  ids_str = ",".join(f"'{subid}'" for subid in subids)                                       
  rs = context.obsdb.query("obs_id IN (%s)" % ids_str)
```
?

sotodlib/mapmaking/utils.py
```diff
@@ -380,14 +393,19 @@ def expand_ids(obs_ids, context=None, bands=None):
...
     if wafers is not None:
         wafs   = astr_tok(subids,":",1)
         subids = subids[np.isin(wafs, wafers)]
     if bands is not None:
         bpass  = astr_tok(subids,":",2)
         subids = subids[np.isin(bpass, bands)]
+    if ots is not None:
+        # Somewhat hacky implementation
+        obs_ids = astr_tok(subids, ":",0)
+        has_ot = np.prod([np.char.find(obs_ids, ot) for ot in ots], 0)
```
iparask, line 416: What is the intent here?

sotodlib/mapmaking/utils.py
```diff
@@ -569,8 +593,11 @@ def rangemat_sum(rangemat):
         res[i] = np.sum(ra[:,1]-ra[:,0])
     return res
 
-def find_usable_detectors(obs, maxcut=0.1, glitch_flags: str = "flags.glitch_flags"):
-    ncut  = rangemat_sum(obs[glitch_flags])
+def find_usable_detectors(obs, maxcut=0.1, glitch_flags: str = "flags.glitch_flags", to_null : str = "flags.expected_flags"):
+    flag = obs[glitch_flags]
+    if to_null != "" and to_null in obs._fields:
```
iparask:
```suggestion
    if to_null and to_null in obs:
```

should be enough.

sotodlib/mapmaking/utils.py
```diff
@@ -749,6 +776,34 @@ def setup_passes(downsample="1", maxiter="500", interpol="nearest", npass=None):
         passes.append(entry)
     return passes
 
+def distribute_tods_ra(obsinfo, nsplit, site="so", weather="typical"):
```
iparask, line 789: How would this work if `site="so_lat"` instead?

sotodlib/mapmaking/utils.py
```diff
@@ -749,6 +776,34 @@ def setup_passes(downsample="1", maxiter="500", interpol="nearest", npass=None):
...
+    weight= obsinfo["n_samples"][order]
+    cum   = putils.cumsum(weight)
+    wtot  = cum[-1]+weight[-1] # = np.sum(weight)
+    owner = np.zeros(len(obsinfo),int)
+    owner[order] = putils.floor(cum*nsplit/wtot)
+    # Make sure none ended up empty. If they did, fall back to unweighted,
+    # which will always succeed
+    if np.any(np.bincount(owner, minlength=nsplit) == 0):
+        owner[order] = np.arange(len(obsinfo))*nsplit/len(obsinfo)
+    return owner
```
iparask: These are `owners`, right? It returns how TODs are in all splits, not a single one.

sotodlib/mapmaking/utils.py
```diff
@@ -749,6 +776,34 @@ def setup_passes(downsample="1", maxiter="500", interpol="nearest", npass=None):
         passes.append(entry)
     return passes
 
+def distribute_tods_ra(obsinfo, nsplit, site="so", weather="typical"):
+    """For each row in obsinfo, assign it to one of nsplit categories
+    such that each category is as compact in RA as possible, and each
+    category has as similar size as possible. Assumes that all obs
+    are at the same site."""
+    if len(obsinfo) <= nsplit:
+        return np.arange(nsplit)[:len(obsinfo)]
```
iparask, line 795: I would prefer to define `owners` up here as
```
owners = np.arange(nsplit)[:len(obsinfo)]
```
and then always return `owners`

sotodlib/mapmaking/utils.py
```diff
@@ -749,6 +776,34 @@ def setup_passes(downsample="1", maxiter="500", interpol="nearest", npass=None):
         passes.append(entry)
     return passes
 
+def distribute_tods_ra(obsinfo, nsplit, site="so", weather="typical"):
+    """For each row in obsinfo, assign it to one of nsplit categories
+    such that each category is as compact in RA as possible, and each
+    category has as similar size as possible. Assumes that all obs
+    are at the same site."""
+    if len(obsinfo) <= nsplit:
```
iparask, line 794: Now that I think about it, I would do something like:
```
owners = np.arange(nsplit)[:len(obsinfo)]
if nsplit > len(obsinfo):
...
return owners
```

it is much cleaner and removes this weird `return`

sotodlib/mapmaking/utils.py
```diff
@@ -749,6 +776,34 @@ def setup_passes(downsample="1", maxiter="500", interpol="nearest", npass=None):
...
+    ctime = (obsinfo["start_time"] + obsinfo["stop_time"])/2
+    ra    = so3g.proj.CelestialSightLine.az_el(ctime, az, el, site=site, weather=weather).coords()[:,0]
+    # Put everything on the same copy of the sky
+    ra    = putils.rewind_compact(ra)
+    # Partition sorted ras equally by weight. Would ideally include
+    # ndet too, but the effective ndet depends on the cuts etc.
+    order = np.argsort(ra)
+    weight= obsinfo["n_samples"][order]
+    cum   = putils.cumsum(weight)
+    wtot  = cum[-1]+weight[-1] # = np.sum(weight)
```
iparask: Why not just `np.sum(weight)`?

sotodlib/mapmaking/utils.py
```diff
@@ -830,3 +885,27 @@ def atomic_db_aux(atomic_db, info: list[AtomicInfo]):
             session.commit()
         except exc.IntegrityError:
             session.rollback()
+
+
+def prune_mpi(comm, ranks_to_keep):
```
iparask: What is the purpose of this function? It is not used anywhere.

sotodlib/site_pipeline/make_ml_map.py
```diff
@@ -25,34 +29,46 @@ def get_parser(parser=None):
...
     parser.add_argument(      "--interpol",   type=str, default="nearest", help="Pmat interpol per pass. ,-sep")
     parser.add_argument("-T", "--tiled"  ,   type=int, default=1, help="0: untiled maps. Nonzero: tiled maps")
     parser.add_argument(      "--srcsamp",   type=str, default=None, help="path to mask file where True regions indicate where bright object mitigation should be applied. Mask is in equatorial coordinates. Not tiled, so should be low-res to not waste memory.")
     parser.add_argument(      "--unit",      type=str, default="uK", help="Unit of the maps")
+    parser.add_argument(      "--maxcut", type=float, default=.3, help="Maximum fraction of cut samples in a detector.")
+    parser.add_argument(      "--no-sidelobe", action="store_true", help="Do not mask Moon/Sun sidelobes")
+    parser.add_argument(      "--sun-mask", type=str, default="/global/cfs/cdirs/sobs/users/sigurdkn/masks/sidelobe/sun.fits", help="Location of Sun sidelobe mask")
+    parser.add_argument(      "--moon-mask", type=str, default="/global/cfs/cdirs/sobs/users/sigurdkn/masks/sidelobe/moon.fits", help="Location of Moon sidelobe mask")
```
iparask, line 43: If I run this anywhere apart from NERSC it will fail. If these inputs are required, remove the defaults and make them `require=True`

sotodlib/site_pipeline/make_ml_map.py
```diff
@@ -25,34 +29,46 @@ def get_parser(parser=None):
...
-    class DataMissing(Exception): pass
+    LoaderError = d1u.LoaderError
+    DataMissing = d1u.DataMissing
 
     warnings.simplefilter('ignore')
     args    = bunch.Bunch(**args)
     SITE    = args.site.lower()
     verbose = args.verbose - args.quiet
     comm    = mpi.COMM_WORLD
+    comm_size = comm.size
```
iparask: Not used, remove

sotodlib/site_pipeline/make_ml_map.py
```diff
@@ -6,9 +8,11 @@ def get_parser(parser=None):
     parser.add_argument("query")
     parser.add_argument("area")
     parser.add_argument("odir")
+    parser.add_argument("preprocess_config", help="Preprocess configuration file")
     parser.add_argument("prefix", nargs="?")
     parser.add_argument(      "--comps",   type=str, default="TQU",help="List of components to solve for. T, QU or TQU, but only TQU is consistent with the actual data")
     parser.add_argument("-W", "--wafers",  type=str, default=None, help="Detector wafer subsets to map with. ,-sep")
+    parser.add_argument("-O", "--ots",  type=str, default=None, help="Optics tubes to map with. ,-sep")
```
iparask, line 16: `-sep` is nowhere to be found. How can I set the separator?

Also why is it not `nargs=+` to accept multiple args?

sotodlib/site_pipeline/make_ml_map.py
iparask: It is also a bit dangerous. We may need an auxiliary function here to make everything return, like:

```python
def exit(comm: MPI.Comm, code: int)->None:
    if comm is not None: # This check might need an expansion
        comm.Abort(code)
    else:
        sys.exit(code)
```

and change all `sys.exit`s with this one.

sotodlib/mapmaking/utils.py
iparask: This function is a perfect candidate for a unit test.

sotodlib/mapmaking/utils.py
```diff
@@ -569,8 +593,11 @@ def rangemat_sum(rangemat):
         res[i] = np.sum(ra[:,1]-ra[:,0])
     return res
 
-def find_usable_detectors(obs, maxcut=0.1, glitch_flags: str = "flags.glitch_flags"):
-    ncut  = rangemat_sum(obs[glitch_flags])
+def find_usable_detectors(obs, maxcut=0.1, glitch_flags: str = "flags.glitch_flags", to_null : str = "flags.expected_flags"):
```
iparask, line 605: Can you add a test?

sotodlib/mapmaking/utils.py
```diff
@@ -380,14 +393,19 @@ def expand_ids(obs_ids, context=None, bands=None):
                         sub_ids.append("%s:ws%d:%s" % (obs_id, si, band))
     return sub_ids
 
-def filter_subids(subids, wafers=None, bands=None):
+def filter_subids(subids, wafers=None, bands=None, ots=None):
```
iparask, line 405: How is this tested?

sotodlib/mapmaking/utils.py
```diff
@@ -322,6 +323,18 @@ def get_subids_file(fname, context=None):
     sub_ids = expand_ids(sub_ids, context=context)
     return sub_ids
 
+def get_obsinfo_subids(subids, context):
```
iparask, line 329: Can we have a test?

sotodlib/site_pipeline/make_ml_map.py
chervias: yes, the PR by Sigurd a couple of weeks ago fixes this

sotodlib/site_pipeline/make_ml_map.py
chervias: yes, @skhrg added this so that in the second pass you don't try to load obs you already know failed in the first pass.

sotodlib/mapmaking/utils.py
chervias: yes I'll remove it. Saianeesh add it at some point where we had a bug but there is no need now.

sotodlib/site_pipeline/make_ml_map.py
chervias: @skhrg this is yours

sotodlib/site_pipeline/make_ml_map.py
skhrg: Yup as Carlos said its to avoid trying to reload (and reprocess) cut obs. 

The reduce honestly isn't needed since each process will only try to same (non overlapping) set of obs each pass. So I'll remove that.

sotodlib/mapmaking/utils.py
skhrg: There have been times in the past where the input obslist had all OTs in it and I have wanted to just map them one at a time. This adds the information for me do that without needing to filter the input files ahead of time.

sotodlib/mapmaking/utils.py
skhrg: Good point.

sotodlib/site_pipeline/make_ml_map.py
skhrg: Yeah, I also would prefer nargs.
This format is simply to copy the style of other existing arguments.

The sep thing in the help string is saying that `,` is the seperator.

If we want we can change this (and other similar args) to just use nargs but I don't think its worth doing that as a part of this PR.

sotodlib/site_pipeline/make_ml_map.py
iparask: @chervias this breaks the ML mapmaker for ACT. Since we have the `site` information from the arguments, shouldn't you pass it as an argument?

sotodlib/site_pipeline/make_ml_map.py
skhrg: There is some niceness to defaults that work easily...

I would advocate for the following:

* Change `sidelobes.get_cuts` to take the path to a directory where it searches for files with the names `{obj_name}.fits` for everything in `object_list`
* Make `object_list` a cli argument that defaults to `[sun, moon]` and pass it into `sidelobes.get_cuts`
* Take in the directory to look for files in as a CLI argument but dont require it or set a default
* If the directory is not passed do something sane to set the default in a what that isn't computer dependent. I would either:
  * Have it assume some relative path like `masks/sidelobe` (lame and probably not a useful default in most cases)
  * Have it grab the path to the metadata directory (by looking at where context says the obsdb is) and assuming a path relative to that (ie: `{metadata_dir}/masks/sidelobe`) and we put the current nominal masks there.

That way we end up with a system that doesn't rely on hardcoded paths but is easy to override as needed.

Probably too involved for this PR as it should be merged soon since its been hanging around for a while, but just throwing that out there as an option for future changes.

For now I think its fine that is will crash outside of NERSC, we can just add a warning to the help message that the defaults are for NERSC or something.

sotodlib/mapmaking/utils.py
skhrg: I imagine the intent was just to not sum over all of `weight`  since `cum[-1]` has already summed over all but the last element of `weight` anyways.

I think this:

```
cum_full = putils.cumsum(weight, endpoint=True)
cum, wtot = cum_full[:-1], cum_full[-1]
```

is the same thing as what is in the code and is a little bit easier to parse.

sotodlib/site_pipeline/make_ml_map.py
chervias: done

sotodlib/mapmaking/utils.py
chervias: done

sotodlib/mapmaking/utils.py
chervias: but we run it with so_lat in the ML mapmaker. This is a general function so it doesn't matter if the default is so. The difference is tiny.

chervias:
@mhasself there are still a couple of pending issues and unit tests, but you can start to have a look. Thanks!

sotodlib/mapmaking/utils.py
iparask: Aren't we supposed to support ACT data as well with SO tools?

mhasself (approved):
Ready when you are.  Please squash on merge, and clean up the commit message (do not include 157 commit messages in the log...).  Thanks!

PR #1235, opened 2025-05-28
Modify mainly source_flags so that we can use multiple sources and taua.

yoshinori-0778:
This PR is for pipeline planet mapmaking, which contains four changes:
- `source_flags` including `planet.py`:
  - These change allows us to use multiple sources and taua like the block below
  - I also modify source selection so that we can cut detectors that is not observing the source. And I add options for cutting detectors based on e.g. the Moon.
  - I add function for making Inverce mask of source mask.
   ```
    - name : "source_flags"
      calc:
        mask: {'shape': 'circle',
              'xyr': [0, 0, 2.0]}
        center_on: ['jupiter', 'tauA'] # list of str
        res: 2 # arcmin
        max_pix: 4000000 # max number of allowed pixels in map
        distance: 0 # max distance of footprint from source in degrees
        inv_flag: True
      save: True
      select: # optional
        kind: "any"
        select_source: ['jupiter']
   ```
- `trim_flag_edge`: I add this new function for triming TOD because I sometimes have a problem when I use `filter_for_source` (pca filter). This problem is due to interplay between gapfill and flags. If we have flags of True at the end of TOD, gapfill try to fill those based on only the prior TOD whereas gapfill generally try to fill based on both prior and posterior TODs. This can degrade the quality of the gapfill.  This function trim both the beginning and end of the TOD so that gapfill is never based on only one side.
- `glitch`: This change allows us to set any flags name.
- `correct_iir_params`: fix a bug 

I checked that these works as expected, but I am happy to modify them more if needed.

### DIFF
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -171,7 +171,6 @@ class GlitchDetection(_FracFlaggedMixIn, _Preprocess):
     Example configuration block::
         
       - name: "glitches"
-        glitch_name: "my_glitches"
         calc:
           signal_name: "hwpss_remove"
           t_glitch: 0.00001
@@ -191,15 +190,15 @@ class GlitchDetection(_FracFlaggedMixIn, _Preprocess):
     name = "glitches"
     _influx_field = "glitch_flags_frac"
 
-    def __init__(self, step_cfgs):
-        self.glitch_name = step_cfgs.get('glitch_name', 'glitches')
-        super().__init__(step_cfgs)
-
     def calc_and_save(self, aman, proc_aman):
         _, glitch_aman = tod_ops.flags.get_glitch_flags(aman,
             merge=False, full_output=True, **self.calc_cfgs
         ) 
-        aman.wrap(self.glitch_name, glitch_aman)
+        if self.calc_cfgs.get('name') is not None:
+            self.wrap_name = self.calc_cfgs['name']
+        else:
+            self.wrap_name = "glitches"
+        aman.wrap(self.wrap_name, glitch_aman)
         self.save(proc_aman, glitch_aman)
         if self.calc_cfgs.get('save_plot', False):
             flag_utils.plot_glitch_stats(aman, save_path=self.calc_cfgs['save_plot'])
@@ -208,7 +207,8 @@ class GlitchDetection(_FracFlaggedMixIn, _Preprocess):
         if self.save_cfgs is None:
             return
         if self.save_cfgs:
-            proc_aman.wrap(self.glitch_name, glitch_aman)
+            #proc_aman.wrap("glitches", glitch_aman)
+            proc_aman.wrap(self.wrap_name, glitch_aman)
  
     def select(self, meta, proc_aman=None):
         if self.select_cfgs is None:
@@ -216,7 +216,7 @@ class GlitchDetection(_FracFlaggedMixIn, _Preprocess):
         if proc_aman is None:
             proc_aman = meta.preprocess
         flag = sparse_to_ranges_matrix(
-            proc_aman[self.glitch_name].glitch_detection > self.select_cfgs["sig_glitch"]
+            proc_aman[self.wrap_name].glitch_detection > self.select_cfgs["sig_glitch"]
         )
         n_cut = count_cuts(flag)
         keep = n_cut <= self.select_cfgs["max_n_glitch"]
@@ -232,10 +232,9 @@ class GlitchDetection(_FracFlaggedMixIn, _Preprocess):
             filename = filename.replace('{obsid}', aman.obs_info.obs_id)
             det = aman.dets.vals[0]
             ufm = det.split('_')[2]
-            plot_signal_diff(aman, proc_aman[self.glitch_name], flag_type='glitches', flag_threshold=self.select_cfgs.get("max_n_glitch", 10), 
+            plot_signal_diff(aman, proc_aman.glitches, flag_type='glitches', flag_threshold=self.select_cfgs.get("max_n_glitch", 10), 
                              plot_ds_factor=self.plot_cfgs.get("plot_ds_factor", 50), filename=filename.replace('{name}', f'{ufm}_glitch_signal_diff'))
-            plot_flag_stats(aman, proc_aman[self.glitch_name], flag_type='glitches', filename=filename.replace('{name}', f'{ufm}_glitch_stats'))
-
+            plot_flag_stats(aman, proc_aman.glitches, flag_type='glitches', filename=filename.replace('{name}', f'{ufm}_glitch_stats'))
 
 class FixJumps(_Preprocess):
     """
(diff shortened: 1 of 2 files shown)


sotodlib/preprocess/processes.py
```diff
@@ -1212,66 +1211,112 @@ class SourceFlags(_Preprocess):
...
         else:
-            source_flags = proc_aman.source_flags
+            if self.calc_cfgs.get('source_flags_name') is not None:
+                source_flags = proc_aman[self.calc_cfgs.get('source_flags_name')]
+            else:
+                source_flags = proc_aman.source_flags
 
-        source_list = np.atleast_1d(self.calc_cfgs.get('center_on', 'planet'))
-        if source_list == ['planet']:
+        select_list = np.atleast_1d(self.select_cfgs["select_source"])
```
msilvafe, line 1299: Give an example of this in the example config entry(ies) in the docstring. Currently I think if you passed in `self.select_cfgs = True` as is done in the current docstring example this will fail bc you're missing `select_source` for that reason to observe backwards compatibility choose a reasonable default--probably all source flags.

sotodlib/preprocess/processes.py
```diff
@@ -1212,66 +1211,112 @@ class SourceFlags(_Preprocess):
...
-                keep = ~has_all_cut(source_flags[source])
+                if self.select_cfgs["kind"] == "any":
+                    keep = has_any_cuts(source_flags[source])
+                elif self.select_cfgs["kind"] == "all":
+                    keep = ~has_all_cut(source_flags[source])
+                elif self.select_cfgs["kind"] == "cut":
+                    keep = ~has_any_cuts(source_flags[source])
+                else:
+                    raise ValueError(f"Unknown kind of selection: {self.select_cfgs['kind']}")
                 meta.restrict("dets", meta.dets.vals[keep])
```
msilvafe, line 1316: Do one restrict at the end. Start with an empty `keep` flag then `keep +=` each of these over the `select_list` then run the restrict on that total keep at the end. Also I think a useful extension here would be not just to cut based on any or all criteria but if > X% of samples are flagged for a detector.

sotodlib/preprocess/processes.py
```diff
@@ -2016,10 +2061,26 @@ class CorrectIIRParams(_Preprocess):
...
+    name = 'trim_flag_edge'
+
+    def process(self, aman, proc_aman, sim=False):
+        flags = aman.flags.get(self.process_cfgs.get('flags'))
+        trimst = np.where(~np.any(flags.mask(), axis = 0))[0][0]
+        trimen = np.where(~np.any(flags.mask(), axis = 0))[0][-1]
+        aman.restrict('samps', (aman.samps.offset + trimst,
+                                aman.samps.offset + trimen))
+        proc_aman.restrict('samps', (proc_aman.samps.offset + trimst,
+                                     proc_aman.samps.offset + trimen))
```
msilvafe, line 2082: We handle this in the glitch detection step by not allowing samples to be flagged within some buffer number of samples from the beginning or end of the TOD could that be an extension to this instead? Also it should be able to compute this from the ranges tuples instead of completely unpacking from the RangesMatrices into boolean masks twice (that conversion can be slow always faster if you can work within Ranges/sparse arrays). Also need to add an example config entry for this.

sotodlib/preprocess/processes.py
```diff
@@ -171,7 +171,6 @@ class GlitchDetection(_FracFlaggedMixIn, _Preprocess):
     Example configuration block::
         
       - name: "glitches"
-        glitch_name: "my_glitches"
```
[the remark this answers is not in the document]
yoshinori-0778: I did test with an old version. I took back this modify.

sotodlib/preprocess/processes.py
```diff
@@ -1212,66 +1211,112 @@ class SourceFlags(_Preprocess):
...
-                                                          center_on=p[0],
+                                                          center_on=center_on,
                                                           res=self.calc_cfgs.get('res', None),
                                                           max_pix=self.calc_cfgs.get('max_pix', None))
 
-            source_aman.wrap(p[0], source_flags, [(0, 'dets'), (1, 'samps')])
+            source_aman.wrap(source, source_flags, [(0, 'dets'), (1, 'samps')])
+
+            if self.calc_cfgs.get('inv_flag') is not None:
+                if self.calc_cfgs['inv_flag']:
```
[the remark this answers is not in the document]
yoshinori-0778: That is good to know. I modified these.

sotodlib/preprocess/processes.py
```diff
@@ -1212,66 +1211,112 @@ class SourceFlags(_Preprocess):
...
+            updated_list = []
+            for isource in SOURCE_LIST:
+                if isinstance(isource, str):
+                    if isource in source_list:
+                        updated_list.append(isource)
+                elif isinstance(isource, tuple):
+                    isource_name = isource[0]
+                    if isource_name in source_list:
+                        updated_list.append(isource)
+            source_list = updated_list
```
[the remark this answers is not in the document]
yoshinori-0778: I made new function `get_source_list_fromstr` in planets.py and change scripts so that new function is used.

sotodlib/preprocess/processes.py
yoshinori-0778: Modified these based on your comments. I made new function `has_ratio_cuts` and `flag_cut_select`. default values of new parameters written in yaml file are set for backward compatibility.

sotodlib/preprocess/processes.py
yoshinori-0778: Modified these so that we calculate these indices with range. New version is much faster than old version by about order of three.
Also very good to know how you handle this in the glitch detection. I agree that we can do the same treatment in the filter_for_sources but I would like to add this class because it need additional study. I could try this type of treatment later.

yoshinori-0778:
I modified scritps based on Max's comments. Thank you Max.

sotodlib/preprocess/processes.py
yoshinori-0778: @mmccrackan I noticed that recent PR https://github.com/simonsobs/sotodlib/pull/1116 modified these part and conflict with my version now. how should I do? This is the first time to have a conflict...
Should I revert my change and make another PR for this?

mmccrackan: I think it should be fine to add a small change here.  Basically you'll just want to pass `in_place=True` into `select()`.  If `True`, return the restricted `meta`.  Otherwise, return the final `keep` after accumulating all the flags you want in it.

yoshinori-0778: Thank you for your comment, I slightly modified my modification to incorporate your modification. Also rebased to avoid conflict.

yoshinori-0778:
I rebased and slightly modified it to avoid conflict and incorporate @mmccrackan's modification https://github.com/simonsobs/sotodlib/pull/1116.

yoshinori-0778:
@msilvafe @mmccrackan This is ready for review again. Would you mind if you do review when you have a time?

msilvafe:
> @msilvafe @mmccrackan This is ready for review again. Would you mind if you do review when you have a time?

Sounds good, I'll take a look this morning. For the future, if you click the "re-request review" button on the top right of the PR I'll get notified that you're ready for me to look over it again.

yoshinori-0778:
I see, I did not know that button, good to know, thank you.
Thank you for taking time for review as well!

msilvafe (approved):
A couple inline comments but I'm approving the changes. Thanks for making all of the requested updates. I think we will benefit from having the updates you added to flagman and more examples of manipulating RangesMatrices without expanding to boolean arrays.

sotodlib/core/flagman.py
```diff
@@ -299,3 +354,23 @@ def sparse_to_ranges_matrix(arr, buffer=0, close_gaps=0, val=True):
         x[i].close_gaps(close_gaps)
     return x
 
+def find_common_edge_idx(flags):
+    """Find the common valid range across multiple RangesMatrix objects.
+
+    Args:
+        flags (RangesMatrix): An instance of so3g.proj.RangesMatrix indicating flagged time ranges.
+    Returns:
```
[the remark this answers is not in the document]
yoshinori-0778: I do not have an situation that we want to find edges that all have True now, so I would like to leave this for now. But thank you for the suggestion.

sotodlib/core/flagman.py
```diff
@@ -278,6 +278,61 @@ def has_all_cut(flag):
...
+      ratio (float or int): The threshold ratio above which a flag is 
+        considered to have too many cuts.
+    """
+
+    if not isinstance(ratio, (float, int)):
+        raise ValueError('Ratio must be float or int')
+    len_samps = flag.shape[-1]
+    return np.array([np.sum(np.diff(x.ranges()))/len_samps > ratio for x in flag])
+
+def flag_cut_select(flags, cut, kind):
```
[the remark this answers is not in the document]
yoshinori-0778: Sure, I changed the name to `invert` and modified corresponding parts.

PR #1236, opened 2025-05-28
band stop filter

ykyohei:
We need band stop filter remove PTC pickup etc, this is missing from fft_filters.


msilvafe (approved):
Looks good, how have you tested this?

Issue #995, opened 2024-10-10
calc_wn is biased


ykyohei:
`fft_ops.calc_wn` takes median of psd as a estimator of white noise level (wn), and this is biased from the average of psd.

The power spectrum density is usually calculated by the `fft_ops.calc_psd` using scipy.welch, and the amount of bias can be different between different observations because we allow different nperseg for different observations.

There are at least two ways to obtain the unbiased estimate of wn.
1. Take an average of the psd in the frequency range without peaks.
2. Take a median of the psd to be robust against peaks and then debias. For example. we set `noverlap = 0` in scipy welch and debias depending on the number of segments. The average PSD of non-overlapping n segments (scipy.welch psd with noverlap = 0) follows chi square distribution with 2*n degrees of freedom, thus correction factor will be `2*n/scipy.stats.chi2.ppf(0.5, 2*n)`. The maximum correction factor for psd is ~1.44 when n=1. The current fft_ops.calc_psd allows overlap between segments and it's more complicated to obtain the correction factor.

Controlling the bias of wn is important for correct understanding of the detector noise performance, and for correct inverse variance weighting for mapmaking.

![Figure 25 (3)](https://github.com/user-attachments/assets/23d924e8-99d6-444e-a01c-c9f27d33e526)

PR #1238, opened 2025-05-29
Debias wn

ykyohei:
This PR corrects the bias of median white noise estimation pointed out in https://github.com/simonsobs/sotodlib/issues/995
Also added unit test for this.


ykyohei:
Merged latest changes of noise fit and confirmed iso preprocess runs. This PR should be non-breaking changes now. 
I moved test_psd from test_tod_ops.py to test_psd.py but it's not changed. I hope it's okay.

sotodlib/preprocess/processes.py
```diff
@@ -356,41 +356,55 @@ class PSDCalc(_Preprocess):
...
     def __init__(self, step_cfgs):
         self.signal = step_cfgs.get('signal', 'signal')
         self.wrap = step_cfgs.get('wrap', 'psd')
 
         super().__init__(step_cfgs)
 
     def calc_and_save(self, aman, proc_aman):
-        freqs, Pxx = tod_ops.fft_ops.calc_psd(aman, signal=aman[self.signal],
-                                              **self.calc_cfgs)
+        full_output = 'full_output' in self.calc_cfgs and self.calc_cfgs['full_output']
```
msilvafe, line 386: This can be `full_output = self.calc_cfgs.get('full_output')` it will populated `full_output` with `None` if not passed which is treated the same as `False` in `if/else` statements so will work for your logic.

sotodlib/tod_ops/fft_ops.py
```diff
@@ -8,13 +8,17 @@
 import numdifftools as ndt
 import numpy as np
 import pyfftw
+import logging
```
msilvafe, line 11: Instead of introducing a logger just to add the warning message use python warnings package i.e.
```
import warnings
...
warnings.warn('calc_wn will be biased. noverlap argument of welch '
                          'needs to be 0 to get unbiased median white noise estimate.')
```

sotodlib/tod_ops/fft_ops.py
```diff
@@ -295,17 +300,38 @@ def calc_psd(
...
-        freqs, Pxx = _calc_psd_subscan(aman, signal=signal, freq_spacing=freq_spacing, **kwargs)
+        if full_output:
+            freqs, Pxx, nseg = _calc_psd_subscan(aman, signal=signal,
+                                                 freq_spacing=freq_spacing,
+                                                 full_output=True,
+                                                 **kwargs)
+        else:
+            freqs, Pxx = _calc_psd_subscan(aman, signal=signal,
+                                           freq_spacing=freq_spacing,
+                                           **kwargs)
```
msilvafe, line 332: does this need to be in an `if/else`? Can't you just pass `full_output=full_output`?

sotodlib/tod_ops/fft_ops.py
```diff
@@ -424,12 +480,28 @@ def calc_wn(aman, pxx=None, freqs=None, low_f=5, high_f=10):
     if pxx is None:
         pxx = aman.Pxx
 
+    if nseg is None:
+        nseg = aman.get('nseg')
+
+    if nseg is None:
+        logger.warning('white noise level estimated by median PSD is biased. '
+                       'nseg is necessary to debias. Need to use following '
+                       'argements in calc_psd to get correct nseg. '
```
msilvafe, line 489: typo "arguments"

msilvafe (approved):
Thanks for adding such extensive tests, will be helpful! Please address my inline comments but will give an approval.

sotodlib/tod_ops/fft_ops.py
ykyohei: This is because returned values are different.

Issue #1215, opened 2025-05-12
Preprocess: proc_aman and aman trimmed differently in PCARelcal


earosenberg:
On this line in PCARelcal:
https://github.com/simonsobs/sotodlib/blob/65d07b78ae01f7f648ab1b782a33b3b045a0e5d2/sotodlib/preprocess/processes.py#L1349
proc_aman samps are trimmed and aman are not. This causes them to get out of sync and creates problems when trying to pull information from proc_aman into aman (e.g. UnionFlags). 
My guess is that the `trim_samps` in `demodulate` was masking this issue previously.

Just adding a line to trim `aman` would solve the problem, but break the rule of "calc_and_save doesn't modify the aman", so pinging preprocess experts (@msilvafe @mmccrackan) for input.

mmccrackan:
Thanks for bringing this up.  It looks like a samps restriction on `aman` was added here:
https://github.com/simonsobs/sotodlib/commit/ffa4c72300a50079f7cc14958b6e336902c307c0

But was removed here probably because it modified `aman` as you say:
https://github.com/simonsobs/sotodlib/commit/60fb595cd4ce8d37eb947a03f6ed0e76905f0952

I guess maybe we could add a temporary filtered signal array (since we don't want to actually lowpass signal in `aman`) into `aman` or `proc_aman` in `process` and maybe delete it in `calc_and_save` once we're done with it so it doesn't get saved.  We could also restrict on `samps` in the `select` function, but that would require `select` to always be enabled.

PR #1242, opened 2025-05-31
Fix aman and proc_aman unequal restriction.

msilvafe:
Addresses issue #1215.


mmccrackan (approved):
As discussed, I agree that this should be fine since there shouldn't be anything depending on `samps` being written out in the relcal functions now.  We save `pca_mode0` temporarily for plotting but it is deleted so as not to save per-band information.

PR #1245, opened 2025-06-03
Add quality report plotting

mmccrackan:
Adds components for quality report plots.  Files adapted from [here](https://github.com/simonsobs/pwg-scripts/tree/master/pwg-qds/quality_reports).

See notebook on site here for how to run: `/so/home/mmccrackan/quality_plots/20250530_quality_plots/20250603_sotodlib_qplots.ipynb`

Config files here:
```
/so/home/mmccrackan/quality_plots/20250530_quality_plots/longterm_cfg.yaml
/so/home/mmccrackan/quality_plots/20250530_quality_plots/generate_report.yaml
```


ahincks:
Plots look nice! Main question is whether we want the code that generates the plots and the webpages to live here in `sotodlib` or whether it belongs in another (perhaps new?) repo that imports `sotodlib`. @mhasself , what is your opinion?

(If it would be better to put them in a repo, and if that were to delay deploying this live, I might suggest that we merge it now and move it later, just so that we can get this up and running.)

PR #1246, opened 2025-06-03
perf: skip shape check when expanding ranges matrix if possible

skhrg:
Adds compatibility for https://github.com/simonsobs/so3g/pull/213

### DIFF
--- a/sotodlib/core/axisman_io.py
+++ b/sotodlib/core/axisman_io.py
@@ -1,5 +1,4 @@
 import inspect
-from functools import partial
 import so3g
 import numpy as np
 import json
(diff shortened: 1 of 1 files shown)


sotodlib/core/axisman_io.py
```diff
@@ -13,6 +15,11 @@
 from .axisman import *
 from .flagman import FlagManager
 
+# Backwards compatibility for before skip_shape_check was added
+RangesMatrix = so3g.proj.RangesMatrix
+if "skip_shape_check" in inspect.signature(so3g.proj.RangesMatrix).parameters:
+    RangesMatrix = partial(so3g.proj.RangesMatrix, skip_shape_check=True)
+
```
mhasself, line 22: Instead of replacing RangesMatrix class with whatever partial(...) would give you, please just create var
```
_rm_fast_kwargs = {}
if "skip_shape_check" ...:
  _rm_fast_kwargs = {'skip_shape_check': True}
```
Then pass those to usual class.

mhasself (changes requested):
One note...

PR #1249, opened 2025-06-04
Add detcal cuts.

msilvafe:
Move nan cuts out of fourier_filter and into new process. Uses meta.det_cal.tau_eff to check for nans instead of signal to be compatible with loading.


mmccrackan (approved):
Okay, pretty sure this should work.  If `det_cal` is included in the context I think that the select in the `load_and_preprocess` and `run` functions should always be populated.

PR #1252, opened 2025-06-06
New atomic db and some changes to the MPI

chervias:
This PR does 2 things
1. The line `executor: Union["MPIPoolExecutor", "ProcessPoolExecutor"]` is changed to  `executor: Union["MPICommExecutor", "ProcessPoolExecutor"]`. @iparask can you have a look? This is what @msilvafe and @mmccrackan are doing for the preprocessing script. I don't understand much, but they told me to change it. However, now you have to run with mpi and 2+ tasks.
2. The writing into the atomic sqlite database is moved out from the parallel function and brought back to the `as_completed_callable` loop, in a similar way to how it is done in the preprocessing script. @mhasself I have talked to you about this issue a couple of times, I would love to hear your thoughts on how we handle it here.

### DIFF
--- a/sotodlib/mapmaking/demod_mapmaker.py
+++ b/sotodlib/mapmaking/demod_mapmaker.py
@@ -626,13 +626,15 @@ def add_weights_to_info(info, weights, split_labels):
         if sub_weights.shape[0] != 3:
             raise ValueError(f"sub_weights has unexpected shape {sub_weights.shape}. First axis should be (3,) for TT, QQ, UU")
         mean_qu = np.mean(sub_weights[1:], axis=0)
-        positive = np.where(mean_qu > 0)
-        sumweights = np.sum(mean_qu[positive])
-        meanweights = np.mean(mean_qu[positive])
-        medianweights = np.median(mean_qu[positive])
-        sub_info.total_weight_qu = float(sumweights)
-        sub_info.mean_weight_qu = float(meanweights)
-        sub_info.median_weight_qu = float(medianweights)
+        pweights = mean_qu[mean_qu > 0]
+        if pweights.size == 0:
+            pweights = [0]
+        sumweights = np.sum(pweights)
+        meanweights = np.mean(pweights)
+        medianweights = np.median(pweights)
+        sub_info.total_weight_qu = sumweights
+        sub_info.mean_weight_qu = meanweights
+        sub_info.median_weight_qu = medianweights
         info[isplit] = sub_info
     return info
 
--- a/sotodlib/mapmaking/demod_mapmaker.py
+++ b/sotodlib/mapmaking/demod_mapmaker.py
@@ -452,17 +452,15 @@ def setup_demod_map(noise_model, shape=None, wcs=None, nside=None,
                                          singlestream=singlestream)
     return mapmaker
 
-def write_demod_maps(prefix, data, info, unit='K', split_labels=None,):
+def write_demod_maps(prefix, data, info, unit='K', split_labels=None):
     """
     Write maps from data into files
     """
     Nsplits = len(split_labels)
     for n_split in range(Nsplits):
-        if np.all(data.wmap[n_split] == 0.0):
-            info[n_split]['valid'] = False
+        info[n_split]["valid"] = np.any(data.wmap[n_split] != 0.0)
+        if info[n_split]["valid"] == False:
             continue
-        else:
-            info[n_split]['valid'] = True
         data.signal.write(prefix, "%s_wmap"%split_labels[n_split],
                           data.wmap[n_split], unit=unit+'^-1')
         data.signal.write(prefix, "%s_weights"%split_labels[n_split],
@@ -542,6 +540,9 @@ def make_demod_map(context, obslist, noise_model, info,
         List of errors from preprocess database. To be used in cleanup_mandb.
     outputs : list
         List of outputs from preprocess database. To be used in cleanup_mandb.
+    info: list
+        List of dictionaries with the info to write into atomic db. It will be
+        the same as the input, but the weights values will be added.
     """
     from ..preprocess import preprocess_util
     #context = core.Context(context)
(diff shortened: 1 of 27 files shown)


sotodlib/mapmaking/demod_mapmaker.py
```diff
@@ -608,9 +605,8 @@ def make_demod_map(context, obslist, noise_model, info,
     info = add_weights_to_info(info, weights, split_labels)
 
     # output to files
-    write_demod_maps(prefix, mapdata, info, split_labels=split_labels, atomic_db=atomic_db, unit=unit)
-
-    return errors, outputs
+    write_demod_maps(prefix, mapdata, info, split_labels=split_labels, unit=unit)
+    return errors, outputs , info
```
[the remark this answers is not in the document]
JBorrow: New return value, info, should also be documented

sotodlib/site_pipeline/make_atomic_filterbin_map.py
```diff
@@ -464,12 +464,17 @@ def main(
...
-            atomic_db=args.atomic_db,) for r in run_list]
+            wn_label=args.wn_label,) for r in run_list]
     for future in as_completed_callable(futures):
         L.info('New future as_completed result')
         try:
-            errors, outputs = future.result()
+            errors, outputs, d_ = future.result()
+            if d_ is None:
+                # this means the map failed and the function returned early
+                continue
```
JBorrow, line 474: Would likely make more sense to check `errors` than the value of `d_`, no? The usual paradigm is:

```
def my_func(x: float):
  errors = []
  if x == 2.0:
    errors.append("x is two!!!")
    return errors, None

  return errors, 2.0 - x

err, val = my_func(4.0)

if err:
  ... Handle error

... Use val
```

JBorrow (commented):
Some minor comments but looks good

sotodlib/site_pipeline/make_atomic_filterbin_map.py
```diff
@@ -464,12 +464,17 @@ def main(
...
+            for n_split in range(len(split_labels)):
+                info3 = mapmaking.AtomicInfo.from_dict(d_[n_split])
+                mapmaking.atomic_db_aux(args.atomic_db, info3)
```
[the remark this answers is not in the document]
JBorrow: Bulk save is not a good idea (it's designed for far more data than this) but using add_all would be better.

sotodlib/mapmaking/demod_mapmaker.py
```diff
@@ -452,32 +452,29 @@ def setup_demod_map(noise_model, shape=None, wcs=None, nside=None,
...
     Write maps from data into files
     """
     Nsplits = len(split_labels)
     for n_split in range(Nsplits):
-        if np.all(data.wmap[n_split] == 0.0):
-            if atomic_db is not None:
-                smutils.atomic_db_aux(atomic_db, info[n_split], valid=False)
+        info[n_split]["valid"] = np.any(data.wmap[n_split] != 0.0)
+        if info[n_split]["valid"] == False:
             continue
```
JBorrow, line 463: Sorry, I mis-understood the control flow here. 
```suggestion
        if not np.any(data.wmap[n_split] != 0.0):
            continue
        
        info[n_split]["valid"] = True
```

sotodlib/mapmaking/demod_mapmaker.py
```diff
@@ -452,32 +452,31 @@ def setup_demod_map(noise_model, shape=None, wcs=None, nside=None,
...
     Nsplits = len(split_labels)
     for n_split in range(Nsplits):
         if np.all(data.wmap[n_split] == 0.0):
-            if atomic_db is not None:
-                smutils.atomic_db_aux(atomic_db, info[n_split], valid=False)
+            info[n_split]['valid'] = False
             continue
+        else:
+            info[n_split]['valid'] = True
         data.signal.write(prefix, "%s_wmap"%split_labels[n_split],
```
iparask, line 466: Since you want to write the data signals if `info[n_split]['valid'] = True` why don't you push them a level in?

That would make the continue (line 463) unnecessary, and make the code a bit easier to read.

sotodlib/mapmaking/demod_mapmaker.py
```diff
@@ -452,32 +452,31 @@ def setup_demod_map(noise_model, shape=None, wcs=None, nside=None,
...
+def write_demod_maps(prefix, data, info, unit='K', split_labels=None):
     """
     Write maps from data into files
     """
     Nsplits = len(split_labels)
     for n_split in range(Nsplits):
         if np.all(data.wmap[n_split] == 0.0):
-            if atomic_db is not None:
-                smutils.atomic_db_aux(atomic_db, info[n_split], valid=False)
+            info[n_split]['valid'] = False
```
iparask, line 462: By looking at the function, without taking its call into account, I thought that info was not needed. But it seems it is updated in place.

Can you mention that in the docstring if it is correct?

chervias: yes, you need to change `info` in place for the non-existing obs. This is for example an obs with wafer ws0 and det_out split doesn't exist, because ws0 is completely inside det_in. So these splits are cut fully and the weights map will be all 0s. valid=False is that, for maps that don't exist. You mean the docstring of `write_demod_maps`?

iparask: Yes!

sotodlib/site_pipeline/make_atomic_filterbin_map.py
iparask: Why not reverse the condition?

```
if d_ is not None:
    list_infos = [] ...
```

Since there is nothing else happening under that if I think reverting it is better.

mhasself (approved):
Glad you found a way!

I imagine it's also possible to create dissociated ORM objects -- i.e. blank ones that aren't attached to a session -- and then add them in on the other side.  But the dict way has many benefits.

PR #1254, opened 2025-06-09
Allow filter chaining

mmccrackan:
Update `FourierFilter` to allow for filter chaining.  Old input of `filter_function` is still supported for backwards compatibility.


msilvafe (approved):
Looks good to me, assuming you've tested backwards compatibility and an example of a chained filter happy to merge. Thanks!!!

Issue #1259, opened 2025-06-13
Un-ending print


AERAdler:
I think this line prints the whole data! Makes for hard to read logs...

https://github.com/simonsobs/sotodlib/blob/master/sotodlib/toast/workflows/sim_sky.py#L137

PR #1260, opened 2025-06-13
print bug fix in sim_sky.py

AERAdler:
See #1259


tskisner (approved):
Thanks for catching this stale debug printing!

PR #1219, opened 2025-05-13
load_book: support for fixed tones, smurfgaps flags, no_headers

mhasself:
- When "special_channel=True" is passed to get_obs, the "untracked" channels are loaded and populate a .tones sub-AxisMan.
- The smurfgaps flags are loaded from the per-stream frame data and put into `.flags.smurfgaps_{stream_id}`.
- When user passes "no_headers=True", the smurf header fields will not be unpacked (so there is no .primary).

PR #1261, opened 2025-06-13
update-smurf-caldbs: error checking, and get_obs(no_headers=False)

mhasself:
update_smurf_caldbs (specifically the det_cal update) started failing when the "don't load headers for obs books" feature was enabled by default (#1219), because it needs bias_lines and those aren't loaded unless you're getting either the signal or the frame headers.  No one noticed for a little bit.

This PR does two things, for det_cal:
1. adds `no_headers=False` to the get_obs.
2. counts successes / failures / crashes, and logs them
3. raises an error, on exit, if there were any individual "crashes".

### DIFF
--- a/sotodlib/site_pipeline/update_smurf_caldbs.py
+++ b/sotodlib/site_pipeline/update_smurf_caldbs.py
@@ -241,7 +241,8 @@ def get_cal_resset(ctx: core.Context, obs_id) -> CalResult:
     for each detector in the observation
     """
     am = ctx.get_obs(
-            obs_id, samples=(0, 1), ignore_missing=True, no_signal=True,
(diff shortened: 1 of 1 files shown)


guanyilun (approved):
great, thanks for the fix! Would it make sense to also report which obs_id has failed at the end?

kmharrington (approved):
yup, works for me

mhasself:
> Would it make sense to also report which obs_id has failed at the end? Because it will show up in prefect alert, it could be made more informative.

Do you mean to add all the crashed obs_ids to the exception text?  That could get crazy.  I think in practice one would immediately want to go to the log to see _what kind of crashes_ they were, anyway.

Issue #1263, opened 2025-06-13
crossover split


ykyohei:
Crossover split works for f090, but this split does not work correctly for f150 because crossover for f150 is L or R.
Need to know which is same as B to correct this split.
https://github.com/simonsobs/sotodlib/blob/6e4f8cce755930bc639e681713ae75c247f2bf87/sotodlib/obs_ops/splits.py#L117-L119

PR #1264, opened 2025-06-13
fix crossover split

ykyohei:
This resolves https://github.com/simonsobs/sotodlib/issues/1263
Daniel confirmed by directly looking at the full wafer GDS and following traces to the labeled bond pads. because we did not find good document for this.
This split was not used in sat-iso null tests.

### DIFF
--- a/sotodlib/obs_ops/splits.py
+++ b/sotodlib/obs_ops/splits.py
@@ -88,7 +88,7 @@ def get_split_flags(aman, proc_aman=None, split_cfg=None):
     '''
     # Set default set of splits
     default_cfg = {'high_gain': 0.115, 'high_tau': 1.5e-3,
-                   'det_A': 'A', 'pol_angle': 35, 'det_top': 'BL', 'right_focal_plane': 0,
+                   'det_A': 'A', 'pol_angle': 35, 'crossover': 'BL', 'right_focal_plane': 0,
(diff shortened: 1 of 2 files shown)


mhasself:
This change matches what I have.  But for terminology: "crossover" seems like a better name for this split than "top/bottom".  It's also historically called the "detector AB" split.  We actually use these "A and B" labels for the two polarization channels in each pixel x band:

- type "A": crossover in ['R', 'T']
- type "B": crossover in ['L', 'B']

The A, B code is available in `det_info.wafer.pol`.  The correspondence with RT / LB was put in carefully:
```
>>> set(rs['dets:wafer.crossover'][rs['dets:wafer.pol']=='A'])
{np.str_('T'), np.str_('R'), np.str_('X')}
>>> set(rs['dets:wafer.crossover'][rs['dets:wafer.pol']=='B'])
{np.str_('B'), np.str_('X'), np.str_('L')}

```

mhasself (approved):

msilvafe:
> Fine with me. @msilvafe put that name here, are you ok with it?

Yes, fine with me as well.

ykyohei:
Renamed and tested, 
This doesn't break the preprocessing with current config det_top threshold will be ignored.
@chervias Could you check?

PR #1268, opened 2025-06-18
Revert "TOAST Scan-Synchronous Signal now requires Stokes weights"

keskitalo:
Reverts simonsobs/sotodlib#1265


tskisner:
I think we can leave this change in, since there should be a compatible toast tag out in a few hours.  However, I'll leave this open until then as a reminder.

tskisner:
Everything is now compatible, so this PR is not needed.

PR #1269, opened 2025-06-20
Add option to run preprocesing with cprofile

mmccrackan:
Adds an optional call to `preprocess_tod` and `multilayer_preprocess_tod` to run with `cprofile` and output profile files.  Also adds `cprofile` decorator.

Might not merge, just adding for keeping track of profiling/benchmarking efforts and having a place for potential related discussions.


mmccrackan:
This is super out of date and Josh has a different implementation for profiling so since this wasn't really intended to be merged, I will close.

Issue #1185, opened 2025-04-15
obsdb: warn and convert double quotes to single quotes


mhasself:
Related to #1119 and #1126; when users construct their own queries they might not know about this.  So .warn and .replace.

Also, check in on #705 ... is that fixed?

Issue #1178, opened 2025-04-09
LAT obsdb has invalid data in some columns, leads to bad arrays


mhasself:
For example:
```
from sotodlib import core
ctx = core.Context('/so/metadata/lat/contexts/basic.yaml')
rs = ctx.obsdb.get()
print(rs['el_throw'])
>> array([None, None, None, ..., 0.0, 0.0, 0.0], dtype=object)
```

The reasonable expectation is that `rs['el_throw']` should be an array of _float_.

We may have no way of encoding NaNs in the db... in which case we have to catch them in the ResultSet conversion-to-array routine.

PR #1271, opened 2025-06-24
Quality of life obsdb/manifestdb/resultset fixes

### LINKED ISSUE #705: Escape the SQLlite queries with colons?
We've written a bunch of tables sqlite tables with column names that include colons ("obs:obs_id" is everywhere). I often run into the issue that the colon inside the string gets sent directly to sqlite and than causes an operation error. I know there's a way to escape it (but that way escapes me right now, hence writing this issue :-D ). Can we add string conditioning to the functions that are generating sqlite commands where this comes up?

Big ones for me are looking at manifest dbs. Running ` db.get_entries(fields=["obs:obs_id"])` gets an sqlite operational error and not a list of obs_ids in the database.

mhasself:
- ObsDb will create indices on main tables.
- ObsDb.query will noisily replace double quotes with single quotes in
  query strings.
- update_obsdb will populate LAT "detector_flavor" (previously was
  leaving null).
- ResultSet -- if a column has a None value, some effort is made to
  cast to the type of the other data in the column.
- core imports .metadata, to give easier access to those things and
  simplify use alongside io.metadata
- ManifestDb.get_entries() will auto-quote fields, so colons in field
  names are easier to work with.

Resolves #705, #1178, #1185.

### DIFF
--- a/sotodlib/site_pipeline/update_obsdb.py
+++ b/sotodlib/site_pipeline/update_obsdb.py
@@ -78,7 +78,8 @@ def telescope_lookup(telescope: str):
         return {"telescope": "satp3", "telescope_flavor": "sat",
                 "tube_flavor": "mf", "detector_flavor": "tes"}
     elif telescope == "lat":
-        return {"telescope": "lat", "telescope_flavor": "lat"}
+        return {"telescope": "lat", "telescope_flavor": "lat",
+                "detector_flavor": "tes"}
     else:
         logger.error("unknown telescope type given by bookbinder")
         return {}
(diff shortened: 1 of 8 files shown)


sotodlib/site_pipeline/update_obsdb.py
AERAdler, line 1: Are we 100% sure no MKIDs will ever get in the LAT? I thought it was one of the SO:UK deliverables.

AERAdler (approved):
Looks good. Thanks Matthew!

PR #1272, opened 2025-06-24
Fix broken pior support

amaurea:
This fixes the broken or incomplete support for priors in the maximum-likelhood mapmaker. This must have worked before, but has broken at some point, so the prior was not being taken into account at all. This is needed when doing source sub-sampling. Without the prior, the equation system becomes degenerate, and the mapmaker does not converge properly. This pull requests restores the missing functionality. I've tested it with a case that used to not work (a LAT scan of the galaxy), and confirmed that it now converges. The test was done on a branch with some more features for reading in LAT data, but I've isolated the actual fix in this pull request.


Bai-Chiang:
The TOAST operator also works.

PR #1274, opened 2025-06-25
Changes to better support SAT reduction with toast

tskisner:
This adds logic to insert the preproc archive in the live config.  It also ensures that the hwp model is only applied once.


AERAdler (approved):
The extra flexibility will come in handy, and I don't see anything that would break other sotodlib uses.

PR #1277, opened 2025-06-26
add process for smurfgaps

ykyohei:
This flag is very useful, and we need to use more.
The default buffer value is arbitrary, I chose this based on ws2 of obs_1727642777_satp3_1111111 and obs_1727991085_satp3_1111111

link to the presentation showing the effectiveness of this flag.
https://simonsobs.atlassian.net/wiki/spaces/PRO/pages/806846478/SAT+LAT+Analysis+Section+1+telecons#26-June-2025-(B-slot-%E2%80%93-8pm-ET)

### DIFF
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -2336,6 +2336,8 @@ class TrimFlagEdge(_Preprocess):
 
 class SmurfGapsFlags(_Preprocess):
     """Expand smurfgaps flag of each stream_id to all detectors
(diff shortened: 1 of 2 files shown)


msilvafe (approved):
It would be great to expand the docstrings to describe how these smurfgaps are constructed from the lower level smurf data but otherwise happy and eager to have this in master + start using it.

ykyohei:
Added some explanations based on discussion in DM1 call. I'm going to merge this later today unless someone stops me!

PR #1279, opened 2025-06-27
Sim filtering iso v2

kwolz:
Small fixes that guarantee simulation filtering is working for ISO v2.

### DIFF
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -1138,7 +1138,7 @@ class SubtractAzSSTemplate(_Preprocess):
     name = "subtract_azss_template"
 
     def process(self, aman, proc_aman, sim=False):
-        process_cfgs = self.process_cfgs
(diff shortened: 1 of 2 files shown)


sotodlib/preprocess/processes.py
```diff
@@ -1138,7 +1138,11 @@ class SubtractAzSSTemplate(_Preprocess):
     name = "subtract_azss_template"
 
     def process(self, aman, proc_aman, sim=False):
-        tod_ops.azss.subtract_azss_template(aman, **self.process_cfgs)
+        process_cfgs = self.process_cfgs
```
mmccrackan, line 1141: You might want to make a copy here since this is a dictionary and this will only be a reference.  We have run into issues with modifying references to the config dictionaries which will affect thing when the function is run again.

mmccrackan (changes requested):

msilvafe (approved):

PR #1286, opened 2025-07-08
TOAST mlmapmaker: Add downweight option

Bai-Chiang:


Bai-Chiang:
What should we set default downweight value in TOAST? Should we disable it by default?

chervias:
> What should we set default downweight value in TOAST? Should we disable it by default?

I think so, we can probably set it to be 1s by default

keskitalo (approved):
This is a valuable addition.  I would prefer defaulting to no downweight but can be convinced otherwise.

PR #1287, opened 2025-07-10
Handle collided resonators in uMUX crosstalk

tpsatt:
Supports handling of collided resonators (very small measured frequency spacing in UFM screening at Princeton -- makes chi get very large) by not adding crosstalk to these TODs, and flagging detectors.


sotodlib/toast/ops/mumux_crosstalk_util.py
```diff
@@ -211,16 +216,29 @@ def pos_to_chi(focalplane, dets, alpha=9.64e-30, tol=1.0):
...
+                    # freq_neighbors[0] is idet1
+                    # frequency neighbors on either side
+                    continue
                 x2, y2 = position_subset[idet2]
                 # Translate frequencies to chi
                 df = freq1 - freq2
                 avg_f = (freq1 + freq2) / 2
                 chi = alpha * np.power(avg_f, 4) * np.power(df, -2) * 1e12
-                if chi > 1:
+                # Check for anomalously high chi in absense of resonator collision
```
keskitalo, line 232: This seems like a hack. What is causing anomalously high chi without collision?

sotodlib/toast/ops/sim_mumux_crosstalk.py
```diff
@@ -173,6 +175,8 @@ def _draw_Phi0(self, obs, focalplane, detectors, vmin=0.3, vmax=1.3):
                 counter=(counter1, counter2),
             )[0]
             v = vmin + x * (vmax - vmin)
+            # Convert from Phi0 to Phi
+            v *= 2 * np.pi
```
keskitalo, line 179: This looks significant... Glad you caught it.

sotodlib/toast/ops/sim_mumux_crosstalk.py
```diff
@@ -285,9 +292,16 @@ def _exec(self, data, detectors=None, **kwargs):
...
                     )
-                    crosstalk += chi * np.sin(
-                        source_squid_phase - target_squid_phase
-                    )
+                    # Add crosstalk if not collided resonator
+                    if not np.isnan(chi):
+                        crosstalk += chi * np.sin(
+                            source_squid_phase - target_squid_phase
+                        )
+                    #else:
```
keskitalo: We can probably remove the debugging statements.

keskitalo (approved):
Critical fixes and readability improvements. Great work!

sotodlib/toast/ops/mumux_crosstalk_util.py
tpsatt: The print statements are a legacy from the original implementation. I can't think of a reason the chi would be anomalously high without a collision other than anomalously high frequencies (i.e. input in the wrong units) which we haven't seen before but could be worth checking against

sotodlib/toast/ops/sim_mumux_crosstalk.py
tpsatt: Just fixed this so the detectors are now flagged (and tested) -- thanks for catching

PR #1288, opened 2025-07-10
AxisManager.save supports flacarray compression

mhasself:
Remember to pip install flacarray!


tskisner (approved):
I think this is great- one small comment.  This gets us another step closer to a L3 test book and context loader.

PR #1289, opened 2025-07-14
mpi option for make_hwp_solution

ykyohei:
Add mpi option to run make_hwp_solution in both site computing and data center.


ykyohei:
@iparask thanks for suggestions, I think I simplified the pipeline.

PR #1291, opened 2025-07-15
mlmapmaker: Fix missing ivar when loading NmatUnit

Bai-Chiang:
This fix missing ivar entry when loading cached `NmatUnit` noise model in TOAST.


chervias (approved):
Ok, I assume this works. Looks good.

PR #1296, opened 2025-07-15
Reuse rfft object in demod_tod

mmccrackan:
Uses the new functionality from #1230 to create an RFFT object up-front for all of the filters in `demod_tod`.  This saves a small amount of time (5-10 seconds) when demodulating.  Note the `.copy()` is required since they will only store references if those are not included.

Tested by saving `demodQ`, `demodU`, and `dsT` with the old version then reloading and comparing them with the values from the new approach.  Confirmed they are exactly zero.

### DIFF
--- a/tests/test_demod_map.py
+++ b/tests/test_demod_map.py
@@ -113,8 +113,6 @@ class DemodMapmakingTest(unittest.TestCase):
         signal_map[1] += Q_stream
         signal_map[2] += U_stream
 
-        print(signal_map.dtype)
-
         # from_map expects and outputs float64
         _ = coords.demod.from_map(tod, signal_map, modulated=True, wrap=True)
         hwp.demod_tod(tod)


sotodlib/hwp/hwp.py
```diff
@@ -665,20 +665,28 @@ def demod_tod(aman, signal=None, demod_mode=4,
...
     demod = tod_ops.fourier_filter(aman, bpf, detrend=None,
-                                   signal_name=signal_name) * phasor
+                                   signal_name=signal_name, rfft=rfft).copy() * phasor
 
     # Filter the demodulated signal
     demod_aman = core.AxisManager(aman.dets, aman.samps)
     demod_aman.wrap("timestamps", aman.timestamps, axis_map=[(0, 'samps')])
     demod_aman.wrap("dsT", aman[signal_name], axis_map=[(0, 'dets'), (1, 'samps')])
-    demod_aman["dsT"] = tod_ops.fourier_filter(demod_aman, lpf, signal_name='dsT', detrend=None)
+    demod_aman["dsT"] = tod_ops.fourier_filter(demod_aman, lpf, signal_name='dsT', detrend=None, rfft=rfft).copy()
```
mhasself, line 681: You can use `demod_aman["dsT"][:] = ...` instead of the copy.

sotodlib/hwp/hwp.py
```diff
@@ -665,20 +665,28 @@ def demod_tod(aman, signal=None, demod_mode=4,
...
     demod_aman.wrap("demodQ", demod.real, axis_map=[(0, 'dets'), (1, 'samps')])
-    demod_aman["demodQ"] = tod_ops.fourier_filter(demod_aman, lpf, signal_name="demodQ", detrend=None) * 2.
+    demod_aman["demodQ"] = tod_ops.fourier_filter(demod_aman, lpf, signal_name="demodQ", detrend=None, rfft=rfft).copy() * 2.
```
mhasself, line 683: Once again, you don't need `.copy()`, because you're multiplying by 2.

However ... consider putting that factor of 2 into phasor, to begin with.  Then just use the same `[:]` trick as for dsT.

sotodlib/hwp/hwp.py
```diff
@@ -665,20 +665,28 @@ def demod_tod(aman, signal=None, demod_mode=4,
...
-        
+
+    # Create a RFFT object to reuse
+    n = fft_ops.find_superior_integer(aman.samps.count)
+    rfft = fft_ops.RFFTObj.for_shape(aman.dets.count, n, 'BOTH')
+
     phasor = np.exp(demod_mode * 1.j * aman.hwp_angle)
     demod = tod_ops.fourier_filter(aman, bpf, detrend=None,
-                                   signal_name=signal_name) * phasor
+                                   signal_name=signal_name, rfft=rfft).copy() * phasor
```
mhasself, line 675: You don't need the `.copy()`, because you're multiplying by phasor.

mhasself (commented):
Yay re-using buffers.

mhasself:
> Wanted to ask a question about preferences here.
> 
> In this new approach, for `dsT`, `demodQ`, and `demodU`, this preserves the incoming `dtype` of `signal` for outputs, which is kind of pointless since the FFT demotes things to `float32` anyway. The test fails since:
> 
> ` _ = coords.demod.from_map(tod, signal_map, modulated=True, wrap=True)`
> 
> outputs `float64` for `signal`, but:
> 
> `results = coords.demod.make_map(tod)`
> 
> wants a `float32`. I was trying to avoid doing `astype` when populating `demod_aman` with `dsT`, `demodQ`, and `demodU` since that incurs a copy of `aman[signal_name]`, `demod.real`, and `demod.imag`. I guess we might need to do `.copy()` for the `tod_ops.fourier_filter` commands or is there a cleaner option that anyone can see?

Seems like `from_map` is to blame here, as float32 is our standard signal type. Perhaps we just fix the test without changing the filtering code.

This seems like something we should warn about though; downgrading to float32 in the filtering could lead to some surprises (in signal only situations mostly).

mhasself (approved):
Looks good to me, thanks!

PR #1297, opened 2025-07-16
Bug fix in get_binned_hwpss

17-sugiyama:
I suggest changing the indent from [line 261](https://github.com/simonsobs/sotodlib/blob/780191b4363bf42ca4bd29c38404b13157618d5b/sotodlib/hwp/hwp.py#L261) to line 265 in `hwp.py`. 
This `else` block should be related to the `if apodize_edges:` block, otherwise `weight_for_signal` will be never assigned.
This change is trivial since almost no one uses `get_hwpss` without the apodization, but I would be happy if someone can review this pull request.


ykyohei (approved):
This looks good! I agree that this has no impact usually.

PR #1298, opened 2025-07-16
Check if frequency cutoffs are in full aman

mmccrackan:
When loading, `update_full_aman` is not called, so it will not contain `frequency_cutoffs` since it is added later.  Adds a check to the `frequency_cutoffs` move to make sure it won't fail.


chervias (approved):
I tested it and seems to be working

PR #1305, opened 2025-07-18
Generate_reports updates

mmccrackan:
Fix to support `numpy` v2 when querying apex pwv data.  Adds per-band yields and a per-platform navigation dropdown to make it easier to go between quality reports.

Doesn't add any new dependencies.

Examples here: https://site.simonsobs.org/~mmccrackan/nav_test/

### DIFF
--- a/sotodlib/qa/quality_reports/report_data.py
+++ b/sotodlib/qa/quality_reports/report_data.py
@@ -97,7 +97,7 @@ class ObsInfo:
     obs_tube_slot: str
     obs_tags: str = ""
     pwv: float = np.nan
-    num_valid_dets: float = np.nan
+    num_valid_dets: str = ""
 
     @classmethod
     def from_obsdb_entry(cls, data) -> "ObsInfo":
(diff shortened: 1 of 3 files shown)


sotodlib/qa/quality_reports/report_data.py
```diff
@@ -229,6 +221,30 @@ def load_pwv(cfg: ReportDataConfig) -> hkdb.HkResult:
...
+    Load the pwv from either the CLASS or APEX radiometer.
+    Merge both datasets when available.
+    """
+    result = load_pwv(cfg)
+    result_apex = get_apex_data(cfg)
+    if result_apex is not None:
+        result_apex = (np.array(result_apex['timestamps']), np.array(0.03+0.84 * result_apex['pwv']))
+
+    if hasattr(result, 'pwv'):
+        combined_times = np.concatenate((result.pwv[0], result_apex[0]))
```
guanyilun, line 240: result_apex can be None here?

guanyilun (approved):
looks good to me except a potential edge case

sotodlib/qa/quality_reports/report_data.py
mmccrackan: Yeah, the logic wasn't quite right there.  Thanks for pointing that out!

PR #1310, opened 2025-07-28
rotate focal plane

ykyohei:
I find this function helpful to study various statistics in ground (az/el) frame before map making.
If we run `rotate_focal_plane` and `rotate_demodQU`, we can get the focal plane and tod in ground frame.


mhasself (approved):
Nicely done!

ykyohei:
I pushed preprocess module that was forgotten to push. This is tested. I'm going to merge this later today.

PR #1314, opened 2025-07-31
Removing the polarization angle dependency in the h_n maps computation

Magwos:
In the same spirit as what it done here https://github.com/hpc4cmb/toast/blob/8cb36a922508d268554bde32b3ce7416d5efe6fe/src/toast/ops/derivatives_weights.py#L205 for the TOD template approach, I propose that the $h_n$ maps are generated without the polarization angle. Instead, the polarization angle could be added by the user instead. 

It would be more convenient in the pixel-based mapmaking processes as well.

If you need more details on the equations behind, I would be happy to provide them!

### DIFF
--- a/sotodlib/toast/ops/h_n.py
+++ b/sotodlib/toast/ops/h_n.py
@@ -87,6 +87,11 @@ class Hn(Operator):
         False, help="If True, do not clear detector pointing matrices after use"
     )
 
+    include_pol_angle = Bool(
+        False,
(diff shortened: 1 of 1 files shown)


keskitalo (changes requested):
This seems like a reasonable change.  I have one ask: please add a boolean trait, `include_pol_angle`, that controls the subtraction of the polarization angle. It can default to `False`.

keskitalo (approved):
Great work! Feel free to merge.

PR #1315, opened 2025-08-01
When loading from a context, cut detectors with NaNs.

tskisner:
Some preprocessing steps may introduce NaN values in some detector timestreams without cutting those detectors.  This change checks for such cases and flags these detectors after logging a warning.


AERAdler (approved):
Thank you, this really butters my toast.

PR #1316, opened 2025-08-05
Add qu rotation function

17-sugiyama:
This PR adds an option to `coords.demod.rotate_demodQU`.
If radial is True, rotate_demodQU converts the telescope coordinate into a radial coordinate, which is suitable to investigate the radial or tangential polarization on the focal plane. 
The detailed descriptions are here.
https://simonsobs.atlassian.net/wiki/spaces/PRO/pages/1183121436/Noise+investigation+in+Qr+Ur+flame

This PR also includes a minor change in `demod.rotate_focal_plane`.
Following the rotation of the focal plane, hwp angle will be also rotated.
This allows the demodulation after the focal plane rotation.

### DIFF
--- a/sotodlib/coords/demod.py
+++ b/sotodlib/coords/demod.py
@@ -204,15 +204,14 @@ def rotate_demodQU(tod, sign=1, offset=0, radial=False, update_focal_plane=True)
     Apply detectors' polarization angle calibration to the HWP demodulated Q and U timestreams to
     place all detectors' Q and U timestreams in a common telescope frame. This updates tod.demodQ
     and tod.demodU, in place.
-    To get Qr Ur timestreams, run `rotate_demodQU(tod, radial=True)`.
+    To get Qr Ur timestreams, run `rotate_demodQU(tod)` and then `rotate_demodQU(tod, radial=True)`.
     To restore the Q U timestreams, run `rotate_demodQU(tod, sign=-1, radial=True)`.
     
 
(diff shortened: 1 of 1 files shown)


ykyohei (changes requested):
Thank you for showing demonstrations. Looks mostly good. I have two requests.

sotodlib/coords/demod.py
```diff
@@ -214,9 +220,17 @@ def rotate_demodQU(tod, sign=1, offset=0, update_focal_plane=True):
...
         sign : int, optional
             A sign factor to control the direction of the rotation (default is +1).
+        radial : bool, optional
+            If True and the Q U timestreams in tod are in a common telescope flame, this function turns Q U
+            into Qr Ur. 
+            
 
     """
-    demodC = ((tod.demodQ + 1j*tod.demodU).T * np.exp( sign*(-2j*tod.focal_plane.gamma + 1j*np.deg2rad(offset)) )).T
+    if radial:
```
ykyohei, line 229: Please add gamma rotation for when both radial and update_focal_plane are True

ykyohei (approved):
Looks good to me. thank you for the changes!
I tested to make map map with Qr Ur coordinate, and it agreed with map without rotation to radial coordinate within the precision of float32 ~ 1e-7. I think this level of difference is expected because we use float32 for signal and focal plane.
Please squash and merge.

PR #1317, opened 2025-08-05
update_det_cal: refactor multiprocessing logic to use concurrent.futures

tdalford:
- Replace multiprocessing.Pool with executor objects
- Switched from apply_async + callback model to submit + as_completed
- Avoid passing non-pickleable axismanager to subprocesses

I believe these updates will fix the hanging update_det_cal flows on prefect. Separate note that [sodetlib PR#468](https://github.com/simonsobs/sodetlib/pull/468) which similarly updates the multiprocessing code inside `tes_param_correction.py` is required to be merged as well to make these changes work.

### DIFF
--- a/sotodlib/site_pipeline/update_det_cal.py
+++ b/sotodlib/site_pipeline/update_det_cal.py
@@ -304,8 +304,6 @@ class ObsInfo:
     ------------
     obs_id: str
         Obs id.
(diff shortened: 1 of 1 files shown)


guanyilun (approved):
Looks good to me!

PR #1332, opened 2025-08-14
Fix error of hwp angle sign

ykyohei:
Apply special correction to hwp angle sign.
`apply_hwp_angle_model(aman, on_sign_ambiguous='fail')` had `ValueError: hwp rotation direction is ambiguous` in following observations, This PR will fix it

The first pid sign error was casued by bug of hwp-pid agent, we fixed the agent to not cause same problam anymore but we haven't applied fix to data. https://github.com/simonsobs/chwp-discussions/discussions/24

The second offcenter sign error is  because satp1 SCR6 has unusual offcentering of hwp. This estimation method depends on the sign of hwp offcentering and this had same sign in all the platforms and cooldowns so far. But this is different in satp1 SCR6 and we need to correct it.

### DIFF
--- a/sotodlib/hwp/__init__.py
+++ b/sotodlib/hwp/__init__.py
@@ -7,6 +7,7 @@ This module contains code for HWP analysis tools
 """
 from .g3thwp import G3tHWP
 from .hwp import (get_hwpss, subtract_hwpss, demod_tod, get_hwp_freq)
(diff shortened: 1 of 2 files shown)


ykyohei:
We handle this correction in hwp_angle_model.
https://github.com/simonsobs/pipeline-metadata-scripts/pull/20

PR #1339, opened 2025-08-20
update_obsdb: fix how list of bad books is maintained

mhasself:
In case of failure, this was erroneously adding the last scanned book to the list instead of the failed one.


AERAdler:
Thanks for catching this embarrassing. error!

PR #1342, opened 2025-08-20
Creating requirements.txt

iparask:
Creating a `requirements.txt` and adding `sodetlib` as it is not available through PyPI.


BrianJKoopman (changes requested):
We should update the installation documentation at the same time.

Issue #1325, opened 2025-08-11
Add returns for process and calc_and_save


mmccrackan:
It would be useful if the `process` and `calc_and_save` functions for preprocessing could return both `aman` and `proc_aman`.  An example use case for this would be downsampling, where we want to completely replace `aman` and `proc_aman` with downsampled versions.  Currently neither function returns so it is hard to do this "in-place."  In this case we can just then do `aman, proc_aman = process(aman, proc_aman, sim)`.

mmccrackan:
Completed in #1345.

PR #1345, opened 2025-08-21
Make preprocess functions return axis managers

mmccrackan:
This is intended to address #1325.  Updates `process` and `calc_and_save` to return both `aman` and `proc_aman`.  This shouldn't occur any significant additional computation time or memory based on monitoring both while running with and without this change.


msilvafe (approved):
Looks good, please update the wiki page to reflect this new behavior: https://simonsobs.atlassian.net/wiki/spaces/PRO/pages/305627401/TOD+Processing+Infrastructure+Introduction#Example-Usages%3A-Running-a-Pipeline-Interactively

Issue #1323, opened 2025-08-11
Update full_aman handling in Pipeline run function


mmccrackan:
Currently, the `full` axis manager, which is the axis manager that stores the contents of `proc_aman` but without any `dets` or `samps` cuts is created inside of the `run` function of the `Pipeline` class, using whatever the `dets` and `samps` of `aman` are (it will be a copy of `aman.preprocess` if that exists).  This is not ideal for the `multilayer_preprocess_tod` function since it means that function always starts with a restricted number of detectors (`aman` gets restricted in `select`).  The `full` axis manager should instead be optionally passable into the `run` function so that we can preserve the original dimensions for each axis.

mmccrackan:
This was completed in #1528.

PR #1346, opened 2025-08-21
Allow pipeline to take input full axis manager

mmccrackan:
Intended to address #1323.  Updates the `Pipeline` class in `pcore` to accept `full` as an input.  Defaults to None such that the previous behavior is maintained.  This allows for multilayer pipelines to preserve the original `detector` and `samps` counts.  The loading function `multilayer_load_and_preprocess` now loops through both of the `init` and `proc` pipeline restrictions now to propagate all of the cuts.

Also includes one small fix for the `query` input for `preprocess_tod` and `multilayer_preprocess_tod`.  The default was originally `None`, but this doesn't work for the file checking line here: https://github.com/simonsobs/sotodlib/blob/9754e46b5f702c2ad746dcc361d534758665ad02/sotodlib/site_pipeline/util.py#L379 so it has been replaced with an empty string.

### DIFF
--- a/sotodlib/preprocess/preprocess_util.py
+++ b/sotodlib/preprocess/preprocess_util.py
@@ -375,6 +375,7 @@ def load_and_preprocess(obs_id, configs, context=None, dets=None, meta=None,
         full_aman = meta.preprocess.copy()
     else:
         full_aman = None
+    logger.info("Restricting detectors on all pipeline processes")
     if (
         'valid_data' in meta.preprocess and
         isinstance(meta.preprocess.valid_data, core.AxisManager)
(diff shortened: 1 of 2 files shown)


sotodlib/preprocess/preprocess_util.py
```diff
@@ -385,12 +394,13 @@ def load_and_preprocess(obs_id, configs, context=None, dets=None, meta=None,
         pipe = Pipeline(configs["process_pipe"], logger=logger)
         aman = context.get_obs(meta, no_signal=no_signal)
         pipe.run(aman, aman.preprocess, select=False)
-        return aman
+        return aman, full_aman
```
msilvafe: Did you look into if making this function now return two outputs will break anyone's existing workflows? And same question applies for `multilayer_load_and_preprocess`

msilvafe (commented):
A couple inline comments to respond to and want to know how you tested this before approving.

PR #1365, opened 2025-09-02
Pypi release

iparask:
Our Pypi release did not run because as it seems when we use `GITHUB_TOKEN` in a workflow any repository changes the workflow makes do not trigger a new workflow run.

This PR makes a change to the Pypi release to allow for a "workflow" based trigger. It checks the events and when the monthly release is run successfully, it gets triggered and releases to pypi.


tskisner (approved):
Looks good!

PR #1369, opened 2025-09-09
update_obsdb: remove auto-update of bad_books_list

mhasself:
This reflects how we want to run, in practice, which is to fail until the bad_books file has been manually updated.

Also adds --limit option, and speeds up directory scanning by using glob instead of walk.


AERAdler (approved):
Is there a prefect alert that tells us when update_obsdb has failed?

Issue #1307, opened 2025-07-25
Tracking: Map Tracking Database


JBorrow:
This is a tracking issue for the Depth 1 and Map Tracking database, informed by the O1 session of the F2F.

For ensuring we can access D1 maps, we need a database that lists the location and creation time of those maps.

For the time domain pipeline, we need a database that connects D1 map entries to their processing status through the time domain pipeline.

For possible pointing offsets, we need a database that connects individual D1 maps to pointing offset measurements.

It would be helpful if...

- We maintained a list of all TODs that were used to create a D1 map
- A record of pipeline information (software and metadata versions) that were used to create a D1 map
- A 'sky coverage' patch for each D1 map (see below)
- We could track information about map quality and 'comments' from visual inspection
- This database was accessible at all computing centers

Map Information
---------------

We should have a table, `depth_one_maps` that contains the following information:

```
map_name -- Unique name (UUIDv7?) for map
map_path -- Non-localized path; use e.g. `DEPTH_ONE_MAPS_PATH=/so/data/maps:/so/data/depth1`
optics_tube
wafer
frequency
ctime
... -- Other map-specific metadata
```

Time Domain
-----------

We should have a table, `time_domain_processing` that contains the following information:

```
map_name -- Foreign key into `depth_one_maps`
processing_start
processing_status
processing_end
... -- Additional metadata for tracking, TBD
```

Pointing Offsets
----------------

We should have a table, `depth_one_pointing_offsets` that contains the following information:

```
map_name -- Foreign key into `depth_one_maps`
ra_offset
dec_offset
....
```

Time-Ordered Data
-----------------

We should have a table, `tod_depth_one` that contains the following information:

```
obs_id
map_name -- Foreign key into `depth_one_maps`
... -- Any TOD-specific metadata we need to remember
```

Pipeline Information
--------------------

We should have a table, `pipeline_information`, that contains the following information:

```
map_name -- Foreign key into `depth_one_maps`
sotodlib_version
map_maker
preprocess_info
...
```

Sky Coverage
------------

I propose that we split the sky into a 2N x N grid of 'sub-patches' such that we can label each D1 map with the patches that it covers. This will be an important optimization when we come back to the D1 maps to recreate light-curves. Each patch can be referred to by a unique identifier.

As such, we should have a table, `depth_one_sky_coverage`, that contains the following information.

```
patch_coverage -- Some integer or string for each patch
map_name -- Foreign key into `depth_one_maps`
...
```

[...]

PR #1370, opened 2025-09-09
Map tracking

Sulla2012:
This PR addresses the tracking issue identified in [Issue 1307](https://github.com/simonsobs/sotodlib/issues/1307). The PR implements all the tables specified in that issue, as well as some basic methods for accessing them.


ahincks:
For the `depth_one_maps` table, I think there should be a start and end C-time, not a single C-time. That will make searching more accurate/powerful and I believe the start and end times should be pretty well defined.

PR #1371, opened 2025-09-11
Add a valid ranges matrix to proc_aman

mmccrackan:
Adds a `dets` x `samps` `valid` `RangesMatrix` at the end of the preprocess pipeline that keeps track of the final number of valid detectors.  Some changes were made to the `load_and_preprocess` functions to accommodate and use this when available.  Tested running through first and second layers, loading only first layer and both first and second.  Loaded existing db from before this change to check for backwards compatibility.  Confirmed same det vals appear in all cases.


mmccrackan:
Just a note on the last commit, but it was just removing an unnecessary check on `fld_init`, so will merge.

PR #1374, opened 2025-09-16
Tau hwp

ykyohei:
Add functions for time constant calibration by hwp spin up down measurements.


sotodlib/hwp/hwp.py
```diff
@@ -695,3 +696,218 @@ def demod_tod(aman, signal=None, demod_mode=4,
         del demod_aman
     else:
         return demod_aman['dsT'], demod_aman['demodQ'], demod_aman['demodU']
+
+
+def tau_model(fhwp, tau, AQ, AU, mode):
+    return (AQ + 1j*AU) * np.exp(1j * mode * 2 * np.pi * tau * fhwp)
```
17-sugiyama, line 702: I have checked the linear fit of tau_eff and tau_hwp, and found that the data points with tau_eff > 6ms are slightly off from the unity.
image

This might be resulted from the approximation of arctangent. 
Let me try `(AQ + 1j*AU) * np.exp( np.arctan(1j * mode * 2 * np.pi * tau * fhwp) )` and see how it changes the result.
image

17-sugiyama (changes requested):

Issue #1358, opened 2025-08-28
coords.helpers.get_footprint not graceful about branch cut


mhasself:
If the data in the obs straddles lon=180, the returned geometry will be full-sky.

The work-around is to tweak wcs_kernel, in advance, to set the branch cut.

But that should be documented more carefully, and it would be nice if get_footprint had a mode that tried to automatically shift the branch cut to a reasonable place.

PR #1379, opened 2025-09-17
Fix angle wrapping in get_footprint

amaurea:
Avoids huge maps when patch straddles geometry limits, and the geometry is being automatically built.

### DIFF
--- a/sotodlib/coords/helpers.py
+++ b/sotodlib/coords/helpers.py
@@ -355,6 +355,10 @@ def get_footprint(tod, wcs_kernel, dets=None, timestamps=None, boresight=None,
         # Works whether rot is a quat or a vector of them.
         asm.Q = rot * asm.Q
     proj.get_planar(asm, output=planar)
+    # planar is now [ndet,nsamp,{ira,idec}] in intermediate
+    # coordinates in radians. These will be 0 at the
+    # reference point, and span at most [-pi,pi] in ira
+    # and [-pi/2,pi/2] in idec
 
     # Sometimes a patch will end up straddling the edege of the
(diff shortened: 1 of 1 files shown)


sotodlib/coords/helpers.py
```diff
@@ -356,6 +356,13 @@ def get_footprint(tod, wcs_kernel, dets=None, timestamps=None, boresight=None,
         asm.Q = rot * asm.Q
     proj.get_planar(asm, output=planar)
 
+    # Unwind each detector in ra to avoid angle jumps
+    planar[:,:,0] = utils.unwind(planar[:,:,0])
+    # Harmonize them. This assumes that the detectors won't
+    # be more than 180° away from each other at any given time
+    offs = (planar[:,0,0]-planar[0,0,0]+np.pi)//(2*np.pi)*(2*np.pi)
```
mhasself, line 363: You added a pi here ... sure it doesn't need to be subtracted again?

mhasself (changes requested):
Thanks.  The adjustment makes sense for cylindrical projections but less obvious that it's well-behaved for zenithal.  I propose either (a) making this optional (with keyword switch `compactify=True`) or (b) only applying this when not a zenithal proj.

Also one sanity check inline.

And a test would be appreciated.

I note this will resolve https://github.com/simonsobs/sotodlib/issues/1358

sotodlib/coords/helpers.py
amaurea: `wcsutils.is_separable` should work. It captures exactly the case where you have a non-oblique cylindrical projection.

mhasself (approved):

PR #1380, opened 2025-09-17
TOAST mlmapmaker: Raise error when there are too few observations

Bai-Chiang:
If there are more TOAST groups than number of observation, some TOAST groups would have no observations in it. And this will cause error. This add a check make sure all TOAST groups have at least one observation.


keskitalo:
Please note the unit test failure.  You need to skip collective operations when there is no communicator available. This happens when running serially without MPI.

Bai-Chiang:
The error message was removed in https://github.com/hpc4cmb/toast/pull/846. After talking with @tskisner , that PR fixed case when there's no detector allocate to the process in TOAST group, but the TOAST group have observations. And rely on operator handle the case when there's no observation in TOAST group.

For the mlmapmaker, if there's no observation, it won't run `mapmaker.add_obs` [1]. And when preparing mapmaking [2] it will complain can't concatenate empty list for SignalCut [3].

[1] https://github.com/simonsobs/sotodlib/blob/355411f9812fefb9ef361e1c0365fb67fae9f7fe/sotodlib/toast/ops/mlmapmaker.py#L705-L711
[2] https://github.com/simonsobs/sotodlib/blob/355411f9812fefb9ef361e1c0365fb67fae9f7fe/sotodlib/toast/ops/mlmapmaker.py#L488
[3] https://github.com/simonsobs/sotodlib/blob/355411f9812fefb9ef361e1c0365fb67fae9f7fe/sotodlib/mapmaking/ml_mapmaker.py#L542

PR #1382, opened 2025-09-18
Update preprocessing Fourier filter docstring

mmccrackan:
Docstring for the Fourier filter in preprocessing had some typos and didn't apply to the signal which caused some confusion when I was using it in the site-pipeline-configs preprocessing configurations.

### DIFF
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -1663,6 +1663,18 @@ class FourierFilter(_Preprocess):
               timeconst: "det_cal.tau_eff"
               invert: True
 
+    Example for passing in a different signal name and wrapping into a new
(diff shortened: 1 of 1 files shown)


sotodlib/preprocess/processes.py
```diff
@@ -1667,9 +1665,7 @@ class FourierFilter(_Preprocess):
 
     Example config file entry for two filters::
 
-        - name: "fourier_filter_chain"
-          wrap_name: "lpf_sig"
-          signal_name: "signal"
```
msilvafe, line 1672: Give one example that passes `signal_name: "demodQ"` and one that passes a wrap name other than the default.

msilvafe (approved):

PR #1386, opened 2025-09-22
Fix edge cases of hwp offcenter correction

ykyohei:
This PR fixes the bug of offcentering correction.
offcentering correction is done by taking average of two encoder data. Previous method was not making sure that two encoder data are aligned in time domain. In some observations (~20), previous correction was taking average of different rotations and leading to incorrect ver3 hwp angle. This was not introducing major issues in observations with constant speed, but this was introducing non-negligible bias in spin up observations.
Details: https://simonsobs.atlassian.net/wiki/spaces/PRO/pages/1248657450/Tau+HWP+debug#250922-validation-of-offcenter-fix

Since the number of observations affected by this issue is small, I plan not to update metadata version. I plan to re-run the solution for those observations.

### DIFF
--- a/sotodlib/hwp/g3thwp.py
+++ b/sotodlib/hwp/g3thwp.py
@@ -644,37 +644,44 @@ class G3tHWP():
         logger.info('Remove offcentering effect from hwp angle and overwrite')
         # Find the first point where two encoder data are available
         # and offcentering is calculable
-        for i in solved["ref_indexes_1"]:
-            if solved['fast_time_1'][i] > solved['fast_time_2'][0]:
-                offcenter_idx1_start = i
-                t1_start = solved['fast_time_1'][i]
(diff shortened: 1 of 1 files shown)


17-sugiyama:
I crosschecked the new angle solution pipeline.
I calculated the hwp angle for the spin up/down measurement and estimated the detector time constant with new angle solution.
This plot shows the difference between tau_eff (biasstep time constant) and tau_hwp (time constant measured by hwp spin up).
The 'v2' and 'v3' are solutions before and after the offcenter correction.
With older hwp angle solution, tau_hwp tends to have ~1 ms offset because of the defect that Kyohei pointed out.
New angle solution does not have this offset.
Thanks to the proper offcenter correction, the uncertainty of the tau_hwp estimation is reduced by ~10%.
image

PR #1390, opened 2025-09-23
ObsFileDb: update diff routine for memory efficiency

mhasself:
With this implementation, peak memory should be driven by the size of the difference between two databases, not the size of the dbs.

In a test run on LAT obsfiledb this reduces peak memory usage from 5GB to 100MB.


tskisner (approved):
Thanks for the work on this!

PR #1393, opened 2025-09-24
Update record_qa for preprocess_lite

mmccrackan:
Updates `record_qa` and related functions to be able to parse preprocess_lite databases which do not have `wafer_info` and instead use `det_cal` bandpasses.

Also adds the ability to add a field name suffix to the end of the noise metric names so we can get out T, Q, and U.  Changes default noise units factor to 1 since we are interested in pW noise values.

### DIFF
--- a/sotodlib/qa/metrics.py
+++ b/sotodlib/qa/metrics.py
@@ -199,17 +199,25 @@ class PreprocessValidDets(PreprocessQA):
     def _process(self, meta):
 
         # add specified tags
-        tag_keys = ["wafer_slot", "tel_tube", "wafer.bandpass"]
+        tag_keys = ["wafer_slot", "tel_tube"]
+
+        if _has_tag(meta.det_info, 'wafer.bandpass'):
+            bandpasses = meta.det_info.wafer.bandpass
+            tag_keys += ["wafer.bandpass"]
+        else:
+            bandpasses = meta.det_cal.bandpass
+            tag_keys += ["det_cal.bandpass"]
+
         tag_keys += [t for t in self._tags if t not in tag_keys]
 
         # record one metric per wafer slot, per bandpass
         # extract these tags for the metric
         tags = []
         vals = []
-        for bp in np.unique(meta.det_cal.bandpass):
+        for bp in np.unique(bandpasses):
             for ws in np.unique(meta.det_info.wafer_slot):
                 subset = np.where(
-                    (meta.det_info.wafer_slot == ws) & (meta.det_cal.bandpass == bp)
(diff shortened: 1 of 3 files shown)


sotodlib/qa/metrics.py
```diff
@@ -368,7 +388,7 @@ def _process(self, meta):
 
         obs_time = [meta.obs_info.timestamp] * len(tags)
         return {
-            "field": self._influx_field,
+            "field": f"{self._influx_field}{'_' + self._field_name.replace("_", "", 1) else ''}",
```
tristpinsm, line 391: Was this string formatting tested? I suspect it won't work. Maybe rather like this?
```suggestion
            "field": f"{self._influx_field}{'_' + self._field_name.replace('_', '', 1) if self._field_name else ''}",
```
Also, is there a reason to restrict the number of `'_'` to remove to one?

Oh now I see this is how it is above.

tristpinsm (commented):
A couple questions:

1. Is `det_info` available in the regular `preprocess` metadata? If so we could just move to that rather than adding a check.
2. I may be missing something, but in the metrics where you are appending `_Q` etc to the field name, is it selecting that polarisation in the code somewhere? Or is that happening implicitly or in the config somehow?

mmccrackan:
> A couple questions:
> 
> 1. Is `det_info` available in the regular `preprocess` metadata? If so we could just move to that rather than adding a check.
> 2. I may be missing something, but in the metrics where you are appending `_Q` etc to the field name, is it selecting that polarisation in the code somewhere? Or is that happening implicitly or in the config somehow?

Yeah `det_info` is available but `wafer` is not included in `preprocess_lite` since `wafer_info` isn't included.  I think we can just point to `meta.det_cal.bandpass` since that is in both.

The T, Q, and U white noise values are stored in different axis managers and are selected by the `noise_aman` input in the metrics config.  Since those can technically be named anything, this just assumes the `field_name` will match what is supposed to be in them.  I suppose we could just take the `noise_aman` name instead of having a different `field_name`, so it would be something like `array_net_noiseT` but I think having the `influx` names be independent of what is specified in the preprocessing config might be better.

sotodlib/qa/metrics.py
mmccrackan: Yeah looks like a typo.

I was removing any added potential extra underscores just for cleanliness of the field names in the influx db, but will just remove that for clarity.

tristpinsm (changes requested):
Found a couple typos.

Is there any reason not to just use a single key, `bandpass` in both cases?

tristpinsm (approved):

PR #1404, opened 2025-10-06
ML Mapmaker fixes

skhrg:
Some minor fixes and QOL updates to the ML mapmaker.
These are mostly non-invasive but the quaternion change in the recentering code needs verification (will actually produce maps now but that could be true even if I messed it up).

### DIFF
--- a/sotodlib/mapmaking/utils.py
+++ b/sotodlib/mapmaking/utils.py
@@ -471,7 +471,7 @@ def evaluate_recentering(info, ctime, geom=None, site=None, weather="typical"):
         if sys == "cel" or sys == "equ":
             return lonlat
         elif sys == "hor":
-            return so3g.proj.CelestialSightLine.az_el(ctime, lonlat[0], lonlat[1], site=site, weather=weather).coords()[0,:2]
+            return so3g.proj.CelestialSightLine.az_el(ctime, -1*lonlat[0], np.pi/2 - lonlat[1], site=site, weather=weather).coords()[0,:2]
         else:
             raise NotImplementedError
     def get_pos(name, ctime, sys=None):
@@ -481,10 +481,16 @@ def evaluate_recentering(info, ctime, geom=None, site=None, weather="typical"):
             elif name == "auto":
                 return np.array([0,0]) # would use geom here
             else:
-                obj = getattr(ephem, name)()
-                djd = ctime/86400 + 40587.0 + 2400000.5 - 2415020
-                obj.compute(djd)
-                return np.array([obj.a_ra, obj.a_dec])
+                try:
+                    planet = coords.planets.SlowSource.for_named_source(
+                        name, ctime)
+                    ra0, dec0 = planet.pos(tod.timestamps.mean())
+                except:
+                    obj = getattr(ephem, name)()
+                    djd = ctime/86400 + 40587.0 + 2400000.5 - 2415020
+                    obj.compute(djd)
+                    ra0, dec0 = obj.a_ra, obj.a_dec
+                return np.array([ra0, dec0])
         else:
             return to_cel(name, sys, ctime, site, weather)
     p1 = get_pos(info["from"], ctime, info["from_sys"])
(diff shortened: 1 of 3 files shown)


sotodlib/mapmaking/utils.py
```diff
@@ -566,8 +577,11 @@ def rangemat_sum(rangemat):
         res[i] = np.sum(ra[:,1]-ra[:,0])
     return res
 
-def find_usable_detectors(obs, maxcut=0.1, glitch_flags: str = "flags.glitch_flags"):
-    ncut  = rangemat_sum(obs[glitch_flags])
+def find_usable_detectors(obs, maxcut=0.1, glitch_flags: str = "flags.glitch_flags", to_null : str = "flags.expected_flags"):
+    flag = obs[glitch_flags]
+    if to_null != "":
+        flag *= ~obs[to_null]
```
chervias: Won't this fail if the obs doesn't have `flags.expected_flags`? We either need to artificially add those empty flags before in the main script or leave the default of `to_null` to be an empty string

sotodlib/site_pipeline/make_ml_map.py
```diff
@@ -71,6 +62,7 @@ def main(**args):
     from sotodlib.core import metadata
     from pixell import enmap, utils, fft, bunch, wcsutils, mpi, bench
     import yaml
+    from mpi4py import MPI
```
chervias: why is this needed? mpi is loaded through pixell, which is also mpi4py

chervias (changes requested):

sotodlib/mapmaking/utils.py
```diff
@@ -481,10 +486,16 @@ def get_pos(name, ctime, sys=None):
...
-                djd = ctime/86400 + 40587.0 + 2400000.5 - 2415020
-                obj.compute(djd)
-                return np.array([obj.a_ra, obj.a_dec])
+                try:
+                    planet = coords.planets.SlowSource.for_named_source(
+                        name, ctime)
+                    ra0, dec0 = planet.pos(tod.timestamps.mean())
+                except:
+                    obj = getattr(ephem, name)()
+                    djd = ctime/86400 + 40587.0 + 2400000.5 - 2415020
```
iparask: What are all these numbers? It would be better if you include them in variables or constants that we can import

sotodlib/site_pipeline/make_ml_map.py
```diff
@@ -380,11 +393,24 @@ class DataMissing(Exception): pass
...
             sys.exit(1)
+           
+        if np.any(nkept_all == 0):
+            group = comm.Get_group()
+            new_group = group.Incl(np.where(nkept_all > 0)[0])
+            new_comm = comm.Create(new_group)
+            if nkept == 0:
+                L.info("No tods assigned to this process. Pruning")
+                sys.exit(0)
+            comm = new_comm
```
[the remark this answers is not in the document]
skhrg: Sounds good

sotodlib/mapmaking/utils.py
```diff
@@ -471,7 +477,7 @@ def to_cel(lonlat, sys, ctime=None, site=None, weather=None):
         if sys == "cel" or sys == "equ":
             return lonlat
         elif sys == "hor":
-            return so3g.proj.CelestialSightLine.az_el(ctime, lonlat[0], lonlat[1], site=site, weather=weather).coords()[0,:2]
+            return so3g.proj.CelestialSightLine.az_el(ctime, -1*lonlat[0], np.pi/2 - lonlat[1], site=site, weather=weather).coords()[0,:2]
```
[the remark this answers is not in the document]
skhrg: Hmm, I am trying to remember my original justification and I think it was just that

`for_horizon` (https://github.com/simonsobs/so3g/blob/master/python/proj/coords.py#L191) does this but `az_el` (https://github.com/simonsobs/so3g/blob/master/python/proj/coords.py#L128) does not. But looking more closely I do see that ultimately that end up calling https://github.com/arahlin/qpoint/blob/9189a303458f4970b6f9168da872e7f499bf2b90/src/qpoint.c#L214 which already applies pi/2 - el, so I have essentially just cancelled that out.

I think this best course of action is to remove this change for now so that the other QOL improvements in this PR can be merged and if I have time to dig deeper into the reentering I may put in another PR down the line.

Does that sound okay to you?

PR #1396, opened 2025-09-29
Add counter 1/f and common mode subtraction to preprocess

17-sugiyama:
This PR modifies the scripts to enable fancy counter 1/f filter and common mode subtraction.

The changes for counter 1/f filter:
The filter function 'counter_1_over_f' is modified to accept the array-like input of parameters.
This enables to apply independent 1/f parameters to individual detectors.

The changes for common mode subtraction:
The PCAFilter is modified to have `signal` and `model_signal` independently.
This enables to estimate PCA model from the down sampled signal, and then subtract it from the signal.
For the fancy operations, I introduced `get_common_noise_params` to estimate knee frequency of the common noise.
The estimated knee frequency is applied to the down sample frequency of `model_signal` .
However, this deployment requires many changes in `fft_ops`.
(As common mode aman doesn't have 'dets' LabelAxis, aman.dets is changed to aman[LabelAxis].)
These changes can be discarded if they are too much.

The following lines are the example of the process pipeline after the demodulation.
I have confirmed the processes run with this config.

```
process_pipe:
    # correct boresight rotation to zero
    - name : "rotate_focal_plane"
      process:
        hwp: True

    # telescope coordinate
    - name : "rotate_qu"
      process:
        sign: 1 
        offset: 0 
        update_focal_plane: True

    # radial coordinate
    - name : "rotate_qu"
      process:
        sign: 1 
        offset: 0 
        radial: True
        update_focal_plane: True

    # calculate psd
    - name: "psd"
      skip_on_sim: False
      signal: "demodQ"
      wrap: "psdQ"
      calc:
        max_samples: 524288 #2**19
        merge: False
        nperseg: 65536 #2**16
        noverlap: 0
        full_output: True
      save: True

    # fit psd
    - name: "noise"
      skip_on_sim: False
      psd: "psdQ"
      fit: True
      calc:
        fwhite: [1, 2]
        f_max: 2
        merge_name: 'noise_fit_statsQ'
      save:
        wrap_name: "noise_fitQ"

    # apply counter 1/f
    - name: "fourier_filter"
      skip_on_sim: False
      signal_name: "demodQ"
      wrap_name: "demodQ"
      process:
        filt_function: "counter_1_over_f"
        filter_params:
          noise_fit_stats: "noise_fit_statsQ"
          
    # get common mode and its fknee
    - name: "get_common_mode"
      noise_fit: True
      f_max: 2
      wrap_name: "demodQ_commonmode"
      calc:
        signal: "demodQ"
[...]

PR #1408, opened 2025-10-07
Junna preproc

msilvafe:
Here I've merged into master #1396 and #1381 from @17-sugiyama to be able to set up a preprocessing flow and test running through 100 observations using this new processing collaboratively.

### DIFF
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -375,7 +375,6 @@ class Jumps(_FracFlaggedMixIn, _Preprocess):
                              plot_ds_factor=self.plot_cfgs.get("plot_ds_factor", 50), filename=filename.replace('{name}', f'{ufm}_jump_signal_diff'))
             plot_flag_stats(aman, proc_aman[name], flag_type='jumps', filename=filename.replace('{name}', f'{ufm}_jumps_stats'))
 
-
 class PSDCalc(_Preprocess):
     """ Calculate the PSD of the data and add it to the Preprocessing AxisManager under the
     "psd" field.
@@ -453,7 +452,6 @@ class PSDCalc(_Preprocess):
 
             plot_psd(aman, signal=attrgetter(f"{self.wrap}.Pxx")(proc_aman),
                      xx=attrgetter(f"{self.wrap}.freqs")(proc_aman), filename=filename, **self.plot_cfgs)
-
 
 class GetStats(_Preprocess):
     """
@@ -1747,6 +1745,7 @@ class FourierFilter(_Preprocess):
                 noise_fit=noise_fit,
                 filter_params=spec.get("filter_params")
             )
+
             ffun = getattr(tod_ops.filters, fname)
             filters.append(ffun(**params))
 
@@ -1948,16 +1947,21 @@ class PCAFilter(_Preprocess):
 
     def __init__(self, step_cfgs):
         self.signal = step_cfgs.get('signal', 'signal')
+        self.model_signal = step_cfgs.get('model_signal', None)
 
         super().__init__(step_cfgs)
 
     def process(self, aman, proc_aman, sim=False):
         n_modes = self.process_cfgs.get('n_modes')
         signal = aman.get(self.signal)
+        if self.model_signal is not None:
+            model_signal = aman.get(self.model_signal)
+        else:
+            model_signal = signal.copy()
         if aman.dets.count < n_modes:
             raise ValueError(f'The number of pca modes {n_modes} is '
                              f'larger than the number of detectors {aman.dets.count}.')
-        model = tod_ops.pca.get_pca_model(aman, signal=signal, n_modes=n_modes)
+        model = tod_ops.pca.get_pca_model(aman, signal=model_signal, n_modes=n_modes)
         _ = tod_ops.pca.add_model(aman, model, signal=signal, scale=-1)
         return aman, proc_aman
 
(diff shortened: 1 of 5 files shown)


sotodlib/preprocess/processes.py
```diff
@@ -1954,16 +1954,21 @@ class PCAFilter(_Preprocess):
 
     def __init__(self, step_cfgs):
         self.signal = step_cfgs.get('signal', 'signal')
+        self.model_signal = step_cfgs.get('model_signal', None)
```
msilvafe: Why is model_signal added? What is this used for?

sotodlib/preprocess/processes.py
```diff
@@ -1974,28 +1979,45 @@ class GetCommonMode(_Preprocess):
...
         save: True
 
+    If ``noise_fit`` is True, the 1/f noise fit parameters of the common mode
+    is wrapped together.
     .. autofunction:: sotodlib.tod_ops.pca.get_common_mode
     """
     name = 'get_common_mode'
+    def __init__(self, step_cfgs):
+        self.noise_fit = step_cfgs.get('noise_fit', False)
+        self.f_max = step_cfgs.get('f_max', None)
```
msilvafe: These are only used in the `calc_and_save` step why do you need to define them globally?

msilvafe (commented):
Some comments which @17-sugiyama and I talked about over zoom.

17-sugiyama:
This is the comparison of the transfer function between sat-iso v3 and this process.
(reference:`/scratch/gpfs/SIMONSOBS/sat-iso/v3/spectra/satp1/south/full-dataset/science/transfer_functions/transfer_function_SATp1_f150_south_science_x_SATp1_f150_south_science.npz`)
image

sotodlib/preprocess/processes.py
```diff
@@ -846,6 +910,16 @@ def process(self, aman, proc_aman, sim=False):
             raise ValueError(f"Entry '{self.process_cfgs['kind']}'"
                               " not understood")
         return aman, proc_aman
+    
+    def select(self, meta, in_place=True):
```
mmccrackan: The `select` function definitions are usually `select(self, meta, proc_aman=None, in_place=True)`.  This also doesn't do the

```
if proc_aman is None:
            proc_aman = meta.preprocess
```

check that is usually done.

sotodlib/preprocess/processes.py
```diff
@@ -1236,6 +1310,21 @@ def process(self, aman, proc_aman, sim=False):
         else:
             tod_ops.azss.get_azss(aman, **self.calc_cfgs)
         return aman, proc_aman
+    
+    def select(self, meta, in_place=True):
```
mmccrackan: See above about `select` function definition.

sotodlib/preprocess/processes.py
```diff
@@ -1236,6 +1310,21 @@ def process(self, aman, proc_aman, sim=False):
...
+        if self.select_cfgs is None:
+            return meta
+        if 'bad_dets' in meta[self.azss_stats_name]:
+            keep = ~meta[self.azss_stats_name]['bad_dets']
+        else:
+            print('No bad_dets field in azss_stats, skipping det selection.')
```
mmccrackan: Preprocessing doesn't typically print when no cuts are applied, but if we want to do that here we should probably use the `preprocess` logger here (and throughout) since `print` statements can have buffering issues.

sotodlib/preprocess/processes.py
```diff
@@ -2108,28 +2207,45 @@ class GetCommonMode(_Preprocess):
     example config file entry::
 
       - name: "get_common_mode"
+        noise_fit: True
```
mmccrackan: These are all calc_and_save parameters right?  Should they be moved into the calc and save keys?

sotodlib/preprocess/processes.py
```diff
@@ -2523,6 +2659,35 @@ def process(self, aman, proc_aman, sim=False):
...
+        self.signal_name_T = step_cfgs.get('signal_name_T', 'dsT')
+        self.signal_name_Q = step_cfgs.get('signal_name_Q', 'demodQ')
+        self.signal_name_U = step_cfgs.get('signal_name_U', 'demodU')
+        super().__init__(step_cfgs)
+
+    def process(self, aman, proc_aman, sim=False):
+        scan_freq = tod_ops.utils.get_scan_freq(aman)
+        hpf_cfg = {'type': 'sine2', 'cutoff': scan_freq, 'trans_width': scan_freq/10}
+        filt = tod_ops.get_hpf(hpf_cfg)
+        aman[self.signal_name_T] = tod_ops.fourier_filter(aman, filt, signal_name=self.signal_name_T, detrend='linear')
```
mmccrackan: It is possible to re-use one fft object to save computation time.  See example in `demod_tod`:

https://github.com/simonsobs/sotodlib/blob/68de048c12e45c333117177f7d54f497837223f9/sotodlib/hwp/hwp.py#L672.

sotodlib/tod_ops/azss.py
```diff
@@ -265,7 +312,14 @@ def get_azss(aman, signal='signal', az=None, azrange=None, bins=100, flags=None,
...
     azss_stats.wrap('binned_signal_sigma', binned_signal_sigma, [(0, 'dets'), (1, 'bin_az_samps')])
     azss_stats.wrap('uniform_binned_signal_sigma', uniform_binned_signal_sigma, [(0, 'dets')])
-
+    if return_det_mask:
+        bad_dets, coverages = _check_azcoverage(aman, flags=flags, az=az, 
+                                                coverage_threshold=coverage_threshold,
+                                                exclude_turnarounds=exclude_turnarounds)
+        azss_stats.wrap('bad_dets', bad_dets, [(0, 'dets')])
+        azss_stats.wrap('az_coverage', coverages, [(0, 'dets')])
+    else:
```
mmccrackan: I don't think `bad_dets` is ever used unless wrapped into `azss_stats` so this `else` seems unnecessary.

sotodlib/tod_ops/fft_ops.py
```diff
@@ -842,86 +885,170 @@ def fit_noise_model(
...
+                    warnings.filterwarnings("error")
+                    try:
+                        Hfun = ndt.Hessian(
+                            lambda params: neglnlike(params, f, p, bin_size=bin_size, **_fixed),
+                            full_output=True,
+                        )
+                        hessian_ndt, _ = Hfun(res["x"])
+                        covout_i = np.linalg.inv(hessian_ndt)
+                    except np.linalg.LinAlgError:
+                        print(
```
mmccrackan: Probably should just use a logger or not output these at all since we better understand them now.

sotodlib/tod_ops/utils.py
```diff
@@ -71,3 +71,51 @@ def get_block_moment(
...
+        in_deg (bool): If True, this returns scan speed in [deg/s].
+
+        wrap (bool): If True, scan speed is wrapped in TOD.
+
+        wrap_name (str): The name to wrap the scan speed.
+    Returns:
+
+        scanspeed (float): The azimuth scan speed.
+    """
+    scanspeed = np.median(np.abs(np.diff(aman.boresight.az))/np.diff(aman.timestamps))
```
mmccrackan: Do we need to cut turnarounds?

mmccrackan (changes requested):
Just some initial comments and questions.

sotodlib/tod_ops/utils.py
msilvafe: I think this would be a good improvement but as is this is biasing the estimate to lower frequency which is probably good for the filtering. So let's just open a new issue to add an optional argument for turnaround masking to this and test it out for the future.

msilvafe: Added issue

msilvafe:
Ok @mmccrackan I think this is ready now

PR #1415, opened 2025-10-14
per tube mode for obs_grouping, to be used in the depth-1 mapmaker

chervias:


sotodlib/mapmaking/obs_grouping.py
```diff
@@ -27,7 +27,7 @@ def __init__(self, msg):
         self.msg = msg
 
 def build_obslists(context, query, mode=None, nset=None, wafer=None,
-        freq=None, ntod=None, tods=None, fixed_time=None, mindur=None, ):
+        freq=None, ntod=None, tods=None, fixed_time=None, mindur=None, per_tube=False ):
     """ 
     Return an obslists dictionary (described in the submodule docstring),
```
mhasself: Since the obslists dictionary is described in the submodule docstring, take this opportunity to explain, there, that the "detset" component is either a wafer_slot value, or a tube_slot value, depending on per_tube.

PR #1416, opened 2025-10-14
Coadd atomic maps site pipeline script

davidvng:
- New site pipeline script to make coadded atomic maps: `make_coadd_atomic_map.py`
- Related modules in `mapmaking`
- The current database reading and writing is temporary solution until #1370 is merged
- Tested on current SATP3 atomic maps database (which is being run automatically on prefect): `/so/site-pipeline/satp3/maps/satp3_atomic_250423m/atomic_db.sqlite`
- Similar to the QA reports, this script can be run for different time intervals (daily, weekly, monthly). Separate config file for each type of interval -> one prefect flow per interval type

Example config file:
```
platform: satp3
start_time: "2025-08-19T12:00:00+00:00"
output_root: /so/home/dnguyen/repos/scripts/maps/outputs/satp3/
interval: weekly
start_weekday: 6
atomic_db: /so/home/dnguyen/repos/scripts/maps/satp3_atomic_db.sqlite
output_db: /so/home/dnguyen/repos/scripts/maps/outputs/satp3/coadd_db.sqlite
bands: ["f090", "f150", "f090+f150"]
split_label: full
geom_file_prefix: /so/site-pipeline/satp3/maps/so_geometry_v20250306_sat
unit: K
overwrite: False
plot: True
```
Note: start time is at noon instead of midnight because we likely want to split daily coadded maps by grouping obs of the same night together rather than cutting them at midnight.

Example output of weekly 20250817_20250824 (coadded map and hits):
Screenshot from 2025-10-14 17-02-43
Screenshot from 2025-10-14 17-02-55

### DIFF
--- a/sotodlib/mapmaking/coadd_mapmaker.py
+++ b/sotodlib/mapmaking/coadd_mapmaker.py
@@ -6,16 +6,29 @@ this will coadd observations within some time interval. This is primarily meant
 to be used by the make_coadd_atomic_map site-pipeline script, which calls
 make_coadd_map.
 """
-__all__ = ['CoaddMapmaker','make_coadd_map','setup_coadd_map','write_coadd_map']
+__all__ = ['CoaddMapmaker','make_coadd_map','setup_coadd_map','write_coadd_map','eplot']
 import numpy as np
-from pixell import enmap
+from pixell import enmap, enplot, utils as putils
 import sqlite3
 import os
 from tqdm import tqdm
 import datetime as dt
 import logging
 
-from . import utils
+from . import safe_invert_div
+
+def eplot(mapdata, fname_prefix, **kwargs): 
+    """Function to help us plot the maps neatly
+    
+    Arguments
+    ---------
+    mapdata : pixell.enmap
+        Map data to plot
+    fname_prefix : str
+        Path prefix to write output plot. Pixell will add '.png' to the end.
+    """
+    plots = enplot.get_plots(mapdata, **kwargs)
+    enplot.write(fname_prefix, plots)
 
 class CoaddMapmaker:
     """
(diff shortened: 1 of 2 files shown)


sotodlib/site_pipeline/make_coadd_atomic_map.py
```diff
@@ -0,0 +1,193 @@
+from argparse import ArgumentParser
+import os
```
chervias: and then use the `__all__` to make only the functions you want visible to the outside

sotodlib/mapmaking/coadd_mapmaker.py
```diff
@@ -0,0 +1,144 @@
+import numpy as np
```
chervias: here at the beginning we need a docstring, similar to the other files in the mapmaking module

chervias (approved):

sotodlib/site_pipeline/make_coadd_atomic_map.py
```diff
@@ -0,0 +1,193 @@
...
+            stop_time FLOAT,
+            PRIMARY KEY (telescope, freq_channel, split_label, prefix_path, geom_file_path, obslist, start_time, stop_time)
+        )
+    '''
+    cur.execute(create_stmt)
+    conn.commit()
+    conn.close()
+
+
+def get_parser(parser: Optional[ArgumentParser] = None) -> ArgumentParser:
```
mmccrackan: This was suggested by Katie for the quality reports, but it might be useful here to expose a few more parameters that are more likely to be changed so that we can run this directly on Prefect.  Maybe at least `start_time` and `stop_time`.

Issue #1367, opened 2025-09-07
det_cal product needs to be split into multiple hdf5 files


mhasself:
The ManifestDb produced by update_det_cal refers into a single hdf5 file.  This leads to risk (HDF5 files are vulnerable to corruption during update, and that would invalidate the entire archive) and annoyance downstream (updates to the archive upset analysis in progress, and transfer interruptions can corrupt the archive copy temporarily).

PR #1417, opened 2025-10-16
Det cal hwpss

ykyohei:
Update det_cal to reanalyze biasstep with hwpss subtraction. We found hwpss subtraction is imporant especially when hwpss is large, i.e. wiregrid, SAT-UHF. This is backward compatible and doesn't modify lat det_cal.
We need discussion but we probably keep normal det_cal running, and we will have another det_cal with hwpss subtraction.

Added special handling for metadata loader error so that this will try again when hwp_angle metadata was not produced before the det_cal.
Added an option to split h5 file based of first N digits of unixtime. Adresses https://github.com/simonsobs/sotodlib/issues/1367

Note: One of the tests are failed at the moment but this is not due to this PR, hopefully fixes by itself. I will rerun later

### DIFF
--- a/sotodlib/site_pipeline/update_det_cal.py
+++ b/sotodlib/site_pipeline/update_det_cal.py
@@ -473,8 +473,8 @@ def biases_flags(bsa, buffer=200):
     Make flags that mask bias steps
 
     Args
-        bsa: BiasStepAnalysis object
-        buffer: buffet to apply on result flags
+        bsa: sodetlib BiasStepAnalysis object
+        buffer: Number of samples to buffer flags
     Returns
         RangesMatrix
     """
@@ -493,7 +493,7 @@ def load_and_reanalyze_bs(bsa, ctx, obs_id):
     Load raw data of biassteps and reanalyze it with hwpss subtraction
 
     Args
-        bsa: BiasStepAnalysis object
+        bsa: sodetlib BiasStepAnalysis object
         ctx: Context object
         obs_id: observation id of bias steps
     """


sotodlib/site_pipeline/update_det_cal.py
```diff
@@ -720,14 +785,23 @@ def handle_result(result: CalRessetResult, cfg: DetCalCfg) -> None:
...
         return
 
     logger.info(f"Adding obs_id {obs_id} to dataset")
     rset = ResultSet.from_friend(result.result_set)
-    write_dataset(rset, cfg.h5_path, obs_id, overwrite=True)
+    h5_path = cfg.h5_path
+    if cfg.h5_unix_digits:
+        name, ext = os.path.splitext(cfg.h5_path)
+        unixtime = obs_id.split('_')[1][:cfg.h5_unix_digits]
+        h5_path = f"{name}_{unixtime}{ext}"
```
msilvafe: Question, no edits required. This is unrelated to hwpss subtraction but you're adding it here to minimize the size of metadata archive files and to make it more in alignment with some of the other metadata formatting?

msilvafe (approved):
Looks good. One question -> please describe how you tested this.

And more of an implementation comment as opposed to specific issue with the modifications in this script but when we start running with the HWPSS subtraction mode turned on I suggest that we run this as a second flow and leave the current detcal unedited so that we generate to versions of the detcal metadata. This is important because detcal is basically the lowest level metadata and everything else relies on it (detmatch, preprocess, etc.) so adding an additional dependency of the HWP angle solution to the context file adds a possible failure to this essential low level metadata. We can then build preprocess_lite off of the standard detcal product but do our science analysis off of the hwpss subtracted detcal product. 

Tagging @ahincks and @mmccrackan as common keepers of the metadata on-site automated flows to just be aware of this change and the subtlety I mention here.

PR #1419, opened 2025-10-17
Add scripts that support SO jobs compatible with toast_run

tskisner:
Toast supports a more flexible way of running jobs where all steps are specified in config files as a sequence of nested Pipeline operators.  This work adds some new scripts for using this mode of running.

- For purely synthetic observing schedules, add a script which writes a simulated SO instrument model to disk in a compatible toast format.

- Add a so-toast-run entrypoint which can generate and run per observation and per wafer-observation tasks with toast_run. The output for each task is redirected to a log file in a per-task directory.  Add unit test for this script which runs a simulation

- Fix simulated focalplane roundtrip to disk- numpy scalars were being dumped instead of python float.

### DIFF
--- a/sotodlib/toast/scripts/so_coadd.py
+++ b/sotodlib/toast/scripts/so_coadd.py
@@ -124,6 +124,7 @@ def get_weights(args, obs):
         axs[-2].axvline(noise_med, color="green", label="Median Noise Magnitude")
         axs[-2].axvline(noise_cut, color="red", label="Cutoff for Coadd")
         axs[-2].legend(loc="best")
+        axs[-2].set_xlabel(r"Spectral Range Noise Magnitude")
 
         # Apply cuts and relative weights
         remove = noise > noise_cut
@@ -133,8 +134,8 @@ def get_weights(args, obs):
 
         # Plot weight values
         counts, bins = np.histogram(weights[keep], bins=50)
-        axs[-1].stairs(counts, bins, color="blue", label="Coadd Weights")
-        axs[-1].legend(loc="best")
+        axs[-1].stairs(counts, bins, color="blue")
+        axs[-1].set_xlabel(r"Extra Inverse Noise Weight (High / Low Ell)")
 
         # The weights expected by the coadd tool are inverse noise weights.
         # If the user does not want relative weighting, set to 1.0


tskisner:
Note that tests will fail until the next toast tag is available.

sotodlib/toast/scripts/so_coadd.py
```diff
@@ -0,0 +1,320 @@
...
+        files[oprops["index"]] = oprops["map"]
+        if oprops["cl"] is None:
+            continue
+        # Load spectra
+        spec = QTable.read(oprops["cl"], format="ascii.ecsv")
+
+        # Compute the inverse noise weights
+        ell_slc = slice(int(ell_min), int(ell_max), 1)
+        spc_mean = np.mean(spec["cl_BB"].to_value(u.uK**2)[ell_slc])
+        weights[oprops["index"]] = spc_mean
```
keskitalo: If I'm reading this correctly, you create and apply weights proportional to the noise variance, not inverse noise variance.

sotodlib/toast/sim_focalplane.py
```diff
@@ -1,7 +1,6 @@
 # Copyright (c) 2018-2023 Simons Observatory.
 # Full license can be found in the top level "LICENSE" file.
-"""Focalplane simulation tools.
-"""
+"""Focalplane simulation tools."""
```
keskitalo: Lots of formatting noise on this file.

tests/test_sim_hwpss.py
```diff
@@ -28,7 +28,7 @@
 from sotodlib.toast.ops import pos_to_chi
 
 
-class SimMuMUXCrosstalkTest(unittest.TestCase):
+class SimHWPSSTest(unittest.TestCase):
```
keskitalo: Good catch!

keskitalo (approved):
Looking forward to putting this to good use.

sotodlib/toast/scripts/so_coadd.py
tskisner: Thanks for looking at this.  The "weights" here are written to the file fed to the toast coadd tool, so they need to be consistent with what is expected there.  That [coadd tool expects noise variance values](https://github.com/hpc4cmb/toast/blob/c256f446eaee8add0db0e7008cb735cc4d7f6a32/src/toast/scripts/toast_healpix_coadd.py#L289-L294), but I agree that when we say "weights" we usually mean that something with higher weight will contribute more to the coadd (i.e. they are inverse noise variance weights).  How about:

1. I'll change the meaning here and also open a small PR on toast to invert the meaning of the values in the file.
2. I'll add a command line option to enable actually writing these weights (currently the outputs are fixed to 1.0).  I think we need to do more testing to see the impact of noise weighting in the coadd step, but having that enabled by a commandline switch (and not be the default) seems good.

tskisner: Summarizing offline discussion:

- The toast coadd script should just take an inverse noise weight which is applied directly to the noise weighted map when accumulating and to the inverse noise covariance when accumulating.  [This is implemented in a PR](https://github.com/hpc4cmb/toast/pull/892).
- The (optional, disabled by default) extra inverse noise weights created by the script in this PR should capture the poorly constrained (by the pixel noise covariance) low-ell noise in each wafer-observation.  It does this by computing the ratio of the (better constrained) high-ell noise to the low-ell noise in polarization.

PR #1420, opened 2025-10-17
TOAST mlmapmaker: Reduce memory when writing div

Bai-Chiang:
Reduce peak memory by writing the per-pixel covariance matrix div one component at a time. So instead of MPI send/receive 3 x 3 x npix matrix, only send/receive npix matrix.

### DIFF
--- a/sotodlib/toast/ops/mlmapmaker.py
+++ b/sotodlib/toast/ops/mlmapmaker.py
@@ -217,7 +217,12 @@ class MLMapmaker(Operator):
         help="Truncate TOD to an easily factorizable length to ensure efficient FFT.",
     )
 
-    write_div = Bool(True, help="Write out the noise weight map")
+    write_div = Unicode(
+        "all",
+        allow_none=True,
+        help="Components (must be 'T', 'QU', 'TQU', 'all', None, or '')"
+    )
+
     write_hits= Bool(True, help="Write out the hitcount map")
 
     write_rhs = Bool(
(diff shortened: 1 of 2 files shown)


Bai-Chiang:
I added `write_div_TT` toggle to write TT component, and made `write_div` to write full matrix (stated in doc). So `write_div_TT` only takes effect when `write_div` is False.

keskitalo:
This is fine. However, if I was designing it, I would probably make `write_div` a `Unicode` trait and allow it to be one of `I, QU, IQU`.

Bai-Chiang:
Ah yeah, that's better.

Bai-Chiang:
Is there a way to set default value of `write_div` to be the same as the `comps` [1]. That is if `comps` is set to `T` then by default `write_div` will also be `T`.

[1] https://github.com/simonsobs/sotodlib/blob/0f8fe217d22e3ee5fc62e11c16b3be096a3b3094/sotodlib/toast/ops/mlmapmaker.py#L89

keskitalo:
To achieve this, you would default `write_div` to `None` and replace it with `comps` in the `exec()` method. User could override the default behavior by supplying a string value for `write_div`. An empty string can be interpreted as "do not write".

keskitalo:
This is looking really good. The unit tests are failing because they still assume `write_div` is a boolean trait: https://github.com/simonsobs/sotodlib/blob/master/tests/test_mapmaker_pointing.py#L142

PR #1426, opened 2025-10-22
Missed a removal of the hwpss destriping template

tskisner:
This was missed in #1418 .


AERAdler (approved):
Tested and approved

tskisner:
The macos / python-3.12 failure is due to the fact that the `autobahn` package does not have wheels for python-3.12 on macos (only 3.13 and 3.14).  This autobahn dependency does not even show up on linux if I do `pipgrip --tree sotodlib`, so I'm not sure what package is pulling in that dependency on macos.  Merging regardless.

PR #1430, opened 2025-10-29
Restructure QA metrics log

davidvng:
Restructures QA metrics log to reduce cardinality. This should resolve the recent memory crashes of influxdb due to `record_qa` flows.

I copied over entries in `qa_metrics_log` into `qa_metrics_log_new` with the restructure (omitting SATP2 obs, since that seemed to initially break things). The new structure swaps field and tag values. In `qa_metrics_log`, setting `'observation'` as the tag blew up the cardinality due to many unique obs_ids. In `qa_metrics_log_new`, the obs_ids are field values of the field `'observation'` with `'metric_type'` as the tag. As a comparison, this reduces the cardinality from 27446 to 32.

I also add the capability to set the measurement name for the log, to allow setting `qa_metrics_log_new` in the configs. Once this is deployed and tested, we should delete the old measurement `qa_metrics_log` and replace with the new one.

### DIFF
--- a/sotodlib/qa/metrics.py
+++ b/sotodlib/qa/metrics.py
@@ -102,7 +102,7 @@ class QAMetric(object):
             raise Exception(f"Metadata does not correspond to obs_id {obs_id}.")
         line = self._process(meta)
         log_tags = {"metric_type": self._influx_field}  # used to identify this entry
-        self.monitor.record(**line, log=self._influx_log, measurement=self._influx_meas, log_tags=log_tags, obs_id=obs_id)
+        self.monitor.record(**line, log=self._influx_log, measurement=self._influx_meas, log_tags=log_tags, qa_metrics=True, obs_id=obs_id)
         self.monitor.write()
 
     def get_new_obs(self):
--- a/sotodlib/site_pipeline/monitor.py
+++ b/sotodlib/site_pipeline/monitor.py
@@ -194,7 +194,7 @@ class Monitor:
 
         return influxdata
 
-    def record(self, field, values, timestamps, tags, measurement, log='obs_process_log', log_tags=None, obs_id=''):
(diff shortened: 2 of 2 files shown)


tristpinsm (commented):
This seems like a sensible change. In principle the `Monitor` is a general-purpose thing that is not specifically for `record_qa`, although in practice it looks like it is sparsely used (`site_pipeline/update_book_plan.py` does at least). Have you checked that these changes will not break these other instances?

davidvng:
> This seems like a sensible change. In principle the `Monitor` is a general-purpose thing that is not specifically for `record_qa`, although in practice it looks like it is sparsely used (`site_pipeline/update_book_plan.py` does at least). Have you checked that these changes will not break these other instances?

@tristpinsm Thanks for pointing that out. I've added a boolean defaulted to False to split how Monitor writes to the influx log.

davidvng:
> I would be hesitant to introduce two ways of logging things here -- in particular since with the latest changes it looks like entries logged with `qa_metrics=False` would fail to be found by the `check` method, no? This would have to be updated as well as the docstrings. If that is the way forward you might want to make these more robust, and make the argument more generic (something like `meas_is_oid` instead of `qa_metrics`).
> 
> Others may want to chime in, but if we can confirm that there are few if any other uses of this code in the pipeline, and it does seem like this structure is more efficient, maybe we should just change the behaviour and fix anywhere it is being used?

`update_book_plan.py` for example, which you've pointed out, actually doesn't have the issue of high cardinality since it doesn't use obs_ids as tag values. However, it still writes a simple `1` value for fields such as `bound` in the log measurement. I also don't see any instances of it querying or checking the log. In the case of QA metrics, we want unique obs_ids as field values since we query it to check if the obs_id has been completed or not.

davidvng:
@tristpinsm @mmccrackan @BrianJKoopman If there's no pushback, I'd like to get this approved and merged so we can get `record_qa` back up and running.

Issue #1440, opened 2025-11-04
On the fly `load_and_preprocess`


skhrg:
It would be very useful (to me at least) if there was an option in `load_and_preprocess` to run the preprocessing on the fly if that observation isn't in the database already.

This would be useful for when we have fresh obs that we wan't to process immediately with the same set of steps as usual without having to wait for preprocess to run (or setting up a temporary database). Also would be useful for testing changes in a config file.

Happy to code this up but pinging @mmccrackan and @msilvafe in case they have thoughts.

msilvafe:
We have a function called preproc_or_load_group that's been used by the mapmaker. Would that work for you?

PR #1441, opened 2025-11-04
Allow preproc_or_load_group to save archive

mmccrackan:
Addresses point made in #1440.

Adds `save_archive` option (default `False`) to `preproc_or_load_group` which will automatically save the archive instead of having to call it separately.

Tested with simple preproc configs in single and multiple layers.

### DIFF
--- a/sotodlib/preprocess/preprocess_util.py
+++ b/sotodlib/preprocess/preprocess_util.py
@@ -913,7 +913,8 @@ def preproc_or_load_group(obs_id, configs_init, dets, configs_proc=None,
         Optional. Whether or not to overwrite existing entries in the preprocess manifest db.
     save_archive :
         Call cleanup_mandb if True to save to the archive and database files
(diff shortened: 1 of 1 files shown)


sotodlib/preprocess/preprocess_util.py
```diff
@@ -911,6 +911,9 @@ def preproc_or_load_group(obs_id, configs_init, dets, configs_proc=None,
         Optional. Logger object or None will generate a new one.
     overwrite: bool
         Optional. Whether or not to overwrite existing entries in the preprocess manifest db.
+    save_archive :
+        Call cleanup_mandb if True to save to the archive and database files
+        in configs_init and configs_proc
```
msilvafe: Add a statement here that says that this argument should not be used if this function is being called from within a multiproc or mpi script.

msilvafe (approved):
1 comment inline.

PR #1442, opened 2025-11-09
Add support for housekeeping in LoadContext operator

tskisner:
- Add traits to LoadContext to specify the data root, hkdb, field list, and aliases for both the site and platform HK data.

- Add a new HKManager instance to observations on load if the site or platform HK locations are specified. This works similar to detdata, shared, and interval data manager classes.

- In addition to raw HK data and timestamp access, the HKManager class has methods to interpolate data on the fly to the observation sample times.

Here are some snippets from a jupyter notebook showing the use of these new traits, and how the `obs.hk` manager can be used:
load_hk_1
load_hk_2
load_hk_3
load_hk_4

### DIFF
--- a/sotodlib/toast/hkmanager.py
+++ b/sotodlib/toast/hkmanager.py
@@ -1,0 +1,206 @@
+# Copyright (c) 2025-2025 Simons Observatory.
+# Full license can be found in the top level "LICENSE" file.
+
+from collections.abc import MutableMapping
+
+import numpy as np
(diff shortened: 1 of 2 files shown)


keskitalo (approved):
An elegant solution. I really appreciate the choice of three (!) interpolating functions.

I've tested this in my work and it does what it promises. I hope we soon get the ability to write the housekeeping data into HDF5 dumps.

PR #1449, opened 2025-11-11
Add discrete cosine transform (DCT) for mlmapmaker

Bai-Chiang:


Bai-Chiang:
new forced commits adapt changes in https://github.com/simonsobs/so3g/pull/226

chervias (approved):
I ran a map with the site pipeline script and it worked. If you have tested the TOAST operator then I'm happy for this to be merged. You can remove the default downweight, I think we won't use it anymore.

Bai-Chiang:
@chervias could you test again now I removed `downweight` from site pipeline.

PR #1452, opened 2025-11-13
add loadpremadeflags

yoshinori-0778:
This PR adds new class of preprocess, which loads flags that made before and stored in aman. This is needed for running simulation of planet mapmaker that Sam and I have developed because it needs flags for a simlated planet that cannot be calculated in current preprocess. [This](https://github.com/simonsobs/pwg-scripts/blob/ys/1105_unified_mapmaker_update/flp/planet_mapmaker/sim.py) is a code in pwg-script corresponding to this update. 
Current way is 
1. load and run some basic preprocess (e.g., pointing_model/fp_flags)
2. calculate and add simulated TOD into aman.signal, and calculate flags for the simlated planet.
3. Run rest of preprocess that run for actual planet mapmaking. 
Since the flags for the simlated planet is needed for these preprocess (e.g., PCA filter) in step 3, I would like to add this.

There could be more smart way, so please let me know if you come up with more smart way.

### DIFF
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -1486,6 +1486,70 @@ class DarkDets(_Preprocess):
         else:
             return keep
 
+class LoadPremadeFlags(_Preprocess):
+    """Load premade flags from aman.
+
+    Saves results in proc_aman under the "premade_flags_name" field of aman.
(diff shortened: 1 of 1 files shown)


yoshinori-0778:
Thanks for the comment, let me think and try that.

PR #1456, opened 2025-11-18
Sidelobe cutter

chervias:
This implements the sidelobe cutter from @amaurea sogma code. The mask is a simple enmap so it should be low-res and full sky to not be super slow. The coords module seems like the natural place, but I'm not completely sure, so open to suggestions.

### DIFF
--- a/sotodlib/coords/sidelobes.py
+++ b/sotodlib/coords/sidelobes.py
@@ -3,7 +3,7 @@ from pixell import enmap, utils, coordsys
 from sotodlib import coords, mapmaking
 import so3g
 
-def sidelobe_cut(obs, args, sidelobe_cutters, object_list=None):
+def get_cuts(aman, args, sidelobe_cutters=None, object_list=None):
     """
     Function to calculate cuts (RangesMatrix) for Sun and Moon
     from a binary sidelobe mask with an arbitrary shape.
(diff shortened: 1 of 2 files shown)


sotodlib/coords/sidelobes.py
```diff
@@ -0,0 +1,96 @@
...
+    args : dict
+        An argument dictionary, which must contains the path to the
+        masks in mask_sun and mask_moon. This is only used for these
+        paths.
+    sidelobe_cutters : dict
+        A dict holding SidelobeCutter for each sun and moon. In the
+        first run this should be empty and will be filled. Subsequent
+        runs will use these precalculated objects.
+    object_list : list, optional
+        A list of the objects we want to mask. If None, sun and moon
```
mhasself: A list of strings (lower case) indicating the objects to mask.

sotodlib/coords/sidelobes.py
```diff
@@ -0,0 +1,96 @@
...
+            # some slightly more than 180 can lead to the numbers drifting over a large range.
+            # It might be better to catch this by just looking for values too close to the poles
+            # explicitly instead of relying on gpu_mm to catch things itself
+            #L.print("Error cutting %s sidelobes for %s: Pointing overflow. Skipping cut" % (self.objname, id), level=2, color=colors.red)
+            return Simplecut(obs.signal.shape[0], obs.signal.shape[1])
+        cuts = so3g.proj.RangesMatrix.from_mask(tod.astype(int)) #here the mask is expected to be int
+        return cuts
+
+def forward(tod, map, pmat):
+    pmat.from_map(dest=tod, signal_map=map, comps="T",)
```
mhasself: This is called exactly once, so does it need to be a function?  If you keep it, make it private (`_forward`).

PR #1460, opened 2025-11-19
Extend TOAST/sotodlib pointing comparison

keskitalo:
Check that TOAST and sotodlib assign same samples to same pixels given the same WCS configuration.

Unit tests will not pass until https://github.com/hpc4cmb/toast/pull/894 is merged.

### DIFF
--- a/tests/test_toast_context.py
+++ b/tests/test_toast_context.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2023-2023 Simons Observatory.
+# Copyright (c) 2023-2025 Simons Observatory.
 # Full license can be found in the top level "LICENSE" file.
 """Test toast context loading."""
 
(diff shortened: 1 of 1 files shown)


keskitalo:
Turns out that the 1/2 pixel offset was only needed before a recent `crpix` data type fix in `so3g`. I removed the offset from this PR.

@Bai-Chiang noticed that the exact weather parameters used in `qpoint` azel->radec conversion impact the TOAST/so3g agreement. SimGround now has a special `so3g_compat_mode` where the pointing conversion assumes typical Atacama weather just like so3g.

Bai-Chiang (approved):
Also it's better change the default `weather` in mlmapmaker
https://github.com/simonsobs/sotodlib/blob/9fd7bebfbf459b1c75d115c1a6df8440c6a0514f/sotodlib/toast/ops/mlmapmaker.py#L205

and add `so3g_compat_mode` to workflow script
https://github.com/simonsobs/sotodlib/blob/9fd7bebfbf459b1c75d115c1a6df8440c6a0514f/sotodlib/toast/workflows/sim_observe.py#L103-L108

PR #1462, opened 2025-11-20
Fix for discarding too short obs for obs_grouping

chervias:


mmccrackan (approved):

mhasself (commented):
When you specify defaults in the function signature, it makes it harder to actually activate those defaults, especially when cascading some argument through multiple functions and possibly from a config file.

For example, if you pass in "None" from the outside (perhaps via a config), then the code will simply fail, because None will be explicitly passed all the way to build_period_obslists, where it's assumed to be float-like.

A more robust approach to such arguments is to have them default to None, in the function signatures, and then in the body of the function that uses the value, set its default value if None was passed in.

PR #1463, opened 2025-11-20
Add h5py to dependencies.

tristpinsm:


BrianJKoopman (approved):
> Hmm, actually I thought `h5py` was already a dependency (but clearly it was not). We definitely use `h5py` in the sotodlib package. Seems good to make the dependency explicit.

Thanks for chiming in @tskisner! I was just noticing the same thing -- that `h5py` is used, but was missing from the dependency list. This sounds good to me now too! I'll go ahead and approve and merge.

> It is also a dependency of `flacarray` which we depend on, so it may just have been always brought in by other dependencies. 

It's actually not listed as a dependency there: https://github.com/hpc4cmb/flacarray/blob/ccec7d70a977d3243cc2285f3fa661d14586bec4/pyproject.toml#L19-L21

Indeed in a clean environment, all `flacarray` pulls in is `numpy`:
```
flacarray==0.3.4
└── numpy [required: Any, installed: 2.3.5]
```

PR #1467, opened 2025-11-24
add phase_to_pW into get_det_bias_flags

yoshinori-0778:
This PR adds phase_to_pW selection into `get_det_bias_flags` because I found the observation below contains phase_to_pW is about 10^48, which makes `signal` inf when I apply phase_to_pW calibarion.

iobsid = 'obs_1726651313_satp3_1111111'
iws = 'ws1'
iband = 'f090'

### DIFF
--- a/sotodlib/tod_ops/flags.py
+++ b/sotodlib/tod_ops/flags.py
@@ -17,7 +17,7 @@ from . import fourier_filter
 
 def get_det_bias_flags(aman, detcal=None, rfrac_range=(0.1, 0.7),
                        psat_range=None, rn_range=None, si_nan=False,
-                       merge=True, overwrite=True,
+                       phase_to_pW=None, merge=True, overwrite=True,
                        name='det_bias_flags', full_output=False):
     """
     Function for selecting detectors in appropriate bias range.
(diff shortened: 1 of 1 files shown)


sotodlib/tod_ops/flags.py
```diff
@@ -125,6 +130,10 @@ def get_det_bias_flags(aman, detcal=None, rfrac_range=(0.1, 0.7),
             msk = ~(np.all([range], axis=0))
             msks.append(RangesMatrix([Ranges.ones_like(x) if Y
                                       else Ranges.zeros_like(x) for Y in msk]))
+            
+        if phase_to_pW is not None:
+            ranges.append(detcal.phase_to_pW >= phase_to_pW[0])
+            ranges.append(detcal.phase_to_pW <= phase_to_pW[1])
 
         msk_names = ['bg', 'r_tes', 'r_frac_gt', 'r_frac_lt']
```
msilvafe: I know this wasn't introduced by you but I think this should be moved up above line 121 and then mask names for each of the criteria `psat_gt`, `psat_lt`, `rn_gt`, `rn_lt`, `phase_to_pw_gt`, phase_to_pw_lt` should be appended to `msk_names` if they're used. This is to be able to break down the contributions to the total det_bias_flags cuts to see which is impacting the cut most.

msilvafe (approved):
One comment but looks good.

sotodlib/tod_ops/flags.py
yoshinori-0778: Thank you for chekcing and comments on this. I modified this part. Could you take a look? If it looks fine,  I will squash the commits and merge it into main.

msilvafe: looks good

PR #1480, opened 2025-12-02
fixed a bug

yoshinori-0778:
Fixed a bug.


yoshinori-0778:
`np.ones_like(aman.dets.count)` made array(1).
This made the t2p fit function failed at `sigma_demod[di]`. 
I can imagine that this intented to make arrays that contains 1 and its length is aman.dets.count (i.e., array(1,1,....,1)).

msilvafe (approved):

PR #1485, opened 2025-12-05
Depth1 mapmaking script refactoring

iparask:
Trying to integrate mapcat to the depth1 mapmaking script, I realized that it was too difficult for me to move around. So I suggest the following refactoring as the first step. I think more cleaning needs to happen as there are functions that are not used at all.

Let me know what you think and how we can make it better.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -61,7 +61,7 @@ jobs:
 
       - name: Install sotodlib
         run: |
-          python3 -m pip install -vv .[tests]
+          python3 -m pip install -vv .[site_pipeline,tests]
 
       - name: Run Serial Tests
         run: |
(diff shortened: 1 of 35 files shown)


mhasself (commented):
There's already a site_pipeline.util, so we're not going to add a .utils :)

Maybe put these functions in `sotodlib.mapmaking.depth1_utils` ?

This PR isn't only whitespace / organizational changes... this version has stopped writing the .empty files.  Is everyone on board with that?

iparask:
With the integration of the DB, there is no reason for the empty files. This is my understanding at least. @chervias can you verify?

> There's already a site_pipeline.util, so we're not going to add a .utils :)
> Maybe put these functions in sotodlib.mapmaking.depth1_utils ?

I thought they would overpopulate the util.py file in site-pipeline, that is why the folder. I can move them under `mapmaking` but that would make `mapcat` a general dependency where now is just a site dependency (see [here](https://github.com/simonsobs/sotodlib/blob/depth1_mapmaker/pyproject.toml#L50-L61)).

PR #1488, opened 2025-12-08
Fix det cal hwpss

ykyohei:
This PR fixes issues found when I ran det_cal with hwpss subtraction for all the old observations.

1. Inconsistent number of channels between raw biasstep npy file and L3 oper book. -> updated load_and_reanalyze_bs to load fixed tones and reindex axismanager.
2. smurfgaps in biases leads to incorrect bias edges. -> Added correction in fill_zeros_biases
3. Check if data of tones is None or not to debug the error of loading fixed tones. Resolves https://github.com/simonsobs/sotodlib/issues/1508

PR #1501, opened 2025-12-15
add get_obs_all_channels

ykyohei:
Added a function that loads both normal and special channels into dets axis.
This is necessary for re-analyzing Biasstep or IV when special channels for example fixed tones are streaming.
This was originally implemented in https://github.com/simonsobs/sotodlib/pull/1488, but implemented this independently for more general purposes.


msilvafe:
Why a new function instead of a `load_fixed_tones` argument?

PR #1511, opened 2026-01-04
Fix PWV handling for quality reports

mmccrackan:
Saw that the recent quality reports had failed due to the way PWV was being handled from the hkdb, so this fixes it and adds better checks.


guanyilun (approved):
looks good to me!

PR #1512, opened 2026-01-07
Fix calling filterbin script through `so-site-pipeline`

sbiquard:
Previously, the `main` function in `make_atomic_filterbin_map.py` would not accept the `nprocs` argument defined in the parser, which would prevent from running it through the `so-site-pipeline` command.

This fixes the problem by using the `main_launcher` utility.


iparask:
What is the error you see? How can we trigger it?

chervias:
We don't use the `so-site-pipeline` command on prefect, we run the script directly. None of that affected?

PR #1515, opened 2026-01-08
Det match updates

mmccrackan:
Work in progress.  Will remove some formatting and try except debug changes before merge.

Some updates for `det_match` to try and address mismatches.  Mainly just prevents switching between assigned bias groups, only allowing a detector to jump between unassigned and assigned.  This change results in det_ids being assigned to readout_ids with the right bias group (or bg=-1) across many tunes.  Also forces `UNRT`, `SQID`, and `BARE` to be unassigned and ensures `match_idx` is reset for each matching iteration as some det_ids had multiple readout_ids with the same `match_idx` (this should not have resulted in any mismatches since `match_idx` is not used for those, so this is mostly just a fix for plotting and bookkeeping).

For the solutions matches, there still seems to be some influence on how the pointing results are handled.  There is a step that merges different pointing through detector matching.  Interestingly, using a too small distance width there results in the exact same mismatches for the one LAT obs_id (`obs_1757469130_lati3_111`) where the mismatch was first seen.

Plot just showing the evolution of det id bias groups from different tunes for the LAT for one wafer.

image

### DIFF
--- a/sotodlib/coords/det_match.py
+++ b/sotodlib/coords/det_match.py
@@ -613,14 +613,14 @@ class MatchParams:
     unassigned_slots: int = 1000
     freq_offset_mhz: float = 0.0
     freq_width: float = 2.
-    dist_width: float =0.01
+    dist_width: float = 0.01
     unmatched_good_res_pen: float = 10.
     good_res_qi_thresh: float = 100e3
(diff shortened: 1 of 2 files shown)


mmccrackan:
Closing in favor of #1581.

Issue #1304, opened 2025-07-18
Reformat unqueriable tags in obsdb


mhasself:
Sometimes obs are acquired with tags that contain problematic characters (brackets, leading digits).

Those should be reformatted before being put in the obsdb.

Also it would be great if such tags got blocked, earlier... but that's not an issue for this repo.

PR #1520, opened 2026-01-12
Wrap 'tags' for ObsDb in single quotes to accomodate special characters

JBorrow:
Accomodates special characters in tag names by wrapping them in single quotes in the query creation.

### DIFF
--- a/sotodlib/core/metadata/obsdb.py
+++ b/sotodlib/core/metadata/obsdb.py
@@ -417,7 +417,7 @@ class ObsDb(object):
             obsdb.query('`bad-tag`=1', tags=['bad-tag'])
 
           For this reason, we generally advise against the use of 
-          non-alphanumeric chacaters in tags.
+          non-alphanumeric characters in tags.
 
         """
         sort_text = ''


mhasself:
Great! Needs a test case... 

Possibly works around #1304 sufficiently to close it...

JBorrow:
Added a test that includes both dashes and emoji in the tag name. Cherry picking that test back to `master` causes the following failure, indicating that this PR does indeed solve the issue:

```
                joins += (f" {join_type} (select distinct obs_id from tags where tag='{t}') as tt{tagi} on "
                          f"obs.obs_id = tt{tagi}.obs_id")
        extra_fields = ''.join([','+f for f in extra_fields])
        q = 'select obs.* %s from obs %s where %s %s' % (extra_fields, joins, query_text, sort_text)
>       c = self.conn.execute(q)
            ^^^^^^^^^^^^^^^^^^^^
E       sqlite3.OperationalError: near "-": syntax error

sotodlib/core/metadata/obsdb.py:446: OperationalError
```

PR #1521, opened 2026-01-13
Preprocessing stats db

mmccrackan:
Based on #1348.  Adds the option to create and write stats to a "stats" ObsDb indexed by the preprocessing config's grouping (i.e. wafer and band).


kmharrington:
What is a "preprocessing stats obsdb"? Do we have that somewhere already?

Issue #1326, opened 2025-08-11
Refactor preprocess load and run functions


mmccrackan:
While we have another issue for refactoring preprocessing (#1083), this one is a bit more specific and short term.  We should:

- Combine `load_and_preprocess` and `multilayer_load_and_preprocess.` Possibly combine `sim` versions too.
- Make `multilayer_preprocess_tod` and `preprocess_tod` both rely on `preproc_or_load_group` to reduce code duplication.  `Multilayer_preprocess_tod` already calls it once for the first layer but this can be expanded.
- Simplify `preproc_or_load_group`: this function is long and has logic that can be simplified for improved readability.   This has already been started in an older branch: #1156.

PR #1529, opened 2026-01-20
Merge preprocess loading functions

mmccrackan:
Addresses first point of #1326 which is the last one remaining to be done.  Should be merged after #1528.

Combines `load_and_preprocess` and `multilayer_load_and_preprocess` into one function so that it matches the usage of `preproc_or_load_group` which handles both cases.  The sim version is kept separate as is the case for `multilayer_preprocess_tod` since it has some extra inputs.


mmccrackan:
I merged in master and think I've dealt with all of the merge problems.  I tested the standard running and loading functions and they seem to work, but haven't tested the sim version.

PR #1530, opened 2026-01-20
Allow scalar value in timeconst filter

ykyohei:
This PR modifies timeconst_filter to allow scalar timeconst argument.
Also adds timeshift filter.


sotodlib/tod_ops/filters.py
```diff
@@ -349,6 +348,21 @@ def gain(freqs, tod, gain=1.):
     """
     return gain * np.ones(len(freqs))
 
+@fft_filter
+def delay(freqs, tod, dt=0, invert=False):
```
mhasself: More generally, this is a time-shift filter.  Let's call it "timeshift", and explain that positive values of dt will cause the signal to be moved to later samples of the array.

PR #1537, opened 2026-01-24
Make get_source_azel compatible with fixed source

ykyohei:
This is useful for me, and it should have been possible.


mhasself (approved):
Great! Thanks.

PR #1543, opened 2026-02-02
gamma from wiregrid

adrien-laposta:
This PR modifies `preprocess_util.multilayer_load_and_preprocess_sim` to load `focal_plane.gamma` from `wiregrid_cal` before loading any simulated TOD into the `AxisManager`.


chervias (approved):
Tested it and seems to work

PR #1546, opened 2026-02-07
Add latest mocks for building docs

mhasself:


sanahabhimani (approved):
Thanks for the fix--I just checked the build on readthedocs for this PR and it's very nice to see all those docstrings come back!

Not sure if you prefer to merge or have the reviewer do so---approving without merging for now.

PR #1548, opened 2026-02-09
Data Packaging Updates for ASO / LF

kmharrington:
Updates to get data packaging ready for the LF data.


sotodlib/io/bookbinder.py
```diff
@@ -1236,6 +1236,10 @@ def get_metadata(self, telescope=None, tube_config={}):
 
         # build detset list in same order as slots
         meta['stream_ids'] = self.book.slots.split(',')
+        if len(meta['stream_ids']) != len(set(meta['stream_ids'])):
+            assert meta['tube_flavor'] == 'lf', "Only LF can have duplicate stream_ids"
+            meta['stream_ids'] = list(set(meta['stream_ids']))
```
mhasself: Is `list(set(...))` guaranteed to always produce the items in the same order?  The comment above says "in same order as slots".  In practice it may, but I bet we're not supposed to depend on it.

I think it would be safer to also wrap in `sorted(...)`.

PR #1550, opened 2026-02-09
Allow ufm selection in update_det_match

mmccrackan:
In order to prevent failures from new ufms being added that do not currently have `detmatch_solutions`, this adds an optional list input to the `update_det_match` config that allows for only running `update_det_match` on the detsets that are from the specified ufms.  Other ufms will not be added to the missing obs_id list.

### DIFF
--- a/sotodlib/site_pipeline/update_det_match.py
+++ b/sotodlib/site_pipeline/update_det_match.py
@@ -134,6 +134,9 @@ class UpdateDetMatchesConfig:
             if self.resonator_set_dir is None:
                 raise ValueError("Must specify resonator_set_dir for solution_type='resonator_set'")
 
+        if self.ufms is not None:
+            self.ufms = [u.lower() for u in self.ufms]
+
 
 class Runner:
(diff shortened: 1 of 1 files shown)


sotodlib/site_pipeline/update_det_match.py
```diff
@@ -169,6 +173,8 @@ def __init__(self, cfg: UpdateDetMatchesConfig):
 
     def get_remaining_detsets(self) -> List[str]:
         detsets_all = set(self.detset_db.get_entries(['dataset'])['dataset'])
+        if self.cfg.ufms:
+            detsets_all = set([d for d in detsets_all if '_'.join(d.split('_')[:2]) in self.cfg.ufms])
```
mhasself: This requires exact case matching -- is all-lower-case standard at this point?

mmccrackan: Yeah all of the `detset` dataset names are lower case, so I've made it convert the inputs to also be lower case.

PR #1565, opened 2026-02-24
D1 mapmaker, adding capability to run on prefect

chervias:


sotodlib/site_pipeline/make_depth1_map.py
```diff
@@ -276,6 +285,12 @@ def main(config_file, defaults=d1u.DEPTH1MAPMAKER_DEFAULTS, **args):
     else:
         raise ValueError(f"Unrecognized noise model '{args['nmat']}'")
 
+    if (args['update_delay'] is not None):
+        # this is only to be used for running automatic maps on prefect
+        args['query'] += f" and duration>{args['min_dur']}"
```
mhasself: Why is this inside the update_delay block?

mhasself: Ohhh, because the argument "query" is double purpose, to _either_ be a filename or a db query string.  Which is not how I would have done things.

You can revert this commit and just do what you had before.

mhasself (approved):

iparask:
> > ok thanks. I won't merge this until I actually test it on incoming data, so we need the LAT to resume obs for that
> 
> It might be a good idea to have a mode that supports testing in such situations. Consider taking the time to add this, as it will help with development now and in the future.
> 
> For example an argument called "--pretend-now-is=...".

Or we can mock what time.time returns instead of adding a new flag. :)

chervias:
@mhasself this is ready to merge. With @iparask we have tested it on tiger and it works fine. Please have a quick look before we merge. thanks

sotodlib/site_pipeline/make_depth1_map.py
```diff
@@ -276,6 +289,20 @@ def main(config_file, defaults=d1u.DEPTH1MAPMAKER_DEFAULTS, **args):
...
+    if not os.path.isfile(args['query']):
+        args['query'] += f" and duration>{args['min_dur']}"
+
+    if (args['update_delay'] is not None):
+        # this is only to be used for running automatic maps on prefect
+        if args['pretend_now_is'] is not None:
+            date_format = "%Y-%m-%d %H:%M:%S"
+            dt_obj = time.strptime(args['pretend_now_is'], date_format)
+            min_ctime = int(time.mktime(dt_obj)) - args['update_delay']*86400
+            args['query'] += f" and timestamp>={min_ctime} and timestamp<={int(time.mktime(dt_obj))} "
```
mhasself: I believe this is sensitive to the timezone of the host system. Better to force it into UTC.

tests/test_depth1_utils.py
```diff
@@ -0,0 +1,129 @@
...
+            self.assertEqual(config[key], value)
+
+    def test_config_file(self):
+        file_config = {"query": "type == 'obs' and subtype == 'cmb' and tube_flavor == 'mf'",
+                       "context": "/so/metadata/lat/contexts/use_this.yaml",
+                       "area": "/so/site-pipeline/lat/maps/so_geometry_f090_lat.fits",
+                       "odir": "/so/site-pipeline/lat/maps/depth1/",
+                       "site": "so_lat",
+                       "preprocess_config": "/config/lat/preprocess_config_cmb_mf.yaml",
+                        "mapcat_database_name": '/so/site-pipeline/lat/maps/depth1/mapcat.sqlite',
```
mhasself: Indentation slip?

tests/test_depth1_utils.py
```diff
@@ -0,0 +1,129 @@
...
+                        "maxiter": 20,
+                        "verbose": True,
+                        "quiet": False,
+                        "tasks_per_group": 4,
+                        "unit": 'K',
+                        "update_delay": 2}
+
+        with tempfile.NamedTemporaryFile(
+            mode="w", suffix=".yaml", delete=False
+        ) as f:
```
mhasself: Instead of needing a "finally" to unlink the file, you can use the NamedTemporaryFile context.  Just pass delete_on_close=False (instead of delete=False); then f.close() after you dump, and it's safe to read.

tests/test_depth1_utils.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def test_defaults(self):
+        config = create_mapmaker_config(args={"area": "a", "context": "c"})
+        for key, value in DEPTH1MAPMAKER_DEFAULTS.items():
+            self.assertIn(key, config)
+            self.assertEqual(config[key], value)
+
+    def test_config_file(self):
+        file_config = {"query": "type == 'obs' and subtype == 'cmb' and tube_flavor == 'mf'",
+                       "context": "/so/metadata/lat/contexts/use_this.yaml",
```
mhasself: This big config block is duplicated several times.  Would be better to have it declared as a module const, or returned by a helper function.

mhasself (changes requested):
Couple suggestions.

sotodlib/site_pipeline/make_depth1_map.py
chervias: done

tests/test_depth1_utils.py
iparask: Done

tests/test_depth1_utils.py
iparask: Done with using `tmp_path` that purest offers.

mhasself (approved):
Thanks!

PR #1572, opened 2026-03-06
Test Jobdb batch locking

mmccrackan:
Changes `lock`, `unlock`, `locked`, and `remove_jobs` to accept lists of jobs and update them all at once.  Still supports single inputs and `locked` yields  a single job if only one job is input.

Updates `preprocess_tod` and `multilayer_preprocess_tod` to batch the writes to the JobDB in the same way as the recent `ManifestDb` update.  Also moves the `get_jobs` call in preprocessing outside of the futures loop.

On a single thread on the Nersc login node just populating a `jobdb` with 1e4 entries, without batching it takes ~230 seconds, whereas running with batches of 100 or so takes ~8 seconds.


msilvafe:
I'm testing on 200 obs on satp1 on tiger (over 3 nodes) now. If that works I will call my testing complete so then just barring @mhasself review for merge.

mmccrackan:
~@mhasself There's still some testing going on to make sure this is doing what we want so I'll leave it as a draft, but I think this PR is probably in a good enough shape for a review of the `jobdb` changes to see if they make sense whenever you have a chance.~

Sorry, didn't see Max's message which says the same thing.

sotodlib/site_pipeline/jobdb.py
```diff
@@ -304,55 +304,103 @@ def get_jobs(self,
...
         """
         if owner is None:
             owner = self._lockstr()
+
+        now = time.time()
+
+        if isinstance(job_ids, int):
+            job_ids = [job_ids]
+
+        job_ids = [(j.id if isinstance(j, Job) else j) for j in job_ids]
```
mhasself, line 326: Either outlaw passing in Job(s) entirely, or accept that a user might pass in a single Job; e.g. `isinstance(job_ids, (int, Job))`

sotodlib/site_pipeline/jobdb.py
```diff
@@ -304,55 +304,103 @@ def get_jobs(self,
...
+            )
+
+            for job in jobs:
+                session.expunge(job)
 
-        if n == 0 or job.lock_owner != owner:
+        locked_jobs = [j for j in jobs if j.lock_owner == owner]
+
+        if n !=len(job_ids) or len(locked_jobs) != len(job_ids):
             raise JobLockedError()
```
mhasself, line 354: This behavior should be revisited -- as implemented here, if one of the many jobs is locked already, the code will lock them all and then raise an error.  This is a sure-fire way for a small problem of a stray locked row to compound into multiple stray locked rows.

Reasonable behaviors:
- best effort -- lock as many jobs as possible: lock and return _only those ones_ (even if empty set) without raising any errors.
- al-or-none; attempt lock of all jobs, but if _any_ are already locked, then _do not lock_ any of them (i.e. rollback()) and then raise the error.

sotodlib/site_pipeline/jobdb.py
```diff
@@ -304,55 +304,103 @@ def get_jobs(self,
...
 
-        return job
+        return locked_jobs[0] if len(locked_jobs) == 1 else locked_jobs
```
mhasself, line 356: As written, if I pass in [job_id] (a list) I will get back a scalar.  That will cause trouble in some cases; better to capture scalarness in a variable at the top of this function, and use that to decide whether to unpack the list in the return.

sotodlib/site_pipeline/jobdb.py
```diff
@@ -304,55 +304,103 @@ def get_jobs(self,
...
+
+    def unlock(self, jobs, merge=True):
+        """Unlock one or more jobs."""
+        if not isinstance(jobs, (list, tuple, set)):
+            jobs = [jobs]
```
mhasself, line 361: Make this check the same as what is done in `lock` (or make lock the same as here).

sotodlib/site_pipeline/jobdb.py
```diff
@@ -304,55 +304,103 @@ def get_jobs(self,
...
+                db_map = {j.id: j for j in db_jobs}
+
+                for job in job_objs:
+                    j1 = db_map[job.id]
+
+                    if j1.lock_owner is None:
+                        raise JobNotLockedError()
+
+                    if j1.lock_owner != job.lock_owner:
+                        raise JobNotOwnedError()
```
mhasself, line 398: As above, error handling in the multiple-jobs should be done with some care.  Perhaps on failure, you make a note and then try the next job?  Then raise if _any_ failures?

In a multi-job case it would be helpful to provide traceback about what job failed user's assumptions here.

sotodlib/site_pipeline/jobdb.py
```diff
@@ -368,26 +416,30 @@ def clear_locks(self, jobs=None):
...
-                q = session.query(Job).filter(Job.id == job_id)
+                q = q.filter(Job.lock == None)  # noqa: E711
 
-            n = q.delete()
+            n = q.delete(synchronize_session=False)
             session.commit()
-        if n == 0:
+
+        if n != len(job_ids):
             raise JobLockedError()
```
mhasself, line 435: Maybe a custom error indicating that "only _n_ of _m_ jobs were successfully deleted"?

sotodlib/site_pipeline/jobdb.py
```diff
@@ -411,30 +463,38 @@ def locked(self, jobs, count=None, owner=None):
...
+            for jid in job_ids:
+                if jid in unlocked_ids:
+                    selected.append(jid)
+                if len(selected) >= limit:
+                    break
+
+            if selected:
+                locked = self.lock(selected, owner=owner)
+
+            yield locked
```
mhasself, line 493: Needs special handling to yield a single job if count is None.

sotodlib/site_pipeline/jobdb.py
```diff
@@ -411,30 +463,38 @@ def locked(self, jobs, count=None, owner=None):
...
+            limit = count if count is not None else 1
+
+            for jid in job_ids:
+                if jid in unlocked_ids:
+                    selected.append(jid)
+                if len(selected) >= limit:
+                    break
+
+            if selected:
+                locked = self.lock(selected, owner=owner)
```
mhasself, line 491: (Depending on how self.lock() error handling is modified --) There's a race condition here that will not necessarily produce _count_ newly locked items.  Although it will be slow on slow databases, you probably need a loop here to keep locking items until you reach count or run out of candidates.

There may be atomic one-liners to do this sort of thing; the original code didn't optimize for that (partly because I don't know sqlalchemy well enough).  But consider something like:
```update jobs set locked={timestamp}, owner={owner} where job_id in (select job_id from jobs where locked=null limit {count})```.

sotodlib/site_pipeline/jobdb.py
mhasself, line 461: Oops, bad docstring ... please change docstring or implement this missing behavior :)

mhasself (changes requested):
Some interface requests -- for these supporting systems, clearly defined and consistent behaviors are very important.

PR #1581, opened 2026-03-17
Det match updates

mmccrackan:
Replaces #1515 which was similar.  This does the following:

- Adds a parameter called `allow_unassigned_to_assigned` which will either allow or prevent matches from bg=-1 in the design or pointing data to bg != -1 in the pointing or tuneset data.  We want to prevent such matches when matching from design to pointing, since the detector types with bg=-1 should always have that value.  When matching between pointing and tunesets, we allow bg = -1 to other bg values since some detectors may be assigned bg = -1 in det_cal due to the conditions of when the pointing tune was taken but may be recovered in later tunesets (this likely won't happen since it would ideally match the pointing to a bg != -1 detector and copy the bg from the design).
- Like #1515, forces `det_types` of `UNRT`, `SQID`, and `BARE` to always be bg = -1.
- Prevents mismatch between assigned bgs (i.e matching bg=4 to bg=6).
- Makes `det_row`, `det_col`, and `det_rhombus` optional for LF.
- Adds `det_match_solutions` code to `coords`.
- Adds `start_time` and `stop_time` to `update_det_match` to simplify epoch handling for detsets.  Future update will add epochs to the databases.

Remainder of LF matching updates will be in a separate PR after testing and confirming it works on better pointing data (also needs #1603).

### DIFF
--- a/sotodlib/coords/det_match.py
+++ b/sotodlib/coords/det_match.py
@@ -196,6 +196,7 @@ def apply_design_properties(smurf_res, design_res, in_place=False, apply_pointin
         'mux_bondpad', 'mux_subband', 'mux_band', 'mux_channel',
         'mux_layout_pos'
     ]
+
     if apply_pointing:
         design_props += ['xi', 'eta', 'gamma']
 
@@ -613,10 +614,11 @@ class MatchParams:
     unassigned_slots: int = 1000
     freq_offset_mhz: float = 0.0
     freq_width: float = 2.
-    dist_width: float =0.01
+    dist_width: float = 0.01
     unmatched_good_res_pen: float = 10.
     good_res_qi_thresh: float = 100e3
     enforce_pointing_reqs: bool = False
+    allow_unassigned_to_assigned: bool = True
 
     assigned_bg_unmatched_pen: float = 100000
     unassigned_bg_unmatched_pen: float = 10000
@@ -706,6 +708,11 @@ class Match:
         m = src_arr['is_north'][:, None] != dst_arr['is_north'][None, :]
         mat[m] = np.inf
 
+        src_no_match = np.isin(src_arr['det_type'], ['NC'])
+        dst_no_match = np.isin(dst_arr['det_type'], ['NC'])
+        mat[src_no_match, :] = np.inf
+        mat[:, dst_no_match] = np.inf
+
         if self.match_pars.enforce_pointing_reqs:
             # Det types of DARK and SLOT are allowed to may or may not have
             # pointing data. For these types of detectors, the cost is left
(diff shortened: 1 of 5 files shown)


sotodlib/coords/det_match.py
```diff
@@ -613,10 +653,11 @@ class MatchParams:
     unassigned_slots: int = 1000
     freq_offset_mhz: float = 0.0
     freq_width: float = 2.
-    dist_width: float =0.01
+    dist_width: float = 0.01
     unmatched_good_res_pen: float = 10.
     good_res_qi_thresh: float = 100e3
     enforce_pointing_reqs: bool = False
+    allow_unassigned_to_assigned: bool = True
```
mhasself, line 662: Please add this to class docstring ... and maybe check that over to make sure nothing else is missing.

sotodlib/coords/det_match_solutions.py
```diff
@@ -0,0 +1,556 @@
+from sotodlib.core import Context, AxisManager, LabelAxis
```
mhasself, line 1: Add submodule docstring explaining briefly what the point of this is.

If not defined elsewhere, explain the basic terminology (solution, pointing, ...).

There's already some overview docs on det_match -- it would be nice to mention this new capability in there, and to autodoc the submodule.

sotodlib/coords/det_match_solutions.py
```diff
@@ -0,0 +1,556 @@
...
+import yaml
+from copy import deepcopy
+from scipy import interpolate
+from tqdm.auto import tqdm, trange
+from copy import deepcopy
+
+import h5py
+
+from importlib import reload
+import yaml
```
mhasself: reload not used; yaml is duplicate import.

sotodlib/coords/det_match_solutions.py
```diff
@@ -0,0 +1,556 @@
...
+    wafer_slots = ["ws0", "ws1", "ws2", "ws3", "ws4", "ws5", "ws6"]
+    results = {ws: get_wafer_solution(cfg, ws, save=True) for ws in wafer_slots}
+    return results
+
+
+if __name__ == "__main__":
+    cfg_file = sys.argv[1]
+    with open(cfg_file, "r") as f:
+        cfg = SolutionsCfg(**yaml.safe_load(f))
+    solve_all(cfg)
```
mhasself, line 589: Instead of having random runnable scripts in the modules, can we make this accessible via update_det_match?  E.g.

``so-site-pipeline update-det-match soln-cfg.yaml --solutions-mode ``

mhasself (approved):
Thanks -- mostly documentation / tidying requests, but I understand if you need to press forward. Great to have this side of things covered in sotodlib now.

mmccrackan:
> Thanks -- mostly documentation / tidying requests, but I understand if you need to press forward. Great to have this side of things covered in sotodlib now.

Okay, fixed imports, updated the docstrings and documentation to add a description of the solutions code and what is meant by solutions and pointing, and added and tested the solutions script to `update_det_match` and the CLI interface.

PR #1587, opened 2026-03-23
Add support for pickling error

keskitalo:


AERAdler:
Didn't really seem to work, both for cases where end of file was reached unexpectedly and where an unrecognised character was found.

keskitalo:
Can you post the stack trace here? There is nothing obvious in the source code that would explain the failure.

keskitalo:
Weird. I created a corrupted pickle file by splitting a valid file into pieces. The exception is caught just as it should be.

I also found another bug: the `sleep()` method was not imported and was causing a separate failure.

Can you update your branch and rerun? If you see the error again, try examining `/global/homes/a/aeadler/toast_dev_env/lib/python3.12/site-packages/sotodlib/toast/ops/utils.py` to see if it actually has the corrected code in place or if it is stale.

PR #1595, opened 2026-04-01
Add axis assignements everywhere

kmharrington:
Using this for other reasons and trying to concatenate axis managers after down sampling. Recognized that there needs to be a lot more axis assignments.


skhrg (approved):
Approving because this looks safe to me but any shot you could add a test before merging (I know I should have added one with the original PR...)

PR #1598, opened 2026-04-06
Epochs

skhrg:
Leaving this as a draft for now because I need to update the matching configs and test.

But it would be useful at this point to get some feedback on how this is shaking out and if there are other features to include at this point.

### DIFF
--- a/sotodlib/utils/epochs.py
+++ b/sotodlib/utils/epochs.py
@@ -3,12 +3,11 @@ from copy import deepcopy
 from dataclasses import dataclass, field
 from itertools import accumulate
 from operator import add, and_, mul, or_
-from typing import Any, Optional, Self, cast
+from typing import Any, Literal, Optional, Self, cast, overload
 
 import yaml
 from deepdiff import DeepDiff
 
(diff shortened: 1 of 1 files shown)


skhrg:
Okay I have tested that with [this file](https://github.com/simonsobs/site-pipeline-configs/blob/inst_times/lat/instrument_times.yaml) things behave as expect (and in particular the behavior I want to streamline making pointing metadata is doing what I want).

skhrg:
Added an additional QOL feature.

I am going to start writing some pointing code using this to streamline some things, so hopefully you don't want any significant API changes @mhasself .

sotodlib/utils/epochs.py
```diff
@@ -0,0 +1,638 @@
+import re
+from copy import deepcopy
+from dataclasses import dataclass, field
+from functools import reduce
+import operator
+from typing import Any, Literal, Optional, Self, cast, overload
+import numpy as np
+
+import yaml
+from deepdiff import DeepDiff
```
mhasself: Add to dependencies in pyproject.

PR #1602, opened 2026-04-08
Change if statement in preprocess select

kmharrington:
I noticed that passing in `select: False` didn't actually cause the select not to happen.  The way to have select not to happen is to not have select in the yaml at all. 

Changing the statement to `if not self.select_cfgs` allows both None and False. 

Should we make this change for all of the preprocess modules?


mmccrackan:
I think it is fine to make this change for all of the processes.  I had started trying to standardize the check for saving in this [PR](https://github.com/simonsobs/sotodlib/pull/1521) for some entries, so we could do something similar.  Some processes like [source flags](https://github.com/simonsobs/sotodlib/blob/49b10351b73bca66dc0fe0929a3797bb6ff9160e/sotodlib/preprocess/processes.py#L1895) might need to have different handling though.

msilvafe:
Yes we should make that change everywhere

mmccrackan:
Added the change throughout for both `select_cfgs` and `save_cfgs`.

msilvafe (approved):

PR #1603, opened 2026-04-10
feat: add support for ASO tubes

skhrg:
ASO OTs need to have their clocking accounted for, this adds that into the template code.

### DIFF
--- a/sotodlib/coords/optics.py
+++ b/sotodlib/coords/optics.py
@@ -4,6 +4,7 @@ Also includes tools for computing transforms between point clouds.
 
 LAT code adapted from code provided by Simon Dicker.
 """
+
 import logging
 from functools import lru_cache, partial
 import numpy as np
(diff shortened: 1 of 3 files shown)


mhasself (approved):
I'm sure this is all fine (you're sure all the rotation sign conventions area right?)... but it's worth explaining somewhere what you mean by "OT" vs. "FP" coords.

Is there an example of the config file syntax?

You could be testing more things, in the tests -- this is all light-weight, and the test would provide a reference for what format the config information should have.

skhrg:
Ah yeah I guess what I call "OT" coordinates here are receiver coordinates [here](https://simonsobs.atlassian.net/wiki/spaces/DM/pages/101974017/Focal+Plane+Coordinates#FP-%E2%86%92-Receiver-Transformation). I will standardize that. 
Syntax is [here](https://github.com/simonsobs/site-pipeline-configs/blob/main/shared/focalplane/optics_tubes.yaml), I can add that to the module docstring. 

The rotation sign convention should be correct, it is consistent with the convention for the array rotations and all values for both were determined from the same solidworks model in the same way.

skhrg:
Will make those changes (and see if I can flesh out the tests) before merging.

PR #1605, opened 2026-04-10
Lat pmchanges

skhrg:
Adds OT floating option, makes fitter settings configurable, merge in latest LAT pointing model.


sotodlib/site_pipeline/solve_pointing_model.py
```diff
@@ -1091,6 +1168,20 @@ def main(config_path: str):
...
+            # Remove OT float parameters
+            if config["float_ots"]:
+                ot_float_aman = core.AxisManager()
+                for ot in np.unique(solver_aman.ot_list):
+                    for par in [f"{n}_{ot}" for n in ["xioff", "etaoff", "rot", "xiscale", "etascale"]]:
+                        if par not in solver_aman.pointing_model._assignments:
+                            continue
+                        ot_float_aman.wrap(par, solver_aman.pointing_model[par])
+                        solver_aman.pointing_model.move(par, None)
+                solver_aman.wrap("ot_float_aman", ot_float_aman)
```
elle-shaw, line 1181: good, I am glad the ot float values will get saved somewhere.

elle-shaw (approved):
Looks good to me, lets merge it.

sotodlib/site_pipeline/solve_pointing_model.py
```diff
@@ -987,12 +1060,16 @@ def main(config_path: str):
...
 
+        if tot_bad == 0:
+            logger.info("No bad points found so not running second fit!")
+    if tot_bad > 0:
         model_solved_params = minimize(
             objective_model_func_lmfit_joint,
-            fit_params,
-            method="nelder",
+            model_solved_params,
+            method=config.get("fit_method", "nelder"),
```
elle-shaw, line 1069: I'm inclined to say that we should just have the default be least squares.  To do leastsq, though I changed the optimization function to append all the residuals instead of minimizing just the chi2 single number. I'd say this is necessary to do on this merging of your branch though.

PR #1607, opened 2026-04-10
Remove source_flags cut

mmccrackan:
Removes an unneeded source flags cut that will cause a crash if all detectors are removed.


msilvafe (approved):
We tested this together so happy to approve + merge.

PR #1608, opened 2026-04-13
Fix D1 mapmaker config create bug

iparask:
Fixed how the args from the arg parser are passed to the `create_mapmaker_config`

I also added a unit test to make sure that the functionality is correct and expected.


chervias (approved):
It works!

Issue #1597, opened 2026-04-01
Legendre azss fit range is only based on first detector


ykyohei:
The azss fitting range is only based on first detector, which is not ideal.
This can lead to nans in fitted azss model or other issues.
https://github.com/simonsobs/sotodlib/blob/f915bfab48483efa7d794c9a139172a5b0977136/sotodlib/tod_ops/azss.py#L151

We should apply similar fix as https://github.com/simonsobs/sotodlib/pull/1596

ykyohei:
Addressed by #1610

PR #1610, opened 2026-04-14
Fix azss legendre fit mask

ykyohei:
This PR addresses the issue of azss legendre fit https://github.com/simonsobs/sotodlib/issues/1597
Tested that we get consistent result as before through preprocess.
Added some unit tests to cover edge cases of azss subtractions.


msilvafe (approved):
Thanks for fixing this bug and significantly reducing redundancy in the code.

PR #1611, opened 2026-04-17
fix: verbosity error

iparask:
The preprocessing `_main` passes a value of `None` for verbosity and this showed that the logger init was only inside `if` without and `else` clause.

This fixes that and makes `ERROR` as the default behavior of the logger.


mmccrackan (approved):
I think this is fine.  I see the site-pipeline version is also the same.  Might want to update that to match this too.

PR #1615, opened 2026-04-20
Fix nmat

amaurea:
The main thing this does is to fix the way eigenvectors are projected out of the measured detector-detector covariance matrix, which is a step in building the noise model. It used to do A-AV(V'AV)"V'A, which is equivalent to setting those vectors to be infinitely certain. But what we actually want to do is to get the equivalent of measuring the covariance of the detector timestreams after projecting out those modes, and the effect of that on the covmat is A+VV'AVV'-AVV'-VV'A, which this pull request changes it to.

Other changes:

* Changed the internal normalization of V and iD in the noise matrix to make it simpler and bring it in line with sogma, but this doesn't have any effect on the result
* Added NmatDebug, which implementes a 1/f noise model with static f_knee and alpha
* Added an argument --iunit for specifying the unit of the TOD returned by load_and_preprocess. This is used in conjunction with the existing --unit argument to transform the loaded to the target unit. This should ideally be a separate pull request, but hopefully it's small enough to sneak through here.

### DIFF
--- a/sotodlib/mapmaking/noise_model.py
+++ b/sotodlib/mapmaking/noise_model.py
@@ -111,6 +111,45 @@ class NmatUncorr(Nmat):
     def from_bunch(data):
         return NmatUncorr(spacing=data.spacing, nbin=data.nbin, nmin=data.nmin, bins=data.bins, ips_binned=data.ips_binned, ivar=data.ivar, window=data.window, nwin=data.nwin)
 
+class NmatDebug(Nmat):
+    def __init__(self, ivar=None, alpha=-3, fknee=2.5, profile=None, bsize=256, downweight=1):
+        self.bsize = bsize
+        self.ivar  = ivar
+        self.alpha = alpha
+        self.fknee = fknee
+        self.profile = profile
+        self.downweight = downweight
+        self.nwin  = 0
+        self.ready = ivar is not None
+    def build(self, tod, srate, **kwargs):
+        nsamp  = tod.shape[1]
+        nblock = nsamp//self.bsize
+        var    = np.median(np.var(tod[:,:nblock*self.bsize].reshape(-1,nblock,self.bsize),-1),-1)
+        with utils.nowarn():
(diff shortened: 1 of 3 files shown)


chervias (approved):
Tested and looks good

sotodlib/mapmaking/noise_model.py
```diff
@@ -187,11 +226,10 @@ def build_fourier(self, ftod, nsamp, srate, nwin=0, **kwargs):
...
         ivar  = np.sum(iD*bsize[:,None],0)/np.sum(bsize)
-        # What about units? I haven't applied any fourier unit factors so far,
-        # so we're in plain power units. From the uncorrelated model I found
-        # that factor of tod.shape[1] is needed
-        iD   *= nsamp
-        iV   *= nsamp**0.5
+        # I originally normalized iD and iV here, but then undid that normalization
+        # when actually applying the model. Let's just skip that
+        #iD   *= nsamp
+        #iV   *= nsamp**0.5
```
mhasself, line 232: I think it's probably better to just delete these things that you've commented out.  Easy enough to find this commit in the future.

It would perhaps be more useful to place a comment here about what the significance of the normalization is, now.  E.g. do the diagonals correspond to inverse var per frequency bin?  Or should I expect them to be equal to the white noise ivar, in the white noise region?  I'm just skimming this so my examples might not make sense :)

mhasself (approved):

sotodlib/mapmaking/noise_model.py
amaurea: The change I made didn't affect the public interface of the class, so e.g. apply() and white() work the same as before. I *think* (but didn't check since it hasn't turned out to be useful to have this) the old normalization gave iD units of inverse variance per sample, so that if you built a noise model from white noise with an rms-per-sample of 10, then iD would be 1e-2. The new version does away with this, and just uses what you get from the plain unnormalized Fourier transform.

PR #1616, opened 2026-04-21
making mapcat a requirement

chervias:


iparask:
Since we're changing `pyproject.toml` can you also remove `numpy-quaternion`. It is already part of pixell's dependencies.

https://github.com/simonsobs/pixell/blob/master/pyproject.toml#L23

iparask (approved):
LGTM

PR #1617, opened 2026-04-21
feat: adding some unittests

iparask:
These are a couple of tests for the ML mapmaker utils. Claude helped me with the `find_usable_detectors` tests.

Can you both have a look and let me know if they make sense?

Thank you


chervias (approved):
looks fine to me

mhasself (approved):

PR #1619, opened 2026-04-23
Fix for smurf_caldbs failures

mmccrackan:
This is to address recent failures on the `update_smurf_caldbs` flows: https://prefect.simonsobs.org/runs/flow-run/019db6ff-db7d-7ea5-8870-48760e00177b


kmharrington (approved):
nice!

PR #1620, opened 2026-04-24
Update source flags handling

mmccrackan:
Updates source flags handling to only create a mask centered on the point source instead of being centered at (0, 0) which causes LAT preprocessing to fail due to the mask being too large.  Thanks to Matthew Hasselfield for helping work through this.


mmccrackan:
Fell behind on this but have now tested with the `demod_mapmaker` and `ml_mapmaker` doing the following:

**Ran a LAT CMB obs_id that failed in the previous version through preprocessing and the `ml_mapmaker` with the new version:**

obs_id: obs_1777232557_lati1_111
wafer: ws0, ws1, ws2
band: f090
planet: Jupiter

Confirmed it no longer fails and Jupiter is successfully masked.

**Ran a LAT CMB obs_id that succeeded in the old version through preprocessing and the `ml_mapmaker` with the new version and compared both versions:**

obs_id: obs_1781618914_lati1_111
wafer: ws0, ws1, ws2
band: f090
planet: Saturn

Subtracted both maps and confirmed they were identical.

**Ran SAT CMB TODs with the Moon in them (thanks to Kyohei for providing a list) through preprocessing and the `demod_mapmaker` with the new version and compared versions:** 

obs_id: obs_1764446058_satp1_1111111
wafer: ws6
band: f150
roll: 0 deg

obs_id: obs_1767953161_satp1_1011111
wafer: ws3
band: f150
roll: 45 deg


Subtracted both maps and confirmed they were identical.

mhasself (approved):

PR #1622, opened 2026-04-26
[Preprocess] Global AzSS Template

msilvafe:
Adds functions to be able to regress a global AzSS template provided by the metadata from the data. Currently a WIP - haven't tested yet just wanted to get initial code up to discuss.

### DIFF
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -1386,6 +1386,94 @@ class AzSS(_Preprocess):
         else:
             return keep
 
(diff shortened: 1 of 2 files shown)


msilvafe:
In discussion with @ykyohei we decided on a few changes needed:
1. Add a pre filter as is done in get_azss
2. We won't save anything to the archive so all steps go in process so that they're applied to both data + sim.
3. Need to add something to extract the right field out of the template metadata currently wafer_band axis labels.
4. Get rid of bad detector treatment so no calc or select needed.

PR #1625, opened 2026-04-29
Fix reindexing of fixed tones

tpsatt:
Loading an observation with fixed tones (`special_channels`) and reindexing was broken because multiple UFMs (`stream_id`s) could share the same band/channel numbers. This fixes that, and fixes a bug where NaNs were not propagated in the reindexing.

### DIFF
--- a/sotodlib/core/axisman.py
+++ b/sotodlib/core/axisman.py
@@ -877,7 +877,7 @@ class AxisManager:
                         shape.append(s)
 
                 new_v = np.empty(shape, dtype=v.dtype)
-                if isinstance(v.dtype, float):
+                if v.dtype == float:
                     # Fill any float arrays with nans
                     # Non float arrays may have weird
                     # behavior for newly added indexes. 
--- a/sotodlib/core/axisman.py
+++ b/sotodlib/core/axisman.py
@@ -882,7 +882,11 @@ class AxisManager:
                     # Non float arrays may have weird
                     # behavior for newly added indexes. 
                     # Oh well.
-                    new_v *= np.nan  
+                    new_v = np.full(shape, np.nan)
+                elif v.dtype == int:
+                    # Fill with -1 values which generally
+                    # correspond ot unassigned (e.g. bias lines)
+                    new_v = np.full(shape, -1)
 
                 for i, index in enumerate(indexes):
                     if np.isnan(index) or not (0 <= index < len(v)):
--- a/sotodlib/core/axisman.py
+++ b/sotodlib/core/axisman.py
@@ -876,7 +876,6 @@ class AxisManager:
                     for s in np.shape(v)[1:]:
                         shape.append(s)
 
-                new_v = np.empty(shape, dtype=v.dtype)
                 if v.dtype == float:
                     # Fill any float arrays with nans
                     # Non float arrays may have weird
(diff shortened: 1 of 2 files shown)


mhasself:
@tpsatt Thanks for working this out.  Since unit tests on this kind of thing are pretty tricky, can you mention in the PR comments here what obs_id(s) you used for testing?

tpsatt:
@mhasself Good point. For testing, I was using `obs_id` `obs_1750380829_lati6_111` which had fixed tones enabled. Before the patch, the data beyond just signal got scrambled after reindexing since the channel matching returned multiple results due to not using `stream_id`. Now, everything lines up, and all of those fields go to empty strings, NaNs, or anomalously large/small ints for fixed tones (which lack that info).

ykyohei (commented):
This looks reasonable to me, and I tested to run update_det_cal site-pipeline with this branch and confirmed that there is no problem, just in case.

This reindexing was used for update_det_cal with hwpss_subtraction for SATs. update_det_cal should not have had issues because it analyzes oper books that is saved per wafer.

sotodlib/core/axisman.py
```diff
@@ -877,7 +877,7 @@ def reindex_axis(self, axis, indexes, in_place=True):
                         shape.append(s)
 
                 new_v = np.empty(shape, dtype=v.dtype)
```
mhasself: Can we use zeros, or zeros() - 1, instead of empty here?

The idea that "empty" will always produce garbagey results that are easy to distinguish from "good numbers" is not accurate.  Sometimes "empty" will give you 0, or -1, or other normal numbers.

So I'd rather we have the _predictability_ of a simple result here, such as zero or -1 for integers, than to rely on whatever garbage is given by empty.

tpsatt: Good point. I've just pushed a change which sets floats to NaN and ints to -1. I think this is minimally invasive since many bad float values in "real" data (e.g. R_ns) are already NaNs, and bad ints (e.g. bias lines) are -1s. This also preserves strings being ''.

tpsatt:
@ykyohei should probably review again since I've tweaked the reindexing function

sotodlib/core/axisman.py
mhasself: Thanks.  Given that new_v is now simply replaced in the int and float specializations, you can skip the .empty creation here and instead do a .zeros in an "else" clause for the type checks.

I really don't like leaving .empty arrays lying around!  Non-determinism.

tpsatt: Sometimes we have strings in the data, so I like that `empty` becomes an empty string. When I tried to put a case in for this, it broke (I think because of how numpy encodes strings, I'm not an expert here)

mhasself: .zeros has always worked on strings for me ... also returns an array of ''s ... What's the dtype that fails?

tpsatt: I see your point now. I've updated this and committed. Thanks!

tskisner:
The unit test failures here are due to an interplay between the latest toast wheels and the so3g wheels.  I'm debugging [in a separate PR](https://github.com/simonsobs/sotodlib/pull/1629).

ykyohei:
I tested update_det_cal again and I didn't find problem.

tpsatt:
Is this ready to merge now that #1629 has converged? (@tskisner)

PR #1626, opened 2026-04-30
Add ScanAlm to the suite of sky signal samplers

keskitalo:


tskisner:
This is fine, but do we have a plan for deprecating the `sotodlib.toast.workflows` package in favor of just using config files with either `toast_run` or `so-toast-run`?

keskitalo:
Yes, that is the long term plan but we are already running MSS3 with `toast_so_sim` and want to switch map sampling for alm sampling.

tskisner (approved):
Looks good to me.

PR #1627, opened 2026-05-01
Account for tube orientation in det match

mmccrackan:
Adds support for #1603 into det match to properly account for ASO tube orientation.

Adds matching support for LF wafers all at once.  If input wafer is `ws.`, it will first check the imprinter file and find the `wafers` and `wafer_slots`.  It will then get the `wafer_info` for each of those wafers, select `det_id's` that are for the correct wafer, store these as a list, then merge all of them at the end.  The loads the full `wafer_info` and puts the `xi`, `eta`, and `wafer_slot` fields in the correct rows to maintain the `det_id` order (order is important) Also adds `ignore_north_south` option since this is not relevant for LF.

Removed the book check for `wafer_slot` as the field is now in the axis managers anyway.  Defaults to `ws.` since that was chosen for LF for when we don't know the wafer.

Now populates `wafer_slot` to the `det_match` metadata since this will likely be needed downstream.

Updated optics zemax file reading to work with the new file that was produced to fix the edge effects on o6.


tskisner:
The unit test failures here are due to an interplay between the latest toast wheels and the so3g wheels.  I'm debugging [in a separate PR](https://github.com/simonsobs/sotodlib/pull/1629).

mhasself (approved):
Sorry for the delay.

There are now lots of "ignore_north_south" getting passed around.  Is there perhaps an easier to capture this (and avoid so much kwarg copying) e.g. just by setting the value of some field, in supporting metadata, to all have the same value?

PR #1628, opened 2026-05-02
Add modes_axis_name option for azss legendre fit to avoid name conflict

ykyohei:
We would like to use different legendre modes for T and demodQU.
Due to the name conflict, this was not possible. This PR resolves it.


msilvafe (approved):
Successfully tested this on the SATp3 V4 paper dataset. 

Just to note. This Legendre fitting method requires more memory than other methods for reference when running on tiger we usually run with 2 CPUs per task (or 56 tasks per node) which allows each task to use 1000GB/56 = 17.9 GB when using Legendre fitting you must use at least 3 CPUs per task (37 tasks per node) which leaves ~27 GB per task.

PR #1629, opened 2026-05-05
Update test runner to use stable toast release

tskisner:


tskisner:
From local tests in a virtualenv, running the test in gdb I get a stacktrace like:
```
Thread 1 "python3" received signal SIGSEGV, Segmentation fault.
0x00007fffe68d480a in daxpy_k_COOPERLAKE () from /home/kisner/sotest/lib/python3.10/site-packages/so3g/libso3g.so
(gdb) where
#0  0x00007fffe68d480a in daxpy_k_COOPERLAKE () from /home/kisner/sotest/lib/python3.10/site-packages/so3g/libso3g.so
#1  0x00007fffc2767e6f in dsytd2_ () from /home/kisner/sotest/lib/python3.10/site-packages/toast/_libtoast.cpython-310-x86_64-linux-gnu.so
#2  0x00007fffc22074b3 in dsytrd_ () from /home/kisner/sotest/lib/python3.10/site-packages/toast/_libtoast.cpython-310-x86_64-linux-gnu.so
#3  0x00007fffc22070fe in dsyev_ () from /home/kisner/sotest/lib/python3.10/site-packages/toast/_libtoast.cpython-310-x86_64-linux-gnu.so
#4  0x00007fffc09657d7 in ?? () from /home/kisner/sotest/lib/python3.10/site-packages/toast/_libtoast.cpython-310-x86_64-linux-gnu.so
#5  0x00007fffe1a11726 in GOMP_parallel () from /home/kisner/sotest/lib/python3.10/site-packages/so3g/../so3g.libs/libgomp-e985bcbb.so.1.0.0
#6  0x00007fffc0965b15 in toast::LinearAlgebra::syev_batched(char, char, int, double*, int, double*, int*, int) ()
   from /home/kisner/sotest/lib/python3.10/site-packages/toast/_libtoast.cpython-310-x86_64-linux-gnu.so
#7  0x00007fffc097b5b7 in toast::cov_eigendecompose_diag(long, long, long, double*, double*, double, bool) ()
   from /home/kisner/sotest/lib/python3.10/site-packages/toast/_libtoast.cpython-310-x86_64-linux-gnu.so
#8  0x00007fffc0a9ddaa in ?? () from /home/kisner/sotest/lib/python3.10/site-packages/toast/_libtoast.cpython-310-x86_64-linux-gnu.so
```
So somehow the `libgomp` bundled with so3g is loaded first and then the toast lapack call ends up using this libgomp rather than the one bundled with the toast wheel.  This should never happen and is one of the features of using cibuildwheel (bundling dependency libs into the wheel with different names seen by the loader).  I'm investigating locally.

tskisner:
The hashes of the vendored `libgomp` in the toast and so3g wheels are identical, which just means that within the build environment these libraries had the same checksums (which makes sense since both are using the same cibuildwheel docker containers for the build).  And so there should in principle be no problem with this scenario (since the two vendored libraries should be identical).

tskisner:
I have been exploring this issue more, but still no resolution.  Since the primary package that has changed is the toast wheel, I'm looking at differences since the last wheel.  Those include:

- Bumped version of cibuildwheel used
- Bumped version of vendored OpenBLAS to 0.3.33
- Small changes to the cmake files

I'll continue to post updates here as I work to resolve this.

tskisner:
I found a workaround for this by reverting the toast wheel builds to use a vendored version of OpenBLAS 0.3.29, which matches the version vendored with so3g.  I will tag a new toast release with this change as a stop-gap.  Going forward, this situation is not sustainable. We need to find a longer-term solution to using threads in compiled extensions in wheels that play nice with compiled extensions in other wheels.  Free-threaded python will improve things, and packages like omp4py:

https://github.com/citiususc/omp4py
https://arxiv.org/html/2411.14887v1

might let us skip compiled code completely.

tskisner:
Also just a reminder:  once `so3g` is on conda-forge, then the compatibility between toast, so3g, scipy, etc will be guaranteed by construction (when using a conda environment instead of wheels).

chervias (approved):
it worked. Thanks @tskisner

PR #1630, opened 2026-05-08
Changed local median signal calculation to global median

ttsan2521:
In sim_mumux_crosstalk.py, changed the local median signal calculation when evaluating dPhi0/dT to a global median calculation. In the original script, it calculates the median of the input signal after an MPI distribution, which put all the detectors into bins or ranks where each rank has a subset of the sample. So all the detectors inside a rank only see a portion of the scan, and the calculated median within each rank may differ from rank to rank, which can cause a scan-like pattern as the crosstalk operator is implemented. A way to go about this is to take the median before the distribution or a global median.

### DIFF
--- a/sotodlib/toast/ops/sim_mumux_crosstalk.py
+++ b/sotodlib/toast/ops/sim_mumux_crosstalk.py
@@ -10,6 +10,7 @@ from toast.ops.operator import Operator
 from toast.timing import function_timer, Timer
 from toast.traits import Bool, Int, Unicode, trait_docs
 from toast.utils import Environment, Logger, unit_conversion
+from toast.mpi import MPI #added this import- Tran
 
 from .mumux_crosstalk_util import detmap_available, pos_to_chi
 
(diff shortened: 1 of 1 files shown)


sotodlib/toast/ops/sim_mumux_crosstalk.py
```diff
@@ -10,6 +10,7 @@
 from toast.timing import function_timer, Timer
 from toast.traits import Bool, Int, Unicode, trait_docs
 from toast.utils import Environment, Logger, unit_conversion
+from toast.mpi import MPI #added this import- Tran
```
keskitalo, line 13: We can leave out the comment here.  If we need to figure out who added which line, there is always `git blame`.

sotodlib/toast/ops/sim_mumux_crosstalk.py
```diff
@@ -205,7 +206,18 @@ def _evaluate_dPhi0dT(self, obs, signal, detectors, rows, Phi0):
             )
             #import pdb
             #pdb.set_trace()
-            median_signal = np.median(signal[row][good])
+            ## Calculate a global median 
+            local_good = signal[row][good]
+            if obs.comm.comm_group is not None:
+                all_good = obs.comm.comm_group.gather(local_good, root=0)
```
keskitalo, line 212: This gets the job done but can be slow if there are lots of detectors and processes. It would be more expedient to compute and store the medians in the beginning of `exec()` before the data are redistributed.

Since I am not running this operator myself, I leave it up to Tran and Toby to decide if they want to develop the more efficient computation.

sotodlib/toast/ops/sim_mumux_crosstalk.py
```diff
@@ -328,6 +340,7 @@ def _exec(self, data, detectors=None, **kwargs):
                 # Propagate flags
                 obs.detdata[self.det_flags][det] |= temp_obs.detdata[self.det_flags][det]
 
+
```
keskitalo, line 343: Looks like an extra line change.

keskitalo (approved):

sotodlib/toast/ops/sim_mumux_crosstalk.py
tpsatt: Confirming that in my environment, the added run time seems to be negligible.

tpsatt:
Apologies for the delay on my end. I've finally been able to get this tested, and I don't see any issues. Added run time seems negligible.

tskisner:
The test failures here are unrelated (due to API changes in `numdifftools` in other parts of the code).  I think you are fine to squash and merge.

PR #1634, opened 2026-05-13
Small fixes for so-toast-run

tskisner:


sotodlib/toast/ops/readout_filter.py
```diff
@@ -78,6 +78,7 @@ def _exec(self, data, detectors=None, **kwargs):
             if self.iir_params not in ob:
                 msg = "Cannot apply readout filter. "
                 msg += f"'{self.iir_params}' does not exist in '{ob.name}'"
+                log.error(msg)
```
keskitalo, line 81: Doesn't `log.error()` already terminate the program? We should decide which is our canonical solution.

sotodlib/toast/ops/readout_filter.py
```diff
@@ -111,27 +124,53 @@ def _exec(self, data, detectors=None, **kwargs):
...
+                ob.comm.comm_group.barrier()
 
     @function_timer
-    def _filter_detectors(self, rate, freq, det_array, iir_props, flags, mask):
+    def _filter_detectors(self, ob, rate, freq, det_array, iir_props, flags, mask):
         # Get the common filter kernel for all detectors
         iir_filter = filters.iir_filter(iir_params=iir_props)(freq, None)
 
+        log = Logger.get()
+        msg = f"DBG rank {ob.comm.group_rank}: {det_array} {flags} {iir_filter}"
```
keskitalo: Leftover debug statement?

keskitalo (approved):
Mostly improved verbosity. Merge away.

sotodlib/toast/ops/readout_filter.py
tskisner: Thanks, fixed.

sotodlib/toast/ops/readout_filter.py
tskisner: `log.error()` just prints messages at ERROR log level, but it uses the toast logger which writes to C stdout with a flush of the low-level filehandle.  Sometimes this message gets through to the log file even when the exception doesn't.

PR #1637, opened 2026-05-15
Implement interpolation for filterbin mapmaker

chervias:


chervias:
Ok, we are scaling back the PR. We won't allow for tod->map with bilinear since for filterbin it depends on the white noise assumptions. We will only keep the map->tod operation to be potentially made with bilinear.

amaurea:
@chervias I think it's more accurate to say that it depends on P'WP being diagonal, which it isn't with bilinear interpolation even if W is diagonal, where W is the weighting matrix. For maximum-likelihood mapaking that doesn't matter, since P'N"P wouldn't be diagonal anyway.

amaurea (approved):
Looks fine. Apparently the script that uses these new arguments is in another repository.

PR #1640, opened 2026-05-21
Make a lightweight script for bootstrapping a dev env

tskisner:
This PR:

- Adds a small script (`conda_dev_setup.sh`) that can create a conda environment for SO development.  This is a lightweight version of soconda, but without module files and without installation of moby2 and the ACT tools.
- Fixes the macos github runner config.


mhasself (approved):
Thanks!  Squash!

PR #1644, opened 2026-05-24
update how HK database delays are processed

kmharrington:
We had a hardcoded 1 hour delay as the acceptable limit for HK database updates compared to when imprinter was searching for new books. This is getting broken with the new HK suprsyncs.

This fixes that and moves the logic for where the error message gets thrown. We'll match the warning / error timeouts to what is used for incomplete observations.

### DIFF
--- a/sotodlib/io/imprinter.py
+++ b/sotodlib/io/imprinter.py
@@ -1231,7 +1231,9 @@ class Imprinter:
             books. Useful as a debugging tool
         incomplete_timeouts: tuple
             hours passed for (attempting to complete observation, raising a error) 
-            if incomplete observations are found in the g3tsmurf database
+            if incomplete observations are found in the g3tsmurf database. Also used
+            as limits to decide if the g3tsmurf database is stale and raise warnings/
+            errors accordingly.
         """
         if not self.build_det:
             return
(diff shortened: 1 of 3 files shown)


sotodlib/io/imprinter.py
```diff
@@ -1262,8 +1264,20 @@ def update_bookdb_from_g3tsmurf(
         final_time = SMURF.get_final_time(
             streams, min_ctime, max_ctime, check_control=True
         )
+        ## this will catch suprsync drops and hk/g3tsmurf databases not being updated
         if final_time < max_ctime:
+            if max_ctime - final_time > 3600*incomplete_timeouts[1]:
```
JBorrow, line 1269: For clarity:
```suggestion
            if (max_ctime - final_time) > (3600 * incomplete_timeouts[1]):
```

sotodlib/io/imprinter.py
```diff
@@ -1262,8 +1264,20 @@ def update_bookdb_from_g3tsmurf(
...
+                raise ValueError(
+                   f"G3tSMURF + HK databases are stale. Last updates were "
+                    f"more than {incomplete_timeouts[1]} hours in the past."
+                )
+            if max_ctime - final_time > 3600*incomplete_timeouts[0]:
```
JBorrow, line 1274:
```suggestion
            if (max_ctime - final_time) > (3600 * incomplete_timeouts[0]):
```

sotodlib/io/load_smurf.py
```diff
@@ -1486,8 +1486,22 @@ def get_final_time(
...
         if session is None:
             session = self.Session()
-        
+        if stop is None:
+            stop = dt.datetime.now().timestamp()
+            
         HK = self.get_HK()
+        last_update = HK.get_last_update()
+        if stop > last_update:
+            if stop > last_update + 2*3600:
```
JBorrow, line 1495:
```suggestion
            if stop > (last_update + 2 * 3600):
```

JBorrow (approved):
Some minor clarity patches but looks good

PR #1646, opened 2026-05-25
Sdw filter for sources pca map

samdayweiss:
Allows option to wrap calcualted pca modes and weights to the axis manager when calling filter_for_sources. We want this for the planet map maker so we can map the modes being removed. Also changes the process.py file accordingly to be able to use this new feature in a preprocess configuration file.

### DIFF
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -2382,6 +2382,7 @@ class FilterForSources(_Preprocess):
           source_flags: "source_flags"
           edge_guard: 10 # Number of samples to make the first and last flags False
           trim_samps: 100
+          pca_wrap: "pca_model" # optional, if provided, the PCA model is wrapped into proc_aman under this name
 
     .. autofunction:: sotodlib.coords.planets.filter_for_sources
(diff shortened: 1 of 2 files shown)


sotodlib/preprocess/processes.py
```diff
@@ -2382,6 +2382,7 @@ class FilterForSources(_Preprocess):
           source_flags: "source_flags"
           edge_guard: 10 # Number of samples to make the first and last flags False
           trim_samps: 100
+          pca_wrap: "pca_model" # optional, if provided, the PCA model is wrapped into proc_aman under this name
```
mmccrackan, line 2385: It is wrapped under `aman` not `proc_aman` right?

mmccrackan (approved):
Looks good to me, just one point regarding the docstring.

sotodlib/preprocess/processes.py
samdayweiss: ah shoot I totally didn't see this comment, yes it is under aman, but isn't that only after proc_aman gets wrapped into aman?

PR #1648, opened 2026-05-27
Sdw filter for sources pca map

samdayweiss:


samdayweiss:
sorry @mmccrackan, fumbled a bit here because I clicked revert which opened up a second PR, but instead I just fixed the doc string here on a new commit, let me know if things look okay

mmccrackan (approved):
Yeah docstring change looks good.

PR #1649, opened 2026-05-28
Improvements to LoadContext and wobble correction operator

tskisner:
- Update LoadContext to load hwp, boresight rotation, and corotator angles by default if they exist.  Additional small fixes.

- Implement new operator to apply wobble correction metadata to the boresight pointing.

### DIFF
--- a/sotodlib/toast/ops/__init__.py
+++ b/sotodlib/toast/ops/__init__.py
@@ -14,6 +14,7 @@ from .sim_catalog import SimCatalog
 from .sim_hwpss import SimHWPSS
 from .sim_source import SimSource
 from .h_n import Hn
+from .hwp_wobble import HWPWobbleCorrect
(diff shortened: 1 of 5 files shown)


AERAdler (commented):
Has the HWP wobble operator been tested?

tskisner:
For testing, I only tested in a notebook with satp1 data.  The implementation copies the [order of quaternion rotations in the code here](https://github.com/simonsobs/sotodlib/blob/9baa1f1c9d9b499a793d60af251d0e33ba1b9374/sotodlib/coords/helpers.py#L779-L797).  I guess I could test this artificially (without access to real data in CI tests) by adding fake wobble parameters in the observation and also manually calling the so3g function.  I'll go do that.

AERAdler (approved):

PR #1651, opened 2026-05-29
Errors as values update book plan

### LINKED ISSUE #1650: Data Packaging: update_book_plan should check all tubes before throwing errors
The finalization time error in `update_book_plan` gets thrown as soon as the script encounters a tube that has stale suprsyncs or a stale HK db. 

But, especially for the LAT, we're looping over tubes and so one crate going down pauses data-packaging for all other tubes.

Update the script / functions so that we check all the tubes, update the monitor, and _then_ throw the error that >=1 of the tubes has gone stale.

JBorrow:
Allows progress on other tubes for `update_book_plan` even when we error out for a specific tube. Fixes #1650.

Also includes updates from [monitor-and-profile-update_g3thk_database](https://github.com/simonsobs/sotodlib/compare/master...monitor-and-profile-update_g3thk_database) which allows for monitoring and profiling of the g3thk_database flow.

### DIFF
--- a/sotodlib/io/bookbinder.py
+++ b/sotodlib/io/bookbinder.py
@@ -1576,7 +1576,10 @@ def get_hk_files(hkdir, start, stop, tbuff=10*60):
             continue
 
         subpath = os.path.join(hkdir, subdir)
-        files.extend([os.path.join(subpath, f) for f in os.listdir(subpath)])
+        files.extend([
+            os.path.join(subpath, f) for f in os.listdir(subpath)
+            if 'suprsync' not in f
+        ])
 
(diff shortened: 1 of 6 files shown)


JBorrow:
My only concern here is whether this causes other problems further down the line, I don't think so as the tubes should be completely independent and they get bound into separate books?

kmharrington (approved):
the timecode and hk books have independent checking to decide if they're ready to be registered, so I don't expect this to cause any errors farther down

JBorrow:
Given that this is running happily, I am ok to merge at this stage.

PR #1653, opened 2026-06-01
fix: sanatize scalars before saving

skhrg:


mhasself (approved):
I guess this is ok ... this does lose information -- maybe someone _wants_ to store a float32 scalar... but let's just go with it.

skhrg:
Agreed... I think it may be worth revisiting the entire save/load code to check if there are any performance bottlenecks (I remember Sigurd had found it to be slower than he wanted), so when we do that maybe we can see if there is a cleaner way of dealing with this (ie. drop the `json` stuff and store things as individual attrs or something).

PR #1655, opened 2026-06-05
Fix early return in webhook loop.

JBorrow:
The webhook loop always returned after the first webhook was sent. This is why alerts weren't propagating to the instrument-specific channels.


kmharrington (approved):
looks good

PR #1656, opened 2026-06-06
Data Packaging: Logging and Autofixing Improvements

kmharrington:
Add a way to autofix books when then azimuth mode has a non-monotonic data point. I think this happens when the ACU agent is updated. 

Add more/improved logging information to g3tsmurf database updates. I've left the tqdm in there for now since prefect suppresses it but I use it sometimes in scratch / dev work.

### DIFF
--- a/sotodlib/io/load_smurf.py
+++ b/sotodlib/io/load_smurf.py
@@ -558,7 +558,7 @@ class G3tSmurf:
         ## files must be updated in sequencial order. otherwise we may end up
         ## with more TuneSets than are necessary
         for i,f in tqdm(enumerate(sorted(files)), disable=(not show_pb)):
-            if i>0 and np.mod(i,25)==0:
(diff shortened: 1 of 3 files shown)


sotodlib/io/load_smurf.py
```diff
@@ -557,7 +557,9 @@ def index_archive(
 
         ## files must be updated in sequencial order. otherwise we may end up
         ## with more TuneSets than are necessary
-        for f in tqdm(sorted(files), disable=(not show_pb)):
+        for i,f in tqdm(enumerate(sorted(files)), disable=(not show_pb)):
+            if i>0 and np.mod(i,25)==0:
```
JBorrow: I think it's ok to print once when you have indexed zero files...


```suggestion
            if (i % 25) == 0:
```

JBorrow (approved):

PR #1659, opened 2026-06-10
Stimulator analysis pipeline

YudaiSeino:
I wrote stimulator analysis code. Main functions are `calc_gain()` and `calc_timeconstant()`. `calc_gain()` calculates stimulator's signal amplitude. `calc_timeconstant()` calculates optical time constant and readout delay.


sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,967 @@
...
+    
+    Args:
+        aman: Axis manager with the detector data.
+        hkdata: The housekeeping data.
+    
+    Returns:
+        bool: True if timing data is valid, False otherwise.
+    """
+    state = np.array(hkdata.data['stimulator-enc.stim_enc.state'][1])
+    t_enc = np.array(hkdata.data['stimulator-enc.stim_enc.timestamps_tai'][1])[state == 0]-37
```
ykyohei: what is this -37?

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,967 @@
...
+        fields=feed_list,
+        start=t_start, 
+        end=t_end,
+    )
+    result = hkdb.load_hk(lspec) 
+
+    return result
+    
+
+def get_downsample_factor(aman, ctx):
```
ykyohei: why don't we simply do

sampling_rate = 1 / np.median(np.diff(aman.timestamps))
downsample_factor = int(4000 / sampling_rate)

Then we don't need `get_downsample_factor_tags`

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,967 @@
...
+    Return:
+        bool: True if data is valid and calculation is performed, False otherwise.
+    """
+    det_mask = np.full(aman.dets.count, False)
+    if idxs is None:
+        det_mask[:] = True
+    else:
+        det_mask[idxs] = True
+
+    vailed_data = True
```
ykyohei: vailed_data -> valid_data same typo in calc_timeconstant

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,967 @@
...
+        List of downsample factor tags in the database.
+    """
+    cursor = ctx.obsdb.conn.execute("SELECT DISTINCT tag FROM tags")
+    all_tags = np.array([row[0] for row in cursor.fetchall()])
+    mask = np.char.find(all_tags,'downsample') != -1
+    
+    return np.array(all_tags)[mask].tolist()
+    
+
+def calc_gain(aman, hkdata, idxs=None, bool_plot=False, bool_save=False, bool_preprocess=True, output_dir=None):
```
ykyohei: I think it's more standard in sotodlib to not to have bool_ suffix in arguments.
For example, plot=False, save=False.

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,967 @@
...
+
+
+    filtering_params = filtering(aman,chopping_freqs,'gain')
+
+
+    model, params_base = get_fit_params(cal_type='gain')
+
+    # Make and get co-added data
+    coadd_data, fit_result = get_dicts('gain')
+    n_bins = 40
```
ykyohei: n_bins should be arguments of calc_gain

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,967 @@
...
+            plot_tod(aman,i_det,coadd_data,fit_result,filtering_params,cal_type='gain')
+            
+            if bool_save:
+                obs_id = aman.obs_info.obs_id
+                if output_dir is not None:
+                    output_dir_ = Path(f'{output_dir}/{ufm}_{obs_id}')
+                    output_dir_.mkdir(parents=True, exist_ok=True)
+                    plt.savefig(f'{output_dir_}/Gain_det{i_det:04d}.png')
+            if bool_plot==False:
+                plt.close(fig)
```
ykyohei: fig does not exist

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,967 @@
...
+                params['t6'].set(value=0,vary=False)
+                result = model.fit(y, params, t=x, weights=1/np.array(yerr)) 
+            else:
+                result = model.fit(y, params, t=x, weights=1/np.array(yerr)) 
+
+            fit_result['fit_coadd'][filt_key]['f1_gain'].append(result)
+
+        if bool_plot==False and bool_save==False:
+            pass
+        else:
```
ykyohei:
```suggestion
        if bool_plot:
```

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,967 @@
...
+        cal_type = 'gain'
+    elif len(aman.stm_ana.t_chopping_change)>2:
+        cal_type = 'gain_timeconstant'
+    else:
+        ValueError('Chopping information is incorrect.')
+
+
+    t_cuts = []
+
+    t_start = 2+aman.stm_ana.t_chopping_change[0]  # 2 seconds to avoid ringing by high-pass filter
```
ykyohei: I suggest to make this 2 seconds buffer to arguments of get_timing_cut

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,967 @@
+import numpy as np
+import matplotlib.pyplot as plt
+import lmfit
+from matplotlib.patches import Wedge
+from matplotlib.collections import PatchCollection
```
ykyohei: These are unused import

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,967 @@
...
+            chopping_freqs[key] = round(aman.stm_ana.chopping_freqs[i])
+        else:
+            chopping_freqs[key] = f
+
+    filtering_params = filtering(aman,chopping_freqs,'timeconstant')
+
+    models, params_bases = get_fit_params(cal_type='timeconstant')
+    
+    coadd_data, fit_result = get_dicts('timeconstant')
+    n_bins = 40
```
ykyohei: n_bins should be arguments of calc_timeconstant

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,967 @@
...
+                    plt.savefig(f'{output_dir_}/Gain_det{i_det:04d}.png')
+            if bool_plot==False:
+                plt.close(fig)
+            
+    fill_data(aman,coadd_data,fit_result,n_bins,cal_type='gain')    
+    
+    return vailed_data
+
+
+def calc_timeconstant(aman, hkdata, idxs=None, bool_plot=False, bool_save=False, bool_preprocess=True, output_dir=None):
```
ykyohei: We don't need bool_ suffix

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,967 @@
...
+            for fit_key in ['fit_amp', 'fit_phase__fix_tau', 'fit_phase__free']:
+                params = params_bases[fit_key].copy()
+                if fit_key == 'fit_amp':
+                    if np.isnan(a0s).all():
+                        fit_result[fit_key][filt_key].append(None)
+                        continue
+
+                    weights = [fit_result['fit_coadd'][filt_key][f_key][-1].params['a0'].stderr if fit_result['fit_coadd'][filt_key][f_key][-1] is not None and fit_result['fit_coadd'][filt_key][f_key][-1].params['a0'].stderr is not None else np.nan for f_key in fit_result['fit_coadd'][filt_key].keys()]
+                    weights = np.array(weights)
+                    #result = models[fit_key].fit(a0s,params,f=f,weights=weights, method='least_squares')
```
ykyohei: Please delete commented out lines!

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,967 @@
...
+                            result = None
+                    else:
+                        #result = models[fit_key].fit(-np.array(t0s)*360,params,f=f, weights=weights, method='least_squares')
+                        result = models[fit_key].fit(-np.array(t0s)*360,params,f=f, method='least_squares')  # No weight for first step analysis
+                fit_result[fit_key][filt_key].append(result)
+        
+        
+        if bool_plot==False and bool_save==False:
+            pass
+        else:
```
ykyohei:
```suggestion
        if bool_plot:
```

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,967 @@
...
+            plot_tod(aman,i_det,coadd_data,fit_result,filtering_params,cal_type='timeconstant')
+
+            obs_id = aman['obs_info']['obs_id']
+            if bool_save:
+                output_dir_ = Path(f'{output_dir}/{ufm}_{obs_id}')
+                output_dir_.mkdir(parents=True, exist_ok=True)
+                plt.savefig(f'{output_dir_}/Tau_det{i_det:04d}.png')
+            if bool_plot==False:
+                plt.close(fig)
```
ykyohei: fig doesn't exist

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,967 @@
...
+
+        if bool_plot==False and bool_save==False:
+            pass
+        else: 
+            plot_tod(aman,i_det,coadd_data,fit_result,filtering_params,cal_type='gain')
+            
+            if bool_save:
+                obs_id = aman.obs_info.obs_id
+                if output_dir is not None:
+                    output_dir_ = Path(f'{output_dir}/{ufm}_{obs_id}')
```
ykyohei: ufm is undefined

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,967 @@
...
+    
+
+    if not 'stm_samps' in aman._axes.keys():
+        stm_samps = core.IndexAxis('stm_samps',t_enc.size)
+        stm_ana = core.AxisManager(stm_samps)
+        aman.wrap('stm_ana',stm_ana,overwrite=True)
+    aman.stm_ana.wrap('t_enc', t_enc, [(0,'stm_samps')], overwrite=True)
+    aman.stm_ana.wrap('t_hk', t_hk, [(0,'stm_samps')], overwrite=True)
+
+    vailed_data = True
```
ykyohei: vailed -> valid

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,967 @@
...
+    if t_enc[-1] < aman.timestamps[-1]:
+        print(f'Invailed HK data. last t_enc: {t_enc[-1]}, last TOD timestamp: {aman.timestamps[-1]}')
```
ykyohei: Invaled -> Invalid

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,967 @@
...
+        chopping_freqs: dictionary of chopping frequencies
+    """
+
+    #Invert IIR filter
+    iirc_filter = tod_ops.filters.iir_filter(aman, invert=True)
+    signal_new = tod_ops.fourier_filter(aman, iirc_filter, signal_name='signal')
+    aman.wrap(f'signal_iirc', signal_new, [(0,'dets'),(1,'samps')], overwrite=True)
+
+    # Make HPFed data
+    hpf_cutoff = 1
```
ykyohei: I suggest to make these hardcoded cutoffs (hpf_cutoff, cutoff_factor, width_fraction) to arguments of filtering so that we can modify easily.

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,967 @@
...
+
+    else:
+        raise ValueError(f"'{cal_type}' is a wrong type. Please specify 'gain' or 'timeconstant'.")
+
+    filtering_params = {'hpf_cutoff': hpf_cutoff, 'lpf_cutoff_factor': cutoff_factor, 'lpf_width_fraction': width_fraction, 'chopping_freqs': chopping_freqs}
+
+    return filtering_params
+
+
+def filtering__with_delay_filter(aman, filter, signal_name_pre, signal_name_new):
```
ykyohei: sine2 filters implemented in sotodlib doesn't have phase delay, and we don't need to apply fourier filter twice to cancel phase delay.
I think filter with delay is not necessary for stimulator so I suggest to remove it.

No need to wrap and overwrite twice.

Also I think the name of two functions (with_delay and without_delay) are flipped.. Why do we have double under score?

sotodlib/stimulator/utils_stimulator.py
```diff
@@ -0,0 +1,76 @@
...
+
+def func_response_phase(f, tau, theta_geo):
+    """
+    Detector response function of phase.
+    
+    Args:
+        f: Chopping frequency [Hz]
+        tau: time constant of a detector in [s]
+        theta_geo: Offset of phase delay [deg]
+        theta: Phase delay of stimulator signal [deg]
```
ykyohei: this is return value!

sotodlib/stimulator/utils_stimulator.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def func_response_phase_with_dt(f, tau, theta_geo, dt):
+    """
+    Detector response function of phase.
+    
+    Args:
+        f: Chopping frequency [Hz]
+        tau: time constant of a detector in [s]
+        theta: Phase delay of stimulator signal [deg]
```
ykyohei: theta is return value!

sotodlib/stimulator/utils_stimulator.py
```diff
@@ -0,0 +1,76 @@
...
+    """
+    Detector response function of amplitude.
+    
+    Args:
+        f: Chopping frequency
+        tau: time constant of a detector in [s]
+        a: Amplitude of sin function for '0' frequency signal
+
+    Return:
+        y: Value of fitting function.
```
ykyohei: amplitude at chopping frequency

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,967 @@
...
+        aman.wrap('det_cal', core.AxisManager(aman._axes['dets']))
+
+    if cal_type == 'gain':
+        idx = np.where(aman.stm_ana.positions.vals == 'heater')[0][0]
+        heater_temp = aman.stm_ana.temps[idx][0]
+        idx = np.where(aman.stm_ana.positions.vals == 'env')[0][0]
+        env_temp = aman.stm_ana.temps[idx][0]
+
+        arr = np.array([float(x.best_values['a0']) if x is not None else np.nan for x in fit_result['fit_coadd']['lpf']['f1_gain']])
+        arr = abs(arr) * (750/(heater_temp-env_temp))
```
ykyohei: what is this 750?

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,967 @@
...
+            'stimulator-thermo.temperatures.Channel_4_T',
+            'stimulator-thermo.temperatures.Channel_6_T',
+            'stimulator-thermo.temperatures.Channel_5_T']
+    temps = {}
+    for position_key in position_keys:
+        temps[position_key] = []
+
+    for key,position_key in zip(keys,position_keys):
+        x = hkdata.data[key][0] - aman.timestamps[0]
+        y = hkdata.data[key][1]+273.15
```
ykyohei: Please note that this convert from degC unit to Kelvin unit

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,967 @@
...
+            else:
+                field = aman.stm_ana[fit_key][filt_key]
+                result = fit_result[fit_key][filt_key]
+                weight_axis = 'ndata_taufit'
+                fill_field(field, result, weight_axis)
+
+
+    # Fill result to aman.det_cal
+    if 'det_cal' not in aman.keys():
+        aman.wrap('det_cal', core.AxisManager(aman._axes['dets']))
```
ykyohei: I think we should't save values under det_cal. 
This is confusing and it might interfere with another det_cal metadata when people try to use both stimulator metadata and det_cal metadata

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,967 @@
...
+    
+    Args:
+        aman: axis manager
+        coadd_data: co-added data
+        fit_result: fit result
+        n_bins: number of bins
+        cal_type: 'gain' or 'timeconstant'. type of calibration
+    """
+    if not 'coadd_data' in aman.stm_ana.keys():
+        bins = core.IndexAxis('bins',n_bins)
```
ykyohei: 'bin' is very general name and this might interfere with other IndexAxis. I suggest to use stimulator specific name.

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,967 @@
...
+        print(f'Invailed HK data. last t_enc: {t_enc[-1]}, last TOD timestamp: {aman.timestamps[-1]}')
+        vailed_data = False
+    else:
+    # Add timing against encoder to axis manager
+        aman.wrap('frac_timing', np.array(frac_timing), [(0,'samps')], overwrite=True)
+
+    return vailed_data
+
+    
+def get_chopping_status(aman):
```
ykyohei: I haven't carefully read through this function, but I'd suggest making the hardcoded numbers arguments of the function, so that they can be adjusted when necessary

ykyohei (commented):
It's great to see the simulator pipeline in sotodlib!

Preliminary review comments. sotodlib follows PEP8, but whitespace is missing after commas in many places. I suggest running a linter and correcting them.

sotodlib/stimulator/stimulator.py
YudaiSeino: These lines exist not to run the plotting function when we don't need to plot figure. I'll change naming of bool argument clearly.

sotodlib/stimulator/stimulator.py
YudaiSeino: It's leap seconds. I changed here. New version will take leap second for each data.

sotodlib/stimulator/stimulator.py
YudaiSeino: It is typical signal temperature 750 K. The unit for the conversion here will be Kelvin/rad in the future.

sotodlib/stimulator/stimulator.py
YudaiSeino: I'll delete both functions.

sotodlib/stimulator/stimulator.py
YudaiSeino: The downsample_factor may not be int but I agree that change.

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,851 @@
...
+
+    t_start = dt_buffer+aman.stm_cal.t_chopping_change[0]
+    t_end = min(t_start+dt_gain, aman.stm_cal.t_chopping_change[1])
+    t_cuts.append((t_start,t_end))
+
+    if cal_type == 'gain_timeconstant':
+        for i in range(len(aman.stm_cal.t_chopping_change)-1):
+            t_start = aman.stm_cal.t_chopping_change[i] + dt_wait
+            if i==0:
+                t_start = 2+aman.stm_cal.t_chopping_change[i]  # 2 seconds to avoid ringing by high-pass filter
```
ykyohei: dt_buffer should be used here!

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,851 @@
...
+            t_end = min(t_start+dt_timeconstant, aman.stm_cal.t_chopping_change[i+1])
+            t_cuts.append((t_start,t_end))
+
+    if cal_type == 'gain':
+        label_freqs = core.LabelAxis('freqs',['f1_gain'])
+    else:
+        keys = ['f1_gain']
+        for i in range(len(aman.stm_cal.t_chopping_change)-1):
+            keys.append(f'f{int(i+1)}')
+        label_freqs = core.LabelAxis('freqs',keys)
```
ykyohei: LabelAxis "freqs" is used for axis of psd etc, please rename to be stimulator specific.

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,851 @@
...
+        aman: axis manager
+        fit_result: fit result
+        n_bins: number of bins
+        cal_type: 'gain' or 'timeconstant'. type of calibration
+    """
+
+    # Fill fit result to axis manager
+    for fit_key in fit_result.keys():
+        if fit_key not in aman.stm_cal.keys():
+            aman.stm_cal.wrap(fit_key, core.AxisManager(aman._axes['dets']))
```
ykyohei: `aman._axes['dets']`  can be simplified as `aman.dets`
We can simplify like this in many places.

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,851 @@
...
+
+from sotodlib import tod_ops, core
+from sotodlib.io import hkdb
+from sotodlib.stimulator.utils_stimulator import func_sines, func_response_amplitude, func_response_phase_with_dt
+
+CHOPPING_FREQS = {'f1': 6, 'f2': 15, 'f3': 33,'f4': 63, 'f5': 93, 'f6': 123, 'f7': 147}
+STM_NORMALIZE_TEMP = 750 # Kelvin. Normalization temperature for stimulator signal temperature.
+
+
+def get_hk(hkdb_cfg, aman=None, t_start=None, t_end=None):
```
ykyohei, line 20: This doesn't necessarily need to be implemented in this PR, but for the site-pipeline of stimulator, we need to implement hk loading from level2 not hkdb.

For example, https://github.com/simonsobs/sotodlib/blob/74fe1ad5d7911c36cb9124458d16c3a154063c55/sotodlib/site_pipeline/calibration/wiregrid.py#L90

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,851 @@
...
+            f = [chopping_freqs[f_key] for f_key in chopping_freqs.keys()]
+            for fit_key in ['fit_amp', 'fit_phase__fix_tau', 'fit_phase__free']:
+                params = params_bases[fit_key].copy()
+                if fit_key == 'fit_amp':
+                    if np.isnan(a0s).all():
+                        continue
+
+                    weights = [fit_result['fit_coadd'][filt_key][f_key][i_det].params['a0'].stderr if not isinstance(fit_result['fit_coadd'][filt_key][f_key][i_det], (int, float)) and fit_result['fit_coadd'][filt_key][f_key][i_det].params['a0'].stderr is not None else np.nan for f_key in fit_result['fit_coadd'][filt_key].keys()]
+                    weights = np.array(weights)
+                    result = models[fit_key].fit(a0s,params,f=f, method='least_squares')
```
ykyohei: In calc_timeconstant, weights are computed but not used for fitting in multiple places.

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,851 @@
...
+            continue
+        t1_enc_tmp = t_enc[i_enc]
+        t2_enc_tmp = t_enc[i_enc+1]
+
+        while t2_enc_tmp < t:
+            i_enc += 1
+            t1_enc_tmp = t_enc[i_enc]
+            t2_enc_tmp = t_enc[i_enc+1]
+
+        frac_timing.append((t - t1_enc_tmp)/(t2_enc_tmp-t1_enc_tmp))
```
ykyohei: I think this function might have a bug. frac_timing has "samps" axis but it can be shorter than samps.

I suggest to change for example,
```
frac_timing = np.full(aman.samps.count, np.nan)
for i, t in enumerate(aman.timestamps):
    if i_enc >= len(t_enc)-2:
            continue
...
    frac_timing[i] = (t - t1_enc_tmp)/(t2_enc_tmp-t1_enc_tmp)
...
aman.wrap('frac_timing', frac_timing, [(0,'samps')])
```

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,851 @@
...
+            filter_cutoff = lpf_cutoff_factor*chopping_freq
+            lpf = tod_ops.filters.low_pass_sine2(filter_cutoff,filter_cutoff*lpf_width_fraction)
+            signal_new = tod_ops.fourier_filter(aman, lpf, signal_name='signal_hpf')
+            aman.wrap(f'signal_lpf_{key}', signal_new, [(0,'dets'),(1,'samps')], overwrite=True)
+
+    else:
+        raise ValueError(f"'{cal_type}' is a wrong type. Please specify 'gain' or 'timeconstant'.")
+
+
+    # Fill filtering parameters to axis manager
```
ykyohei, line 626: I think these are not necessarily need to be saved. 
Site pipeline configuration file will have these information.

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,851 @@
...
+                weight_axis = 'ndata_taufit'
+                fill_field(field, result, weight_axis)
+
+
+    # Fill result for general people
+    if cal_type == 'gain':
+        idx = np.where(aman.stm_cal.positions.vals == 'heater')[0][0]
+        heater_temp = aman.stm_cal.temps[idx][0]
+        idx = np.where(aman.stm_cal.positions.vals == 'env')[0][0]
+        env_temp = aman.stm_cal.temps[idx][0]
```
ykyohei: We can do this.
`env_temp = aman.stm_cal.temps[aman.stm_cal.positions.vals == 'env'][0][0]`

Just out of curiosity, why don't we correct temperature drift for the timeconstant?

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,851 @@
...
+        end=t_end,
+    )
+    result = hkdb.load_hk(lspec)
+
+    return result
+
+
+def calc_gain(aman, hkdata, idxs=None, n_bins=40, preprocessing=True):
+    """
+    Calculate the gain of the detectors.
```
ykyohei, line 165: It would be helpful to add brief description.

For each detector, the raw signal is deconvolved using the IIR filter, and then either high-pass filtered (hpf), or both low-pass and high-pass filtered (lpf) to form a bandpass around the stimulator's chopping frequency. 
The filtered signal is co-added into chopper phase bins.
In former case (hpf) coadded signal is fit to a seven-harmonic sine model to obtain its fundamental amplitude (a0). In the latter case (lpf), the co-added waveform is instead fit to a single-mode sine curve.
Gain is the amplitude (a0) normalized to a typical temperature of 750 K by the measured heater/env temperatures. Results are written into aman.stm_cal.

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,851 @@
...
+            fit_result['fit_coadd'][filt_key]['f1_gain'][i_det] = result
+
+    fill_data(aman, fit_result, cal_type='gain')
+
+    return valid_data
+
+
+def calc_timeconstant(aman, hkdata, idxs=None, n_bins=40, preprocessing=True):
+    """
+    Calculate the time constant of the detectors.
```
ykyohei: Brief description will be helpful

Calculate the time constant of the detectors and the readout delay.

For each detector, filters the raw signal (HPF/LPF) around the stimulator's chopping frequency, co-adds it into chopper phase bins, and fits the co-added waveform to a seven harmonics sine model to get its fundamental amplitude (a0) and phase delay (t0) per chopping frequency.

Fits amplitude (a0) vs. chopping frequency to a single-pole response model to get the time constant (tau), then
fits phase delay (t0) vs. chopping frequency with tau fixed to get the readout delay (dt). Additionallyy fits phase delay (t0) vs. chopping frequency with both timeconstant and readout delay as free parameters.
Results are written into aman.stm_cal

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,851 @@
+import numpy as np
+import lmfit
+import ruptures as rpt
+import astropy
+
+from sotodlib import tod_ops, core
+from sotodlib.io import hkdb
+from sotodlib.stimulator.utils_stimulator import func_sines, func_response_amplitude, func_response_phase_with_dt
+
+CHOPPING_FREQS = {'f1': 6, 'f2': 15, 'f3': 33,'f4': 63, 'f5': 93, 'f6': 123, 'f7': 147}
```
ykyohei: These values are used for chopping frequencies and measured chopping frequency is not used for analysis. Control of chopping frequency is probably very accurate, but why do we do this? 
Using measured chopping frequencies can be important especially for timeconstant.

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,851 @@
+import numpy as np
+import lmfit
+import ruptures as rpt
```
ykyohei, line 3: This is new dependency of sotodlib. We need to add this to optional site-pipeline dependencies. https://github.com/simonsobs/sotodlib/blob/74fe1ad5d7911c36cb9124458d16c3a154063c55/pyproject.toml#L51-L52

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,851 @@
...
+
+    Returns:
+        bool: True if timing data is valid, False otherwise.
+    """
+
+    t_astropy = astropy.time.Time(hkdata.data['stimulator-enc.stim_enc.state'][0][0], format='unix')
+    leap_seconds = t_astropy.unix_tai - t_astropy.unix
+
+    state = np.array(hkdata.data['stimulator-enc.stim_enc.state'][1])
+    t_enc = np.array(hkdata.data['stimulator-enc.stim_enc.timestamps_tai'][1])[state == 0]-leap_seconds
```
ykyohei: It got correct value but I note that I got this warning when I tested this at tiger3.
`WARNING: IERSStaleWarning: leap-second file is expired. [astropy.utils.iers.iers]`

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,851 @@
...
+                fit_result[fit_key][filt_key][i_det] = result
+
+    fill_data(aman, fit_result, cal_type='timeconstant')
+
+    return valid_data
+
+
+def get_encoder_timing(aman, hkdata):
+    """
+    Get timing against encoder start time t0.
```
ykyohei: Can you add explanation for t_enc, t_hk, and frac_timing?

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,851 @@
...
+
+    if not valid_data:
+        return valid_data
+
+    chopping_freqs = {}
+    for i,(key,f) in enumerate(CHOPPING_FREQS.items()):
+        if round(aman.stm_cal.chopping_freqs[0]) != f:
+            chopping_freqs[key] = round(aman.stm_cal.chopping_freqs[i])
+        else:
+            chopping_freqs[key] = f
```
ykyohei: Why do we used hardcoded frequency instead of measured frequency?

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,851 @@
...
+
+            # Edit a0 and t0 result
+            for i in range(len(t0s)):
+                if a0s[i] < 0:
+                    t0s[i] = t0s[i]-0.5
+                if i > 0:
+                    if t0s[i] < t0s[i-1]:
+                        t0s[i] = t0s[i] + 1
+            a0s = np.abs(a0s)
+            t0s = np.array(t0s)
```
ykyohei, line 237: What is this doing?

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,851 @@
...
+    aman.stm_cal.filtering_params.wrap('lpf_cutoff_factor', lpf_cutoff_factor, overwrite=True)
+    aman.stm_cal.filtering_params.wrap('lpf_width_fraction', lpf_width_fraction, overwrite=True)
+
+    if cal_type == 'gain':
+        aman.stm_cal.filtering_params.wrap('chopping_freqs_gain', np.array(list(chopping_freqs.values())), overwrite=True)
+    elif cal_type == 'timeconstant':
+        aman.stm_cal.filtering_params.wrap('chopping_freqs_tau', np.array(list(chopping_freqs.values())), overwrite=True)
+
+
+def get_dicts(aman, cal_type):
```
ykyohei: get_dicts and fill_data look redundant.
We can initialize axis manager that will contain results instead of dictionary here.
Then we don't need fill_data?

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,851 @@
...
+    else:
+    # Add timing against encoder to axis manager
+        aman.wrap('frac_timing', np.array(frac_timing), [(0,'samps')], overwrite=True)
+
+    return valid_data
+
+
+def get_chopping_status(aman, min_step_duration=10,penalty=100,delta_t=5,wait_t=5):
+    """
+    Get chopping frequency and timing when the chopping speed changes.
```
ykyohei: Can you add brief descriptions of t_chopping_change and chopping_freqs here?

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,851 @@
...
+    aman.stm_cal.wrap('t_enc', t_enc, [(0,'stm_samps')], overwrite=True)
+    aman.stm_cal.wrap('t_hk', t_hk, [(0,'stm_samps')], overwrite=True)
+
+    valid_data = True
+    if t_enc[-1] < aman.timestamps[-1]:
+        print(f'Invalid HK data. last t_enc: {t_enc[-1]}, last TOD timestamp: {aman.timestamps[-1]}')
+        valid_data = False
+    else:
+    # Add timing against encoder to axis manager
+        aman.wrap('frac_timing', np.array(frac_timing), [(0,'samps')], overwrite=True)
```
ykyohei: I think frac_timing should be wrapped to aman.det_cal.

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,851 @@
...
+        params_base.add('a4',value=1e-4)
+        params_base.add('a5',value=1e-5)
+        params_base.add('a6',value=1e-5)
+        params_base.add('t0',value=0.5,min=0,max=1)
+        params_base.add('t1',value=0.5,min=0,max=1)
+        params_base.add('t2',value=0.5,min=0,max=1)
+        params_base.add('t3',value=0.5,min=0,max=1)
+        params_base.add('t4',value=0.5,min=0,max=1)
+        params_base.add('t5',value=0.5,min=0,max=1)
+        params_base.add('t6',value=0.5,min=0,max=1)
```
ykyohei: How about doing like this for both gain and timeconstant.
```
amps_init = [1e-3, ...]

for i, a in enumerate(amps_init):
    params_base.add(f'a{i}', value=a)
    params_base.add(f't{i}"', value=0.5, min=0, max=1)
```

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,851 @@
...
+        bool: True if data is valid and calculation is performed, False otherwise.
+    """
+    det_mask = np.full(aman.dets.count, False)
+    if idxs is None:
+        det_mask[:] = True
+    else:
+        det_mask[idxs] = True
+
+    valid_data = True
+    if preprocessing:
```
ykyohei: preprocessing look common between gain and timeconstant.
I think it would be cleaner and simpler to make independent preprocessing function.

For example:
```
def preprocessing(aman, hkdata, idxs=None, n_bins=40, cal_type='all'):

        valid_data = get_encoder_timing(aman, hkdata)  # Get timing against encoder t0
        aman.stm_cal.wrap('sampling_rate', 1 / np.median(np.diff(aman.timestamps)), overwrite=True)
        get_chopping_status(aman)
        get_timing_cut(aman)
        get_signal_temp(aman,hkdata)
        chopping_freqs = <both gain and timeconstant>
        get_coadd_data(aman, "all", n_bins, det_mask)

       # fit_coadd with sine function is also common
        do sine fitting.
```

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,851 @@
...
+        lpf = tod_ops.filters.low_pass_sine2(filter_cutoff,filter_cutoff*lpf_width_fraction)
+        signal_new = tod_ops.fourier_filter(aman, lpf, signal_name='signal_hpf')
+        aman.wrap('signal_lpf', signal_new, [(0,'dets'),(1,'samps')], overwrite=True)
+
+    elif cal_type == 'timeconstant':
+        for key, chopping_freq in chopping_freqs.items():
+            filter_cutoff = lpf_cutoff_factor*chopping_freq
+            lpf = tod_ops.filters.low_pass_sine2(filter_cutoff,filter_cutoff*lpf_width_fraction)
+            signal_new = tod_ops.fourier_filter(aman, lpf, signal_name='signal_hpf')
+            aman.wrap(f'signal_lpf_{key}', signal_new, [(0,'dets'),(1,'samps')], overwrite=True)
```
ykyohei: This actually only saves lpf at highest chopping frequency!!!
I guess we actually do not need two versions of filtering, and it will simplify the pileline and outputs.

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,851 @@
...
+                    params['t4'].set(value=0,vary=False)
+                    params['t5'].set(value=0,vary=False)
+                    params['t6'].set(value=0,vary=False)
+                    result = models['fit_coadd'].fit(y, params, t=x, weights=1/np.array(yerr))
+                else:
+                    result = models['fit_coadd'].fit(y, params, t=x, weights=1/np.array(yerr))
+
+                fit_result['fit_coadd'][filt_key][f_key][i_det] = result
+
+            a0s = [fit_result['fit_coadd'][filt_key][f_key][i_det].best_values['a0'] if not isinstance(fit_result['fit_coadd'][filt_key][f_key][i_det], (int, float)) else np.nan for f_key in fit_result['fit_coadd'][filt_key].keys()]
```
ykyohei: I wonder why we need to check instance in this line, and many other lines? what is the expected value that is not (int, float)?

ykyohei (commented):
Thanks! it's getting much better!
I guess I found few bugs, please check.

sotodlib/stimulator/stimulator.py
YudaiSeino: Do you mean parameters for smurf IIR filter? Sorry, I don't understand your comment.

ykyohei: I mean lines from 564 to 576 are okay to drop.
I think "filtering_params" axismanager will be redundant with pipeline configuration.

sotodlib/stimulator/stimulator.py
YudaiSeino: I wrapped to aman.stm_cal

sotodlib/stimulator/stimulator.py
ykyohei: this was my misunderstanding, these are saved as different name.

sotodlib/stimulator/stimulator.py
YudaiSeino: I'll keep those here because plotting function will use these values.

sotodlib/stimulator/stimulator.py
YudaiSeino: This is shifting half cycle of phase when sine wave amplitude is minus. Theoretically, the calculation here is not required if we can set (sine wave amplitude) > 0. Unfortunately, the fitting fails sometimes with this limit. I didn't fined good condition for the fitting yet.

ykyohei: OK, Please wite a comment of what you explained.
We don't need to change it but 
c0 * cost + s0 *sint instead of a0 * sin(t + t0) will not have this problem.

sotodlib/stimulator/stimulator.py
YudaiSeino: I made new functions to remove redundant.

sotodlib/stimulator/stimulator.py
YudaiSeino: Here, I used chopping frequency for filtering. I want to fix filtering frequency.

sotodlib/stimulator/stimulator.py
YudaiSeino: It was bug. I fixed it.

sotodlib/stimulator/stimulator.py
YudaiSeino: It is to check if value is nan or not. I changed code and these parts are not needed anymore.

sotodlib/stimulator/stimulator.py
YudaiSeino: This function will be added in automation code.

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,1006 @@
...
+
+    if valid_timeconstant:
+        filter_freqs = {}
+        for key, freq in [
+            (key, freq)
+            for key, freq in zip(
+                aman.stm_cal.chopping_freq_key.vals, aman.stm_cal.chopping_freqs
+            )
+            if key != "f1_gain"
+        ]:
```
ykyohei:
```suggestion
        for key, freq in zip(
            aman.stm_cal.chopping_freq_key.vals, aman.stm_cal.chopping_freqs
        ):
            if key != "f1_gain":
               continue
```

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,1006 @@
...
+        fields=feed_list,
+        start=t_start,
+        end=t_end,
+    )
+    result = hkdb.load_hk(lspec)
+
+    return result
+
+
+def preprocessing(aman, hkdata, idxs=None, n_bins=40, delete_filtered_tod=True):
```
ykyohei, line 60: It's much more memory efficient to do filtering, coadd, fit and delete filtered tod for gain.
And then at each chopping frequency do filtering, coadd, fit and delete filtered tod.

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,1006 @@
...
+        ]:
+            if round(freq) != CHOPPING_FREQS[key]:
+                filter_freqs[key] = round(freq)
+            else:
+                filter_freqs[key] = CHOPPING_FREQS[key]
+        filtering(aman, filter_freqs, "timeconstant")
+
+    # Make and fit co-added data
+    model, params_base = get_fit_params(cal_type="coadd")
+    initialize_aman(aman, "coadd", model, n_bins)
```
ykyohei: init_type should be changed depending on valid_gain and valid_timeconstant?

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,1006 @@
...
+        )
+    elif cal_type == "timeconstant":
+        aman.stm_cal.filtering_params.wrap(
+            "filter_freq_tau",
+            np.array(list(filter_freqs.values())),
+            overwrite=True,
+        )
+
+
+def initialize_aman(aman, init_type, model, n_bins=None):
```
ykyohei, line 653: I guess this needs to allow init_type='gain'?

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,1006 @@
...
+    valid_data = True
+    if t_enc[-1] < aman.timestamps[-1]:
+        print(
+            f"Invalid HK data. last t_enc: {t_enc[-1]}, last TOD timestamp: {aman.timestamps[-1]}"
+        )
+        valid_data = False
+    else:
+        # Add timing against encoder to axis manager
+        # aman.stm_cal.wrap('frac_timing', np.array(frac_timing), [(0,'samps')], overwrite=True)
+        aman.wrap("frac_timing", np.array(frac_timing), [(0, "samps")], overwrite=True)
```
ykyohei: I think frac_timing is better to be wrapped in stm_cal, because in the site_pipeline I think we will save aman.stm_cal to hdf5 file.

pyproject.toml
```diff
@@ -58,11 +58,12 @@ site_pipeline = [
   "pandas",
   "alphashape",
   "slack-sdk",
+  "mapcat",
   "requests",
   "tilemaker",
   "GitPython",
   "sqlmodel",
-  "pyinstrument"
+  "ruptures"
```
ykyohei: This has conflict with latest master branch, please resolve conflict.

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,1006 @@
...
+        "stm_tau", aman.stm_cal["fit_amp"]["lpf"]["tau"], [(0, "dets")], overwrite=True
+    )
+    aman.stm_cal.wrap(
+        "readout_delay",
+        aman.stm_cal["fit_phase__fix_tau"]["lpf"]["dt"],
+        [(0, "dets")],
+        overwrite=True,
+    )
+
+    if "stm_gain" in aman.stm_cal._fields:
```
ykyohei:
```suggestion
    if "stm_gain" in aman.stm_cal:
```

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,1006 @@
...
+        "chopping_freqs",
+        np.array(chopping_freqs),
+        [(0, "chopping_freq_key")],
+        overwrite=True,
+    )
+
+
+def get_signal_temp(aman, hkdata):
+    """
+    Get signal temperature for chopping frequency.
```
ykyohei:
```suggestion
    Get stimulator signal temperature at each chopping frequency.
```

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,1006 @@
...
+
+        refined_bkps.append(left + idx + 1)
+
+    # Add last point
+    refined_bkps.append(x_ave.size)  # index +1
+
+    t_chopping_change = x_ave[np.array(refined_bkps) - 1]
+    t_chopping_change = np.concatenate([[0], t_chopping_change], 0)
+
+    aman.stm_cal.wrap("t_chopping_change", t_chopping_change, overwrite=True)
```
ykyohei, line 439: I think this has chopping_freq_key axis?
```suggestion
    aman.stm_cal.wrap("t_chopping_change", t_chopping_change, [(0, "chopping_freq_key")], overwrite=True)
```

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,1006 @@
...
+        get_coadd_data(aman, "timeconstant", n_bins, det_mask)
+        fit_coadd_data(aman, "timeconstant", det_mask, model, params_base)
+
+    # Delete filtered TOD data if requested.
+    # This operation will run for each filtering frequency in near future to save memory usage.
+    if delete_filtered_tod:
+        keys_to_delete = [key for key in aman._fields if key.startswith("signal_")]
+        for key in keys_to_delete:
+            if key.startswith("signal_"):
+                del aman[key]
```
ykyohei:
```suggestion
        keys_to_delete = [key for key in aman._fields if key.startswith("signal_")]
        for key in keys_to_delete:
                aman.move(key, None)
```
We use move to delete field properly. startwith is called twice.

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,1006 @@
...
+        params_base: base parameters for fitting
+    """
+    if cal_type == "all":
+        freq_keys = aman.stm_cal.coadd_data["hpf"]
+    if cal_type == "gain":
+        freq_keys = ["f1_gain"]
+    elif cal_type == "timeconstant":
+        freq_keys = CHOPPING_FREQS.keys()
+    else:
+        raise ValueError(
```
ykyohei, line 904: all will raise error
```suggestion
    if cal_type == "all":
        freq_keys = aman.stm_cal.coadd_data["hpf"]
    elif cal_type == "gain":
        freq_keys = ["f1_gain"]
    elif cal_type == "timeconstant":
        freq_keys = CHOPPING_FREQS.keys()
    else:
        raise ValueError(
```

sotodlib/stimulator/stimulator.py
```diff
@@ -0,0 +1,1006 @@
...
+
+        temps[position_key] = np.array(temps[position_key])
+
+    temps["env"] = (temps["chopper_rear"] + temps["chopper_front"] + temps["air"]) / 3
+    position_keys.append("env")
+
+    label_positions = core.LabelAxis("positions", position_keys)
+    aman.stm_cal.add_axis(label_positions)
+    arr = [x for x in temps.values()]
+    arr = np.array(arr)
```
ykyohei:
```suggestion
    arr = np.array(list(temps.values()))
```

ykyohei (commented):
Thanks for all the changes, almost there!

`overwrite=True` is used in many places. I think these were convenient when developping pipeline but these are not really necessary. I'll let you decide but and it might be safer to remove.

sotodlib/stimulator/stimulator.py
YudaiSeino: It doesn't has chopping_freq_key axis.

sotodlib/stimulator/stimulator.py
YudaiSeino: I just forgot to reverse it during debugging, thanks!

pyproject.toml
YudaiSeino: I deleted by mistake.

sotodlib/stimulator/stimulator.py
YudaiSeino: No

sotodlib/stimulator/stimulator.py
YudaiSeino: No,

ykyohei:
Looks good, thank you!

ykyohei (approved):

PR #1664, opened 2026-06-12
Push loop over fields for HK down into so3g to avoid duplicate file reads

JBorrow:
We were only providing `HKArchive.simple` a single field at a time, and each time it had to read the entire `.g3` file. This PR pushes the fields down as a list so they can all be extracted in a single file read. It provides binary-identical results in my testing.


kmharrington (approved):
brilliant lol

PR #1667, opened 2026-06-18
io.load_book handles new spt3g/so3g data structure

mhasself:
Cf. https://github.com/simonsobs/so3g/pull/229

This modifies load_book to work with existing and future proposed so3g/spt3g versions.

Also introduces tests (which demonstrate how to write ~equivalent obs books in the new paradigm).

Draft while sorting out performance issues and completing testing.

@tskisner


tskisner:
Thanks for working on this!

PR #1669, opened 2026-06-18
Complete level 2 cleanup reorganization

kmharrington:
I'm going to admit to having used Claude to help reorganize this. It's going to substantially reduce the number of times we run through the file system and databases as we are cleaning things up. (I don't think this will be enough to shorten LATs runtime but it's a start). 

This changes how the cleanup loops work:

Before:
```
for each timecode:
    check_timecode_complete()
for each timecode:
    check_timecode_complete()
    delete_staged()
for each timecode:
    check_timecode_complete()
    delete_level2()
```

After
```
for each timecode:
    check_timecode_complete()
    if ready:
        delete_staged()
    if ready:
        delete_level2()
```

Other changes
* add option to set a maximum allowable runtime (although only checked once per loop)
* add dry run print outs


kmharrington:
I've now tested and run this over everything. Ready to deploy.

PR #1670, opened 2026-06-19
Map-making with fixed tones

tpsatt:
Pushing my local work on implementing fixed tones into the mapmaking noise model.

Fixed tones (or "probe tones") are tones sent out by the SMuRF system which live between the resonance frequencies of our resonators (which are coupled to detectors). They therefore come out of SMuRF, pass through all of our cabling, enter and exit the cryostat, and return to SMuRF unperturbed by on-sky signal. They measure changes to the readout chain, for example via thermal changes to the coaxial cables.

The idea here is to incorporate these into the ML mapmaker noise model in case they are of some utility (perhaps they can help filter out low-frequency noise). The procedure has been to include them in `obs.signal` but set their `T=P=0` in the pointing model so they become part of the noise term. Some of this code is a bit hacky to accommodate getting the tones into the AxisManager's signal field (shoutout Claude).

For initial runs, I'm using median detector values for `abscal` and `relcal` so we can scale the fixed tone TODs to the correct units. This ignores tone tracking, but it may be appropriate given the scope of this project. I also do the basic detrending to emulate what happens to preprocessed signal.

### DIFF
--- a/sotodlib/core/axisman.py
+++ b/sotodlib/core/axisman.py
@@ -877,7 +877,7 @@ class AxisManager:
                         shape.append(s)
 
                 new_v = np.empty(shape, dtype=v.dtype)
-                if isinstance(v.dtype, float):
+                if v.dtype == float:
                     # Fill any float arrays with nans
                     # Non float arrays may have weird
                     # behavior for newly added indexes. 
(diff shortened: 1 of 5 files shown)


tpsatt:
For anyone who stumbles upon this PR, I've written up some info here: https://simonsobs.atlassian.net/wiki/spaces/PRO/pages/2057732109/Fixed+Tone+Map-Making

PR #1672, opened 2026-06-22
Added core profiling module and used across data-package scripts

JBorrow:
This PR adds a new profiling utility module inside the `site_pipeline` module that can allow for easy profiling of data packaging and pipeline scripts that use the `main()` signature. This works within Prefect. Documentation is included.


mmccrackan:
Do you think there is any utility in moving this to a lower level of `sotodlib` (like `core/util.py` maybe) so it can be used in other parts of the codebase?  Seems like it might be useful to have that.  I had started doing something like that in a much older PR when looking into preprocessing performance but didn't really intend on merging that.

JBorrow:
Happy to move it if/when others want to use the functionality... I think let's keep it in here for now as it's only currently used for the site pipeline flows. The writing directly to files etc. is less compatible with the 'library' code rather than the 'pipeline script' style.

mmccrackan (approved):

msilvafe:
Hey @JBorrow we've been in need of some standard profiling tools in the main repo for awhile. Thank you for adding this! Can you comment on how or if we might use this for profiling some of the parallel site-pipeline scripts such as preprocess_tod, multilayer_preprocess_tod, make_atomic_filterbin_map, make_tau_hwp, update_det_cal etc. I only briefly glanced at this and see that the examples aren't covering any parallel scripts. If not supported I don't mean this comment to hold up merging this PR as it'll be broadly useful regardless just thinking about what I'd need to extend this for my workflows!

JBorrow:
I believe these should work automatically (especially the pyinstrument ones) for parallel scripts. I've not extended this much further than it is now just because I haven't needed it; I'd like to get this PR in and then you can give it a go on your parallel scripts?

PR #1674, opened 2026-06-23
Make bin_signal compatible with 1d-array

ykyohei:
This is for binning 1d-array such as thermometer data to az.
This doesn't affect binning of 2d array.

There was an issue where, when the range was narrower than the bin_by value, the edge bins returned the wrong value due to np.digitize and np.clip. 
This PR also fixes this issue. This error should not impact most analyses.

Added unit test for binning.

### DIFF
--- a/sotodlib/tod_ops/binning.py
+++ b/sotodlib/tod_ops/binning.py
@@ -93,7 +93,7 @@ def bin_signal(aman, bin_by, signal=None,
         if (not is_1d) and flags.shape == (aman.dets.count, aman.samps.count):
             flag_is_2d = True
             m_2d = base_valid[None, :] & ~flags.mask()
-        elif is_id and flags.shape == (aman.samps.count, ):
+        elif flags.shape == (aman.samps.count, ):
             flag_is_2d = False
             m = base_valid & ~flags.mask()
         else:
@@ -106,25 +106,27 @@ def bin_signal(aman, bin_by, signal=None,
 
         # prepare binned signal array
         binned_signal = np.full(nbins, np.nan, signal_dtype)
+        binned_signal_squared_mean = np.full(nbins, np.nan, signal_dtype)
         binned_signal_sigma = np.full(nbins, np.nan, signal_dtype)
 
-        bin_counts = np.bincount(
+        bin_counts_dets = np.bincount(
             bin_indices[m], weights=weight_for_signal[m], minlength=nbins)
-        mcnts = bin_counts > 0
+        mcnts = bin_counts_dets > 0
         binned_signal[mcnts] = np.bincount(
             bin_indices[m], weights=signal[m] * weight_for_signal[m], minlength=nbins
-        )[mcnts] / bin_counts[mcnts]
-        binned_signal_squared_mean = np.bincount(
+        )[mcnts] / bin_counts_dets[mcnts]
+        binned_signal_squared_mean[mcnts] = np.bincount(
             bin_indices[m], weights=(signal[m] * weight_for_signal[m]) ** 2,
             minlength=nbins
-        )[mcnts] / bin_counts[mcnts]
+        )[mcnts] / bin_counts_dets[mcnts]
         binned_signal_sigma[mcnts] = np.sqrt(
-            np.abs(binned_signal_squared_mean - binned_signal[mcnts] ** 2)
-        ) / np.sqrt(bin_counts[mcnts])
+            np.abs(binned_signal_squared_mean[mcnts] - binned_signal[mcnts] ** 2)
+        ) / np.sqrt(bin_counts_dets[mcnts])
 
     else:
         # prepare binned signal array
         binned_signal = np.full([aman.dets.count, nbins], np.nan, signal_dtype)
+        binned_signal_squared_mean = np.full([aman.dets.count, nbins], np.nan, signal_dtype)
         binned_signal_sigma = np.full([aman.dets.count, nbins], np.nan, signal_dtype)
         bin_counts_dets = np.full([aman.dets.count, nbins], np.nan)
 
--- a/sotodlib/tod_ops/binning.py
+++ b/sotodlib/tod_ops/binning.py
@@ -16,6 +16,7 @@ def bin_signal(aman, bin_by, signal=None,
         The array by which signal is binned. It should has the same samps as aman.
     signal : str, optional
         The name of the signal to be binned. Defaults to aman.signal if not specified.
+        Either 1D array of shape ``(samps,)`` or 2D array of shape ``(dets, samps)``.
     range : list or None
         A list specifying the bin range ([min, max]). Default is None, which means bin range
         is set to [min(bin_by), max(bin_by)].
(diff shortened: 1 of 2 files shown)


sotodlib/tod_ops/binning.py
```diff
@@ -55,43 +72,64 @@ def bin_signal(aman, bin_by, signal=None,
...
-                                      ) / np.sqrt(bin_counts[mcnts])
-        bin_counts_dets = np.tile(bin_counts, (aman.dets.count, 1))
 
+    # Drop NaN bin_by and out-of-range samples so that np.digitize + np.clip
+    # do not silently pile them into the first/last bin.
+    base_valid = (
+        np.isfinite(bin_by)
+        & (bin_by >= range[0])
+        & (bin_by <= range[1])
+    )
```
msilvafe, line 85: When `bins` is passed as an explicit edge sequence, `np.histogram_bin_edges` ignores range so the mask and the bin edges can disagree how about using bin_edges[0/-1] instead of range[0/1]?

Also bin_by needs to be an np array for >= and <= logic to work need to enforce this if want to allow input to be a list too.

sotodlib/tod_ops/binning.py
```diff
@@ -55,43 +72,64 @@ def bin_signal(aman, bin_by, signal=None,
...
+            m_2d = base_valid[None, :] & ~flags.mask()
         elif flags.shape == (aman.samps.count, ):
             flag_is_2d = False
-            m = ~flags.mask()
+            m = base_valid & ~flags.mask()
         else:
             raise ValueError('flags should have shape of (`dets`, `samps`) or (`samps`,)')
 
+    if is_1d:
+        if weight_for_signal.shape != (aman.samps.count,):
```
msilvafe, line 103: "array like" in docstring implies this could be a list but .shape will fail for that enforce to be a np array.

sotodlib/tod_ops/binning.py
```diff
@@ -55,43 +72,64 @@ def bin_signal(aman, bin_by, signal=None,
...
+        )[mcnts] / bin_counts_dets[mcnts]
+        binned_signal_squared_mean[mcnts] = np.bincount(
+            bin_indices[m], weights=(signal[m] * weight_for_signal[m]) ** 2,
+            minlength=nbins
+        )[mcnts] / bin_counts_dets[mcnts]
+        binned_signal_sigma[mcnts] = np.sqrt(
+            np.abs(binned_signal_squared_mean[mcnts] - binned_signal[mcnts] ** 2)
+        ) / np.sqrt(bin_counts_dets[mcnts])
+
+    else:
```
msilvafe, line 126: Do you think you could just convert to 2D with np.atleast_2d() and then compress back to 1d at the end that would reduce some duplicate blocks in the if/else. Some examples of this in tod_ops/utils.py L50-74, and tod_ops/jumps.py L100-146

sotodlib/tod_ops/binning.py
```diff
@@ -55,43 +72,64 @@ def bin_signal(aman, bin_by, signal=None,
...
-                                                 )[mcnts]/bin_counts[mcnts]
-            binned_signal_squared_mean[i][mcnts] = np.bincount(bin_indices, weights=(signal[i]*weight_for_signal)**2, minlength=nbins
-                                                 )[mcnts]/bin_counts[mcnts]
-            
-        binned_signal_sigma[:, mcnts] = np.sqrt(np.abs(binned_signal_squared_mean[:,mcnts] - binned_signal[:,mcnts]**2)
-                                      ) / np.sqrt(bin_counts[mcnts])
-        bin_counts_dets = np.tile(bin_counts, (aman.dets.count, 1))
 
+    # Drop NaN bin_by and out-of-range samples so that np.digitize + np.clip
+    # do not silently pile them into the first/last bin.
```
msilvafe, line 80: Since you raise this as a specific concern and decided to add test coverage can you add a test for this particular failure?

msilvafe (changes requested):
Couple inline comments, feel free to push back/reply/resolve any and hit re-review when ready. I'll be fast next time around (I think)!

sotodlib/tod_ops/binning.py
ykyohei: Right, I changed as suggested.

sotodlib/tod_ops/binning.py
ykyohei: Unified 1d and 2d binning as suggested.

sotodlib/tod_ops/binning.py
ykyohei: Added a test, confirmed this test fails without the fix of this PR

ykyohei (commented):
Made suggested changes, thanks for the review.

Additionally, I modified bin_signal to accept arbitrary samps length of signal array. This won't affect previous bahavior and helpful for binning house keeping data which has different sampling rate from detector's data. Test for this new feature is also added `test_400_arbitrary_length_of_signal`

In addition to the unit test, tested this with real data and confirmed that this reproduces exact same results, when range is wide enough.

sotodlib/tod_ops/binning.py
```diff
@@ -36,81 +40,102 @@ def bin_signal(aman, bin_by, signal=None,
...
-        bin_counts_dets = np.full([aman.dets.count, nbins], np.nan)
         if isinstance(flags, str):
             flags = aman.flags.get(flags)
-        if flags.shape == (aman.dets.count, aman.samps.count):
-            flag_is_2d = True
-            m_2d = ~flags.mask()
-        elif flags.shape == (aman.samps.count, ):
-            flag_is_2d = False
-            m = ~flags.mask()
+        if (not is_1d) and flags.shape == (aman.dets.count, nsamps):
```
msilvafe: why use a mix of `aman.dets.count` and `nsamps`, why not `ndets` and `nsamp`? Same question for L105.

sotodlib/tod_ops/binning.py
```diff
@@ -36,81 +40,102 @@ def bin_signal(aman, bin_by, signal=None,
...
 
     signal_dtype = signal.dtype
-    
-    if weight_for_signal is None:
-        weight_for_signal = np.ones(aman.samps.count, signal_dtype)
-        
+
     # get bin_edges
     bin_edges = np.histogram_bin_edges(bin_by, bins=bins, range=range,)
     bin_centers = (bin_edges[1] - bin_edges[0])/2. + bin_edges[:-1] # edge to center
```
msilvafe: I realize you didn't touch these lines in this PR but if `bins` passed to `np.histogram` is non-uniform then these bin centers will be wrong since they're computed only from the first bin edges. Can just change to `np.diff(bin_edges)/2 + bin_edges[:-1]`?

sotodlib/tod_ops/binning.py
```diff
@@ -13,9 +14,12 @@ def bin_signal(aman, bin_by, signal=None,
...
     bin_by : array-like
-        The array by which signal is binned. It should has the same samps as aman.
-    signal : str, optional
-        The name of the signal to be binned. Defaults to aman.signal if not specified.
+        The array by which signal is binned. Any length is allowed, but it must be consistent
+        with `signal` (and `flags` and `weight_for_signal`, if specified).
+    signal : str or array-like optional
+        signal to be binned or its name. Defaults to aman.signal if not specified.
+        Either 1D array or 2D array with shape ``(nsamps)`` or``(dets, nsamps)``,
+        where nsamps is length of bin_by.
```
msilvafe: I think you should specify that the signal array needs to be floats and maybe add a check for this since you initialize/fill arrays with np.nan only float are supported.

msilvafe (approved):
Alright, thanks for addressing all of my previous comments. I'm approving but left a few inline comments which you can take-or-leave before merge.

sotodlib/tod_ops/binning.py
ykyohei: Okay I will do `bin_centers = (bin_edges[1:] + bin_edges[:-1]) / 2`

sotodlib/tod_ops/binning.py
ykyohei: ndets is fine. I will change to use ndets.

PR #1675, opened 2026-06-23
Apply pointing correction to SAT sim

ykyohei:
Apply both boreight pointing model and pointing wobble to SAT simulation. wobble is optional.
Another PR of bb-awg-scripts: https://github.com/simonsobs/bb-awg-scripts/pull/21

Tested this using boresight pointing of obs_1731438172_satp3_1111111, ws1, f090
The middle one is difference of no wobble and wobble input only. We see residual in small scale as expected.
The bottom one is difference of no wobble and wobble correction both input and output (mapmaking). SInce these two scans slgihtly different region of sky, we see residual at the edge of map, as expected.

map
diff1
diff2


sotodlib/preprocess/preprocess_util.py
```diff
@@ -923,19 +930,26 @@ def multilayer_load_and_preprocess_sim(obs_id, configs_init, configs_proc,
...
+                logger.info("Apply pointing wobble")
+                sight = get_deflected_sightline(aman)
+            else:
+                sight = None
             aman.move("signal", None)
 
-            logger.info("Reading in simulated map")
-            demod_mm.from_map(aman, sim_map, wrap=True, modulated=True, interpol=interpol)
+            demod_mm.from_map(aman, sim_map, wrap=True, modulated=True,
+                              interpol=interpol, sight=sight)
```
adrien-laposta, line 952: Just for curiosity, does this also need to be propagated when we do the inverse operation, i.e. tod->map ? In which case we'll have to modify the small function in the filtering script

ykyohei: Exactly! Inverse tod->map operation is PRed here https://github.com/simonsobs/bb-awg-scripts/pull/21

sotodlib/preprocess/preprocess_util.py
```diff
@@ -853,7 +856,11 @@ def multilayer_load_and_preprocess_sim(obs_id, configs_init, configs_proc,
         processing simulations (e.g. to provide a T2P template)
     interpol: str
         Optional. The sub-pixel interpolation to use in from_map
- 
+    apply_wobble: bool
+        If true, apply pointing wobble to boreight pointing.
+        This only works whrn all detectors belong to a single wafer_slot
```
yoshinori-0778: typo: whrn --> when

sotodlib/preprocess/preprocess_util.py
yoshinori-0778: Looks great, thanks for making this!
One question, which might related to Adri's question, is that is this bb-awg-scripts using map making? Wobble correction is already in [here](https://github.com/simonsobs/sotodlib/blob/6a424e17f264402dfd3fa8883a5c1f89be4f023e/sotodlib/mapmaking/demod_mapmaker.py#L301-L311), and I though we have used this to make an atmic map.

ykyohei: Yeah I was confused as well. The data pipeline which you linked had wobble option already, but the tf pipeline is different and it is in awg repo and it didn't have wobble correction option, so I made a PR.

yoshinori-0778: I see, bb-pwg-scripts is used for TF. That makes sense. Thank you!

yoshinori-0778 (approved):

adrien-laposta (approved):

sotodlib/preprocess/preprocess_util.py
adrien-laposta: We might want to merge the two if possible (in the future) so that we don't duplicate code and use a single mapmaking function.

ykyohei: The reason I didn't directly integrate wobble to mapmaking function is because it only works when all the dets belong to single wafer band.

Also I think it might be cleaner to integraye wobble to pointing model, not pointing matrix.

PR #1676, opened 2026-06-23
Check data quality per observation at the site

GoureeSankar:
Contains a python script for implementing data quality check and associated config file.  
Note: Path to the mapmaker on pwg-scripts, given inside the config file under mapmaker_path needs to be modified.


sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,341 @@
...
+import scipy
+from pixell import enmap
+from sotodlib.site_pipeline.utils.pipeline import main_launcher
+import importlib
+import sys
+from pathlib import Path
+
+warnings.filterwarnings('ignore')
+
+# sys.path.append('/home/gs8865/work/scripts/pwg-scripts/flp/planet_mapmaker')
```
mmccrackan, line 16: Fine to remove the commented lines from when you were running this in your own setup.

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,341 @@
...
+
+    if rad_fwhm_min is None:            
+        if band == 'f090':
+            rad_fwhm_min = 20
+        else:
+            rad_fwhm_min = 15
+    try:
+        rad_prof = calc_rad_profile(solved_maps[0])  # solved_maps[0] is imap
+    except Exception as e:
+        print(f'{obs_id} failed fit. Skipping.')
```
mmccrackan, line 181: Can we use a logger instead of print statements throughout the code?

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,341 @@
...
+        return False, None
+    try:
+        profile = 10*np.log10(rad_prof['profile'])
+        radius = rad_prof['radius']
+        _3db = np.nanargmin(np.abs(profile+3)) #finding minimum distance to -3 dB
+        fwhm = radius[_3db]*2        
+        
+    except Exception as e:
+        print(e)
```
mmccrackan, line 190: Maybe add the full traceback (example: https://github.com/simonsobs/sotodlib/blob/75b33658771344fd2458b12b684e72c6516f4bf0/sotodlib/preprocess/preprocess_util.py#L50)

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,341 @@
...
+        print(f'{obs_id} does not have fwhm near -3db.')    
+        return False, None
+    
+    if fwhm > rad_fwhm_max or fwhm < rad_fwhm_min:
+        print(f'{obs_id} has poor fwhm fit: {fwhm}') 
+        return False, None 
+    if obs_id == 'obs_1724567567_satp1_1111111' or obs_id == 'obs_1724553289_satp1_1111111':
```
mmccrackan, line 197: We probably don't want hardcoded observations to skip in the actual site pipeline scripts.  Is there a way you can filter these out up front?  Or can we just run them anyway and tell they are bad from the outputs

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,341 @@
...
+        print(f'{obs_id} skipped! -- seeing double!')
+        return False, None
+    if obs_id == 'obs_1732418203_satp3_1111111':
```
mmccrackan, line 200: Same as above comment about hardcoded obs_ids.

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,341 @@
...
+        fwhm_max (float, optional): Max fitted fwhm in arcminutes. Defaults to None.
+        fwhm_min (float, optional): Min fitted fwhm in arcminutes. Defaults to None.
+        mask_rad (float, optional): Radius [arcmin] to compare fit to observed. Defaults to None.
+    """
+
+    size = len(maps)
+    q_scores = np.empty(size)
+    popts = np.empty([size, 7])
+    if fwhm_max is None:
+        if band == 'f090':
```
mmccrackan, line 221: Should you have options for f220/280 and f030/f040 when we start making maps with those?  Might be able to define dictionaries at the top of the code perhaps.

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,341 @@
...
+        obs_id = filename.rsplit("_", 1)[0]
+        solved_maps = enmap.read_fits(m)
+
+        success, rad_prof = map_selection(band, obs_id, solved_maps=solved_maps,
+                                              rad_fwhm_max=fwhm_max, rad_fwhm_min=fwhm_min)
+        
+        if success:
+            try:  
+                popt, _, = fit_gaussian_to_beam(solved_maps)
+            except Exception as e:
```
mmccrackan, line 250: Same as comment above about the full traceback.

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,341 @@
...
+                print(f'{obs_id} failed 2D Gauss fit: {e}')
+                quality = np.inf
+                popt= np.full(7,np.nan)
```
mmccrackan, line 253: I'm guessing this code isn't planned on being used for the LAT so you will only ever expect 7 wafers.  Might want to define a variable somehwere and replace the 7's throughout with that.

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,341 @@
...
+
+    return parser
+
+def main(config_file: str, wafer: str, freq_channel: str, obs_id: str, verbosity: int):
+    configs = read_configs(config_file)
+    
+    repo_root = Path(configs["mapmaker_path"])
+    module_path = repo_root / "planet_mapmaking.py"
+    
+    sys.path.insert(0, str(repo_root))
```
mmccrackan, line 306: I know I suggested this, but just wanted to confirm you tested the importlib approach works in your own setup?

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,341 @@
...
+
+
+def get_parser(parser=None):
+    if parser is None:
+        parser = argparse.ArgumentParser()
+    parser.add_argument('--config_file', type=str, help='path to config file')
+    parser.add_argument('--freq_channel', type=str, help='frequency channel of observation')
+    parser.add_argument('--wafer', type=str, help='wafer of observation')
+    parser.add_argument('--obs_id', type=str, default=None, help='obs id to be mapped')
+    # parser.add_argument('--nproc', type=int, default=2, help='Number of processes to use')
```
mmccrackan, line 295: Was this meant to be used for the mapmaker?  If not passing it, is there a default (we need to be careful about resources on the site).

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,341 @@
...
+    
+    qscores, popts = calc_map_quality(maps, band)
+    threshold_quality = 0.05
+    if qscores[0] > threshold_quality:
+        print( f'Observation {obs_id} quality is not good enough '
+           f'(RMSE={qscores[0]:.4f})')
+    else: 
+        print(f'{obs_id} is good')
+    
+# if __name__ == '__main__':
```
mmccrackan, line 335: Fine to remove.

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,341 @@
...
+    return q_scores, popts
+
+
+def get_parser(parser=None):
+    if parser is None:
+        parser = argparse.ArgumentParser()
+    parser.add_argument('--config_file', type=str, help='path to config file')
+    parser.add_argument('--freq_channel', type=str, help='frequency channel of observation')
+    parser.add_argument('--wafer', type=str, help='wafer of observation')
+    parser.add_argument('--obs_id', type=str, default=None, help='obs id to be mapped')
```
mmccrackan, line 294: Typically for the site pipeline scripts, we want to run on all recent obs_ids that haven't been processed yet as it isn't really possible to pass in a single obs_id at a time and have it run automatically.  We also don't want to re-run on ones that are already done.  So I think passing in an obsdb query string is probably the right way to go with this (i.e. select planet obs only).  We have a function that will accept a query and return a list of obs_ids for site-pipeline scripts here:

https://github.com/simonsobs/sotodlib/blob/75b33658771344fd2458b12b684e72c6516f4bf0/sotodlib/site_pipeline/utils/obsdb.py#L6

We would probably to remove the freq and wafer inputs (except as optional overrides maybe) too otherwise we'll need to have multiple separate flows for each one of them.  In preprocessing, we have a function that when given a "grouping" like wafer_slot and wafer.bandpass it will find all available options for each obs_id:

https://github.com/simonsobs/sotodlib/blob/75b33658771344fd2458b12b684e72c6516f4bf0/sotodlib/preprocess/preprocess_util.py#L363

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,341 @@
...
+
+    maps_dire_band = os.path.join(map_path, band, "map")
+    matching_files = [os.path.join(maps_dire_band, f) for f in os.listdir(maps_dire_band) if f.startswith(obs_id) and f.endswith(".fits")]
+
+    if len(matching_files) == 0:
+        raise FileNotFoundError(f"No map found for obs_id={obs_id}")
+    map_file = matching_files[0]
+    maps = [map_file]    
+    
+    qscores, popts = calc_map_quality(maps, band)
```
mmccrackan, line 327: We'll want this to save the results somehow.  I'm thinking a [ManifestDb](https://github.com/simonsobs/sotodlib/blob/75b33658771344fd2458b12b684e72c6516f4bf0/sotodlib/core/metadata/manifest.py#L319) like the other metadata products might be best.  You will have some number of hdf5 files containing the output information and a matching ManifestDb sqlite file that tells you which file and dataset the obs_id is in.  You can build a [ResultSet](https://github.com/simonsobs/sotodlib/blob/75b33658771344fd2458b12b684e72c6516f4bf0/sotodlib/core/metadata/resultset.py#L6) from you outputs and save using [write_dataset](write_dataset).  See update_det_cal for an example:

https://github.com/simonsobs/sotodlib/blob/75b33658771344fd2458b12b684e72c6516f4bf0/sotodlib/site_pipeline/update_det_cal.py#L826

sotodlib/site_pipeline/satp1_detcen_jupiter_sitecheck.yaml
```diff
@@ -0,0 +1,497 @@
+context_file: "/home/gs8865/work/files/contexts/satp1/use_this_local_251215_tmatm.yaml" # path to the context file, same as 251215 for comparing this with 2025/12/15 result
```
mmccrackan, line 1: We can discuss offline, but we'll put configs into https://github.com/simonsobs/site-pipeline-configs.  I can handle that once I work out all the paths, so you can remove this file.

mmccrackan (changes requested):
Overall looks solid, just some starting comments mostly on how it integrates into the site-pipeline work flow.

sotodlib/site_pipeline/compute_data_quality.py
GoureeSankar: yes!

sotodlib/site_pipeline/compute_data_quality.py
GoureeSankar: i dont think it is being used for the particular function, can remove it for individual obsid mapmakings.

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,556 @@
...
+            idx = (r > r_bins[i]) & (r < r_bins[i+1])
+            ravg = np.nanmean(map[idx])
+            vars = np.nanstd(map[idx])
+            
+            rad = np.append(rad,(r_bins[i] + r_bins[i+1])/2)
+            amps = np.append(amps,ravg)
+            stdevs = np.append(stdevs, vars)
+    
+    
+    #take peak for normalization as maximum pixel value within 2 arcmin of r=0, so as to avoid degeneracy with FWHM in fitting
```
mmccrackan: Add space after #.

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,556 @@
...
+    pos = maps.posmap()
+    ra  = pos[1]
+    dec = pos[0]
+    max_pix = imap.max()
+    Tp0 = [max_pix,0,0,0.003,0.003,0.0,0]
+    popt, pcov = scipy.optimize.curve_fit(twoD_Gaussian, [ra, dec], imap.ravel(),Tp0)
+    
+    return popt, pcov
+
+def get_errors(e):
```
mmccrackan: Fix indents.

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,556 @@
...
+        band (str): 'f090' or 'f150'
+        solved_maps (ndmap): Map to check beam properties
+        rad_fwhm_max (float) : max fwhm [arcmin] for map selection
+        rad_fwhm_min (float) : min fwhm [arcmin] for map selection
+    Returns:
+        True/False: bool
+    """
+    logger = logging.init_logger("compute_data_quality", verbosity=verbosity)
+    
+    # if rad_fwhm_max is None:
```
mmccrackan: Fine to remove commented outdated code.

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,556 @@
...
+            popt=np.full(n_params, np.nan)
+            
+        q_scores[idx] = quality
+        popts[idx] = popt
+    
+    return q_scores, popts
+
+def add_to_failed_cache(obs_id, band, ws, msg, failed_cache_file):
+    """Cache one failed (obs_id, band, wafer) combination."""
+    logger = logging.init_logger("compute_data_quality")
```
mmccrackan: Can this just be passed in to the function?

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,556 @@
...
+    parser.add_argument('--failed_cache_file', type=str, default='failed_obsids.yaml', help='Path to store failed obsids')
+
+    return parser
+
+def main(config_file: str, query: str, update_delay: int, verbosity: int, h5_path: str, 
+         h5_unix_digits: int, index_path: str, failed_cache_file: str, planet_tags: Optional[List[str]] = None,):
+
+    configs = read_configs(config_file)
+    context = core.Context(configs["context_file"])
+    map_path =  configs['mapmaking']['map']['save_dire']
```
mmccrackan: Probably should use the more standard `save_dir` rather than `site_dire` here and in other places.

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,556 @@
...
+    module_path = repo_root / "planet_mapmaking.py"
+    
+    sys.path.insert(0, str(repo_root))
+    spec = importlib.util.spec_from_file_location("planet_mapmaking", module_path)
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    
+    logger = logging.init_logger("compute_data_quality", verbosity=verbosity)
+    
+    obsid_list = get_obslist(context = context, query = query, update_delay = update_delay, tags = planet_tags, planet_obs = True)
```
mmccrackan: Don't need spaces before and after the equals sign.

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,556 @@
...
+                obs = context.obsdb.get(str(obs_id), tags=True)
+                tags = obs["tags"]
+                wafers = [tag for tag in tags if tag.startswith("ws")]
+
+                if not wafers:
+                    raise ValueError(
+                        f"No wafer tags found for obs_id={obs_id}. Available tags: {tags}"
+                    )
+
+            except KeyboardInterrupt:
```
mmccrackan: Just curious but why do we need to check for `KeyboardInterrupt` here?

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,556 @@
...
+
+    if not success:
+        if msg is None:
+            msg = "unknown error"
+        return add_to_failed_cache(obs_id, band, wafer, msg, failed_cache_file)
+
+    obsid_result = np.array(
+        [(obs_id, rmse, wafer, band, True)],
+        dtype=[
+            ("obs_id", "U50"), ("rmse", "f8"), ("wafer", "U10"),
```
mmccrackan: We should use `dets:wafer.bandpass` and `dets:wafer_slot `instead of band and wafer. Also can you try loading and obs_id with your metadata added as a context file entry to make sure that works correctly?

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,556 @@
...
+import importlib
+import sys
+import traceback
+from pathlib import Path
+from typing import Optional, List
+from sotodlib.site_pipeline.utils import logging
+from sotodlib.site_pipeline.utils.obsdb import get_obslist
+from sotodlib.io.metadata import write_dataset, ResultSet
+
+warnings.filterwarnings('ignore')
```
mmccrackan: What was this guarding against?

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,556 @@
...
+
+# running the planet mapmaker will produce the maps and database, and store it in a location specified by the paths in the config file
```
mmccrackan: Can the planet mapmaker load from an existing preprocess database?  In other words, if preprocessing has already been run, can the mapmaker find that database, load it, and make a map from it or does it always need to run it for the first time.  If it can load it, it might be useful (but not strictly necessary for now) to separate out the preprocessing as a separate Prefect flow.

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,556 @@
...
+
+    Args:
+        map (ndmap): 2D ndmap to profile
+        binsize (float, optional): radial bin size
+        normalize (bool, optional): Normalizes profile to center pixel
+    Returns:
+        rad_prof (dictionary): Keys of 'radius', 'profile', and 'bin_stdev'
+    """
+    rad_prof = {'radius':[], 'profile':[],'bin_stdev':[]}    
+    rad = np.array([])
```
mmccrackan: It might be slightly more efficient to pre-allocate these with `np.empty` or `np.full` instead of appending I think.

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,556 @@
...
+
+    return rad_prof
+
+def fit_gaussian_to_beam(maps):
+    imap = maps[0]
+    pos = maps.posmap()
+    ra  = pos[1]
+    dec = pos[0]
+    max_pix = imap.max()
+    Tp0 = [max_pix,0,0,0.003,0.003,0.0,0]
```
mmccrackan: Are the sigmas hardcoded for a particular band?

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,556 @@
...
+                )
+                continue
+
+            # Process each wafer independently.  This guarantees that each
+            # (obs_id, band, wafer) has its own map-quality calculation and
+            # that a failure in one wafer does not affect the others.
+            for ws in wafers:
+                cache_key_name = f"{obs_id}_{band}_{ws}"
+
+                if cache_key_name in failed_obsids:
```
mmccrackan: You might want to add on `overwrite` option in the config file that allows you to re-run failed cases just in case you discover a bug or want to change parameters.

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,556 @@
...
+                        f"{obs_id} is known to be bad for {band} and {ws}. "
+                        "Skipping this wafer."
+                    )
+                    continue
+
+                try:
+                    logger.info(f"Processing obs_id={obs_id}, band={band}, wafer={ws}")
+
+                    dets = {"wafer_slot": [ws], "wafer.bandpass": band}
+                    module.planet_mapmake_eachobs(config_file, obs_id, dets, verbosity)
```
mmccrackan: Similarly, you might want to check first if the map file already exists and only re-run it if you explicitly decide to overwrite it.  You could just then load it without re-running the mapmaker.  This would be a different case compared to when the outputs were bad.

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,556 @@
...
+
+    if os.path.exists(failed_cache_file):
+        with open(failed_cache_file, "r") as f:
+          d = yaml.safe_load(f) or {}
+    else:
+        d = {}
+
+    failed_obsids = set(d.keys())
+
+    for obs_id in obs_ids:
```
mmccrackan: This is not filtering by those that are already in your output database right?  So it will re-run them if they didn't fail?

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,556 @@
...
+                        raise FileNotFoundError(f"No map found for obs_id={obs_id}, band={band}, wafer={ws}")
+                    if len(matching_files) > 1:
+                        raise RuntimeError(f"Found multiple maps for obs_id={obs_id}, band={band}, "
+                            f"wafer={ws}: {matching_files}")
+
+                    map_file = matching_files[0]
+                    qscores, _ = calc_map_quality([map_file], band, configs)
+                    quality_score = float(qscores[0])
+
+                    threshold_quality = 0.05
```
mmccrackan: Should this be a config option?

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,556 @@
...
+    scheme.add_exact_match("obs:obs_id")
+    scheme.add_exact_match("band")
+    scheme.add_exact_match("wafer")
+    scheme.add_data_field("success")
+    scheme.add_data_field("rmse")
+    scheme.add_data_field("dataset")
+    return core.metadata.ManifestDb(index_path, scheme=scheme)
+
+
+def handle_result(obs_id, rmse, wafer, band, success, h5_path, h5_unix_digits, index_path, failed_cache_file, msg=None):
```
mmccrackan: This doesn't save the fit parameters themselves.  Might be useful to keep those around.

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,556 @@
...
+        filename=relpath, replace=True, )
+    return True
+
+
+def get_parser(parser=None):
+    if parser is None:
+        parser = argparse.ArgumentParser()
+    parser.add_argument('--config_file', type=str, help='path to config file')
+    parser.add_argument('--query', type=str, default=None, help='query string for obsdb')
+    parser.add_argument('--update_delay', type=int, default=None, help='days to subtract from current time')
```
mmccrackan: Better wording might be `number of days to subtract from the current time to set the minimum ctime.`

sotodlib/site_pipeline/compute_data_quality.py
mmccrackan: One other question, but how does the mapmaker handle a case where the wafer is missing in the obs_id?  It just keeps running but doesn't output the map?

mmccrackan (changes requested):
Some additional comments, which are mainly just formatting suggestions and some questions.  Might want to see if you can pass it through a python linter just to clear up some extra/missing spaces.

I think the only major things are about the output file format, adding an overwrite option, and the question about whether it re-runs already completed obs or not.

mmccrackan:
One other comment but can you move the Yaml config file to a draft PR over at https://github.com/simonsobs/site-pipeline-configs and remove it here?  Fine to leave the paths as they are until we finalize them.

sotodlib/site_pipeline/compute_data_quality.py
GoureeSankar: Right now, the mapmaker is designed in such a way to execute the preprocessing and then do the mapmaking. I believe this can be changed, but it might not be possible anytime soon. There are no immediate plans to do this, but I can ask them.

sotodlib/site_pipeline/compute_data_quality.py
GoureeSankar: Probably multiple things which might cause subtle warnings. One instance might be where only information about xxxx detectors out of total yyyy detectors is available. I can't think of specific examples right now, but had that in case somethings popped up which didn't really affect the rest of the process. I can remove it if you want.

sotodlib/site_pipeline/compute_data_quality.py
GoureeSankar: Right now, it is assumed to be the same for both MF bands. But as it is just an initial guess which goes into the scipy minimizer, it does not matter if a value is hardcoded at this stage as that will be replaced by the best-fit value in popt.

sotodlib/site_pipeline/compute_data_quality.py
GoureeSankar: At the moment, yes. It will re-run unless it is not in the failed_cache. 
If you don't want to re-run the successful obsids, one way is to modify the input obsid query to remove those observations. Otherwise, I can modify it in two ways: if it is already in the successful db, do not run it, or give a command which will run even the successful obsids. I don't think it is straightforward to give it a list of obsids and tell it to re-run only a subset of them, even if they are all successful. (not described in the most elegant way, but I hope you get what I meant - happy to chat more about this)

sotodlib/site_pipeline/compute_data_quality.py
GoureeSankar: I copied it from the add_to_failed_cache function which I took update_det_cal. Can get rid of it if you think it is unnecessary. It just prevents an observation from being mislabeled as bad.

sotodlib/site_pipeline/compute_data_quality.py
GoureeSankar: I don't think this is something to worry about. 
`dets = {"wafer_slot": [ws], "wafer.bandpass": band} `

It is the same ws and band that is being passed into handle_result, so there is actual logical change happening here when if we change it

sotodlib/site_pipeline/compute_data_quality.py
GoureeSankar: Sorry, but I don't get what you mean... there is no `site_dire` here...?

sotodlib/site_pipeline/compute_data_quality.py
GoureeSankar: >  how does the mapmaker handle a case where the wafer is missing in the obs_id

I don't get what you mean here. `{"wafer_slot": [ws], "wafer.bandpass": band}` is passed into the mapmaker with the obsid. Inside the mapmaker, there is the call `  meta = context.get_meta(obs_id, dets = wafer_info)`, so if the wafer is not there, then this would return an error and the code would flow to the Exception part.

GoureeSankar: Also, I have modified the code to do the following:
If there are known bad tuple in failed cache, it is skipped entirely.
If the map already exists, then don't run the mapmaker. Just calculate/load its quality.
If the map doesn't exist, then run the mapmaker, find the newly generated map, then calculate quality.
If we want to regenerate existing maps intentionally, added an overwrite_maps flag.

sotodlib/site_pipeline/compute_data_quality.py
mmccrackan: Sorry I meant `save_dir` not `site_dir`.  Fine to leave as is I think if you prefer.

sotodlib/site_pipeline/compute_data_quality.py
mmccrackan: I don't think there's an issue having it here, but we'll be running this in a Prefect flow, so it won't ever really get triggered in our typical usage of this code.

sotodlib/site_pipeline/compute_data_quality.py
mmccrackan: I think maybe we can comment this out when running on site at least for a while just so it is easier to see any errors with deployment.

The missing detectors warning you mention is when some of the metadata doesn't have entries for all detectors.  For some metadata products that indicates something needs updating (though that is outside the scope of your code).

sotodlib/site_pipeline/compute_data_quality.py
mmccrackan: Usually want to be able to add the metadata generated from these scripts to a context file and load it.  In such a case we want the names in the output files to match the naming convention of other metadata so it knows where to add them when you load.

sotodlib/site_pipeline/compute_data_quality.py
mmccrackan: Okay, in that case perhaps you could add it as a config option with these as the default in the code so that the code can be used as is when/if we start running it on f220/280 and f030/f40?

sotodlib/site_pipeline/compute_data_quality.py
mmccrackan: Regarding 

`how does the mapmaker handle a case where the wafer is missing in the obs_id`

Yeah I was just wondering if it silently skipped things or was guaranteed to throw an error.  This sounds fine and I see the error message is added to the `failed_cache` so we can track why things fail.  Also good that transient metadata errors are not added to the `failed_cache`.

sotodlib/site_pipeline/compute_data_quality.py
mmccrackan: Yeah the ideal way to do this is to load your output file, search if an obs, wafer, band entry is in it and skip it unless `overwrite` is `True`.

mmccrackan:
One other comment but we should probably rename the file to something that is more planet map specific.

Also if you could add a short entry to the docs, that would be awesome: https://github.com/simonsobs/sotodlib/blob/master/docs/site_pipeline.rst

sotodlib/site_pipeline/compute_data_quality.py
```diff
@@ -0,0 +1,556 @@
...
+                    dets = {"wafer_slot": [ws], "wafer.bandpass": band}
+                    module.planet_mapmake_eachobs(config_file, obs_id, dets, verbosity)
+
+                    maps_dire_band = os.path.join(map_path, band, "map")
+                    if not os.path.isdir(maps_dire_band):
+                        raise FileNotFoundError(f"Map directory does not exist: {maps_dire_band}")
+
+                    # The existing calc_map_quality code assumes filenames end
+                    # in _<wafer>.fits, so require that exact tuple match here.
+                    matching_files = []
```
mmccrackan: The filenames from the mapmaker should have a standard format for a given obs_id, wafer, and band right? Can't you just construct that string here and look for it instead of searching and restricting from all the files every time?

sotodlib/site_pipeline/compute_data_quality.py
GoureeSankar: Ok, noted!

sotodlib/site_pipeline/compute_data_quality.py
GoureeSankar: Ok, done

sotodlib/site_pipeline/compute_data_quality.py
GoureeSankar: Included it like that

sotodlib/site_pipeline/compute_data_quality.py
GoureeSankar: Ok, I will take it out then

PR #1677, opened 2026-06-24
Fix numpy failure case in det_cal

mmccrackan:
Some obs_ids are silently failing on the LAT with the following error:

```
obs_1782017305_lati3_111: "Traceback (most recent call last):\n  File \"/usr/local/lib/python3.12/site-packages/sotodlib/site_pipeline/update_det_cal.py\"\
  , line 629, in get_cal_resset\n    if not ridx:  # Channel doesn't exist in IV analysis\n\
  \           ^^^^\nValueError: The truth value of an empty array is ambiguous. Use\
  \ `array.size > 0` to check that an array is not empty.\n"
```

This just adds a fix so that it will continue and not fail as was done for `update_smurf_caldbs` in #1619.

### DIFF
--- a/sotodlib/site_pipeline/update_det_cal.py
+++ b/sotodlib/site_pipeline/update_det_cal.py
@@ -626,7 +626,7 @@ def get_cal_resset(cfg: DetCalCfg, obs_info: ObsInfo,
                 continue
 
             ridx = np.where((iva.bands == band) & (iva.channels == chan))[0]
-            if not ridx:  # Channel doesn't exist in IV analysis
+            if ridx.size == 0:  # Channel doesn't exist in IV analysis
                 continue
 
             ridx = ridx[0]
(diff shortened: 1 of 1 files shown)


ykyohei (commented):
I tested update_det_cal  for this obsevation at tiger3 and it didn't have any error and I'm bit confused.
Is it possible that this is due to race conditions of site pipeline..?

mmccrackan:
> I tested update_det_cal for this obsevation at tiger3 and it didn't have any error and I'm bit confused. Is it possible that this is due to race conditions of site pipeline..?

Multiple `update_det_cal` flows aren't permitted to run at the same time nor can they run at the same time as `update_obsdb` or `update_smurf_caldbs` so I wouldn't have thought it was a race condition.

I see many more of these kinds of failures in the `failed_obsid.yaml` file as well.

ykyohei (approved):
Okay I learned that this used to be just DeprecationWarning but from numpy 2.2.0 this leads to ValueError.
Thanks for flagging this!
https://numpy.org/devdocs/release/2.2.0-notes.html#expired-deprecations

mmccrackan:
> Okay I learned that this used to be just DeprecationWarning but from numpy 2.2.0 this leads to ValueError. Thanks for flagging this! https://numpy.org/devdocs/release/2.2.0-notes.html#expired-deprecations

Great, I am going to remove the `obs_ids` with this error from the `failed_obs_ids` file so they will re-run.

PR #1678, opened 2026-06-24
Add sqlite timeout to obsdb

mmccrackan:
Copies #1496 and adds the ability to specify the timeout for the `obsdb` sqlite connection through the same environment variable.  This may help alleviate the locked `obsdb` failures we are having for the `update_obsdb` Prefect flow.


mhasself (approved):
Thanks.

This envvar is now used 3 times in code, but does not appear in any docs or docstrings.  I think doesn't need to be your job but we should https://github.com/simonsobs/sotodlib/issues/1682

PR #1680, opened 2026-06-25
apodize with flags

ykyohei:
Added an option to apodize with flags to apodize_cosine.
There was a bug when we pass a window argument, added test.

### DIFF
--- a/sotodlib/tod_ops/apodize.py
+++ b/sotodlib/tod_ops/apodize.py
@@ -91,11 +91,12 @@ def get_apodize_window_from_flags(aman, flags, apodize_samps=200):
 
 
 def apodize_cosine(aman, signal_name='signal', apodize_samps=1600, in_place=True,
-                   apo_axis='apodized', window=None):
+                   apo_axis='apodized', window=None, flags=None):
     """
     Function to smoothly filter the timestream to 0's on the ends with a
-    cosine function. If window is provided, multiply the window function to 
-    aman[signal_name].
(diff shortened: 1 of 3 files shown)


msilvafe:
Can you share some more details on the bug?

sotodlib/tod_ops/apodize.py
```diff
@@ -91,25 +91,30 @@ def get_apodize_window_from_flags(aman, flags, apodize_samps=200):
...
+            window = get_apodize_window_from_flags(aman, flags, apodize_samps)
+        else:
+            window = get_apodize_window_for_ends(aman, apodize_samps)
 
     if in_place:
-        aman[signal_name] *= w
+        aman[signal_name] *= window
     else:
         aman.wrap_new(apo_axis, dtype='float32', shape=('dets', 'samps'))
-        aman[apo_axis] = aman[signal_name]*w
```
ykyohei, line 114: Ya, sorry. This is very minor bug.
When window is passed, w is undefined and this function did not work. I changed from w to window.

msilvafe: Were we just not using this before or was there a case where we were trying to use it but nothing was being applied?

ykyohei: This was not used through preprocess. I'm not sure if this is used but I noticed this when I added flags.

msilvafe: Gotcha, thanks!

msilvafe (approved):

PR #1681, opened 2026-06-26
Memory saving for stimulator calculation

YudaiSeino:
I change the way to make coadd data to save memory usage. New code calculate coadd data and delete filtered tod for each  chopping frequency. I also simplified some code to wrap new axis manager fields.

### DIFF
--- a/sotodlib/stimulator/stimulator.py
+++ b/sotodlib/stimulator/stimulator.py
@@ -109,53 +109,28 @@ def preprocessing(aman, hkdata, idxs=None, n_bins=40, delete_filtered_tod=True):
         else:
             valid_timeconstant = True
 
+    # Co-add data and fit it for each frequency
     model, params_base = get_fit_params(cal_type="coadd")
     initialize_aman(aman, "coadd", model, n_bins)
 
-    # Calculate filtering frequencies
-    if valid_gain:
-        filter_freqs = {}
-        if round(aman.stm_cal.chopping_freqs[0]) != CHOPPING_FREQS["f1"]:
-            filter_freqs["f1_gain"] = round(aman.stm_cal.chopping_freqs[0])
+    for freq_key in aman.stm_cal.chopping_freq_key.vals:
+        freq = aman.stm_cal.chopping_freqs[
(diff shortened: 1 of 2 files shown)


sotodlib/stimulator/stimulator.py
```diff
@@ -795,21 +704,82 @@ def ensure_wrapped(parent_field, key, arr=None, axis=None):
...
+    if "signal_iirc" not in aman:
+        iirc_filter = tod_ops.filters.iir_filter(aman, invert=True)
+        signal_new = tod_ops.fourier_filter(aman, iirc_filter, signal_name="signal")
+        aman.wrap("signal_iirc", signal_new, [(0, "dets"), (1, "samps")])
+
+    # Make HPFed data
+    if "signal_hpf" not in aman:
+        hpf = tod_ops.filters.high_pass_sine2(hpf_cutoff, hpf_width)
+        signal_new = tod_ops.fourier_filter(aman, hpf, signal_name="signal_iirc")
+        aman.wrap("signal_hpf", signal_new, [(0, "dets"), (1, "samps")])
```
ykyohei: signal_iirc is coadded but it's actually not fitted. 
If we do not need signal_iirc, we can remove it from coadd and use filter chain and save memory bit more.

```
filt = tod_ops.filters.FilterChain([iirc_filter, hpf])
signal_new = tod_ops.fourier_filter(aman, filt, signal_name="signal")
aman.wrap("signal_hpf", signal_new, [(0, "dets"), (1, "samps")])
```
I think raw unfiletered signal is also not used, if it's fine we can overwite raw signal by high pass filtered one to make it even more efficient.

ykyohei (commented):
This is great improvents, thanks!
This made me remember that we can use filter chain for fft filter, and we can make it more efficient.

sotodlib/stimulator/stimulator.py
YudaiSeino: I deleted raw signal tod. I'll change to make iir inverted coadd data optionally.

ykyohei (approved):
Looks good, thank you!

PR #1685, opened 2026-06-29
Ignore archive by default when running site preproc in toast load

tskisner:
This has caused many headaches when loading data into toast with site preproc configs that specify an archive.  In most cases we want to run that preproc on the fly.  Default to that case.


keskitalo (approved):
Great!

PR #1688, opened 2026-07-02
Change default apodization to C1

ykyohei:
The docstring says we apodize by cosine taper but I have realized that it has not been standard one.
I'm not sure if it's okay to change the default, but this PR changes the default to standard cosine taper (hann window).

```
if apo_type == 'C1':
    cosedge = 0.5 * (np.cos(np.linspace(0, np.pi, apodize_samps)) + 1)
elif apo_type == 'old_default':
    cosedge = np.cos(np.linspace(0, np.pi/2, apodize_samps))
```

The calculation of weight_for_signal for hwpss and azss were redundant as get_apodize_window_from_flags.
They are simplified.

apodize_C1

### DIFF
--- a/sotodlib/tod_ops/apodize.py
+++ b/sotodlib/tod_ops/apodize.py
@@ -8,7 +8,12 @@ def get_apodize_window_for_ends(aman, apodize_samps=1600, apo_type='C1'):
     Args:
         aman: An axismanager
         apodize_samps (int): Number of samples to apply the cosine taper to at each end.
-        apo_type: Type of apodization, default is C1.
+        apo_type: (str): Type of apodization window applied to the edges. Options are:
+            - ``'C1'``: Standard cosine (Hann) taper, i.e. a half-cosine that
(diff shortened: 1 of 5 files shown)


sotodlib/tod_ops/apodize.py
```diff
@@ -1,25 +1,30 @@
...
 
-def get_apodize_window_for_ends(aman, apodize_samps=1600):
+def get_apodize_window_for_ends(aman, apodize_samps=1600, apo_type='C1'):
     """
     Generate an apodization window using a cosine taper at the beginning and end.
 
     Args:
         aman: An axismanager
         apodize_samps (int): Number of samples to apply the cosine taper to at each end.
+        apo_type: Type of apodization, default is C1.
```
msilvafe: I think describe what "C1" is and what "old_default" is here in some more detail then for all other functions that use these anodization arguments you can just refer to the docstring of this function.

msilvafe (approved):
Thanks for correcting and simplifying. I didn't realize we were doing something non-standard!

Just one inline comment about docstrings.

Issue #788, opened 2024-03-30
`get_turnaround_flags` needs to handle no az motion


msilvafe:
Currently if the Az is stationary `get_turnaround_flags` errors out with an error like this:

```
  File "/so/home/msilvafe/.local/lib/python3.10/site-packages/sotodlib/preprocess/processes.py", line 647, in process
    tod_ops.flags.get_turnaround_flags(aman, **self.process_cfgs)
  File "/so/home/msilvafe/.local/lib/python3.10/site-packages/sotodlib/tod_ops/flags.py", line 199, in get_turnaround_flags
    part_slices_ta_masked = [slice(None, np.where(_ta_flag)[0][0]), slice(np.where(_ta_flag)[0][-1], None)]
IndexError: index 0 is out of bounds for axis 0 with size 0
```

We need to catch this sort of error and return flags that are all false (or appropriately instantiated). A known satp1 obs that fails in this way is: `obs_1706255231_satp1_1111111` and can be used for testing.

mmccrackan:
For more recent versions of `get_turnaround_flags` the error output of observations where the Az does not vary significantly is:

`"/global/homes/m/mmccrack/so_home/repos/20240819_catchup/sotodlib/sotodlib/tod_ops/flags.py", line 216, in get_turnaround_flags`
    `peaks, _ = find_peaks(np.abs(matched), height=peak_threshold, distance=rel_distance_peaks*approx_samps_onescan)`
`"/global/u1/m/mmccrack/so_home/env/soconda_20240903_0.1.4/lib/python3.11/site-packages/scipy/signal/_peak_finding.py", line 941, in find_peaks
    raise ValueError('distance must be greater or equal to 1')`

This occurs because `np.ptp` is used to find the largest distance between max and min of Az, but this is small for stationary Az observations and may not meet the requirement `rel_distance_peaks*approx_samps_onescan` be greater than 1.  Some might be a few samples bigger though and would not fail.

msilvafe:
I think this is actually ok to fail, if one actually wanted to analyze a stare observation they they wouldn't be calling turnaround flagging.

PR #1689, opened 2026-07-02
BUG FIX: Updated get_turnaround_flags so that it can catch stationary…

aashrita:
… observations that may trickle through


aashrita:
This fix requires all the turnaround flag blocks to be updated with a threshold -- @mmccrackan had suggested that it throw an error if a value is not included so i chose not to make it optional. Can update this if you guys want a default value (e.g. 0.1 degree).

ykyohei (commented):
az_throw is included in obsdb, I think it may be simpler to filter out observations with no azthrow by the obsdb query.

aashrita:
@ykyohei this is an old commit and from a misunderstanding on what Michael had suggested. I'm recommitting what is better fix.

Also the stationary scan that trickled through for satp2 had an az_throw -- it was just very small from jitter in az

sotodlib/tod_ops/flags.py
```diff
@@ -203,6 +206,18 @@ def get_turnaround_flags(aman, az=None, method='scanspeed', name='turnarounds',
     """
     if az is None : 
         az = aman.boresight.az
+
+    # Get the throw (convert to degrees).
+    az_throw = np.ptp(az) * 90 / np.pi
+
+    # If the throw is below the threshold, just output empty flags
+    if az_throw < az_throw_threshold:
+        ta_flag = Ranges.from_bitmask(np.zeros(aman.samps.count, dtype=bool))
```
mmccrackan, line 215: I think this can be simplified to just use RangesMatrix.zeros?

mmccrackan (changes requested):
I think this will work fine and it is nice that it will bypass any of the older failures we had when the az didn't vary enough (https://github.com/simonsobs/sotodlib/issues/788).

I guess the only issue here is that this may require preprocessing to be re-run from scratch.  If you take an existing db, it won't necessarily cut those obs since the turnaround flags were already saved in `proc_aman`.  Even though they are calculated in `process` again (I assume that was to make them available to the mapmaker) those will be zeros in `aman`, not `proc_aman` and it won't be cut on the `select` step.  It might be easier to do the az throw check in the `select` step so existing obs_ids will be cut at the load stage.

sotodlib/tod_ops/flags.py
aashrita: Yes i can do the RangesMatrix.zeros and check it.

On the select thing, yes I can add that to avoid reruns but we should keep the fix in the flagging function because it fixes the bug _in_ the flagging itself.

sotodlib/preprocess/processes.py
```diff
@@ -1526,6 +1527,19 @@ def process(self, aman, proc_aman, sim=False, data_aman=None):
...
+            return meta
+        if proc_aman is None:
+            proc_aman = meta.preprocess
+    
+        ta = proc_aman[self.save_name].turnarounds
+    
+        if np.all(count_cuts(ta) == 0):
+            meta.restrict('dets', meta.dets.vals[:0])
+    
+        return meta
```
msilvafe: Look at how other methods implement this but you're not using in_place. If False you should return keep otherwise restrict as you do here.

sotodlib/tod_ops/flags.py
```diff
@@ -203,6 +206,18 @@ def get_turnaround_flags(aman, az=None, method='scanspeed', name='turnarounds',
     """
     if az is None : 
         az = aman.boresight.az
+
+    # Get the throw (convert to degrees).
+    az_throw = np.ptp(az) * 90 / np.pi
```
msilvafe, line 229: Why 90/pi and not 180/pi? Is this half of the throw? i.e. `ptp/2`?

msilvafe (changes requested):
A couple inline comments, will need to merge in master and resolve conflicts especially since #1713 touched this code a bunch. And still should resolve the one comment from @mmccrackan on using `RangesMatrix.zeros`

sotodlib/tod_ops/flags.py
aashrita: ya, full form np.ptp(az)/2 * 180/pi just bc that's def of throw and i just simplified. lmk if we have some weird other definition of throw tho that i should adhere to :P

msilvafe: That makes sense, thanks.

sotodlib/preprocess/processes.py
aashrita: done in latest commit

sotodlib/tod_ops/flags.py
aashrita: done in latest commit

mmccrackan (approved):

msilvafe (approved):

sotodlib/tod_ops/flags.py
```diff
@@ -218,6 +221,18 @@ def get_turnaround_flags(aman, az=None, method='scanspeed', name='turnarounds',
...
+
+    # If the throw is below the threshold, just output empty flags
+    if az_throw < az_throw_threshold:
+        ta_flag = RangesMatrix.zeros(aman.samps.count)
+        ta_exp = RangesMatrix([ta_flag for i in range(aman.dets.count)])
+        if method == 'scanspeed':
+            return ta_exp, ta_exp, ta_exp
+        return ta_exp
```
ykyohei: Sorry this needs to return three RangesMatrix now, with any methods. Because we changed in https://github.com/simonsobs/sotodlib/pull/1713

```suggestion
        return ta_exp, ta_exp, ta_exp
```
Can we add logger.info("az_throw is below threshold. Abort computing turnaround flags" or something here?

ykyohei (changes requested):

sotodlib/tod_ops/flags.py
```diff
@@ -161,7 +161,8 @@ def get_turnaround_flags(aman, az=None, method='scanspeed', name='turnarounds',
                          merge=True, merge_lr=True, overwrite=True,
                          t_buffer=2., az_buffer=None, smooth_seconds=0.5,
                          kernel_size=400, peak_threshold=0.1, rel_distance_peaks=0.3,
-                         truncate=False, qlim=1, merge_subscans=True, turnarounds_in_subscan=False):
+                         truncate=False, qlim=1, merge_subscans=True, turnarounds_in_subscan=False,
+                         az_throw_threshold=0.):
```
ykyohei, line 168: Do you want to change the default value of az_throw_thresold to 1, to be consistent with example config?

sotodlib/tod_ops/flags.py
```diff
@@ -209,6 +210,8 @@ def get_turnaround_flags(aman, az=None, method='scanspeed', name='turnarounds',
         (Optional). Also merge an AxisManager with subscan information.
     turnarounds_in_subscan : bool
         (Optional). Turnarounds are included as part of a subscan.
+    az_throw_threshold : float
+        (Optional). Minimum azimuth throw required to attempt turnaround flagging. Returns no turnarounds if below threshold.
```
ykyohei: Please describe unit of degrees.

sotodlib/tod_ops/flags.py
aashrita: No, default should have no change/cuts.

sotodlib/tod_ops/flags.py
aashrita: done.

sotodlib/tod_ops/flags.py
aashrita: added logging and removed the if statement in new commit.
i just removed the if statement because i originally added it in to account for the differen number of return values earlier. now that is no longer an edge case to worry about so it is removed.

ykyohei (approved):
Thanks, looks good to me. Please squash and merge.

PR #1690, opened 2026-07-03
Add option to ignore h5 context manager in preprocessing

mmccrackan:
This allows preprocessing to bypass the h5 context manager when running off of site.

### DIFF
--- a/sotodlib/core/util.py
+++ b/sotodlib/core/util.py
@@ -53,19 +53,22 @@ class H5ContextManager:
         Number of times to retry opening the file if it is locked.
     delay : int or float, optional
         Delay in seconds between retries.
+    override_temp : bool
+        Override creation of temporary file and just use original file.
     **kwargs :
         Additional keyword arguments passed to `h5py.File`.
     """
 
     temp_suffix = ".temporary"
 
-    def __init__(self, filename, mode="r", max_attempts=3, delay=5, **kwargs):
+    def __init__(self, filename, mode="r", max_attempts=3, delay=5, override_temp=False, **kwargs):
         self.filename = filename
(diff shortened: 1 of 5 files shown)


tskisner:
@mmccrackan, just checking- have you found any performance issue with the new temp file usage?  Or is this bypass just to avoid the copy when you can already guarantee that only a single process is accessing an HDF5 file?

mmccrackan:
> @mmccrackan, just checking- have you found any performance issue with the new temp file usage? Or is this bypass just to avoid the copy when you can already guarantee that only a single process is accessing an HDF5 file?

It seems when running SAT preprocessing, this incurs a big slowdown due to needing to copy both files simultaneously.  I need to look into why it didn't seem to affect things when testing on LAT ASO data on Nersc, but I'm guessing it was just due to having less data overall + single node usage.  We won't need it for running preprocessing off site so this was just to allow that.  Longer term I think we should add an `override_copy` option to the H5ContextManager to ignore copying as that will be more general for off-site usage.

tskisner:
Ok, the main use case was long-running rsync at site during simultaneous appending to the files.  Maybe there are other cases where h5 files are being appended while another process is reading, but I'm not sure.  Agree about having a disable switch in the context manager.

msilvafe:
Ya, I'll need to test this more. I had to roll back to before this commit to get my jobs to run without piling up a bunch of temp files.

mmccrackan:
Downgrading to draft for now as this will be replaced by #1703, so no review is required, but will keep available for those using it until the other PR is merged.

mmccrackan:
No longer needed.

PR #1691, opened 2026-07-07
warn or raise error when merge/wrap inconsistent axes

ykyohei:
Now I think this was probably intended behavior, but merge and wrap always silently intersect and restrict axes.
Added an on_mismatch to warn or raise error when the coverage of axes with same name are different.

This silent intesection/truncation caused issues in analysis several times, for example when we calculate psd with different length of trimmed data, or after downsampling.
We need to rename axes in such cases but I think default warning (or error) is helpful..

Example:
```
WARNING: sotodlib.core.axisman: Axis mismatch on merge/wrap causes truncation: axis 'dets' of self changes LabelAxis(3:np.str_('det0'),np.str_('det1'),np.str_('det2')) -> LabelAxis(2:np.str_('det0'),np.str_('det1')).
WARNING: sotodlib.core.axisman: Axis mismatch on merge/wrap causes truncation: axis 'samps' of self changes OffsetAxis(1000:None+0) -> OffsetAxis(900:None+0).
```

### DIFF
--- a/sotodlib/core/axisman.py
+++ b/sotodlib/core/axisman.py
@@ -1079,8 +1079,10 @@ class AxisManager:
         axes_out = self.intersection_info(self, *amans)
 
         # Apply the axis-mismatch policy before modifying anything.
-        assert on_mismatch in (None, 'warn', 'error')
-        if on_mismatch is not None:
+        if on_mismatch is None:
+            on_mismatch = 'intersect'
+        assert on_mismatch in ('intersect', 'warn', 'error')
+        if on_mismatch in ('warn', 'error'):
             report = []
             for name, ax in axes_out.items():
                 for src, aman in [('self', self)] + [
--- a/sotodlib/core/axisman.py
+++ b/sotodlib/core/axisman.py
@@ -1,3 +1,4 @@
+import logging
 import numpy as np
 from collections import OrderedDict as odict
 
@@ -11,6 +12,8 @@ except ImportError:
 import so3g
 
 from .util import get_coindices
+
+logger = logging.getLogger(__name__)
 
 
 class AxisInterface:
@@ -635,7 +638,7 @@ class AxisManager:
     # Add and remove data while maintaining internal consistency.
 
     def wrap(self, name, data, axis_map=None,
-             overwrite=False, restrict_in_place=False):
+             overwrite=False, restrict_in_place=False, on_mismatch='warn'):
         """Add data into the AxisManager.
 
         Arguments:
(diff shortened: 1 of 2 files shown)


sotodlib/core/axisman.py
```diff
@@ -635,7 +638,7 @@ def concatenate(items, axis=0, other_fields='exact'):
     # Add and remove data while maintaining internal consistency.
 
     def wrap(self, name, data, axis_map=None,
-             overwrite=False, restrict_in_place=False):
+             overwrite=False, restrict_in_place=False, on_mismatch='warn'):
```
mhasself: Change default to None, and pass through.

In the future we might add more flexibility to on_mismatch -- e.g. to accept is as a dict, where a keyword is specified for particular axes.  But expecting a string is fine for now.

sotodlib/core/axisman.py
```diff
@@ -1042,9 +1052,12 @@ def intersection_info(*items):
                         ax, False)
         return axes_out
 
-    def merge(self, *amans, restrict_in_place=False):
+    def merge(self, *amans, restrict_in_place=False, on_mismatch='warn'):
```
mhasself: Change default to "None", and then check/set default in main function body.

sotodlib/core/axisman.py
```diff
@@ -1064,6 +1077,28 @@ def merge(self, *amans, restrict_in_place=False):
 
         # Get the intersected axis descriptions.
         axes_out = self.intersection_info(self, *amans)
+
+        # Apply the axis-mismatch policy before modifying anything.
+        assert on_mismatch in ('intersect', 'warn', 'error')
+        if on_mismatch != 'intersect':
+            report = []
```
mhasself: Instead of this new logic, try to use the output of "self.intersection_info" to do this test.  You can probably compare old / new using `==` (they all have `__eq__` defined).

mhasself (changes requested):
Thanks -- this will be very useful.  As discussed, here are a couple of implementation detail requests.

Most importantly I'd like to leave the default behavior as 'intersect', for now.

sotodlib/core/axisman.py
ykyohei: Thanks for the suggestion. I've updated the logic to use `self.intersection_info` and `__eq__`

sotodlib/core/axisman.py
```diff
@@ -1064,6 +1077,25 @@ def merge(self, *amans, restrict_in_place=False):
 
         # Get the intersected axis descriptions.
         axes_out = self.intersection_info(self, *amans)
+
+        # Apply the axis-mismatch policy before modifying anything.
+        assert on_mismatch in (None, 'warn', 'error')
+        if on_mismatch is not None:
```
mhasself:
```suggestion
        if on_mismatch is None:
            on_mismatch = 'truncate'
        assert on_mismatch in ('truncate', 'warn', 'error')
        if on_mismatch in ('warn', 'error'):
```

mhasself (approved):
Please make "truncate" an option, but then convert None to "truncate" (the default).

Otherwise this looks good, thanks!

sotodlib/core/axisman.py
ykyohei: I'll make option name 'intersect' to be consistent with original descriptions.

PR #1694, opened 2026-07-11
Add other_fields to loader

ykyohei:
Motivation is to load preprocess archive per observation easily.
The problem was preprocess archive is usually computed per (band, wafer) and each has scalar fields that cannot be concatenated.
One solution was to drop them by `drop_fields `, but number of fields to drop is large for ISO preprocess archive and it can increase. 

This PR adds `other_fields` of aman.concatenate to loader to 'drop' them all at once or pick 'first'.
https://github.com/simonsobs/sotodlib/blob/88a3230ce63cb4430dea9a5bcec0b28787fc3cf0/sotodlib/core/axisman.py#L508-L516

### DIFF
--- a/sotodlib/core/metadata/loader.py
+++ b/sotodlib/core/metadata/loader.py
@@ -322,6 +322,9 @@ class SuperLoader:
         assert(len(results) > 0)
         if len(results) == 1:
             result = results[0]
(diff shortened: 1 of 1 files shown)


mhasself (approved):
Looks good to me!

PR #1697, opened 2026-07-15
Change turnound flags kernel shape for two_leg turnarounds

mjrand:
Changed the matched filter kernel shape so the kernel shape better matches the expected shape of the az velocity in two_leg turnarounds. The current kernel shape for turnaround flags does not properly find the midpoint of turnarounds because of the new two_leg turnaround shape. Here is a comparison of the matched peak for the current and new kernels with a two_leg turnaround:
image
image

Here is a comparison of the kernel shapes before and after this change:

image

The two changes are: 

1. Increasing the width of the kernel. 1500 samples is ~7.5 seconds, which is the approximate length of a turnaround with scan_v = 0.8 and scan_a = 0.25. We probably need to test this with many different scan speeds to make sure it works. Namely does this still work at scan_v = 0.5? Two_leg turnarounds are never run on planet scans so we may never need to look at higher acceleration scans.
2. Adding a 0 velocity center that spans ~1/5 of the turnaround (this may be larger than the 0 velocity portion of the turnaround. This 0 velocity center is vital in ensuring the flag is placed at the center of the turnaround. Otherwise the `find_peaks` function will find the center of the first/second leg and choose either "somewhat" arbitrarily. 

If the kernel size is too small (i.e. we use the original kernel_size=400 with the new kernel shape) the match will also fail:
image

Here is the comparison of matches using the new and current kernels on a standard turnaround with scan_v = 1.5.
image

image

The new kernel does move the flag by ~0.1-0.2 seconds, but that may not be a huge problem?

At scan_v = 0.8 the difference is even less noticeable:
image
image

Finally, I tried this on a Jupiter scan. The new kernel does seem to have a bit of an issue with finding the center of this standard turnaround.
image
image

However, the new shape works perfectly fine as long as we're using the original kernel size (400 indexes instead of 1500):
image

So we may want to feed in the kernel size based on the scan parameters. The kernel size should ~match the turnaround length which should always be `2 * scan_v / scan_a * 200`. This should let the new kernel shape handle even standard turnarounds. I think trying to feed in the turnaround method to switch between kernel shapes is a little overkill as we'll always have to ensure the kernel size matches the turnaround length anyways. The new kernel shape seems to do find with finding the center of standard turnarounds as long as the kernel size is correct!

Please try this out yourself and let me know if it works for you.

### DIFF
--- a/sotodlib/tod_ops/flags.py
+++ b/sotodlib/tod_ops/flags.py
@@ -156,7 +156,7 @@ def get_det_bias_flags(aman, detcal=None, rfrac_range=(0.1, 0.7),
 
 def get_turnaround_flags(aman, az=None, method='scanspeed', name='turnarounds',
                          merge=True, merge_lr=True, overwrite=True, 
-                         t_buffer=2., kernel_size=400, peak_threshold=0.1, rel_distance_peaks=0.3,
+                         t_buffer=2., kernel_size=1500, peak_threshold=0.1, rel_distance_peaks=0.3,
                          truncate=False, qlim=1, merge_subscans=True, turnarounds_in_subscan=False):
     """
     Compute turnaround flags for a dataset.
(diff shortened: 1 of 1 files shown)


msilvafe:
However you end up fixing this flagging it needs to recover the same turnaround flags as before for our standard turnarounds without requiring parameter tweaking by hand. I tend to agree with @skhrg that a simpler flagging than a matched filter may work here and it seems like @tskisner already has something implemented in toast which may work.

PR #1700, opened 2026-07-29
Add option to select on fraction of TOD flagged for glitches

mmccrackan:
Adds `max_t_frac` to `GlitchDetection` select which just cuts a detector if the fraction of samples flagged by glitches exceeds that value.  Uses the built in `has_ratio_cuts` function.  The original selection is still supported and can be run by itself or in conjunction with `max_t_frac`.

Tested on:

obs_id = 'obs_1783371748_lati1_111'
dets = {
    "wafer_slot": "ws0",
    "wafer.bandpass": "f090"
}

Confirmed number of cut dets (32 for `max_t_frac`=0.01) is equivalent when running the calculation manually by summing up flagged samples and dividing by total samps.

### DIFF
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -199,7 +199,7 @@ class GlitchDetection(_FracFlaggedMixIn, _Preprocess):
     Saves retsults in proc_aman under the "glitches" field.
 
     Data section should define either a glitch significant "sig_glitch" and a maximum
-    number of glitches "max_n_glitch" or a maximum fraction of the TOD samples that is
+    number of glitches "max_n_glitch" and/or a maximum fraction of the TOD samples that is
     allowed to be flagged by glitches.
 
     Example configuration block::
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -13,7 +13,8 @@ from sotodlib.hwp import hwp, hwp_angle_model
 import sotodlib.coords.planets as planets
 
 from sotodlib.core.flagman import (has_any_cuts, has_all_cut,
-                                   count_cuts, flag_cut_select,
+                                   count_cuts, has_ratio_cuts,
+                                   flag_cut_select,
                                    sparse_to_ranges_matrix)
 
 from sotodlib.preprocess import preprocess_util as pp_util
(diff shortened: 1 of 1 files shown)


sotodlib/preprocess/processes.py
```diff
@@ -197,8 +198,9 @@ class GlitchDetection(_FracFlaggedMixIn, _Preprocess):
 
     Saves retsults in proc_aman under the "glitches" field.
```
msilvafe, line 199: typo "retsults"

sotodlib/preprocess/processes.py
```diff
@@ -197,8 +198,9 @@ class GlitchDetection(_FracFlaggedMixIn, _Preprocess):
...
 
-    Data section should define a glitch significant "sig_glitch" and a maximum
-    number of glitches "max_n_glitch."
+    Data section should define either a glitch significant "sig_glitch" and a maximum
```
msilvafe, line 201: I think it should be "data selection" also since you spell out "max_n_glitch" and "sig_glitch" here also quote "max_t_frac" i.e. maximum fraction of the TOD samples that is allowed to be flagged by glitches "max_t_frac"

sotodlib/preprocess/processes.py
```diff
@@ -250,11 +253,18 @@ def select(self, meta, proc_aman=None, in_place=True):
...
-        n_cut = count_cuts(flag)
-        keep = n_cut <= self.select_cfgs["max_n_glitch"]
+        if "sig_glitch" in self.select_cfgs and "max_n_glitch" in self.select_cfgs:
+            flag = sparse_to_ranges_matrix(
+                proc_aman[self.glitch_name].glitch_detection > self.select_cfgs["sig_glitch"]
+            )
+            n_cut = count_cuts(flag)
+            keep = n_cut <= self.select_cfgs["max_n_glitch"]
+        else:
+            keep = np.ones(proc_aman.dets.count, dtype=bool)
```
msilvafe, line 263: This `if` statement will silently fall through to `keep = np.ones(...)` for any partial or misspelled config, with no error or log line. I don't think we want that.

sotodlib/preprocess/processes.py
```diff
@@ -250,11 +253,18 @@ def select(self, meta, proc_aman=None, in_place=True):
...
+
+        if "max_t_frac" in self.select_cfgs:
+            above_ratios = has_ratio_cuts(proc_aman[self.glitch_name].glitch_flags, self.select_cfgs["max_t_frac"])
+            keep = keep & ~above_ratios
```
msilvafe, line 267: I think this can be equivalently achieved with one less line as flag_cut_select already does the boolean inversion.

```
if "max_t_frac" in self.select_cfgs:
    keep = keep & flag_cut_select(
        proc_aman[self.glitch_name].glitch_flags,
        self.select_cfgs["max_t_frac"],
    )
```

But this is pretty minor, take it or leave it.

msilvafe (changes requested):
Some inline comments.

sotodlib/preprocess/processes.py
mmccrackan: I changed it to make all entries mandatory since it should be easy to pick values that will always be True.

msilvafe: But this breaks backwards compatibility right? By now requiring `max_t_frac`. I'm more hesitant now to do that than in the past as we have some many versions of preprocessing and so many tutorials and things out in the world now being used.

mmccrackan: Ah yeah, overlooked that.  I guess I can make `max_t_frac` default to 1.0 and keep the others as required.

msilvafe: Ya that sounds good, thanks.

msilvafe (approved):
Make change for backwards compatibility and feel free to merge.

PR #1701, opened 2026-07-29
Save det_cal nan flags

mmccrackan:
Saves det_cal `NaN` flags into the preprocess axis manager and optionally into `aman.flags` in the same way as `fp_flags`.  Some minor docstring updates to `fp_flags` too.

This is a breaking change for older preprocess configs.


sotodlib/tod_ops/flags.py
```diff
@@ -1021,6 +1025,51 @@ def get_focalplane_flags(aman, merge=True, overwrite=True, invalid_flags_name='f
...
+        [Ranges.ones_like(x) if Y else Ranges.zeros_like(x) for Y in flag_invalid_det_cal]
+    )
+
+    if merge:
+        if invalid_flags_name in aman.flags and not overwrite:
+            raise ValueError(f"Flag name {invalid_flags_name} already exists in aman.flags")
+        if invalid_flags_name in aman.flags:
+            aman.flags[invalid_flags_name] = msk_invalid_det_cal
+        else:
+            aman.flags.wrap(invalid_flags_name, msk_invalid_det_cal, [(0, 'dets'), (1, 'samps')])
```
msilvafe, line 1068: It feels like flags.wrap should take in the overwrite arg and handle this logic internally but won't make u change that for this.

sotodlib/preprocess/processes.py
```diff
@@ -2124,31 +2124,42 @@ class DetcalNanCuts(_Preprocess):
...
     """
     name = 'detcal_nan_cuts'
 
     def __init__(self, step_cfgs):
-        self.save_name = None
+        self.save_name = "det_cal_nan_flags"
 
         super().__init__(step_cfgs)
 
+    def calc_and_save(self, aman, proc_aman):
```
msilvafe, line 2139: `calc_and_save` has no `if self.calc_cfgs is None: return` but this class previously had NO calc phase at all — the old config schema was `select: {fields: [...]}` alone, with no `calc:` block required. So this is a breaking change right for all of our old configs. Now in order to do select you must add in the calc and save block. Can we do this in a way that doesn't break maybe adding in a `if calc_cfgs is None: Return` here and then keeping the select logic off of the proc_aman info only if it exists?

sotodlib/tod_ops/flags.py
```diff
@@ -1021,6 +1025,51 @@ def get_focalplane_flags(aman, merge=True, overwrite=True, invalid_flags_name='f
 
     return msk_invalid_fp
 
+
+def get_det_cal_nan_flags(aman, fields, merge=True, overwrite=True, invalid_flags_name="det_cal_nan_flags"):
```
msilvafe, line 1029: I think the defaults should probably be merge=False, overwrite=False why'd you choose this instead?

sotodlib/tod_ops/flags.py
```diff
@@ -1021,6 +1025,51 @@ def get_focalplane_flags(aman, merge=True, overwrite=True, invalid_flags_name='f
...
+    """
+
+    x = Ranges(aman.samps.count)
+    flag_invalid_det_cal = np.any(
+        [np.isnan(getattr(aman.det_cal, field)) for field in fields],
+        axis=0,
+    )
+
+    msk_invalid_det_cal = RangesMatrix(
+        [Ranges.ones_like(x) if Y else Ranges.zeros_like(x) for Y in flag_invalid_det_cal]
```
msilvafe, line 1059: This construction is used 6 (L98-99, L146-147, L623-624, L715-716, L761-762, L1016) other times throughout the module. We should consider adding a helper function for this.

msilvafe (approved):
Gonna approve but I have some inline comments most you can take or leave but want addressed or comment replies for 2 of the 4: 1) the defaults from merge/overwrite, and 2) backwards compatibility.

sotodlib/tod_ops/flags.py
mmccrackan: Mostly just because many of the others were like that, such as `get_focalplane_flags`, `get_subscans`, `get_source_flags`, etc, but I do see a number functions hav them both as `False`.  Should we use this PR to standardize them across the `tod_ops` flagging functions?

sotodlib/tod_ops/flags.py
mmccrackan: Yeah, I can do that in another PR.

sotodlib/tod_ops/flags.py
msilvafe: I think yes and we should standardize to False/False.

mmccrackan: I reverted these to what they were before this PR so as not to break backwards compatibility (`flag_turanrounds` would need `merge`, `overwrite`, and `merge_lr` specified in the configs as an example).

msilvafe (approved):

PR #1703, opened 2026-07-29
Add ENV variable to control atomic hdf update

mmccrackan:
Adds an environment variable to control whether or not the `H5ContextManager` will do an atomic update or not.  If not set it will just overwrite the original file as it did before.

Since the `update_det_cal` flow is on compute21 with just a few other flows like `update_det_match`, `update_smurf_caldbs`, and `update_obsdb` (preprocessing is on different compute nodes), we can try setting the ENV variable on there and see how it goes.  We can also just set it directly in `update_det_cal` if `run_method` is `site`, which calls a different sub function and should keep it scoped.


tskisner (approved):
Looks good, thanks for implementing!

PR #1704, opened 2026-07-30
Fix nseg of psd and reduce warnings

ykyohei:
FIx nseg of psd, when the samps is smaller than max_samples, the nseg was wrong.
We didn't have issue usually, because samps is longer than default max_samples ~ 1310 seconds.
Also reduced the warning of wn bias. Now it only warns when we use median method without nseg.

### DIFF
--- a/sotodlib/tod_ops/fft_ops.py
+++ b/sotodlib/tod_ops/fft_ops.py
@@ -356,7 +356,7 @@ def calc_psd(
         if kwargs["nperseg"] > max_samples:
             nseg = 1
         else:
-            nseg = int(max(n_samps, max_samples) / kwargs["nperseg"])
+            nseg = int((stop - start) / kwargs["nperseg"])
 
         freqs, Pxx = welch(signal[:, start:stop], fs, **kwargs)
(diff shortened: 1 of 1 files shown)


sotodlib/tod_ops/fft_ops.py
```diff
@@ -364,7 +356,7 @@ def calc_psd(
         if kwargs["nperseg"] > max_samples:
             nseg = 1
         else:
-            nseg = int(max_samples / kwargs["nperseg"])
+            nseg = int((stop - start) / kwargs["nperseg"])
```
msilvafe, line 359: Should we also have an elif case for `elif kwargs["nperseg"] > stop-start: nseg = 1`. If nperseg is > max_samples and nperseg > stop - start then nseg = 0 as currently written.

msilvafe (approved):

sotodlib/tod_ops/fft_ops.py
ykyohei: Ah thanks, I'll change these to following, then all the cases should be covered.
```
if kwargs["nperseg"] > stop - start:
    nseg = 1
else:
    nseg = int((stop - start) / kwargs["nperseg"])
```

PR #1705, opened 2026-08-03
Add script for export of specific observation IDs

tskisner:
When debugging code on a small number of observations, it can be useful to copy that data to a local system.  This adds a script to do that export.


mhasself (approved):
I am huge fan of this!

Can you add something to the docs?  Perhaps under Command Line Tools.  A stub is fine, just letting people know it exists.

I want to be really clear: I am not asking for a whole big write-up.  The comments in the script are already extensive.  But it would be good to get them onto readthedocs through an :automodule: or whatever.  Similarly, I find it helpful to have the argparse dumped into the docs; there's a plugin we use for that, that is quite painless.

sotodlib/scripts/so_data_export.py
```diff
@@ -0,0 +1,520 @@
...
+    """
+    if "obsdb" not in ctxt:
+        raise RuntimeError("Context has no obsdb key")
+    obsdb = ctxt["obsdb"]
+    if "obsfiledb" not in ctxt:
+        raise RuntimeError("Context has no obsfiledb key")
+    obsfiledb = ctxt["obsfiledb"]
+
+    rel_obsdb = os.path.relpath(obsdb, args.local_metadata)
+    tele_name, _ = os.path.split(rel_obsdb)
```
jsyleung, line 296: I think this breaks for LAT data. This returns `lat` instead of, e.g., `lati1` which is needed for correct string matching in `sync_obs_g3book()`. Result is that the obs is not found and so the **obs data files are not synced**.

jsyleung (changes requested):
This is very useful—thanks. I’ve even hacked it to sync the preprocessing archives as well. Might this be a feature you want to add? The problem is the directory structure, since you’re using relative paths, but I can assist with that.

There is a **critical bug** for LAT data that needs to be fixed though—see below.

sotodlib/scripts/so_data_export.py
jsyleung: Afaik `lati1` only appears in the `obs_id` so it might be easier to write a new function to parse that instead

tskisner: I just pushed a small change and tested that it works with `obs_1777601145_lati1_111` and context file `/global/cfs/cdirs/sobs/metadata/lat/contexts/preprocess_local.yaml`.  So I believe that is fixed.

tskisner:
> I am huge fan of this!
> 
> Can you add something to the docs? Perhaps under Command Line Tools. A stub is fine, just letting people know it exists.
> 
> I want to be really clear: I am not asking for a whole big write-up. The comments in the script are already extensive. But it would be good to get them onto readthedocs through an :automodule: or whatever. Similarly, I find it helpful to have the argparse dumped into the docs; there's a plugin we use for that, that is quite painless.

Documentation added.

jsyleung (approved):
Confirmed this works for LAT and SATs. You even added the `mset = set(mfiles)` line I forgot last time!

ISO preprocessing archives live in an upstream directory so will still need to be hacked in, but that’s a completely separate issue.

Thanks for this, Ted!

PR #1706, opened 2026-08-04
Adding unfied planet mapmaker using for MF-SATs

yoshinori-0778:
Adding general functions in coords.planets.py and functions related to planet mapmaking in mapmaking.planet_mapmaker.py.

Now is a very very draft, but I confirmed that a planet map was made. Will arrange these more appropriately.


sotodlib/mapmaking/planet_mapmaker.py
```diff
@@ -0,0 +1,1752 @@
...
+        band: frequency band, e.g., 'f090'
+        ws: wafer slot, e.g., 'ws0'
+        telescope: telescope name, e.g., 'satp3'
+        params: (amplitude in arcmin, phase in radian) of wobble
+    Returns:
+        deflection quaternions
+    """
+    # This is based on TOD fits summarized here: https://docs.google.com/presentation/d/1DPBeMX9ZPHwQyi9-s5orfRZIYcRLXaejFWOtlsflgUU/edit?usp=sharing
+    # amplitude below is in arcmin, phase in radian
+    amp_satp1 = {'f090': {'ws0': 0.3945,
```
kmharrington, line 811: these sorts of hard-codings I would move to a version-controlled configuration file.

yoshinori-0778: Thanks for the comment. I have used wobble metadata numbers as `params` here so I can use version-contorlled params. If you want to remove these hard core and set assert, I am happy to do that.

sotodlib/site_pipeline/make_planet_map.py
```diff
@@ -0,0 +1,138 @@
...
+
+    # Make runlist for each observation/wafer/band
+    runlist = []
+    for obs_id in obslist:
+        # find targeted wafers in scan
+        obs = context.obsdb.get(f'{obs_id}', tags=True)
+        tags = obs['tags']
+        subs = 'ws'
+        if configs['query'].get('all_wafers', False):
+            obs_wafers = [f'ws{i}' for i in range(7)]
```
mhasself: Perhaps use obs['wafer_slots_list'] for this.

sotodlib/site_pipeline/make_planet_map.py
```diff
@@ -0,0 +1,138 @@
...
+                obs_wafers = [f'ws{i}' for i in specific_wafers]
+            else:
+                obs_wafers = [i for i in tags if subs in i]
+        
+        bands = configs['query'].get('bands', bands)
+        for band in bands:
+            # add function for overwrite
+            for wafer in obs_wafers: 
+                if configs['overwrite']:
+                    irunlist = {'obs_id':obs_id, 'wafer_info': {'wafer_slot': wafer, 'wafer.bandpass': band}}
```
mhasself: Construct this data one time (instead of duplicating 3 times).

sotodlib/site_pipeline/make_planet_map.py
```diff
@@ -0,0 +1,138 @@
+# To make atmic planet map in detector-centered coordinate system
```
mhasself: Use a full submodule docstring `""" ... """` here, and explain in more detail what the script does.  It's also a good place to put in an example of a configuration file.  The docstring here is easy to have rendered in readthedocs, so it can be the "one place" this script is explained.

sotodlib/coords/planets.py
```diff
@@ -882,3 +882,190 @@ class MmTimer(coords.Timer):
             'P': P,
             'det_weights': det_weights,
             }
+
+def xieta0(q):
```
mhasself, line 886: Ah, xieta decomposition won't work (as you say in docstring), but I think this would?:
```
 phi, theta, gamma = so3g.proj.quat.decompose_iso(q)
 return so3g.proj.rotation_iso(phi, theta, 0.)
```

mhasself (changes requested):
Thanks for all this work! As requested, here is some general guidance (but not an exhaustive review yet):

- Please include an example config file as part of documentation.
- Format code and docs to less than 120 columns.
- There are also some "whitespace" violations -- might be worth just running a source formatter on the new files you're contributing (but don't mass-reformat the existing files like planets.py!)
- Consider making the deflection/wobble code a little more general, and moving into coords module somewhere.
- You have functions for getting and archiving and processing PWV.  Are those needed, if instead we can provide PWV through obsdb? I think it's good to minimize the number of different ways we pull data from APEX / Toco radiometer.
- Use "single" instead of "each" for functions that process one of something (e.g. `get_each_instrument_P`).  "Each" implies that it is looping, internally, over a bunch of things.
- Remove default paths from planet_mapmaker.py funcs

Much of the new code in planets.py is basically dealing with the fact that SlowSource is not really good enough, at this point.  It would be nice to take your functions `calc_planet_*` and instead create a replacement for SlowSource, e.g. AzElSource, that provides an az, el interpolator for any kind of source with sufficient precision for beams work.  (Basically I want to get rid of SlowSource in favor of something that works better, without a huge performance penalty for simple cases.)

mhasself:
> Thanks for all this work! As requested, here is some general guidance (but not an exhaustive review yet):
>  ...

One more thing: utility functions (such as the ones you've added in `coords.planets`) should have unit tests.  (It's not necessary to duplicate all the calculations; rather the tests just make sure there aren't typos and that different ways of asking for the same thing will yield the same answer.)

sotodlib/mapmaking/planet_mapmaker.py
yoshinori-0778: Done, I remove hard-coding part and move this into coords.helper as Matthew commented.

yoshinori-0778:
THanks for the comments @mhasself 

>Please include an example config file as part of documentation.
Could you more clarify where I should inculde? Should I add example config files in this mapmaking directory?

>You have functions for getting and archiving and processing PWV. Are those needed, if instead we can provide PWV through obsdb? I think it's good to minimize the number of different ways we pull data from APEX / Toco radiometer
If there is PWV database, we do not need this. Is that available somewhere? I am happy to incorporate it.

sotodlib/site_pipeline/make_planet_map.py
yoshinori-0778: Sure will do that for all functions. Ans sorry I just found this is a good place to put in an example of a configuration file. I will make sub drectory and put int the example in it.

sotodlib/site_pipeline/make_planet_map.py
yoshinori-0778: Sorry what do you mean? I think there are no dupulicate runlist.. Did you mean I should make this cleaner, like write in one line instead using three `for` iteration?

yoshinori-0778:
Future task for me: run all combination of detector/boresignt/planet center and wafer/detector based deflection correction.

sotodlib/coords/planets.py
yoshinori-0778: Here I would like to set gamma angle to be zero in xi-eta coordinate system, not psi (=gamma-phi in eq12 of [this](https://simons1.princeton.edu/docs/tod2maps_docs/coord_sys/coord_sys_LATEST.pdf)) angle in spherical coordinate system. So I would like to keep this.

kmharrington:
> > You have functions for getting and archiving and processing PWV. Are those needed, if instead we can provide PWV through obsdb? I think it's good to minimize the number of different ways we pull data from APEX / Toco radiometer
> > If there is PWV database, we do not need this. Is that available somewhere? I am happy to incorporate it.

@mhasself  A note on this. I discovered recently that the extended obsdbs running in your user account do not include calibration observations. It would be good to have those expanded and to make them official. Otherwise we'll keep having parallel implementations pop up.

PR #1707, opened 2026-08-05
Wire grid: add circular mean and fix axis bug

### LINKED ISSUE #1699: Wire grid: Issue in averaging angle
### Averaging angle
Current code uses naive mean instead of circular mean. It will fail when the angle is near the edge of the period.
https://github.com/simonsobs/sotodlib/blob/b876f9e10c8221e4c2f1da44b154ce73c98de8d8/sotodlib/site_pipeline/calibration/wiregrid.py#L751-L753

See also
https://github.com/simonsobs/pwg-scripts/issues/92

### Wrong axis
And this is another bug:
https://github.com/simonsobs/sotodlib/blob/b876f9e10c8221e4c2f1da44b154ce73c98de8d8/sotodlib/site_pipeline/calibration/wiregrid.py#L744
Here, `_i` should be `:`.

satoru99:
Fix for #1699

### DIFF
--- a/sotodlib/site_pipeline/calibration/wiregrid.py
+++ b/sotodlib/site_pipeline/calibration/wiregrid.py
@@ -15,6 +15,36 @@ from sotodlib.io import hk_utils
 from sotodlib.site_pipeline.utils.logging import init_logger
 logger = init_logger('wiregrid', 'wiregrid: ')
 
+def circular_mean(angle, error=1.0, axis=None, period=np.pi * 2):
+    """
+    Circular mean of angle
+
+    Parameters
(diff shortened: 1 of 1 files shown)


mhasself (approved):
Domain experts should approve but this looks good to me! @ykyohei

ykyohei (approved):
Looks good to me as well, thanks for fixing it!
It's fine to mix two sotodlib versions for the currently running wiregrid site-pipeline metadata. We just need to consistenly reanalyze everything it in next version.
Please make sure to log this fix in metadata logging page. @YutaroKASAI

PR #1708, opened 2026-08-05
Change fitting method of stimulator data

YudaiSeino:
I changed fitting option of lmfit for time-constant to default because previous method didn't work well for slow time-constant detector (LF).

### DIFF
--- a/sotodlib/stimulator/stimulator.py
+++ b/sotodlib/stimulator/stimulator.py
@@ -171,7 +171,7 @@ def calc_gain(aman):
     aman.stm_cal.wrap("stm_gain", arr, [(0, "dets")], overwrite=True)
 
 
-def calc_timeconstant(aman, idxs=None):
+def calc_timeconstant(aman, idxs=None, method="least_squares", fit_kws=None):
     """
     Calculate the time constant of the detectors and the readout delay.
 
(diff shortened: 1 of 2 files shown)


ykyohei (commented):
Interesting that this fixes the issue for slow detectors. Did you confirm that this didn't affect faster MF UHF detectors?

Okay to merge but I think it would be nice to make fitting parameters configurable.

YudaiSeino:
Thank you for the comment! I made it configurable. Actually, this change also improved fitting result for MF/UHF.
I also fixed a small bug for phase calculation (t0s). And I fixed plotting format a bit.

sotodlib/stimulator/stimulator.py
```diff
@@ -171,7 +171,7 @@ def calc_gain(aman):
     aman.stm_cal.wrap("stm_gain", arr, [(0, "dets")], overwrite=True)
 
 
-def calc_timeconstant(aman, idxs=None):
+def calc_timeconstant(aman, idxs=None, method="least_squares", fit_kws=None):
```
ykyohei: We would like to make "leastsq" as default method because you found it performs better?

ykyohei (commented):
Thanks for the changes! I have one comment.

sotodlib/stimulator/stimulator.py
YudaiSeino: Opps, I'm stupid, it's not correct:(

ykyohei (approved):
Thanks, please squash when merge.

PR #1709, opened 2026-08-06
Add joint demodulated Q/U Fourier Nmat filter

susannaaz:
Adds tod_ops.nmat_filter.apply_joint_qu_nmat_filter, which whitens and jointly analyzes the demodulated Q and U detector streams. Modes inconsistent with independent noise, per a Marchenko-Pastur plus Tracy-Widom threshold, are inverse-covariance weighted in Fourier space.

Included as the `JointQUNmatFilter` preprocess step, which supports `use_data_aman` so the noise operator can be estimated from a real-data snapshot rather than from a signal-only simulation for transfer function purposes.


sotodlib/tod_ops/nmat_filter.py
```diff
@@ -0,0 +1,523 @@
...
+            raise ValueError("model_tod and tod detector axes must match in order")
+        if not np.array_equal(model_tod.timestamps, tod.timestamps):
+            raise ValueError("model_tod and tod timestamps must match")
+    if tod.samps.count < 4:
+        raise ValueError("At least four samples are required")
+    if "timestamps" not in tod._fields:
+        raise ValueError("timestamps are required to define Fourier frequencies")
+    dts = np.diff(tod.timestamps)
+    if not np.all(np.isfinite(dts)) or not np.all(dts > 0):
+        raise ValueError("timestamps must be finite and strictly increasing")
```
ykyohei, line 271: I think these checks are not really necessary.

sotodlib/tod_ops/nmat_filter.py
```diff
@@ -0,0 +1,523 @@
...
+    With ``operator="nmat"``, the selected modes are combined and their
+    frequency-dependent power E(f), together with the residual detector-
+    diagonal power D(f), is measured in approximately logarithmic bins. The
+    model ``N(f) = D(f) + V E(f) V.T`` is inverted with the Woodbury identity.
+    Working in white-noise-normalized coordinates gives a dimensionless
+    operator with approximately unit response to white noise. This matches the
+    detector-vector noise operator used by the maximum-likelihood mapmaker,
+    normalized for preprocessing.
+
+    The legacy ``operator="mode_shrink"`` applies only correlated-mode
```
ykyohei, line 171: Remove "The legacy"?

sotodlib/tod_ops/nmat_filter.py
```diff
@@ -0,0 +1,523 @@
...
+        arr = tod[signal]
+    elif isinstance(signal, np.ndarray):
+        arr = signal
+    else:
+        raise TypeError(f"{label} must be a field name or ndarray")
+    expected = (tod.dets.count, tod.samps.count)
+    if arr.shape != expected:
+        raise ValueError(f"{label} has shape {arr.shape}; expected {expected}")
+    if np.iscomplexobj(arr):
+        raise ValueError(f"{label} must be real-valued")
```
ykyohei, line 126: it seems fine to drop these

ykyohei (commented):
I haven't read everything yet, but I have one suggestion. 

I believe we can save the relatively small set of statistics required to reproduce the NMat filter and use them for the pre-processing run using the archive. This is similar to how we apply the 1/f counter filter using the saved values (wn, fknee, alpha). Then, we do not need to handle model_aman for this.

sotodlib/tod_ops/nmat_filter.py
```diff
@@ -0,0 +1,523 @@
...
+        evals, evecs = np.linalg.eigh(cov)
+        order = np.argsort(evals)[::-1]
+        evals = evals[order]
+        evecs = evecs[:, order]
+
+        keep = np.ones(evals.size, dtype=bool)
+        if singleness_max is not None:
+            keep &= np.max(np.abs(evecs), axis=0) < singleness_max
+
+        gamma = valid_idx.size / n_samples
```
msilvafe, line 390: Why is `cov` normalized by `nb` and `gamma` normalized by `n_samples` (different by a factor of 2).

sotodlib/tod_ops/nmat_filter.py
```diff
@@ -0,0 +1,523 @@
...
+            filtered *= noise_sigma[valid, None]
+            ft[valid_idx, i0:i1] = filtered
+
+    if in_place:
+        q_out, u_out = q, u
+    else:
+        q_out = np.empty_like(q)
+        u_out = np.empty_like(u)
+    fft.irfft(ft[:ndet], q_out, normalize="phys")
+    fft.irfft(ft[ndet:], u_out, normalize="phys")
```
msilvafe, line 505: From what I can tell pixell fft.irfft normalize argument should be a boolean.

sotodlib/tod_ops/nmat_filter.py
```diff
@@ -0,0 +1,523 @@
...
+        if mode_vecs:
+            vecs = np.concatenate(mode_vecs, axis=1)
+            pinv_vecs = np.linalg.pinv(vecs)
+        else:
+            vecs = np.empty((valid_idx.size, 0), dtype=float)
+            pinv_vecs = np.empty((0, valid_idx.size), dtype=float)
+
+        for i0, i1 in profile_bins:
+            model_xbin = np.array(model_ft[valid_idx, i0:i1], copy=True)
+            model_xbin *= inv_sigma[:, None]
```
msilvafe, line 442: `model_xbin` is computed here and in the loop above at L378 since `bins` and `profile_bins` both cover `[fmin, fmax]`, that's redundant. Separately, `model_xbin` and `target_xbin` here are copies of the same array when `model_tod` is omitted (the default). A little more efficient would be to compute the normalized slice(s) up front and pull out the slices you need from one common copy in memory in each loop, and skip the second copy entirely when `model_ft is ft`. This is pretty minor though take it or leave it.

sotodlib/tod_ops/nmat_filter.py
```diff
@@ -0,0 +1,523 @@
...
+    return_diagnostics : bool
+        Return mode and frequency-profile diagnostics.
+    store_mode_diagnostics : bool
+        Also retain per-bin eigenvalues/eigenvectors in diagnostics.
+
+    Returns
+    -------
+    (filtered_Q, filtered_U), diagnostics_or_None
+    """
+    if mode_selection not in ("mp_tw", "fixed"):
```
msilvafe, line 231: If you're going to check all of the arguments like this you might consider a package that does this more natively. One implementation of this already in use in sotodlib is the `dataclass` (i.e. coords/det_match_solutions.py, io/hkdb.py, site_pipeline/update_bgmap_delay.py, site_pipeline/update_det_match.py, coords/fp_containers.py, and qa/quality_reports/plots.py) I think it would look something like this:

```
from dataclasses import dataclass
from typing import Optional, Tuple, Union

@dataclass
class NmatFilterConfig:
    fmin: float = 0.0015
    fmax: float = 0.2
    noise_band: Tuple[float, float] = (0.5, 1.75)
    bin_width_hz: Optional[float] = 0.2
    mode_selection: str = "mp_tw"
    mp_significance: Union[float, str] = 0.999
    n_modes: int = 1
    n_modes_max: Optional[int] = None
    singleness_max: Optional[float] = None
    operator: str = "nmat"
    profile_n_bins: int = 40
    profile_min_nfreq: int = 20
    profile_diagonal_floor: float = 0.05
    psd_scale: float = 1.0
    psd_min_weight: float = 0.0

    def __post_init__(self):
        if self.mode_selection not in ("mp_tw", "fixed"):
            raise ValueError("mode_selection must be 'mp_tw' or 'fixed'")
        if self.operator not in ("nmat", "mode_shrink"):
            raise ValueError("operator must be 'nmat' or 'mode_shrink'")
        if not np.isfinite(self.psd_scale) or not 0 <= self.psd_scale <= 1:
            raise ValueError("psd_scale must be finite and in [0, 1]")
        if not 0 <= self.psd_min_weight <= 1:
            raise ValueError("psd_min_weight must be in [0, 1]")
        if self.n_modes_max is not None and self.n_modes_max < 0:
            raise ValueError("n_modes_max must be non-negative")
        if self.n_modes < 0:
            raise ValueError("n_modes must be non-negative")
        if self.singleness_max is not None and not 0 < self.singleness_max <= 1:
            raise ValueError("singleness_max must be in (0, 1]")
        if not np.isfinite(self.profile_diagonal_floor) or self.profile_diagonal_floor <= 0:
            raise ValueError("profile_diagonal_floor must be positive")
        if self.mp_significance not in _TW_BETA1_QUANTILES:
            raise ValueError(f"Unsupported mp_significance {self.mp_significance!r}")
        if self.profile_n_bins < 1:
            raise ValueError("profile_n_bins must be positive")
        if self.profile_min_nfreq < 1:
            raise ValueError("profile_min_nfreq must be positive")
```

Then in place of lines 231-248 you'd instantiate that class
```
   NmatFilterConfig(
        fmin=fmin,
        fmax=fmax,
        noise_band=noise_band,
        bin_width_hz=bin_width_hz,
        mode_selection=mode_selection,
        mp_significance=mp_significance,
        n_modes=n_modes,
        n_modes_max=n_modes_max,
        singleness_max=singleness_max,
        operator=operator,
        profile_n_bins=profile_n_bins,
        profile_min_nfreq=profile_min_nfreq,
        profile_diagonal_floor=profile_diagonal_floor,
        psd_scale=psd_scale,
        psd_min_weight=psd_min_weight,
    )
```

But I'd agree that it's worth thinking through the logic of which checks are really necessary.

sotodlib/tod_ops/nmat_filter.py
```diff
@@ -0,0 +1,523 @@
...
+        model_xbin *= inv_sigma[:, None]
+        cov = np.real(model_xbin.dot(np.conj(model_xbin.T))) / nb
+        evals, evecs = np.linalg.eigh(cov)
+        order = np.argsort(evals)[::-1]
+        evals = evals[order]
+        evecs = evecs[:, order]
+
+        keep = np.ones(evals.size, dtype=bool)
+        if singleness_max is not None:
+            keep &= np.max(np.abs(evecs), axis=0) < singleness_max
```
msilvafe, line 388: I could imagine this singleness metric is actually a pretty useful thing to diagnose detector behavior and possibly use for detector level cuts.

sotodlib/tod_ops/nmat_filter.py
```diff
@@ -0,0 +1,523 @@
...
+    while i0 < stop:
+        upper_hz = min(fmax, freqs[i0] + bin_width_hz)
+        i1 = min(stop, int(np.searchsorted(freqs, upper_hz, side="left")))
+        i1 = max(i0 + 1, i1)
+        bins.append((i0, i1))
+        i0 = i1
+    return bins
+
+
+def _build_profile_bins(freqs, fmin, fmax, n_bins=40, min_nfreq=20):
```
msilvafe, line 89: `_build_profile_bins` reimplements binning/merge logic that already exists twice elsewhere: `sotodlib/mapmaking/utils.py`'s `makebins` and `sotodlib/tod_ops/fft_ops.py`'s `log_binning`. Can you check if either of these will do the job for you?

sotodlib/tod_ops/nmat_filter.py
```diff
@@ -0,0 +1,523 @@
...
+        raise TypeError(f"{label} must be a field name or ndarray")
+    expected = (tod.dets.count, tod.samps.count)
+    if arr.shape != expected:
+        raise ValueError(f"{label} has shape {arr.shape}; expected {expected}")
+    if np.iscomplexobj(arr):
+        raise ValueError(f"{label} must be real-valued")
+    return arr
+
+
+def apply_joint_qu_nmat_filter(
```
msilvafe, line 130: Consider splitting this into a fit step (everything derived from model_tod: whitening, mode detection, $$D(f)$$, $$E(f)$$) and an apply step (everything derived from tod). Keeping around the mode_shrink option instead of just nmat might make this more complicated to generalize like that so could drop that for now if you want or spend some extra time to propagate both through the split/save. This allows us to save relevant quantities computed from the data and not recompute them when running on the sims.

Ideally then you could do a construction like:

```
def apply_joint_qu_nmat_filter(tod, model_tod, signal_Q,...):
    pkg = fit_joint_qu_nmat_operator(model_tod, signal_Q, signal_U, fmin=fmin, ...)
    return apply_nmat_operator(tod, pkg, signal_Q, signal_U, psd_scale=psd_scale, ...)
```

sotodlib/preprocess/processes.py
```diff
@@ -2330,6 +2330,65 @@ def process(self, aman, proc_aman, sim=False, data_aman=None):
         _ = tod_ops.pca.add_model(aman, model, signal=signal, scale=-1)
         return aman, proc_aman
 
+
+class JointQUNmatFilter(_Preprocess):
```
msilvafe, line 2334: With the suggested split in nmat to a build and apply then you can add a calc_and_save/save, so we can save info in proc_aman some attributes maybe worth storing: `valid_channels` (axis),  `modes` (axis), `profile_bins` (axis), `noise_sigma`, `vecs`, `profile_bin_edges_hz`, `residual_power`, `profile_diagonal_floor`,  `mode_power`

And I think it's worth tracking why a bin ended up with however many modes it did — was it MP,  singleness, or n_modes_max limited? So maybe in nmat_filter you can also compute and return stats like: 

```
above_mp_edge = int(np.sum(evals > lambda_plus))        # MP lambda_plus doesn't exist you'd have to have the lower level get MP threshold function return this to use. 
above_threshold = int(np.sum(evals > threshold))         # Threshold
failed_singleness = int(np.sum((evals > threshold) & ~keep))  # singleness
n_before_cap = selected.size

if n_modes_max is not None:
    selected = selected[:int(n_modes_max)]
n_capped_by_max = n_before_cap - selected.size # Capped by max modes
```
Could do actual boolean arrays to store instead of ints too.

sotodlib/tod_ops/nmat_filter.py
```diff
@@ -0,0 +1,523 @@
+"""Joint demodulated Q/U Fourier Nmat filtering."""
```
msilvafe, line 1: It'd be great if this module had a helper function which would reconstruct the pre/post filtered TOD or what was actually subtracted from the TOD. So that we can overplay the filtered modes with the data and compare this filter to others we've used in TOD space.

msilvafe (changes requested):
This is an awesome addition @susannaaz thanks for implementing this so quickly and the initial results are super promising. I have a few comments but we should talk over this review to decide what you want to implement now vs later (or at all) as I know this is something you want to get to running on more data ASAP.

The main structural thing I want to flag is as written, apply_joint_qu_nmat_filter always derives the operator (whitening, covariance, mode detection, $$D(f)$$, $$E(f)$$) and applies it in the same call, and nothing about the fitted operator is ever saved. I would like there to be a way to fit the operator on a real observation and reapply that same fit to a signal-only sim without redoing the entire real-data analysis.

I've broken out 4 inline comments walking through one way to address this: 1) splitting `apply_joint_qu_nmat_filter` into a fit step (everything derived from real data) and an apply step (everything derived from the target TOD), 2) adding `calc_and_save`/`save` so a sim run can reload rather than refit it, 3) a small utility for reconstructing what's actually being subtracted from a given detector for debugging, and 4) some stats on how many modes get selected vs. rejected and why (MP threshold, singleness veto, or the `n_modes_max` cap).

There are a few more inline comments which I think I'd like to get a response on but take-it-or-leave-it in terms of actually implementing anything in code.

sotodlib/tod_ops/nmat_filter.py
susannaaz: Agreed, and this fits better with how the rest of the pipeline works. I've split it into a fit step and an apply step: `joint_qu_nmat_model` computes the operator and saves it to `proc_aman`, and `joint_qu_nmat_filter` reads it back and applies it, i.e. the same pattern as `noise` -> `fourier_filter` with the (wn, fknee, alpha)
fit. Reloading from the archive is now the default and no `model_aman` is
involved.

What gets stored, on the `dets` / `nmat_modes` / `nmat_profile` / `nmat_bins` axes so it follows detector restriction like any other product: `sigma_Q`, `sigma_U`, `valid_Q`, `valid_U`, `vecs_Q`, `vecs_U`, `residual_power_Q/U` (D(f)), `mode_power` (E(f)), the profile bin edges in Hz, and the per-bin selection stats.

sotodlib/tod_ops/nmat_filter.py
susannaaz: Dropped `mode_shrink` altogether, so that paragraph is gone

sotodlib/tod_ops/nmat_filter.py
susannaaz: Removed - `_resolve_signal` is a one-liner now.

susannaaz:
> This is an awesome addition @susannaaz thanks for implementing this so quickly and the initial results are super promising. I have a few comments but we should talk over this review to decide what you want to implement now vs later (or at all) as I know this is something you want to get to running on more data ASAP.
> 
> The main structural thing I want to flag is as written, apply_joint_qu_nmat_filter always derives the operator (whitening, covariance, mode detection, D ( f ) , E ( f ) ) and applies it in the same call, and nothing about the fitted operator is ever saved. I would like there to be a way to fit the operator on a real observation and reapply that same fit to a signal-only sim without redoing the entire real-data analysis.
> 
> I've broken out 4 inline comments walking through one way to address this: 1) splitting `apply_joint_qu_nmat_filter` into a fit step (everything derived from real data) and an apply step (everything derived from the target TOD), 2) adding `calc_and_save`/`save` so a sim run can reload rather than refit it, 3) a small utility for reconstructing what's actually being subtracted from a given detector for debugging, and 4) some stats on how many modes get selected vs. rejected and why (MP threshold, singleness veto, or the `n_modes_max` cap).
> 
> There are a few more inline comments which I think I'd like to get a response on but take-it-or-leave-it in terms of actually implementing anything in code.

Thanks @msilvafe , I've implemented all four structural points, and the rest is answered inline. Summary:

1. Split into `fit_joint_qu_nmat_operator` (everything from the model TOD) and `apply_joint_qu_nmat_operator` (everything from the target). `apply_joint_qu_nmat_filter` is now a thin wrapper over the two.
2. New `joint_qu_nmat_model` preprocess step with `calc`/`save` writes the operator into `proc_aman`; `joint_qu_nmat_filter` reloads it, so a sim run never refits. A live-fit path is still reachable via `use_data_aman: True`, but only so I can reproduce pre-split runs.
3. Added `get_nmat_subtraction`, which returns what the filter removed in TOD space for overplotting against the input.
4. Per-bin selection stats are stored.

I also dropped `mode_shrink` as you suggested.

sotodlib/tod_ops/nmat_filter.py
susannaaz: That factor of 2 is deliberate, and the two have to differ. A complex Fourier sample carries two real degrees of freedom, so `gamma` uses `n_samples = 2 nb`. But the whitening normalises `mean|x|^2 = 1` per *complex*
sample, and `|x|^2` already sums the real and imaginary parts. Dividing `Re(x x^H)` by `nb` is therefore what puts the diagonal of `cov` at 1, which is the scale the MP edge `(1 + sqrt(gamma))^2` is defined on.

Quick check with p=60, nb=200 white channels: `Re(x x^H)/nb` gives `mean(diag) = 1.0000` and a max eigenvalue of 1.850 against an MP edge of 1.925. Normalising by `2 nb` instead would put the diagonal at 0.5 and the eigenvalues nowhere near the edge.

I've added a comment at that line to make this more obvious!

sotodlib/tod_ops/nmat_filter.py
susannaaz: You're right, changed to `normalize=True`. Note that, it would still behave correctly if `"phys"` is a non-empty string, so it was still "true" and nothing already run is affected.

sotodlib/tod_ops/nmat_filter.py
susannaaz: Done. The band is whitened once up front and both loops take slices of that one array, so there's no second copy and no repeated normalisation. With the fit/apply split the target is only copied in the apply step, where it genuinely is a different array.

sotodlib/tod_ops/nmat_filter.py
susannaaz: I went with your last suggestion and cut the checks back instead. After the split most of them belonged to arguments that no longer travel together, and the ones that were really just restating the signature are gone. What's left is the handful where a bad value fails somewhere unhelpful much later. (But happy to add `NmatFilterConfig` if you'd prefer it!)

sotodlib/tod_ops/nmat_filter.py
susannaaz: Agreed! It's now kept rather than thrown away: `mode_singleness` stores the value per selected mode, and `n_failed_singleness` per bin counts modes that passed the MP threshold but were vetoed by it. Both are in the saved operator, so they're available per observation without rerunning.

Using it for detector-level cuts seems worth looking at separately (a mode with singleness near 1 is essentially one detector, so it should be a fairly direct pointer at a misbehaving channel)

sotodlib/tod_ops/nmat_filter.py
susannaaz: Checked both:

`log_binning` doesn't fit: it averages PSD values into log bins, whereas I need the index ranges themselves.

`makebins` does - same merge logic, down to deleting the second-to-last edge when the last bin is short, but as is two things bite with the defaults:

1. `freq2ind` rounds with `ceil`, which pushes the lowest edge up by one Fourier mode. at the bottom of the band that single mode carries a large share of the 1/f power and can change the in-band power
2. when the whole range is shorter than `nmin`, the trailing-edge deletion leaves no bins at all.

With `rfun=np.floor`, the endpoints clamped to the requested range, and a guard for the short-range case, it reproduces the intended binning, so I've switched to it and dropped the local merge loop. In-band power now agrees with the previous implementation.

One other caveat: `sotodlib.mapmaking` imports `tod_ops`, so the import has to be deferred inside the function to avoid a cycle. That works but it's the wrong way round dependency-wise. If you'd rather not have `tod_ops` reach into `mapmaking`, the alternative is to move `makebins` somewhere neutral. W can do that instead, just didn't want to touch mapmaking in this PR but let me know

sotodlib/tod_ops/nmat_filter.py
susannaaz: Done, essentially the construction you wrote. Now `fit_joint_qu_nmat_operator` returns an AxisManager holding the whitening sigma, the mode vectors, D(f), E(f), the profile bin edges and the selection stats. `apply_joint_qu_nmat_operator` takes that plus a target TOD. It raises if the operator's detector axis doesn't match the target's, and warns if the operator was fit over a different number of detectors than it's being applied to.

I dropped `mode_shrink` as yousuggested, it wasn't being used.

Added a test that the two-step form reproduces the single call exactly, and one that fitting on one observation and applying to another leaves the model TOD untouched.

sotodlib/preprocess/processes.py
susannaaz: Both done.

`joint_qu_nmat_model` has `calc`/`save` and writes the operator into `proc_aman` under a configurable `wrap_name` (default `nmat_qu`). Stored: `sigma_Q/U`, `valid_Q/U`, `vecs_Q/U`, `residual_power_Q/U`, `mode_power`, `mode_singleness`, `profile_lo_hz`, `profile_hi_hz`, `profile_diagonal_floor`, `fmin`, `fmax`, `nsamp`, `n_dets_fit`. Per-detector quantities are on the `dets` axis so restriction propagates.

For the selection I've stored, per mode bin: `n_above_lambda_plus`, `n_above_threshold`, `n_failed_singleness`, `n_capped_by_max`, `n_selected`, plus `threshold`, `lambda_plus`, `gamma` and the bin edges in Hz. So you can see whether a bin was limited by the bulk edge, the Tracy-Widom term, the singleness veto or the cap. `get_marchenko_pastur_threshold` now takes `return_edge=True` to give `lambda_plus` alongside the threshold.

I kept these as counts rather than boolean arrays, the arrays would be n_channels long per bin and I don't have a use for the per-mode detail yet, but I think should be easy to change if you want it.

sotodlib/tod_ops/nmat_filter.py
susannaaz: Added `get_nmat_subtraction(tod, operator, ...)`, which returns `(Q - F Q, U - F U)` in time domain, i.e. exactly what the filter took out, so it can be overplotted on the input or compared against other filters in TOD space. It shares the apply path, so it can't drift from what the filter actually does, and there's a test asserting it equals input minus filtered.

sotodlib/tod_ops/nmat_filter.py
chervias: Hi @susannaaz I think you might have already figured out how this works for the ML mapmaker. So the bins in the ACT noise model are normally defined manually, and at the beginning, the first bin will start on freq = 0Hz, always by default. The thing about the last bin I don't remember where that came from, but we can ask Sigurd.

sotodlib/tod_ops/nmat_filter.py
```diff
@@ -80,444 +97,517 @@ def _build_freq_bins(freqs, fmin=None, fmax=None, bin_width_hz=None):
...
+def _resolve_signal(tod, signal):
+    return tod[signal] if isinstance(signal, str) else signal
+
+
+def _stack_rfft(q, u):
+    """rFFT of the Q/U stack, shape (2 ndet, nfreq)."""
+    ndet, nsamp = q.shape
+    nfreq = nsamp // 2 + 1
+    real_dtype = np.result_type(q.dtype, u.dtype)
+    complex_dtype = np.complex64 if real_dtype.itemsize <= 4 else np.complex128
```
msilvafe: Why have these different complex datatypes?

sotodlib/preprocess/processes.py
```diff
@@ -2361,6 +2330,141 @@ def process(self, aman, proc_aman, sim=False, data_aman=None):
...
+        self.signal_Q = step_cfgs.get("signal_Q", "demodQ")
+        self.signal_U = step_cfgs.get("signal_U", "demodU")
+        self.save_name = None
+        super().__init__(step_cfgs)
+
+    def process(self, aman, proc_aman, sim=False, data_aman=None):
+        cfgs = dict(self.process_cfgs)
+        model_name = cfgs.pop("nmat_model", "nmat_qu")
+        fit_cfgs = cfgs.pop("fit", None)
+
```
msilvafe, line 2618: self.process_cfgs should already be a dict, also `in_place` should also be popped as you hardcode that in L2462

sotodlib/tod_ops/nmat_filter.py
```diff
@@ -0,0 +1,627 @@
...
+            np.array([freqs[i0] for i0, _ in profile_bins]), [(0, "nmat_profile")])
+    op.wrap("profile_hi_hz",
+            np.array([freqs[i1 - 1] for _, i1 in profile_bins]), [(0, "nmat_profile")])
+    for key, val in stats.items():
+        op.wrap(key, np.asarray(val), [(0, "nmat_bins")])
+
+    op.wrap("fmin", float(freqs[band_lo]))
+    op.wrap("fmax", float(freqs[band_hi - 1]))
+    op.wrap("profile_diagonal_floor", float(profile_diagonal_floor))
+    op.wrap("nsamp", int(nsamp))
```
msilvafe, line 402: fmin, fmax, nsamp wrapped but not used again when op is loaded by other functions. Use them in the loaded functions instead of recomputing or what was your thought on their intended purpose?

sotodlib/preprocess/processes.py
```diff
@@ -2361,6 +2330,141 @@ def process(self, aman, proc_aman, sim=False, data_aman=None):
...
+            logger.info(
+                "Fitting the Nmat operator from the supplied AxisManager "
+                "(%d detectors)", model_aman.dets.count
+            )
+            operator = tod_ops.nmat_filter.fit_joint_qu_nmat_operator(
+                model_aman,
+                signal_Q=self.signal_Q,
+                signal_U=self.signal_U,
+                **(fit_cfgs or {}),
+            )
```
msilvafe, line 2638: We should probably remove this block now, or when do you think we should go down this branch?

sotodlib/tod_ops/nmat_filter.py
```diff
@@ -0,0 +1,627 @@
...
+
+    bins = _build_freq_bins(freqs, fmin=fmin, fmax=fmax, bin_width_hz=bin_width_hz)
+    profile_bins = _build_profile_bins(
+        freqs, fmin=fmin, fmax=fmax,
+        n_bins=profile_n_bins, min_nfreq=profile_min_nfreq,
+    )
+
+    n0, n1 = noise_band
+    if n0 < 0 or n1 <= n0:
+        raise ValueError("noise_band must satisfy 0 <= low < high")
```
msilvafe: This should be checked against against the demodulation lowpass filter and not use any frequencies above the lowpass. That was the intention of adding `proc_aman["frequency_cutoffs"]`. But the appropriate place for this is probably in processes.py though so you can use info in proc_aman. There's an existing example in `Noise.calc_and_save` in `processes.py` Lines 745-755. Here you can at least check against the nyquist frequency `i.e. n1 must be < freqs[-1]`

sotodlib/tod_ops/nmat_filter.py
```diff
@@ -0,0 +1,627 @@
...
+        core.IndexAxis("nmat_modes", nmodes),
+        core.IndexAxis("nmat_profile", nprof),
+        core.IndexAxis("nmat_bins", len(bins)),
+    )
+
+    def _split(arr_over_channels, fill=0.0):
+        """Scatter a valid-channel array back to (dets,) Q and U halves."""
+        full = np.full(2 * ndet, fill, dtype=float)
+        full[valid_idx] = arr_over_channels
+        return full[:ndet], full[ndet:]
```
msilvafe: This function seems to never be used.

msilvafe (changes requested):
Ok one more round of inline comments to reply to. Much simpler this time hit-rereview for me when ready and I'll do next round approve fast. Will try to checkout this branch and start running some data with it now. Thank you!!

sotodlib/preprocess/processes.py
msilvafe: Left it in for now.

sotodlib/tod_ops/nmat_filter.py
msilvafe: Left in with some clarification on using it as a check.

msilvafe (approved):

PR #1710, opened 2026-08-07
perf: switch tags from selectin to joined

skhrg:
Massively speeds up my queries.


mhasself (approved):
I admit this seems like a good idea...

PR #1711, opened 2026-08-08
use as_completed in make_wg_cal

ykyohei:
We simply forgot to use `as_completed_callable`.
If we do this, we can get results in order of completion rather than submission, which is more efficient.


YutaroKASAI (approved):
Thank you!

PR #1713, opened 2026-08-10
Asymmetric turnarounds masking

ykyohei:
Updated turnaround flags to accept assymmetric buffer. 
We can set buffer either with time (seconds) or azimuth (degree).
This is mainly for masking systematics of turnaround heating which is known to be assymmetric.

Additionally modified az method to also wrap and return scan flags.
Added missing samps axis for right_scan, left_scan, turnarounds flags.


sotodlib/tod_ops/flags.py
```diff
@@ -313,28 +368,39 @@ def get_turnaround_flags(aman, az=None, method='scanspeed', name='turnarounds',
...
+            raise ValueError("Flag name left/right_flag already exists in aman.flags.")
+        else :
+            if "left_scan" in aman.flags:
+                aman.flags["left_scan"] = left_flag
+            else :
+                aman.flags.wrap("left_scan", left_flag, [(0, 'samps')])
+            if "right_scan" in aman.flags:
+                aman.flags["right_scan"] = right_flag
+            else :
+                aman.flags.wrap("right_scan", right_flag, [(0, 'samps')])
```
msilvafe, line 388: These are wrapping flags that were created before the `if truncate:` block so I think they won't match the samps shape if it went through the `truncate = True` block.

Since it seems like this truncate branch could cause some issues worth adding a test case for it in `test_turnaround_flags.py`

sotodlib/tod_ops/flags.py
```diff
@@ -313,28 +368,39 @@ def get_turnaround_flags(aman, az=None, method='scanspeed', name='turnarounds',
...
-    if method == 'scanspeed':
-        ta_exp = RangesMatrix([ta_flag for i in range(aman.dets.count)])
-        left_exp = RangesMatrix([left_flag for i in range(aman.dets.count)])
-        right_exp = RangesMatrix([right_flag for i in range(aman.dets.count)])
-        return ta_exp, left_exp, right_exp
-    
+    ta_exp = RangesMatrix([ta_flag for i in range(aman.dets.count)])
+    left_exp = RangesMatrix([left_flag for i in range(aman.dets.count)])
+    right_exp = RangesMatrix([right_flag for i in range(aman.dets.count)])
+    return ta_exp, left_exp, right_exp
```
msilvafe, line 405: preprocess `FlagTurnarounds` `calc_and_save` method has a `if self.calc_cfgs['method'] == 'az':` option that's now broken bc it expects just `ta_exp` to be returned.

sotodlib/tod_ops/flags.py
```diff
@@ -247,18 +292,44 @@ def get_turnaround_flags(aman, az=None, method='scanspeed', name='turnarounds',
...
         _right_flag = np.zeros(aman.samps.count, dtype=bool)
-        
+
         dt = np.mean(np.diff(aman.timestamps))
-        nbuffer_half = int(t_buffer/dt//2)
-        
+        if (t_buffer is None) and (az_buffer is None):
+            nbuffer_before, nbuffer_after = 0, 0
+        if (t_buffer is not None) and (az_buffer is not None):
+            raise ValueError('Either t_buffer or az_buffer must be specified')
```
msilvafe: I think this value error reads confusing. I think you want to say "Only t_buffer or az_buffer can be specified, not both.".

sotodlib/tod_ops/flags.py
```diff
@@ -203,18 +215,51 @@ def get_turnaround_flags(aman, az=None, method='scanspeed', name='turnarounds',
...
-    
+
+        if az_buffer is None:
+            buffer_before, buffer_after = 0., 0.
+        elif np.isscalar(az_buffer):
+            buffer_before, buffer_after = az_buffer / 2., az_buffer / 2.
+        elif len(az_buffer) == 2:
+            buffer_before, buffer_after = az_buffer
+        else:
+            raise ValueError
```
msilvafe: Add a descriptive error message.

sotodlib/tod_ops/flags.py
```diff
@@ -203,18 +215,51 @@ def get_turnaround_flags(aman, az=None, method='scanspeed', name='turnarounds',
...
+            az_smooth = az
+        daz = np.gradient(az_smooth)
+        if buffer_before == 0 and buffer_after == 0:
+            _ta_flag = np.logical_or(az < lo, az > hi)
+        else:
+            # top/bottom turnarounds, approaching/leaving
+            top_appr = (daz > 0) & (az > hi - buffer_before)
+            top_leav = (daz < 0) & (az > hi - buffer_after)
+            bot_appr = (daz < 0) & (az < lo + buffer_before)
+            bot_leav = (daz > 0) & (az < lo + buffer_after)
```
msilvafe: Did you intentionally leave off the daz = 0 case. Or stated another way why not daz <= 0 and daz >= 0 here?

msilvafe (changes requested):
Looks good thank you! Other than the unit tests did you test at all?

sotodlib/tod_ops/flags.py
ykyohei: You are right, this bug was fixed and verified by unittest now

sotodlib/tod_ops/flags.py
ykyohei: Thanks for catching this. There are some probabilies that daz gets zero and not flagged as turnarounds. I changed these to daz <= 0 and daz >= 0.

msilvafe (approved):

PR #1714, opened 2026-08-10
Minor bug fix in obtaining HWPSS mask

17-sugiyama:
This PR is for the minor bug fix of `get_psd_mask` in `fft_ops`.

Although `get_psd_mask` allows `hwp_freq` parameter, it was not applied to the calculation.
I have also found that `get_mask_for_hwpss` did not return proper masks when input `hwp_freq` was negative.

These issues hasn't been a problem when we use pipeline scripts, but is a problem when we want to use it for a special situation.
I have confirmed that fixed `get_psd_mask` works with the following inputs:

- `hwp_freq` = None
- `hwp_freq` = 2
- `hwp_freq` = -2

### DIFF
--- a/sotodlib/tod_ops/fft_ops.py
+++ b/sotodlib/tod_ops/fft_ops.py
@@ -661,7 +661,8 @@ def get_psd_mask(aman, psd_mask=None, f=None,
         psd_mask = psd_mask.mask()
 
     if mask_hwpss:
-        hwp_freq = hwp.get_hwp_freq(aman.timestamps, aman.hwp_solution.hwp_angle)
+        if hwp_freq is None:
+            hwp_freq = hwp.get_hwp_freq(aman.timestamps, aman.hwp_solution.hwp_angle)
         psd_mask = psd_mask | get_mask_for_hwpss(f, hwp_freq, max_mode=max_hwpss_mode, width=hwpss_width)
     if mask_peak:
         psd_mask = psd_mask | get_mask_for_single_peak(f, peak_freq, peak_width=peak_width)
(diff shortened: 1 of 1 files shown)


ykyohei (changes requested):
Thanks for fixing this!

> These issues hasn't been a problem when we use pipeline scripts, but is a problem when we want to use it for a special situation.

Can you explain bit more on this point? I guess this is because hwpss mask is only important for raw psd noise modeling and we didn't use noise model of raw psd?

I guess it might be better to change get_hwp_freq to always return positive frequency. Do we need signed hwp_frequency? 
In any case, please update its docstring to "this returns signed hwp frequency". Another option is to add `signed=False` argument to get_hwp_freq.
https://github.com/simonsobs/sotodlib/blob/316ae44508b89bf9ebe18632383bb00a0b387615/sotodlib/hwp/hwp.py#L571-L580

17-sugiyama:
> These issues hasn't been a problem when we use pipeline scripts, but is a problem when we want to use it for a special situation.

Regarding this statement, actually we do the raw psd fitting in the iso-sat pipeline, which you can find [here](https://github.com/simonsobs/iso-sat/blob/b1e0b67f1ef04ab60ffa61adf3e9f8897976e6cf/v4/preprocessing/satp1/preprocessing_config_260702_init.yaml#L291-L314).
`fit: True` represents that it estimates `white_noise` by the 1/f noise fit, and `mask: True` represents it automatically produces and applies the HWPSS mask.
The HWPSS mask estimated here is always valid.

And sorry for the confusion, I guess `get_hwp_freq` always returns a positive value.
My point is that current `get_mask_for_hwpss` accepts negative values, but returns array filled with False (nothing is masked). I feel like to take `abs(hwp_freq)` or insert `assert hwp_freq >0` in `get_mask_for_hwpss`.

ykyohei (approved):
Got it, thanks for the clarification. These changes look good to me.

PR #1716, opened 2026-08-12
Update get_obslist file handling

mmccrackan:
Just a minor change to make it more clear that `get_obslist` accepts either a file input or obsdb query. Came up when #1676 was being tested.

### DIFF
--- a/sotodlib/site_pipeline/utils/obsdb.py
+++ b/sotodlib/site_pipeline/utils/obsdb.py
@@ -1,6 +1,7 @@
 """Observation database query utilities for site_pipeline."""
 
 import time
+import os
 
 
 def get_obslist(context, query=None, obs_id=None, min_ctime=None, max_ctime=None,  # multilayer_preprocess_tod, preprocess_obs, preprocess_tod, update_preprocess_plots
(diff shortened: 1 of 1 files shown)


msilvafe:
I'm a bit confused by this PR...what exactly is this solving? In SAT-ISO we already routinely pass paths to a txt file with a list of obsids in it and it's handled just fine through the block:

```
try:
    with open(query, "r") as fname:
        return [context.obsdb.get(line.split()[0]) for line in fname]
except FileNotFoundError:
```

mmccrackan:
> I'm a bit confused by this PR...what exactly is this solving? In SAT-ISO we already routinely pass paths to a txt file with a list of obsids in it and it's handled just fine through the block:
> 
> ```
> try:
>     with open(query, "r") as fname:
>         return [context.obsdb.get(line.split()[0]) for line in fname]
> except FileNotFoundError:
> ```

Mostly it was just to make it more obvious in the code that this function accepts a text file as input (wasn't in the docstring).  I figured I'd just improve the handling of file inputs at the same time instead of just having it do a try except.  There were no bugs in the implementation.

msilvafe (approved):

PR #1717, opened 2026-08-12
Support crosstalk sims with real observations

tskisner:
- Support real metadata names as seen in a typical observation.
- If using real observations, do not require DetMap.
- Refactor crosstalk utils to extract code that is common between the simulated and real observing cases.
- In SimSSO operator, apply detector cuts if they exist.
- When using real data, translate band names to expected format
- Add det_mask trait and handle detectors which are flagged as invalid.
- Update detector flags if saturated
- Add support for per-detector elevation.  Remove calculation of loading based on the TOD.

### DIFF
--- a/sotodlib/toast/ops/mumux_crosstalk_util.py
+++ b/sotodlib/toast/ops/mumux_crosstalk_util.py
@@ -9,74 +9,102 @@ import numpy as np
 try:
     import pandas as pd
     from detmap.data_io.solution_select import AvailableSolutions
+
     available_solutions = AvailableSolutions()
     detmap_available = True
-except:
+except ImportError:
     detmap_available = False
 
-mappings = {}
 
-# SAT1 MF wafers  = w25-w31
-# SAT2 MF wafers  = w32-w38
-# SAT3 HF wafers  = w06-w12
-# SAT4 LF wafers  = w42-w48
-# LAT LF wafers  = w39-w41
-# LAT MF wafers  = w13-w24
-# LAT HF wafers  = w00-w05
+def wafer_chis(dets, pos, freq, is_north, mux_band, bond_pad, alpha, collision):
+    """Given detector properties on one wafer, compute the crosstalk.
 
-# Arbitrary mapping between wafer slots and array names
-wafer_to_array = {
-    "dummy00" : "Cv4",  # LAT
-    "dummy01" : "Cv5",
-    "w13" : "Mv6",   # LAT  1/12
-    "w14" : "Mv7",   # LAT  2/12
-    "dummy04" : "Mv9",
-    "w15" : "Mv11",  # LAT  3/12
-    "w16" : "Mv12",  # LAT  4/12
-    "w25" : "Mv13",  # SAT1 1/7
-    "w26" : "Mv14",  # SAT1 2/7
-    # "Mv15",  # LAT (missing)
-    "w17" : "Mv17",  # LAT  5/12
-    "w27" : "Mv18",  # SAT1 3/7
-    "w28" : "Mv19",  # SAT1 4/7
-    "w29" : "Mv22",  # SAT1 5/7
-    "w30" : "Mv23",  # SAT1 6/7
-    "w31" : "Mv24",  # SAT1 7/7
-    "w18" : "Mv25",  # LAT  6/12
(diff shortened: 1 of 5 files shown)


tskisner:
I will rebase this work after #1630 is merged.

sotodlib/toast/ops/mumux_crosstalk_util.py
```diff
@@ -104,36 +132,95 @@ def pos_to_ind(pos, bp, pol, mapping, tol=1.0, verbose=False):
...
+    # SAT1 MF wafers  = w25-w31
+    # SAT2 MF wafers  = w32-w38
+    # SAT3 HF wafers  = w06-w12
+    # SAT4 LF wafers  = w42-w48
+    # LAT LF wafers  = w39-w41
+    # LAT MF wafers  = w13-w24
+    # LAT HF wafers  = w00-w05
+
+    # Arbitrary mapping between wafer slots and array names
+    wafer_to_array = {
```
tpsatt, line 165: We could probably be less arbitrary here now that everything is deployed. Are you able to point me to the mapping of e.g. "w13" to the OT and wafer slot number? If so, I can rearrange this

tskisner: Sounds good to try and make that more realistic, but perhaps you could do that as a small separate PR once this is merged?

tskisner:
Ok, @tpsatt  this is ready for re-review.  I updated the unit test so that it is not run unless jbolo is found and the appropriate environment variables are set.  For example, to run the unit test locally, ensure that `DetMap` and `JBolo` are installed and then do something like:
```
JBOLO_PATH="/path/to/jbolo" \
JBOLO_MODELS_PATH="/path/to/bolocalc-so-model" \
OMP_NUM_THREADS=4 \
pytest -s sotodlib/tests/test_sim_mumux_crosstalk.py
```

tpsatt:
Confirming that the unit tests work, and that this produces sensible CMB sims for the LAT D56 region 
Screenshot 2026-09-11 at 3 55 56 PM

sotodlib/toast/ops/sim_mumux_crosstalk.py
```diff
@@ -248,95 +290,154 @@ def _exec(self, data, detectors=None, **kwargs):
...
                     )
-                    # Add crosstalk if not collided resonator or saturated detector
-                    if not (np.isnan(chi) or np.isnan(dPhi0dT[det_source])):
+
+                    # Add crosstalk if not collided resonator
+                    if not np.isnan(chi):
                         crosstalk += chi * np.sin(
                             source_squid_phase - target_squid_phase
                         )
+                        det_msg += f"\n  {det_source} (phi0={Phi0[det_target]:0.2e}, "
```
tpsatt: Should the Phi0 here be `det_source`?

sotodlib/toast/ops/sim_mumux_crosstalk.py
```diff
@@ -166,64 +192,80 @@ def _evaluate_dPhi0dT(self, obs, signal, detectors, rows, Phi0, pwv, elevation):
...
-            #pdb.set_trace()
-            ## Calculate a global median 
-            local_good = signal[row][good]
-            if obs.comm.comm_group is not None:
-                all_good = obs.comm.comm_group.gather(local_good, root=0)
-                if obs.comm.group_rank == 0:
-                    median_signal = np.median(np.concatenate(all_good))
+        for det in detectors:
+            raw_band = focalplane[det]["band"]
+            if raw_band == "DARK" or raw_band[0] == "f":
```
tpsatt: Should this handle the rest of the detector types like `NC` and `SQID`?

sotodlib/toast/ops/sim_mumux_crosstalk.py
tskisner: Good catch, fixed.

sotodlib/toast/ops/sim_mumux_crosstalk.py
tskisner: For real data, the starting detectors list should only include optical and dark (other channels will be pruned when loading from the context).  However, for completeness, I added a set of band types to check.

tpsatt (approved):
Great changes

PR #1718, opened 2026-08-13
Use JBolo to compute detector parameters for crosstalk

tpsatt:
Previously, we've used hardcoded values for detector parameters like atmospheric loading when computing dPhi0/dT in the umux crosstalk operator. Since we want to test different atmosphere conditions, and these heavily influence such properties, I've updated the operator to run a jbolo simulation for each observation using the actual simulated observing PWV and elevation.

In a test of 8 hours of simulated observing with the LAT, I found no noticeable change in execution time (the sims are very fast). I also confirmed that since the observation was at PWV just under 1 mm, the effect was at the ~uK level in the resulting maps.

One semi-open question: JBolo requires that two environment variables be set to access the right models, e.g.:
```
export JBOLO_PATH=/global/cfs/cdirs/sobs/users/tpsatt/umux_xtalk/jbolo
export JBOLO_MODELS_PATH=/global/cfs/cdirs/sobs/users/tpsatt/umux_xtalk/bolocalc-so-model
```
There's a check for these environment variables being set, but do we want to put these directories in some shared SO location on Perlmutter? Or leave it up to each user to have their own install?


tpsatt:
Also a question for @ttsan2521: this uses the observation elevation, but given the SAT's large field of view, the detector loading will actually vary rather significantly across the focal plane. Do we want to implement this? Is there an easy way to read off the "effective" elevation of a detector from the `obs` object?

I'm not worried about this for my LAT sims.

tskisner:
We could easily expand the detector pointing in Az/El and use the per-detector elevation...

sotodlib/toast/ops/sim_mumux_crosstalk.py
```diff
@@ -10,73 +11,36 @@
 from toast.timing import function_timer, Timer
 from toast.traits import Bool, Int, Unicode, trait_docs
 from toast.utils import Environment, Logger, unit_conversion
+import jbolo.jbolo_funcs as jf
+from jbolo.utils import load_sim
```
tskisner: We should put a try / except ImportError around this so that just importing sotodlib.toast does not error if jbolo is not available (similar to DetMap check).  Then when the operator exec is called, we can check that we have jbolo.

sotodlib/toast/ops/sim_mumux_crosstalk.py
```diff
@@ -267,9 +262,13 @@ def _exec(self, data, detectors=None, **kwargs):
             input_data = det_data.data.copy() * det_scale
             output_data = det_data.data  # just a reference
 
+            # Get observation parameters for dPhi0dT calculation
+            pwv = obs.telescope.site.weather.pwv.to_value(u.um)
+            el  = obs["scan_el"].to_value(u.degree)
```
tskisner, line 294: Here is where we could instead compute the mean elevation of each detector, and instead pass that to dPhi0dT.  That would require some more changes though, including passing the detector pointing operator as a trait and then using it.  We can do that in a separate PR.

tskisner (changes requested):

sotodlib/toast/ops/sim_mumux_crosstalk.py
```diff
@@ -206,14 +176,33 @@ def _evaluate_dPhi0dT(self, obs, signal, detectors, rows, Phi0):
...
+            P_atm_ref = P_atm_refs[band]
+            efficiency = efficiencies[band]
+            P_sat = P_sats[band]
+
+            # Compute conversion
             P_atm = efficiency * bandpass.optical_loading(det, median_signal)  # W
             P_opt += P_atm - P_atm_ref
             dPdT = bandpass.kcmb2w(det)  # K_CMB -> W
             R_TES = R_FRAC * R_BOLO
+            assert P_sat > P_opt # Check for saturated detectors
```
keskitalo: This is a useful check, but I wonder if we should handle failures more gracefully. Some options:
1. This is a critical error and we produce an informative error message (preferably with `log.error()`)
2. This is not critical, but we must flag the current detector as bad.

sotodlib/toast/ops/sim_mumux_crosstalk.py
```diff
@@ -231,6 +220,12 @@ def _exec(self, data, detectors=None, **kwargs):
                 "SimMuMUXCrosstalk cannot be run on subsets of detectors"
             )
 
+        # Check for JBOLO data path
+        if 'JBOLO_PATH' not in os.environ.keys() or 'JBOLO_MODELS_PATH' not in os.environ.keys():
```
keskitalo, line 251: JBOLO_PATH is not used anywhere in this file. Is it internal to JBolo? What happens to JBOLO_MODELS if JBOLO_MODELS_PATH is not set?

sotodlib/toast/ops/sim_mumux_crosstalk.py
keskitalo: It would also require separate JBolo calls for every detectors instead of every band. That may be prohibitively expensive.

keskitalo (approved):
Looks great! Please look over the comments and see if they merit more work.

sotodlib/toast/ops/sim_mumux_crosstalk.py
tpsatt: JBolo seems to be decently fast, and there was no noticeable increase in simulation time upon running it for every observation. But this would be an increase of several hundred times, so perhaps we'd have to try.

Is there any easy way to get a detector's elevation, or would we have to write a small helper?

sotodlib/toast/ops/sim_mumux_crosstalk.py
tpsatt: Good point. Just pushed a commit to handle this like we handled `DetMap`.

sotodlib/toast/ops/sim_mumux_crosstalk.py
tpsatt: This is a good point. I just pushed a commit to flag these detectors (and let their dPhi0dT be NaN, then skip crosstalk computation in this case)

sotodlib/toast/ops/sim_mumux_crosstalk.py
tpsatt: Just double checked and it's internal to JBolo. The sim will fail if this value is not set

ttsan2521:
Yes, I'm interested in implementing per-detector elevation since the SAT has a large FOV. Please let me know how, and then I can test this on the SAT sim. If it's a straightforward/easy implementation, then we can try that in this PR. If it's a bit more work, then perhaps we can consider it a separate PR?

tskisner (approved):
Thanks for the import guard!

tskisner:
@tpsatt , the unit test failures here are unrelated, and fixed in another PR.  I think it is fine to merge this.  I have another PR that reorganizes the utilities source to handle real metadata, and will rebase that once this is merged.

PR #1722, opened 2026-08-20
entry point fix from profiling changes

kmharrington:
Claude noticed the profiling update broke this. @JBorrow is this the way this one script should be fixed?


JBorrow (approved):
Yes, looks good. Sorry that I missed this.

PR #1724, opened 2026-08-25
feat: add func to propagate motion of latr to sky

skhrg:


sotodlib/coords/optics.py
```diff
@@ -547,6 +547,59 @@ def LAT_focal_plane(aman, zemax_path, x=None, y=None, pol=None, roll=0, tube_slo
...
+    sec2el, sec2xel = LAT_optics(zemax_path)
+    el0 = sec2el(np.array([0.0]), np.array([0.0]))
+    xel0 = sec2xel(np.array([0.0]), np.array([0.0]))
+    el = sec2el(np.array([sec[0]]), np.array([sec[1]]))
+    xel = sec2xel(np.array([sec[0]]), np.array([sec[1]]))
+
+    rot = np.deg2rad(roll + 180.0)
+    def to_xieta(el, xel):
+        xrot = xel * np.cos(rot) - el * np.sin(rot)
+        yrot = el * np.cos(rot) + xel * np.sin(rot)
```
mhasself, line 595: Probably fine, but a simple flat-sky rotation like this is exact if you go to xieta coords first.  In lonlat it's an approximation.

mhasself (approved):

sotodlib/coords/optics.py
skhrg: Good point. But if we change that we should do it in a separate PR since this happens in another spot that would change what the templates look like.

PR #1726, opened 2026-08-28
feat: add option to rotate out roll

skhrg:


sotodlib/coords/planets.py
```diff
@@ -578,6 +584,8 @@ def compute_source_flags(tod=None, P=None, mask=None, wrap=None,
         of pixels for the mask map.  This is to catch cases where an
         incorrect source has been passed in, for example, leading to a
         weird map footprint
+    unroll: If True then when constructing P rotate the coordinates such that
+            scans go left-right even when roll != 0.
```
mhasself: Fix indentation

mhasself (approved):
Thanks!  It looks right but please confirm (for the record)
- default behavior of unrolling remains the same as before
- you've tested this (though that sign for roll is what I would have guessed)

skhrg:
Confirming both to be true, unless there is a need to test on SAT data as well?

mhasself:
> Confirming both to be true, unless there is a need to test on SAT data as well?

Nah, go for it.

PR #1729, opened 2026-09-01
Add bandpass specific preproc noise select

mmccrackan:
Adds the ability to pass in a dictionary for the `noise` process stage `select` function parameters `min_noise`, `max_noise`, and `max_fknee`.  This is to avoid the LAT preprocessing from needing different configs, since noise is currently the only process that needs different values.  Maintains backwards compatibility and will throw TypeError if not a `float`, `int`, or `dict`.

### DIFF
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -908,7 +908,7 @@ class Noise(_Preprocess):
 
         if _has_tag(meta.det_info, "wafer.bandpass"):
             bandpasses = meta.det_info.wafer.bandpass
-        elif _has_tag(meta, "det_cal"):
+        elif _has_tag(meta, "det_cal.bandpass"):
             bandpasses = meta.det_cal.bandpass
         else:
             bandpasses = None
--- a/sotodlib/preprocess/processes.py
+++ b/sotodlib/preprocess/processes.py
@@ -729,7 +729,7 @@ class Noise(_Preprocess):
             wn_est: noise
             fixed_param: 'wn'
             binning: True
-            fit_method: log_curve_fit #or likelihood 
+            fit_method: log_curve_fit # or likelihood
             curve_fit_kwargs:
                 maxfev: 20000
           save: True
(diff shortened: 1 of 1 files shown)


sotodlib/preprocess/processes.py
```diff
@@ -729,12 +732,13 @@ class Noise(_Preprocess):
...
             binning: True
-            fit_method: log_curve_fit #or likelihood 
+            fit_method: log_curve_fit # or likelihood
             curve_fit_kwargs:
                 maxfev: 20000
           save: True
           select:
-            max_noise: 2000
+            max_noise: 2000 # can also be dict of bandpass specific values
+            max_fknee: 7
```
msilvafe: Add example of dict entry in docstring.

sotodlib/preprocess/processes.py
```diff
@@ -894,13 +898,39 @@ def select(self, meta, proc_aman=None, in_place=True):
...
         keep = np.ones_like(wn, dtype=bool)
+
         if "max_noise" in self.select_cfgs.keys():
-            keep &= (wn <= np.float64(self.select_cfgs["max_noise"]))
+            if isinstance(self.select_cfgs["max_noise"], (int, float)):
+                keep &= (wn <= np.float64(self.select_cfgs["max_noise"]))
+            elif isinstance(self.select_cfgs["max_noise"], dict):
+                for k, v in self.select_cfgs["max_noise"].items():
+                    mask = (meta.det_cal.bandpass == k)
+                    keep[mask] &= (wn[mask] <= v)
```
msilvafe: Why use `det_cal` instead of `det_info` here? Also this requires a context with some non-zero metadata so I think we need to check for this and raise some suitably clear error. Maybe like if det_info in meta then use det_info bandpass, else if det_cal in meta then use det_cal, else raise a value error that neither det_cal nor det_info exist in meta

msilvafe (changes requested):

sotodlib/preprocess/processes.py
mmccrackan: Yeah was handling `bandpass` in an incomplete manner.  It will now try to get `det_info.wafer.bandpass` first and `det_cal.bandpass` second.  If it doesn't find either it just sets it to `None` since we may not need it.  Trying to run the bandpass selects will then error out cleanly if it is `None`.

msilvafe (approved):

PR #1732, opened 2026-09-03
Improve noise select type handling

mmccrackan:
Adds `str` to the select type check introduced in #1729 since it wouldn't actually catch cases like 2e-6.


yoshinori-0778 (approved):

yoshinori-0778:
THanks!

PR #1736, opened 2026-09-08
Add planet mapmaker v2

yoshinori-0778:
This is updated PR of https://github.com/simonsobs/sotodlib/pull/1706.
Sorry I ended up making new pr. I probably make mistake when I merged lots of commits.

Almost of all comments by Katie and Matthew are addressed. So I would like to review again.

### DIFF
--- a/sotodlib/coords/helpers.py
+++ b/sotodlib/coords/helpers.py
@@ -739,7 +739,60 @@ class ScalarLastQuat(np.ndarray):
             return so3g.proj.quat.G3VectorQuat(temp)
         raise ValueError("Can only convert 1- or 2-d arrays to G3.")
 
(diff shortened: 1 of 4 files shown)


yoshinori-0778:
Request to @mhasself:
Could you comment on this? I am happy to modify the scripts if I can use that database.
> A note on this. I discovered recently that the extended obsdbs running in your user account do not include calibration observations. It would be good to have those expanded and to make them official. Otherwise we'll keep having parallel implementations pop up.

yoshinori-0778:
Questions:

Build and Test shows error because Database written in mapcat is not exsist in the one we are using in this build test. How should I manage this? Can I avoid this by removing `from planet_mapmaker.py import *` in __init__.py ?

PR #1739, opened 2026-09-08
Don't fail preprocessing on group error

mmccrackan:
There is a case on site-pipeline where the `get_groups` loop fails due to groups seemingly being empty.  This was previously set to raise when this happened for debugging, but I think we can let it continue to run if it is just a single obs_id.  This is probably just due to some metadata not being in place yet, but the added error from `get_groups` will help us confirm that.


msilvafe (approved):

msilvafe:
Should the same fix be propagated to multilayer_preprocess_tod and _sim?

mmccrackan:
> Should the same fix be propagated to multilayer_preprocess_tod and _sim?

I think this will only be temporary to confirm if it is indeed a `get_groups` issue and has only ever seemed to occur for preprocess_lite I think.  If so, we can remove the logging and just continue.  I tried loading the obs_id  that caused the flow to fail and ran `get_groups` manually but it seemed to work fine.

PR #1741, opened 2026-09-09
Support "non-linear" / "roll-dependent" focal plane distortions

mhasself:
These changes are needed to account for the non-linear distortions due
to the LAT's off-axis mirrors; otherwise only a single value of the
"roll" can be treated in the existing framework.

Introduces the idea of "focal_plane_template", a TOD member like
focal_plane that will be used as the source to compute the
(model-corrected) focal_plane.

This changes the apply_pointing_model interface a bit but probably not
in a way that users will notice.

This framework can also be used to support detailed refraction corrections, in the future.


amaurea (approved):
Looks fine. I checked the arc implementation, and it matches mine. Thanks for impementing this. Is the zemax model updated, or does it still just cover the central tube?

mhasself:
> Looks fine. I checked the arc implementation, and it matches mine. Thanks for impementing this. Is the zemax model updated, or does it still just cover the central tube?

Thanks.  The zemax version applies to the full focal plane.

Also @amaurea mentioned elsewhere that I should probably not import private functions from other submodules, and that `roll_dist_model` should perhaps be a string instead of an int (which makes sense to me).  So I may apply those changes before merging.

skhrg:
Looks good to me. Two nitpicky comments:

1. Seems like we make a ton of copies of the the focal plane? Is this strictly needed? I know you have the `in_place` arg for the distortion but it would be nice if I could just keep the same buffer the and modify it in place always (but copy over the template before distorting or something). I suppose this only really matters when we fit the pointing mode so I can always work around it somehow.
2. `dist_model‎` needs to be an int, but it seems inevitable that its going to get cast to float somewhere by someone, can we cast it to be safe? Also why not do `dist_model = params.get('roll_dist_model', 0)`?

mhasself:
> Looks good to me. Two nitpicky comments:
> 
>     1. Seems like we make a ton of copies of the the focal plane? Is this strictly needed? I know you have the `in_place` arg for the distortion but it would be nice if I could just keep the same buffer the and modify it in place always (but copy over the template before distorting or something). I suppose this only really matters when we fit the pointing mode so I can always work around it somehow.

Ah, ya I can probably remove a bunch of that.  This was sort of left over from when I hadn't figured out how I wanted to do the _template handling.  The "in_place" option is not worth the complexity it adds.

>     2. `dist_model‎` needs to be an int, but it seems inevitable that its going to get cast to float somewhere by someone, can we cast it to be safe? Also why not do `dist_model = params.get('roll_dist_model', 0)`?

Ok, latest code has this as a string.  How do you feel about that?  And I'll change it to `dist_model = params['roll_dist_model']`.  The default val of "none" is included in the lat_v1 and lat_v2 defaults.

skhrg:
> Ok, latest code has this as a string.  How do you feel about that?  And I'll change it to `dist_model = params['roll_dist_model']`.  The default val of "none" is included in the lat_v1 and lat_v2 defaults.

I think an int is preferable to a string, if only because then when you pass this to an optimizer and `int` can just be left fixed but a string will require a wrapper that splices the value in.

mhasself:
This has now been tested against the sogma implementation, and we're in good agreement.

Although @skhrg prefers numerical parameters, I don't really want to go down the road where we build in a reluctance towards including strings.

I think I'm just waiting for some final agreement about the "optics" model details, then will merge if there are no other issues.

skhrg:
> Although @skhrg prefers numerical parameters, I don't really want to go down the road where we build in a reluctance towards including strings.

I'll get over it

> I think I'm just waiting for some final agreement about the "optics" model details, then will merge if there are no other issues.

Your coordinates here: 
https://docs.google.com/presentation/d/1ilqBhN-2FgRzHW_oIUqLXktYzxEJxzzv-7EsOMWA2aY/edit?slide=id.g3fa9381d77e_0_52#slide=id.g3fa9381d77e_0_52
check out to me. So I think its probably good to go.

skhrg (approved):

PR #1743, opened 2026-09-11
Stimulator pipeline code

dixilo:
This PR adds pipline code for the stimulator.
The `make_stm_cal.py` is the main script for this pipeline, where it makes HDF5 output files as output products of stimulator's gain and time constant analysis using `stimulator.py`. 
Small modification on the `stimulator.py` is intended to fix the assignment so that concatenation of the AxisManager works properly.
Modification on `pyproject.toml` should be needed since `stimulator.py` uses `ruptures` package.


sotodlib/site_pipeline/make_stm_cal.py
```diff
@@ -0,0 +1,320 @@
...
+    try:
+        ctx = core.Context(context_path, metadata_list=metadata_list)
+        meta = ctx.get_meta(obs_id, ignore_missing=True)
+
+        if hasattr(meta, 'det_cal') and hasattr(meta.det_cal, 'r_frac'):
+            meta.restrict(
+                'dets',
+                (meta.det_cal.r_frac > r_frac_min)
+                & (meta.det_cal.r_frac < r_frac_max),
+            )
```
ykyohei: Please delete data selection from this pipeline. We can apply this later and we should not apply in this stage.

pyproject.toml
```diff
@@ -24,6 +24,7 @@ dependencies = [
   "pixell>=0.23.0",
   "pytest",
   "scikit-image",
+  "ruptures",
```
ykyohei: Delete this. ruptures is already added to `project.optional-dependencies: site-pipeline` as optional dependency.

sotodlib/site_pipeline/make_stm_cal.py
```diff
@@ -0,0 +1,320 @@
...
+                        output_keys.append('gain')
+                    if obs_type in ('time_constant', 'gain_and_timeconstant'):
+                        output_keys.append('timeconstant')
+
+                    for key in output_keys:
+                        field_name = 'stm_gain' if key == 'gain' else 'stm_timeconstant'
+                        sub = core.AxisManager(result.dets)
+                        sub.wrap(field_name, result[field_name], [(0, 'dets')])
+
+                        h5_fn = f'stm_{key}_{oid}.h5'
```
ykyohei: Making hdf5 file for each observation will produce many small files, which is not very efficient and is not recommended by sotodlib.
I would suggest to split the file based on several digits of unixtime like following.
I think 5 digits would be good for this.
https://github.com/simonsobs/sotodlib/blob/cce97f0c9e77a6281d8c14aa0db58211fc1afb13/sotodlib/site_pipeline/update_det_cal.py#L95-L98

sotodlib/site_pipeline/make_stm_cal.py
```diff
@@ -0,0 +1,320 @@
...
+    stm_config_dict,
+    metadata_list,
+    obs_id,
+    obs_type,
+    n_split,
+):
+    """Process one stimulator observation.
+
+    Dispatches to calc_gain / calc_timeconstant based on obs_type.
+    Returns (obs_id, result AxisManager) where result holds stm_gain and/or
```
ykyohei: Description of return values are not correct.

ykyohei (changes requested):
Thanks for the PR! I have some comments and inlince comments.

1. To make this metadata easier to use, we need to make additional ManifestDb for the CMB or other calibration observations.
I think `update_bgmap_delay` is good example.
https://github.com/simonsobs/sotodlib/blob/cce97f0c9e77a6281d8c14aa0db58211fc1afb13/sotodlib/site_pipeline/update_bgmap_delay.py#L1156-L1171

2. `readout_delay` and `stm_gain_with_tau_correction` are also important calibration parameters that users will use, but it's not included in "stm_gain" and "stm_timeconstant" archives now.

3. I'm not sure if we need to make "stm_gain" and "stm_timeconstant" archives. For example, we can use "load_fields" to load only stm_gain, stm_tau, readout_delay etc from stm_cal archive.
https://github.com/simonsobs/sotodlib/blob/cce97f0c9e77a6281d8c14aa0db58211fc1afb13/sotodlib/core/metadata/loader.py#L850-L857

4. Could you also make (draft) PR of config files for this pipeline in site-pipeline-configs repository?
https://github.com/simonsobs/site-pipeline-configs/tree/main/lat

sotodlib/site_pipeline/make_stm_cal.py
dixilo: I see, I will remove the data selection.

pyproject.toml
dixilo: Thanks for pointing this out. I removed the `ruptures` from [project] section

sotodlib/site_pipeline/make_stm_cal.py
dixilo: Thanks for pointing this out. I modified the description according to the modification of the return values.

sotodlib/site_pipeline/make_stm_cal.py
dixilo: Thnaks for the information.
I assume it will be better for LAT to also have tube slot information so I'm going to rewrite file path like the following:
```python
oid_spl = oid.split('_')
unix = oid_spl[1][:5]
tube_slot = oid_spl[2]
h5_fn = f'stm_cal_{tube_slot}_{unix}.h5'
h5_path = os.path.join(output_dir, h5_fn)
```
Please let me know if you have any opinion on that.

ykyohei: This sounds reasonable to me. I suggest to make unix digits configurable, though.

dixilo:
Hi Kyohei,

Thanks for the review comments!
I made a modification according to your suggestions.

1. ManifestDb structure
I made additional helper functions and database operation codes so that ManifestDbs will have information associating regular observatoins and stimulator calibration.
The code structure was changed from the `update_bgmap_delay` (especially I didn't used codes like `get_cal_obsids`) since it calls DBs massively. Updating DBs at the end of `_main` function still requires ~5 min (measured with `tiger` with 1 CPU).

2. `readout_delay` and `stm_gain_with_tau_correction`
ManifestDBs for these values are also prepared. As I will mention later there are no longer separate h5 files.

3. Archve structure
I made a modification so that it would no longer create separate archives.
Since users can get associated stimulator's `obs_id` from ManifestDb and load the necessary section of the HDF5 file in the curernt Db structure, it would be necessary.

4. Yes I made a PR for the pipeline config as well:
https://github.com/simonsobs/site-pipeline-configs/pull/282

ykyohei (commented):
Thanks for all the changes. I suggest to reduce the number of manifest dbs from 4 to 2.
- stm_gain.sqlite
- stm_time_constant.sqlite
- stm_gain_with_tau_correction.sqlite -> drop
- stm_readout_delay.sqlite -> drop

There are two reasons.
stimulator time constant and readout delay should always exist together.
I also think observation with `time_constant` tag is supposed to produce both gain and time_constant, because the gain will be calculated with lowest chopping frequency part of data. I should confirm to @YudaiSeino.
https://github.com/simonsobs/sotodlib/blob/cce97f0c9e77a6281d8c14aa0db58211fc1afb13/sotodlib/stimulator/stimulator.py#L122-L126

Then the context file will be
```
- db: '.../lat_stimulator_cal_260913m/stm_gain.sqlite'
    label: stm_gain
    load_fields:
        - stm_gain
- db: '.../lat_stimulator_cal_260913m/stm_time_constant.sqlite'
    label: stm_tau
    load_fields:
        - stm_tau
        - readout_delay
        - stm_gain_with_tau_correction
```

For the `update_obs_corresp`, if I understand correctly this runs over all the observations everytime, is this intentional? Do we only need to run new obsevations because this runs after the stimulator data processing?

PR #1744, opened 2026-09-11
feat: ignore vector groups when determining output structure

skhrg:
Mainly so I can include `dets['readout_id'] = [id0, id1, ...]` without things crashing when I want to debug stuff.

### DIFF
--- a/sotodlib/preprocess/preprocess_util.py
+++ b/sotodlib/preprocess/preprocess_util.py
@@ -1001,7 +1001,8 @@ def find_db(obs_id, configs, dets, context=None, logger=None):
     if context is None:
         context = core.Context(configs["context_file"])
     group_by = np.atleast_1d(configs['subobs'].get('use', 'detset'))
-    cur_groups = [list(np.fromiter(dets.values(), dtype='<U32'))]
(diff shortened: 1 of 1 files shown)


sotodlib/preprocess/preprocess_util.py
```diff
@@ -1097,7 +1098,8 @@ def get_preproc_group_out_dict(obs_id, configs, dets, context=None, subdir='temp
     if context is None:
         context = core.Context(configs["context_file"])
 
-    cur_groups = [list(np.fromiter(dets.values(), dtype='<U32'))]
+    # cur_groups = [list(np.fromiter(dets.values(), dtype='<U32'))]
+    cur_groups = [str(val) for val in dets.values() if np.isscalar(val)]
     group_by = np.atleast_1d(configs['subobs'].get('use', 'detset'))
     newpath = f'{subdir}/{obs_id}'
     for cg in cur_groups[0]:
```
mmccrackan, line 1105: I think this line and any others that refer to `cur_groups[0]` will to be updated since your new code doesn't return a list of lists, but just a single list.

mmccrackan (changes requested):

PR #1745, opened 2026-09-13
feat: add fast batch update func

skhrg:
Using the new `update_jobs` function (with the index added) is ~50x faster than the pattern in `unlock` for a single job for me.


mhasself (commented):
Recalling that I'm not that into sqlalchemy  -- why all the manual updating of the fields?  The jobdb.unlock uses session.merge -- shouldn't that work here?  Merge them all, commit, expunge them all.  Return the updated, expunged records.

skhrg:
Honestly we are hitting the limits of my sqlalchemy too... I noticed that a manual update of just `jstate` was way faster so I tried a full manual update and it was also faster.
I imagine there is some subtlety here in that `merge` does a bunch of ORM stuff, so if we want to use the detached `job` after this `merge` is probably safer. But realistically you want to update the db on disk once and never touch the job again in the same session.

PR #1746, opened 2026-09-14
move field of proc_aman

ykyohei:
We would like to remane or remove field of proc_aman.


msilvafe (approved):
Thanks for this quick add. Looks good.

PR #1748, opened 2026-09-16
Change ACU Flag Behavior

kmharrington:
@mmccrackan  found that the acu_drops flag on the LAT has been looking _really_ bad since May. 

It's the co-rotator data stream and not the az or elevation streams. Luckily, the flags are all written separately. This PR changes the behavior to not load the co-rotator flag automatically. This will also prevent the SAT boresight flag from loading.

Open invitation for anyone to make the more configurable and less hardcoded.


mhasself (approved):
Seems fine to me -- might be good to include `flags.acu_drops` explicitly in the commit message, though -- e.g. "Changes how flags.acu_drops is computed".

mmccrackan (approved):

PR #1749, opened 2026-09-16
Fix wafer mappings for crosstalk calculation

tpsatt:
This makes the mapping of wafers in TOAST to actual UFMs deployed in the same position in the field. A few wafers in the field don't have DetMap solutions, so I've picked wafers of the same passband which sit a few spots apart. There's only one UHF wafer for each of SATp2 and the LAT, so those solutions are doing heavy duty.

One consideration is we don't have DetMap solutions for any LF wafers, so LF sims are not supported by this operator.

Also updated the keys of non-optical detectors to reflect all possibilities.


tskisner (approved):
Looks good, thanks for the improvements!

ttsan2521 (approved):
Looks good to me.

PR #1750, opened 2026-09-16
Allow turnaround flags to change wrap name.

msilvafe:
Unfortunately our plan to use `move` to do 2 turnaround steps didn't work because the behavior on load isn't the same as when run the first time and saved. @chervias found this issue:

During the original preprocessing run:
1. The first turnaround step calculates `turnaround_flags`.
2. move renames it to `turnaround_flags_asym`.
3. The second turnaround step calculates a new `turnaround_flags`.

During mapmaking, the pipeline loads saved metadata and skips `calc_and_save`, but still executes `process`:
1. Both fields are loaded from the archive.
2. `move` removes `turnaround_flags`, overwriting `turnaround_flags_asym`.
3. The second turnaround step processes the TOD but does not recreate the saved metadata field.
4. `combine_flags` requests `turnaround_flags.turnarounds` and raises the reported `AttributeError`.

Here I've just resolved this by updating `FlagTurnarounds` to save under different `wrap_name` that way we can remove the `move` process.


ykyohei (approved):
Thanks for the fix, and sorry for the trouble.

